PR #6, opened 2020-08-06
cli: Add empty neofs-cli app structure

realloc:
In the following release `neofs-cli` will be used to directly manage NeoFS Node.
All required definitions and interfaces are also moving from `neofs-api` to
`neofs-node` repository, so it's more convinient to have `neofs-cli` here.

Signed-off-by: Stanislav Bogatyrev <stanislav@nspcc.ru>

### DIFF
--- a/Dockerfile.cli
+++ b/Dockerfile.cli
@@ -1,0 +1,21 @@
+FROM golang:1.14-alpine as basebuilder
+RUN apk add --update make bash
+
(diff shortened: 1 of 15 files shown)


alexvanin (approved):
Looks good, but I would like to merge it after `neofs-api v2.0.0` will be released, so we can look at the list of modules once again.

PR #11, opened 2020-08-24
Initialize transition to v2

cthulhu-rider:


alexvanin:
In next PR remaining services will return back with neofs-api v2.0 protocol definition. Jindo is a work-in-progress branch.

PR #17, opened 2020-09-02
Adopt local object storage for latest SDK

cthulhu-rider:
## Summary

- Implement `Object` and `Address` types in `core/object`;

- Make local object storage to use types and functions from `core`.


alexvanin:
With improved blob-storage we will store metadata as objects, not the raw bytes.

PR #28, opened 2020-09-15
Make storage node configurable

alexvanin:
To run storage node at dev-env environment it should have configurable parameters. To keep `cfg` structures we can read configuration from yaml files and unmarshal it into `cfg` structures. Configurable fields must be public, though.

Example of `config.yml` will be available at dev-env.


alexvanin:
Pull request have been updated with `viper` configuration. This way we can read both `yaml` files and use env variables. Later on config section might be changed according to #29

cthulhu-rider (approved):

realloc (approved):

PR #31, opened 2020-09-17
Implement container placement traverser

cthulhu-rider:


pkg/services/object_manager/placement/traverser.go
```diff
@@ -0,0 +1,182 @@
...
+	if t.rem == 0 || t.nextI == len(t.vectors) {
+		return nil
+	}
+
+	addr, err := network.AddressFromString(t.vectors[t.nextI][t.nextJ].NetworkAddress())
+	if err != nil {
+		// TODO: log error
+	}
+
+	if t.nextJ++; t.nextJ == len(t.vectors[t.nextI]) {
```
alexvanin, line 100: Is this going to work correctly if two vectors have different length?

cthulhu-rider: It should work with any two-dimensional slice. Do you see any problems here?

alexvanin: Okay, I see now how it works.

Another question: what if there is a placement rule like 
```
PUT 2 in HDD
PUT 1 in SSD
...
SELECT 6 Nodes .. As HDD
SELECT 4 Nodes .. As SSD
```
Does traverser work with single replica statement? In this case traverser will be created twice, one for HDD with `t.rem == 2` and one for SSD with `t.rem == 1`.

pkg/services/object_manager/placement/traverser.go
```diff
@@ -66,59 +67,106 @@ func NewTraverser(opts ...Option) (*Traverser, error) {
...
+	} else if cfg.policy == nil {
+		return nil, errors.Wrap(errNilPolicy, invalidOptsMsg)
 	}
 
 	ns, err := cfg.builder.BuildPlacement(cfg.addr, cfg.policy)
 	if err != nil {
 		return nil, errors.Wrap(err, "could not build placement")
 	}
 
+	ss := cfg.policy.GetSelectors()
```
alexvanin, line 80: if `cnt` is a required amount of copies to place in one vector (replica), then it should be `ss := cfg.policy.GetReplicas()` and then `cnt = int(ss[i].GetCount())`

alexvanin (approved):

PR #32, opened 2020-09-21
Add basic ACL checker in object service pipeline

alexvanin:


pkg/services/object/acl/basic.go
```diff
@@ -0,0 +1,398 @@
...
+)
+
+var (
+	ErrMalformedRequest  = errors.New("malformed request")
+	ErrUnknownRole       = errors.New("can't classify request sender")
+	ErrUnknownContainer  = errors.New("can't fetch container info")
+	ErrBasicAccessDenied = errors.New("access denied by basic acl")
+)
+
+// NewBasicChecker is a constructor for basic basicACLHelper checker of object requests.
```
cthulhu-rider, line 56: Seems like `basicACLHelper` is a typo

pkg/services/object/acl/basic.go
```diff
@@ -0,0 +1,398 @@
...
+	info.requestRole = role
+	info.operation = op
+	info.owner = cnr.GetOwnerID()
+
+	return info, nil
+}
+
+func getContainerIDFromRequest(req interface{}) (id *refs.ContainerID, err error) {
+	defer func() {
+		// if there is a NPE on get body and get address
```
cthulhu-rider, line 305: Can it really arise?

pkg/services/object/acl/basic.go
alexvanin: Fixed.

pkg/services/object/acl/basic.go
alexvanin: Sure, one can send request with `nil` owner id. I don't mind to have panic recovery directly in network/transport/object/grpc though. In this case recovery there can be omitted. What do you think?

cthulhu-rider: Current code inside function body does not seem to panic cause of this since used getters are NPE-safe.

alexvanin: They are not NPE-safe unfortunately

```
func (r *GetRequestBody) GetAddress() *refs.Address {
	if r != nil {
		return r.addr
	}

	return nil
}
```

cthulhu-rider:
```
func TestNPE(t *testing.T) {
	var req *object2.GetRequest

	require.NotPanics(t, func() {
		req.GetBody().GetAddress().GetContainerID().GetValue()
	})
}

--- PASS: TestNPE (0.00s)
PASS
```

alexvanin: Oh, I see, they all have that nil check inside. Okay, I'll remove this recovery but still wonder about recovery at network/transport/object/grpc just in case.

alexvanin: Fixed in https://github.com/nspcc-dev/neofs-node/pull/32/commits/6812073152511f371bc572397518743cc790cd97

cthulhu-rider (approved):

PR #42, opened 2020-09-24
object/transformer: Fix incorrect checksums of parent object

cthulhu-rider:
In previous implementations object size limiter left checksums of parent
object payload unset. This was caused by the closure that written calculated
checksums to the child object instead of parent one.

Construct payload checksum writers in separate function in order to prevent
closure bug.


alexvanin (approved):
Don't forget to remove `#[xx]` from commit message.

PR #54, opened 2020-09-28
[#45] object/search: Add filtering parent objects

### LINKED ISSUE #45: Container's object listing ignored split objects
When searching for objects from particular container, split objects are ignored, because they do not exist as a single object.

## Expected Behavior
Container listing returned all objects: regular objects, split objects, split object parts.

## Current Behavior
Split object are ignored

## Possible Solution
On each node serving search request for container listing, additionally list all OIDs from Parent fields and merge those lists in response.

cthulhu-rider:
In previous implementation object.Search services allowed to search only
physically stored objects. This limitation did not allow getting the ID of
the split object.

Extend search execution logic with parent object filtering. Parent objects
that passed filters are now included in the result

Closes #45.

### DIFF
--- a/pkg/services/object/search/local.go
+++ b/pkg/services/object/search/local.go
@@ -26,8 +26,6 @@ type searchQueryFilter struct {
 }
 
 func (s *localStream) stream(ctx context.Context, ch chan<- []*objectSDK.ID) error {
-	idList := make([]*objectSDK.ID, 0)
-
 	filter := &searchQueryFilter{
 		query: s.query,
 		ch:    ch,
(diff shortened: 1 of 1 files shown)


alexvanin:
With this fix search request returns both split objects and their _virtual_ parents in one response by default?

cthulhu-rider:
> With this fix search request returns both split objects and their _virtual_ parents in one response by default?

@alexvanin, exactly (if _virtual_ object matches the query).

alexvanin:
Don't we want to have filter that returns only non-splitted and _virtual_ objects? Something like `Filter{name:"Object.Split", match:StringEqual, val:"NULL"}` or something like that? This query would be quite helpful for container listening in data audit and gate applications.

cc: @cthulhu-rider @realloc

realloc:
> Don't we want to have filter that returns only non-splitted and _virtual_ objects? Something like `Filter{name:"Object.Split", match:StringEqual, val:"NULL"}` or something like that? This query would be quite helpful for container listening in data audit and gate applications.

Having "service" objects to be filtered out from the search result would be also useful for GUI and protocol gateways. Maybe we should add some flag or option to implicitly alter search query for that?

alexvanin (approved):

alexvanin:
Extra filters discussed at https://github.com/nspcc-dev/neofs-api-go/issues/163 and will be implemented in separate PR.

PR #57, opened 2020-09-29
services/object: Sign requests with session key within user session

cthulhu-rider:
## Summary

- fix linking object validation;

- combine common service parameters in a separate pkg;

- add session token to common parameters;

- sign client requests with private session key within user session.


cthulhu-rider:
The signature of requests in the Put service will be added as needed in a separate PR.

alexvanin (approved):
LGTM

PR #58, opened 2020-09-30
services/object: Implement Delete service

cthulhu-rider:


pkg/core/object/fmt.go
```diff
@@ -35,6 +35,10 @@ func (v *FormatValidator) Validate(obj *Object) error {
 		return errNilCID
 	}
 
+	if err := v.validateContent(obj.GetType(), obj.GetPayload()); err != nil {
```
alexvanin, line 38: Format validator invoked on header of the input object. Header does not contain payload, so this validation will always fail for tombstone object.

alexvanin (changes requested):

pkg/core/object/fmt.go
cthulhu-rider: Fixed.

alexvanin (approved):

PR #73, opened 2020-10-05
Use request owner public key in eACL check

alexvanin:
Classifier fetches public key of the request owner and owner itself. Extended ACL check should rely on this public key, because it might be extracted from session token.


realloc (approved):
lgtm

PR #75, opened 2020-10-05
Fix order of update peer state method arguments

alexvanin:
This method has node status first and public key second.


realloc (approved):
Just push such minor stuff directly to `master` =)

PR #90, opened 2020-10-14
[#46] Make SELECT attribute optional

### LINKED ISSUE #46: Parse placement policies without specific attributes
Some placement policies do not specify selection attributes e.g. "put two copies in container of 6 nodes with SSD". This can be presented as

```
REP 2
SELECT 6 FROM F
FILTER StorageType EQ SSD as F
``` 

Take generic policy "put two copies in container of 6 nodes":

```
REP 2
SELECT 6 FROM *
```

## Expected Behavior

These queries parsed with selectors:

```
SelectorExample1 {           SelectorExample2 {
  name:     ""                 name:     ""
  count     6                  count     6
  clause    SAME               clause    SAME
  attribute ""                 attribute ""
  filter    "F"                filter    "*"
}                            }

```

## Current Behavior
Placement policy parser grammar requires `IN` token in selectors to specify bucket, that makes all these queries invalid.
```
2:10: unexpected token "FROM" (expected "IN")
```

## Possible Solution
Make `IN` token optional in placement policy parser grammar.

cc: @realloc @fyrchik

fyrchik:
Fix #46.

### DIFF
--- a/pkg/policy/grammar.ebnf
+++ b/pkg/policy/grammar.ebnf
@@ -15,7 +15,7 @@ CbtStmt ::= 'CBF', Number1 (* container backup factor *)
 
 SelectStmt ::=
     'SELECT', Number1,    (* number of nodes to select without container backup factor *)
-    'IN', Clause?, Ident, (* bucket name *)
+    ('IN', Clause?, Ident)?, (* bucket name *)
     FROM, (Ident | '*'),  (* filter reference or whole netmap *)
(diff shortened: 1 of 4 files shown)


alexvanin:
> Works great except one thing.

The issue is that default clause is a `DISTINCT`, but all nodes with empty select attribute are virtually placed in same bucket. We've decided to fix this in neofs-api-go by virtually placing each node in separate bucket. It makes sense because in the end each node is unique (with unique private key).

PR #93, opened 2020-10-14
Add gRPC Reflection service

im-kulikov:
Proposal to add gRPC Reflection service that can be enabled by settings `grpc.enable_reflect_service`.

By default, service is disabled and can be enabled only by `grpc.enable_reflect_service`

Read more about at
https://github.com/grpc/grpc-go/blob/master/Documentation/server-reflection-tutorial.md

> gRPC Server Reflection provides information about publicly-accessible gRPC services on a server, and assists clients at runtime to construct RPC requests and responses without precompiled service information. It is used by gRPC CLI, which can be used to introspect server protos and send/receive test RPCs.

Signed-off-by: Evgeniy Kulikov <kim@nspcc.ru>

### DIFF
--- a/cmd/neofs-node/grpc.go
+++ b/cmd/neofs-node/grpc.go
@@ -5,6 +5,7 @@ import (
 	"net"
 
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/reflection"
 )
 
 func initGRPC(c *cfg) {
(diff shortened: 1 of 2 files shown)


cmd/neofs-node/grpc.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"net"
 
+	"google.golang.org/grpc/reflection"
+
```
alexvanin, line 8: Do we need that empty line there?

alexvanin (commented):
While it is configurable, don't see any issues with that. Don't forget to update commit message with `[#93]` header.

cmd/neofs-node/grpc.go
im-kulikov: Nope, my fault

im-kulikov: fixed

alexvanin (approved):

PR #112, opened 2020-10-21
Set CreationEpoch field value for new objects

### LINKED ISSUE #60: Object pipeline doesn't set creation_epoch field in object.
## Expected Behavior
New objects have non zero `creation_epoch` field. 

## Current Behavior
Objects have zero in `creation_epoch` field.

## Steps to Reproduce
1. Put object.
2. Head object.
3. Check `creation_epoch` field.

cthulhu-rider:
Closes #60.

### DIFF
--- a/cmd/neofs-node/netmap.go
+++ b/cmd/neofs-node/netmap.go
@@ -5,10 +5,31 @@ import (
 	netmapGRPC "github.com/nspcc-dev/neofs-api-go/v2/netmap/grpc"
 	crypto "github.com/nspcc-dev/neofs-crypto"
 	"github.com/nspcc-dev/neofs-node/pkg/morph/event"
+	netmapEvent "github.com/nspcc-dev/neofs-node/pkg/morph/event/netmap"
 	netmapTransportGRPC "github.com/nspcc-dev/neofs-node/pkg/network/transport/netmap/grpc"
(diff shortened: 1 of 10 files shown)


cmd/neofs-node/netmap.go
```diff
@@ -29,6 +50,11 @@ func initNetmapService(c *cfg) {
 			),
 		),
 	)
+
+	c.cfgNetmap.state = newNetworkState()
+	addNewEpochNotificationHandler(c, func(ev event.Event) {
+		c.cfgNetmap.state.setCurrentEpoch(ev.(netmapEvent.NewEpoch).EpochNumber())
+	})
```
alexvanin, line 57: One more thing: update network state with epoch value from sidechain at `bootstrapNode()` function. This way node will be ready to create objects right from the start.

cthulhu-rider: What component and function should I use in `bootstrapNode`?

alexvanin: I think network map wrapper missing `Epoch` method implementation to get current epoch value. Fix it and then invoke this method in `bootstrapNode()`. 

```go
epoch := c.cfgNetmap.wrapper.Epoch()
c.cfgNetmap.state.setCurrentEpoch(epoch)
```

alexvanin (approved):

PR #113, opened 2020-10-21
Process bearer token in ACL service

alexvanin:
If bearer token is presented in the request then check if it is a valid one and then use it to process extended ACL checks.

There are several steps to check if bearer token is valid:
- it should be signed correctly,
- it should be signed by container owner,
- request sender should be specified in `token.OwnerID` field if it is not empty,
- it should have correct lifetime (check will be implemented later).

### DIFF
--- a/cmd/neofs-node/object.go
+++ b/cmd/neofs-node/object.go
@@ -353,6 +353,7 @@ func initObjectService(c *cfg) {
 					eacl.WithMorphClient(c.cfgObject.cnrClient),
 					eacl.WithLogger(c.log),
 				),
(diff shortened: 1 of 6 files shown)


cthulhu-rider (commented):
Сhecks look correct, but did you forget to check the permission to use bearer token `basicACL.BearerAllowed()`?

alexvanin:
> Сhecks look correct, but did you forget to check the permission to use bearer token `basicACL.BearerAllowed()`?

Yes, forgot about that, thank you!

cthulhu-rider (approved):

Issue #122, opened 2020-10-23
Fit Fixed16 balance from sidechain in decimal response


alexvanin:

cthulhu-rider:
Found that if the balance exceeds maximum `int64` value, we receive stack item with `IntegerT` and cut int64 value (e.g. `923 * 10 ^ 16` -> `-9216744073709551616`. 

At the same time `IntFromStackItem` utility function calls [Int64](https://golang.org/pkg/math/big/#Int.Int64) method, that returns undefined value if integer overflows int64. As a solution, we can return `*big.Int` value from `IntFromStackItem` function.

alexvanin:
> Found that if the balance exceeds maximum `int64` value, we receive stack item with `IntegerT` and cut int64 value (e.g. `923 * 10 ^ 16` -> `-9216744073709551616`.
> 
> At the same time `IntFromStackItem` utility function calls [Int64](https://golang.org/pkg/math/big/#Int.Int64) method, that returns undefined value if integer overflows int64. As a solution, we can return `*big.Int` value from `IntFromStackItem` function.

While it is correct, the roots of the issue are a bit deeper.

First of all, both neo-go and neo runtimes are capable to work with 256-bit integers. However, both of them use JSON to encode and decode invocation arguments and stack items of invocation result. In [RFC 7159](https://tools.ietf.org/html/rfc7159) section 6 there is a _recommendation_ to encode numbers in range `[-(2**53)+1, (2**53)-1]`. 

Neo node [throws error](https://github.com/neo-project/neo/blob/cd988e85222f4d9df614f969668d2a915094b2e2/src/neo/SmartContract/JsonSerializer.cs#L97) if the number is out of this bound, while neo-go node can return any `big.Int` number. However neo-go node can't pass `big.Int` number as an invocation parameter, it works only with  `int64` numbers.

---
Back to the balance contract. Now it has `Fixed16` precision. While it can correctly process `Fixed16` inside the contract, `balanceOf` and `transfer` methods can't return or get a number bigger than `2e53-1`. In decimal integer values it will be:
```
# Max decimal integer value
Fixed8:  90 071 992
Fixed12: 9 007
Fixed16: 9
```

With lower precision, we can limit the max size of a single `Deposit` or `Withdraw` transaction so it will produce `transfer` invocation within the bound. However several `Deposit` still can accumulate quite a big number to break the bound. 

[...]

PR #126, opened 2020-10-27
Use dynamic precision convertations in inner ring

alexvanin:
Related to #122


alexvanin:
Added `precision.Convert()` for in-place conversions. This could be useful in `accounting.Balance` when `big.Int` value from RPC node should be packed into `Decimal` value with `int64` value.

cthulhu-rider (approved):

PR #127, opened 2020-10-28
Reduce returned balance precision in accounting.Balance

### LINKED ISSUE #122: Fit Fixed16 balance from sidechain in decimal response

alexvanin:
Closes #122 

Before patch with `Fixed16` precision in balance smart contract:
1. Deposit 900 GAS
1. Deposit 900 GAS
1. `neofs-cli -c config.yml accounting balance`: -44.6744073709551586

After patch:
1. Deposit 900 GAS
1. Deposit 900 GAS
1. `neofs-cli -c config.yml accounting balance`: 1800.00000000

### DIFF
--- a/pkg/morph/client/util.go
+++ b/pkg/morph/client/util.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"encoding/binary"
+	"math/big"
 
 	sc "github.com/nspcc-dev/neo-go/pkg/smartcontract"
 	"github.com/nspcc-dev/neo-go/pkg/vm/stackitem"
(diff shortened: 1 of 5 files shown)


pkg/morph/client/util.go
```diff
@@ -157,6 +158,21 @@ func IntFromStackItem(param stackitem.Item) (int64, error) {
 	}
 }
 
+// BigIntFromStackItem receives numerical value from the value of a smart contract parameter.
+func BigIntFromStackItem(param stackitem.Item) (*big.Int, error) {
+	switch param.Type() {
+	case stackitem.IntegerT, stackitem.ByteArrayT:
+		i, err := param.TryInteger()
```
cthulhu-rider, line 165: Seems like we can do just
```
func BigIntFromStackItem(param stackitem.Item) (*big.Int, error) {
  return param.TryInteger()
}
```
This is at odds with the approach used in other functions, so it can be done separately in a complex.

cthulhu-rider (approved):

pkg/morph/client/util.go
alexvanin: Seems like a good option! Thank you, fixed.

Issue #125, opened 2020-10-27
Use new "well-known" search related filters in node


alexvanin:
In https://github.com/nspcc-dev/neofs-api/pull/80 and https://github.com/nspcc-dev/neofs-api-go/issues/182 there are new well-known filters `CHILDFREE` and `split.parent`. Adopt in neofs-storage.

PR #128, opened 2020-10-28
Implement primary metabase

cthulhu-rider:
Implement bolt-based metabase that is going to be used in local object storage.

- [x] Put

- [x] Get

- [x] Select

- [x] Delete


alexvanin (approved):
LGTM! 

Don't forget to remove `[#xx]` from commit headers. With new neofs-api-go update tests are failed because of #125, #129. Will fix it soon in separate PR.

PR #132, opened 2020-10-29
Netmap cleaner processor for inner ring

alexvanin:


cthulhu-rider (changes requested):
Logic seems correct.
I suggest to replace proto compiled `github.com/nspcc-dev/neofs-api-go/v2/netmap/grpc.NodeInfo` usage with [types](https://github.com/nspcc-dev/neofs-api-go/blob/352e99d9b91a90f66a18d330a23454719b11bdcb/pkg/netmap/node_info.go#L10) from SDK.

cthulhu-rider (approved):

PR #139, opened 2020-11-03
Emit GAS to storage nodes

alexvanin:


pkg/innerring/processors/alphabet/process_emit.go
```diff
@@ -9,15 +13,61 @@ func (np *Processor) processEmit() {
...
+	}
+
+	ln := len(networkMap.Nodes)
+	if ln == 0 {
+		np.log.Debug("empty network map, do not emit gas")
+
+		return
+	}
+
+	gasPerNode := util.Fixed8(np.storageEmission / uint64(ln))
```
cthulhu-rider, line 53: If `np.storageEmission < ln` then `gasPerNode = 0`, is it correct behavior?

alexvanin: That is a good point. The short answer: I don't know. 

In practice this should be very unlikely, because `np.storageEmission` is a `Fixed8` value. In my internal tests we can spend about 1~2 GAS in an epoch for storage nodes. So `gasPerNode` will be less than zero if there are >1000000 nodes in the network map. 

In this case spending 0 GAS is an predictable behavior. If there are more nodes in the network than storage emission value, then economic model should adapt: either increase storage emission or tranfer GAS to limited set of nodes. 

Both of these options have pros and cons and needs to be discussed. I think I'll create issue on that.

alexvanin: https://github.com/nspcc-dev/neofs-node/issues/143

cthulhu-rider (approved):

PR #141, opened 2020-11-03
Fix double emission on asset mint

alexvanin:
LRU cache is used as storage of previous sidechain GAS emissions at asset mint. It will purge old records and those accounts will be able to emit some sidechain GAS once again after `threshold` epochs.

### DIFF
--- a/pkg/innerring/processors/neofs/process_assets.go
+++ b/pkg/innerring/processors/neofs/process_assets.go
@@ -34,12 +34,34 @@ func (np *Processor) processDeposit(deposit *neofsEvent.Deposit) {
 		np.log.Error("can't transfer assets to balance contract", zap.Error(err))
 	}
 
-	// fixme: send gas when ^ tx accepted
-	// fixme: do not send gas to the same user twice per epoch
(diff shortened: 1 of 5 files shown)


pkg/innerring/processors/neofs/process_assets.go
```diff
@@ -34,12 +34,34 @@ func (np *Processor) processDeposit(deposit *neofsEvent.Deposit) {
...
+	receiver := deposit.To()
+
+	// check if receiver already received some mint GAS emissions
+	// we should lock there even though LRU cache is already thread save
+	// we lock there because GAS transfer AND cache update must be atomic
+	np.mintEmitLock.Lock()
+	defer np.mintEmitLock.Unlock()
+
+	val, ok := np.mintEmitCache.Get(receiver.String())
+	if ok && val.(uint64) >= curEpoch-np.mintEmitThreshold {
```
cthulhu-rider, line 47: Can we use  `val.(uint64)+np.mintEmitThreshold >= curEpoch` expression since overflow is much less likely in it?

alexvanin: Sure. Fixed.

cthulhu-rider (approved):

PR #142, opened 2020-11-03
Bring in some metabase optimizations

cthulhu-rider:
### Disk space
Storage of the address list as a bucket in the header index is replaced by storage as a leaf with a serialized list of addresses. This significantly reduces the space consumption for indexes (`~350MB` vs `~4.5GB` in same scenario).
### Select type
Inclusive selection (at which objects are included as index traversal) outperformed the exclusive one (in which objects are excluded from the set of all stored objects after index traversal) in most tests performed.

### DIFF
--- a/pkg/local_object_storage/metabase/put.go
+++ b/pkg/local_object_storage/metabase/put.go
@@ -1,6 +1,9 @@
 package meta
 
 import (
+	"bytes"
+	"encoding/gob"
+
 	objectSDK "github.com/nspcc-dev/neofs-api-go/pkg/object"
 	v2object "github.com/nspcc-dev/neofs-api-go/v2/object"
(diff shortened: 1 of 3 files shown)


cthulhu-rider:
With the assistance of @alexvanin the following bugs were found (which will be fixed in place):

1. If object

- does not match the filter with index `i`;

-  object satisfies all filters starting from index `j > i`

then the given object will be added to the final selection. But it should not be since selected objects must match all filters (`AND`).

2. If object

- matches all filters except the one with index `i`;

- does not have the header with the key from filter `i`

then the given object will be added to the final selection. But it should not be since missing filtering header is equivalent to filter mismatch.

3. If object

- matches search query;
 
- marked for deletion

then the final selection is empty.

alexvanin (approved):

Issue #81, opened 2020-10-05
Add netmap module to neofs-cli


realloc:
Basic set of  operations to implement:

- Dump netmap for latest Epoch or for Epoch number set in option, if available.
  Should be saved to file or printed to `stdout` in different formats including JSON and protobuf.

- Get latest Epoch number
  
- Request [LocalNodeInfo](https://github.com/nspcc-dev/neofs-api/blob/master/netmap/service.proto#L19) from particular host

- Filter netmap from file or `stdin` using set of filters. List of filters can
  be set in binary, JSON or SQL-like format

alexvanin:
Netmap service is not implemented on storage node yet, so this issue is blocked for a while. There are `accounting`, `object`, `session` and `container` services at neofs storage node at the moment.

alexvanin:
Two out of four proposed operation make sense now.

1. :x: There is no RPC to fetch netmap from node.
2. :heavy_check_mark: CLI will get epoch number from response meta header of any request (e.g. `netmap.LocalNodeInfo` )
3. :heavy_check_mark: Just to mention, that `LocalNodeInfo` will miss pretty Location \ Region \ City attributes that set by inner ring nodes from locodes (they are available only in network map stored in netmap contract).
4. :x: Not sure about the filters, but it can print out container nodes from network map, built with the help of https://github.com/nspcc-dev/neofs-node/issues/94. I suggest to do that in separate issue.

alexvanin:
Fixed in #145

PR #145, opened 2020-11-05
Implement netmap module in CLI

cthulhu-rider:
Relates #81.

- [x] `epoch` command to receive current epoch number

- [x] `nodeinfo` command to receive local node info

### DIFF
--- a/cmd/neofs-cli/modules/netmap.go
+++ b/cmd/neofs-cli/modules/netmap.go
@@ -3,10 +3,10 @@ package cmd
 import (
 	"bytes"
 	"context"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 
-	"github.com/mr-tron/base58"
 	"github.com/nspcc-dev/neofs-api-go/pkg/netmap"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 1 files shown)


cmd/neofs-cli/modules/netmap.go
```diff
@@ -19,13 +28,79 @@ var netmapCmd = &cobra.Command{
...
+
+		fmt.Println(buf)
+
+		return
+	}
 
-	// Cobra supports Persistent Flags which will work for this command
-	// and all subcommands, e.g.:
-	// netmapCmd.PersistentFlags().String("foo", "", "A help for foo")
+	fmt.Println("key:", base58.Encode(i.PublicKey()))
```
alexvanin, line 99: I would like to encode it in hex. It is only 33 byte long and we already using hex encoding for hashes, checksums and verbose output of generated private key. We use base58 mostly for identifiers such as object ID, container ID, owner ID.

alexvanin (approved):

cmd/neofs-cli/modules/netmap.go
alexvanin: Fixed with extra commit.

PR #155, opened 2020-11-05
Update neofs-api-go with refactored pkg/netmap

alexvanin:
Refactored pkg/netmap package provides JSON converters for `NodeInfo` and `PlacementPolicy` structures, that has been used
by client applications. It also updates `Node` and `NodeInfo` structure itself, which is a part of `grpc <-> v2 <-> pkg` conversion chain now.


alexvanin:
Open after https://github.com/nspcc-dev/neofs-api-go/pull/189

cthulhu-rider (approved):

Issue #161, opened 2020-11-09
Tombstone does not include parent ID


alexvanin:
Tombstone body contains object ID that should be removed. This is helpful when object was split into several parts, so tombstone contains ID's of every smaller part. However it does not include removed object ID itself.

Now meta-base stores virtual object IDs of big objects in separate index. If object ID is not presented in tombstone, then it is not marked on meta-base, then it will be present in search responses after delete which is incorrect.

PR #162, opened 2020-11-09
Change handling of deleted objects

### LINKED ISSUE #158: Mark removed objects in meta-storage on RPC request
Now all delete operations happen with some sort of delay. `object.Delete` invocation generates tombstone and node broadcast it to all container nodes. Some of container nodes store tombstone but every container nodes put list of removed objects from tombstone into a GC queue. After a some time GC takes object IDs from the queue, marks them  as removed in meta-storage and removes them from object storage.

With new meta-storage it can be implemented without GC timeout. By receiving tombstone node can mark all removed object IDs in meta-storage index. Then GC will take batch of object IDs asynchronously from index of removed objects, remove it from object storage and rebuild index.

cthulhu-rider:
Closes #158.

### DIFF
--- a/cmd/neofs-node/config.go
+++ b/cmd/neofs-node/config.go
@@ -368,7 +368,9 @@ func initLocalStorage(c *cfg) {
 	)
 	fatalOnErr(err)
 
-	c.cfgObject.metastorage = meta.NewDB(boltDB)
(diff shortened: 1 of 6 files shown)


alexvanin (approved):
Seems good. I'll test that together with #161 fix.

PR #166, opened 2020-11-09
Use async pools in object service handlers

alexvanin:


cmd/neofs-node/config.go
```diff
@@ -93,6 +94,13 @@ const (
 
 	cfgReBootstrapEnabled  = "bootstrap.periodic.enabled"
 	cfgReBootstrapInterval = "bootstrap.periodic.interval"
+
+	cfgObjectPutPoolSize       = "pool.object.put"
```
cthulhu-rider, line 98: I suggest to underline `size` in config path.

cthulhu-rider (approved):

cmd/neofs-node/config.go
alexvanin: > I suggest to underline `size` in config path.

How about `pool.size.object.put`?

cthulhu-rider: Seems good. Alternative: `pool.object.put.size`.

alexvanin: Fixed.

PR #180, opened 2020-11-18
Basic ACL check for inner ring requests

### LINKED ISSUE #178: ACL - Inner Ring PUT/GET operation to the private container
In the previous NeoFS version attempt to PUT/GET an object into a private container from the Inner Ring key was impossible.

In the current version of the NeoFS ACL, it is possible to PUT not only for the container nodes (SYS group) but also for the inner ring nodes.

Do we think this behavior is correct for the current version of NeoFS?

For example for container with basic ACL 0x1C8C8CCC:

```
$ neofs-cli --rpc-endpoint s01.neofs.devenv:8080 --key <INNER RING KEY> object put --file 31bec1f6-9d84-4d92-b9ca-1ed97b91319a --cid 6fWrinptxkQWfNr7qFExmHViz4Wb5ctVMdZFBQ1QhAVy 
[31bec1f6-9d84-4d92-b9ca-1ed97b91319a] Object successfully stored
  ID: 4mo44oJ4KeG6w3YM7HA9XmrLXBoEaT8mN8wAqgabCbi6
  CID: 6fWrinptxkQWfNr7qFExmHViz4Wb5ctVMdZFBQ1QhAVy

```

alexvanin:
Closes #178

### DIFF
--- a/pkg/services/object/acl/basic_helper.go
+++ b/pkg/services/object/acl/basic_helper.go
@@ -121,6 +121,21 @@ func (a basicACLHelper) SystemAllowed(op eacl.Operation) bool {
 	return false
 }
 
+// InnerRingAllowed returns true if the operation is allowed by ACL for
+// InnerRing nodes, as part of System group.
+func (a basicACLHelper) InnerRingAllowed(op eacl.Operation) bool {
+	switch op {
+	case eacl.OperationSearch, eacl.OperationRangeHash, eacl.OperationHead:
+		return true
+	default:
+		if n, ok := order[op]; ok {
+			return isLeftBitSet(a, opOffset+n*bitsPerOp+bitSystem)
+		}
+
+		return false
+	}
+}
+
 // AllowSystem allows System group the n-th operation in ACL.
 func (a *basicACLHelper) AllowSystem(op eacl.Operation) {
 	if n, ok := order[op]; ok {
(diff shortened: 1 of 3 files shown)


pkg/services/object/acl/basic_helper.go
```diff
@@ -121,6 +121,21 @@ func (a basicACLHelper) SystemAllowed(op eacl.Operation) bool {
 	return false
 }
 
+// InnerRingAllowed returns true if inner ring node as a subset of system group
```
realloc, line 124: InnerRingAllowed returns true if the operation is allowed by ACL for InnerRing nodes, as part of System group.

realloc (approved):

pkg/services/object/acl/basic_helper.go
```diff
@@ -121,6 +121,21 @@ func (a basicACLHelper) SystemAllowed(op eacl.Operation) bool {
...
+// allowed the n-th operation is set in ACL.
+func (a basicACLHelper) InnerRingAllowed(op eacl.Operation) bool {
+	switch op {
+	case eacl.OperationSearch, eacl.OperationRangeHash, eacl.OperationHead:
+		return true
+	default:
+		if n, ok := order[op]; ok {
```
cthulhu-rider, line 131: Can we use `SystemAllowed` here?

alexvanin: `SystemAllowed` allows `Put` and `Get` request for private containers, unfortunately.

cthulhu-rider (approved):

pkg/services/object/acl/basic_helper.go
alexvanin: Fixed.

PR #188, opened 2020-11-19
Define shard modes

cthulhu-rider:
- define the enumeration of `Shard` modes;

- implement methods to set shard mode on `Shard` and `StorageEngine`.

Working with particular mode values will be implemented later.

### DIFF
--- a/pkg/local_object_storage/shard/mode.go
+++ b/pkg/local_object_storage/shard/mode.go
@@ -1,0 +1,42 @@
+package shard
+
+// Mode represents enumeration of Shard work modes.
+type Mode uint32
+
+// TODO: more detailed description of shard modes.
+
+const (
+	_ Mode = iota
+
+	// ModeActive is a Mode value for active shard.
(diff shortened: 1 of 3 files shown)


pkg/local_object_storage/shard/mode.go
```diff
@@ -0,0 +1,54 @@
...
+	// ModeInactive is a Mode value for inactive shard.
+	ModeInactive
+
+	// ModeReadOnly is a Mode value for read-only shard.
+	ModeReadOnly
+
+	// ModeFault is a Mode value for faulty shard.
+	ModeFault
+
+	// ModeEvacuate is a MOde value for evacuating shard.
```
realloc, line 23: typo. Mode

realloc (approved):

pkg/local_object_storage/shard/mode.go
cthulhu-rider: thx fixed

PR #189, opened 2020-11-19
localstorage: Implement stack of DumpInfo methods

cthulhu-rider:


pkg/local_object_storage/shard/info.go
```diff
@@ -0,0 +1,23 @@
+package shard
+
+import (
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor"
+	meta "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/metabase"
+)
+
+// Info groups the information about Shard.
+type Info struct {
```
realloc, line 9: Weight should also be here

cthulhu-rider: Added

alexvanin (approved):

PR #198, opened 2020-11-23
Configure dial timeouts in Object services

### LINKED ISSUE #193: Configure timeouts for each operation in object service
In neofs-node prototype we had timeout configuration values for all object service requests. If remote node is unresponsive or has poor connection, then timeouts are really helpful. We should add them back.

cthulhu-rider:
Closes #193.

### DIFF
--- a/pkg/network/cache/client.go
+++ b/pkg/network/cache/client.go
@@ -27,10 +27,8 @@ func NewSDKClientCache() *ClientCache {
 	}
 }
 
-// Get function returns existing client or creates a new one. Consider passing
-// connection options to specify details for client, but don't forget that two
-// different set of options should provide two different clients.
-func (c *ClientCache) Get(key *ecdsa.PrivateKey, address string) (*client.Client, error) {
+// Get function returns existing client or creates a new one.
+func (c *ClientCache) Get(key *ecdsa.PrivateKey, address string, opts ...client.Option) (*client.Client, error) {
 	id := uniqueID(key, address)
 
 	c.mu.RLock()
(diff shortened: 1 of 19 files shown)


pkg/network/cache/client.go
```diff
@@ -30,7 +30,7 @@ func NewSDKClientCache() *ClientCache {
 // Get function returns existing client or creates a new one. Consider passing
 // connection options to specify details for client, but don't forget that two
 // different set of options should provide two different clients.
```
alexvanin, line 32: Lets remove `Consider passing connection options...` comment now, since `Get` function now supports options.

pkg/services/object/search/remote.go
```diff
@@ -19,6 +19,8 @@ type remoteStream struct {
 	addr *network.Address
 
 	clientCache *cache.ClientCache
+
+	clientOpts []client.Option
```
alexvanin, line 23: This field is not set in the code. See `pkg/services/object/search/streamer.go:163`

pkg/services/object/put/remote.go
```diff
@@ -29,13 +29,18 @@ type remoteTarget struct {
 	obj *object.Object
 
 	clientCache *cache.ClientCache
+
+	clientOpts []client.Option
```
alexvanin, line 33: This field is also should be set at `pkg/services/object/put/streamer.go:146`

alexvanin (changes requested):

pkg/network/cache/client.go
cthulhu-rider: Fixed, thx. For different set of options we will use different `ClientCache` instances.

pkg/services/object/search/remote.go
cthulhu-rider: Fixed, thx.

pkg/services/object/put/remote.go
cthulhu-rider: Fixed, thx.

alexvanin (approved):

PR #199, opened 2020-11-23
Refactored metabase

alexvanin:
- [x] Inhume
- [x] Put
- [x] Get
- [x] Exists
- [x] Select
- [x] ToMoveIt (plus Movable, plus DoNotMove)
- [x] IsSmall

Edit1: Couple more upcoming changes and it should be ready to merge:
- [x] Refactor exist check to run it from `meta.Select` method.
- [x] Make one `db.Update` in `meta.Put` instead of two (less wasted iops)
- [x] Fix getting headers of virtual objects
- [x] Support obejct ID search in `meta.Select`

There won't be `Delete` implementation since it was not implemented in original metabase and It takes a bit of time to implement index rebuilding routine accurately. 

There also won't be `splitID` support. It will be supported after #208 will be merged to master and then rebased in localstorage branch.

### DIFF
--- a/pkg/local_object_storage/metabase/v2/put.go
+++ b/pkg/local_object_storage/metabase/v2/put.go
@@ -18,7 +18,10 @@ type (
 	}
 )
 
-var ErrUnknownObjectType = errors.New("unknown object type")
+var (
+	ErrUnknownObjectType          = errors.New("unknown object type")
+	ErrIncorrectBlobovniczaUpdate = errors.New("updating blobovnicza id on object without it")
+)
 
 // Put saves object header in metabase. Object payload expected to be cut.
 // Big objects have nil blobovniczaID.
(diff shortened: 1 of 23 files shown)


pkg/local_object_storage/metabase/v2/put.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	// index blobovniczaID if it is present
+	if blobovniczaID != nil {
+		result = append(result, namedBucketItem{
+			name: smallBucketName(addr.ContainerID()),
+			key:  objectKey(addr.ObjectID()),
+			val:  []byte(blobovniczaID.String()),
+		})
+	}
```
alexvanin, line 94: It seems like `blobovniczaID` makes sense only for non parent objects. Probably this code should be moved up in `!isParent` branch.

alexvanin:
Added `Delete` and `CleanUp` methods.

alexvanin:
Added `SplitID` index support.

cthulhu-rider (approved):

PR #201, opened 2020-11-24
Fix missing error when getting a nonexistent eACL table

cthulhu-rider:
- return error from `EACL` method of morph client wrapper if eACL is not set for container;

- do not check eACL version in container CLI.


alexvanin (approved):
LGTM

Issue #215, opened 2020-11-26
Bearer token for HEAD operation with Filter is not applied


anatoly-bogatyrev:
Bearer token for HEAD operation with Filter is not applied.
1) Public container
2) Created bearer token:

```
{
  "body": {
    "eaclTable": {
      "containerID": {
        "value": "EZqdi6vZXETAvKPxMLyjnb4muBLDCvEzABbmDPekhAk4"
      },
      "records": [
        {
          "operation": "GET",
          "action": "ALLOW",
          "filters": [
            {
              "headerType": "OBJECT",
              "matchType": "STRING_EQUAL",
              "key": "$Object:objectID",
              "value": "AW8PfM9EBmnap9HvLDvfUzkphZt3v2ctjXd5y595keng"
            }
          ],
          "targets": [
            {
              "role": "USER"
            }
          ]
        },
        {
          "operation": "PUT",
          "action": "ALLOW",
          "filters": [
            {
              "headerType": "OBJECT",
              "matchType": "STRING_EQUAL",
              "key": "$Object:objectID",
              "value": "AW8PfM9EBmnap9HvLDvfUzkphZt3v2ctjXd5y595keng"
            }
          ],
          "targets": [
            {
              "role": "USER"
            }
          ]
        },
        {
          "operation": "HEAD",
          "action": "ALLOW",
          "filters": [
            {
              "headerType": "OBJECT",
              "matchType": "STRING_EQUAL",
              "key": "$Object:objectID",
              "value": "AW8PfM9EBmnap9HvLDvfUzkphZt3v2ctjXd5y595keng"
            }
          ],
          "targets": [
            {
              "role": "USER"
            }
          ]
        },
        {
          "operation": "DELETE",
          "action": "ALLOW",
          "filters": [
            {
              "headerType": "OBJECT",
              "matchType": "STRING_EQUAL",
              "key": "$Object:objectID",
              "value": "AW8PfM9EBmnap9HvLDvfUzkphZt3v2ctjXd5y595keng"
            }
          ],
          "targets": [
            {
              "role": "USER"
            }
          ]
        },
        {
          "operation": "SEARCH",
          "action": "ALLOW",
          "targets": [
            {
              "role": "USER"
            }
          ]
        },
        {
          "operation": "GETRANGE",
          "action": "ALLOW",
          "filters": [
            {
              "headerType": "OBJECT",
              "matchType": "STRING_EQUAL",
              "key": "$Object:objectID",
[...]

PR #223, opened 2020-12-01
[#215] placement: Fix local and single-success placement traversal

cthulhu-rider:
In previous implementation placement traverser processed incorrectly with
local placement build. Also entity incorrectly traversed the placement
vectors for fixed number read operations until success. The erroneous
behavior was due to the use of a vector number of successes instead of
a scalar number in these scenarios.


anatoly-bogatyrev (approved):
Testcase has been passed as expected, without "incomplete object put" error.
Thanks.

alexvanin (approved):

PR #230, opened 2020-12-03
Support SplitInfo in storage engine

alexvanin:


pkg/local_object_storage/metabase/v2/put.go
```diff
@@ -305,3 +335,87 @@ func updateBlobovniczaID(tx *bbolt.Tx, addr *objectSDK.Address, id *blobovnicza.
...
+	default:
+		return nil, ErrIncorrectRootObject // should never happen
+	}
+
+	return info, nil
+}
+
+// mergeSplitInfo ignores conflicts and rewrites `to` with non empty values
+// from `from`.
+func mergeSplitInfo(from, to *objectSDK.SplitInfo) *objectSDK.SplitInfo {
```
cthulhu-rider, line 397: This function can return no result since `from` is a pointer. I am also developing a code that implements exactly the same function , so I propose to share it in a common package (not in this PR).

cthulhu-rider (approved):

PR #231, opened 2020-12-03
Fetching objects from storage engine for policer

alexvanin:
Propose to add:

- `Containers` method to metabase which returns available `[]container.ID` 
- `List` method to shard which makes empty select with each `container.ID` from `metabase.Containers()` and returns `[]object.ID`
- `List` method to storage engine which goes through shards and runs `shard.List()` method and then returns `[]object.ID`. It also has `limit` parameter.

### DIFF
--- a/pkg/local_object_storage/engine/select.go
+++ b/pkg/local_object_storage/engine/select.go
@@ -65,6 +65,44 @@ func (e *StorageEngine) Select(prm *SelectPrm) (*SelectRes, error) {
 	}, nil
 }
 
+// List returns `limit` available physically storage object addresses in engine.
+// If limit is zero, then returns all available object addresses.
+func (e *StorageEngine) List(limit uint64) (*SelectRes, error) {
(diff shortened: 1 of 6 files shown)


pkg/local_object_storage/engine/select.go
```diff
@@ -65,6 +65,44 @@ func (e *StorageEngine) Select(prm *SelectPrm) (*SelectRes, error) {
 	}, nil
 }
 
+// List returns `limit` available physically storage object addresses in engine.
+// If limit is zero, then returns all available object addresses.
+func (e *StorageEngine) List(limit uint64) (*SelectRes, error) {
+	addrList := make([]*object.Address, 0)
```
cthulhu-rider, line 71: Correct, but we also can allocate some approximate capacity (`limit`).

alexvanin: That's right, thank you. Fixed.

cthulhu-rider (approved):

PR #250, opened 2020-12-14
service/policer: Don't shrink node list at unknown error

alexvanin:
Every unknown error must not decrease shortage counter and must not exclude faulty node from the node list, because this list will be used later for replication.


realloc (approved):

anatoly-bogatyrev (approved):
The issue with the replication has been fixed. Testcase passed.

PR #265, opened 2020-12-18
Inner ring audit processor

alexvanin:


pkg/innerring/processors/audit/process.go
```diff
@@ -0,0 +1,137 @@
...
+
+		sg = append(sg, result...)
+
+		break // we found storage groups, so break loop
+	}
+
+	return sg
+}
+
+func storageGroupSearchFilter() object.SearchFilters {
```
cthulhu-rider, line 122: Can we initialize immediately `var sgFilter = storagegroup.SearchQuery()`?

cthulhu-rider (approved):

pkg/innerring/processors/audit/process.go
alexvanin: Sure. Fixed.

PR #301, opened 2021-01-11
Process salt from GetRangeHashRequest

### LINKED ISSUE #290: Salt from GetRangeHashRequest is not applied
Function SaltXOR is unused, so it seems that node does not salt range data in GetRangeHash.

## Your Environment
* Version used: **0.14.0-rc.1**

cthulhu-rider:
Closes #290.

### DIFF
--- a/cmd/neofs-cli/modules/object.go
+++ b/cmd/neofs-cli/modules/object.go
@@ -432,7 +432,9 @@ func getObjectHash(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
-	salt, err := hex.DecodeString(cmd.Flag(getRangeHashSaltFlag).Value.String())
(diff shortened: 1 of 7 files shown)


cmd/neofs-cli/modules/object.go
```diff
@@ -427,6 +432,11 @@ func getObjectHash(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
+	salt, err := hex.DecodeString(cmd.Flag(getRangeHashSaltFlag).Value.String())
```
alexvanin, line 435: It would be nice to trim `0x` prefix from salt value just in case.

alexvanin (approved):

cmd/neofs-cli/modules/object.go
cthulhu-rider: Done

PR #304, opened 2021-01-12
Use list of NEO endpoitns in neofs-node config

### LINKED ISSUE #157: Specify pool of NEO endpoints in config
Both inner ring and storage node configurations have NEO endpoint fields. It would be nice to support pool of addresses.

```yaml
# Pure YAML:
morph:
  endpoint: 
    - https://morph1.nspcc.ru
    - https://morph2.nspcc.ru
    - https://morph3.nspcc.ru

# With ENV support:
morph:
  endpoint_0: https://morph1.nspcc.ru
  endpoint_1: https://morph2.nspcc.ru
  endpoint_2: https://morph3.nspcc.ru

# With priorities:
morph:
  endpoint_0: 
    url: https://morph1.nspcc.ru
    priority: 1.0
  endpoint_1: 
    url: https://morph2.nspcc.ru
    priority: 0.5
  endpoint_2: 
    url: https://morph3.nspcc.ru
    priority: 0.5
```

Neofs-node can choose random address and use it in runtime if it is active. This way we can uniformly distribute load of RPC nodes that will be present in preinstalled configuration.

### What's wrong with runtime connection pool?
Instead of choosing one active address before startup, neofs-node can keep pool of active connections, healthcheck it in runtime and rotate connections whenever it needs. This increases stability of RPC node failures but there is a tricky part.

While connection pool works fine with clients, neofs-node also uses event listener to fetch new notifications from blockchain. Therefore neofs-node should multiplex several listeners to provide single stable channel of events. This is not trivial task and requires connections to several RPC nodes simultaneously. The other option: rotate listeners the same way as client connection, but we can have a bit of downtime during rotation, which is bad.

So the convenient way to address this problem: choose one endpoint and work with it. As soon as there are problems listener will close connection and client will return an error. In this case we can restart / reinitialize application with new single RPC endpoint.

alexvanin:
Closes #157 

Configuration will look like this:

```yaml
morph:
  endpoint:
    - http://seed1.neo.org:20333
    - http://seed2.neo.org:20333
    - http://seed3.neo.org:20333
  notification:
    endpoint:
      - ws://seed1.neo.org:20333/ws
      - ws://seed-go.nspcc.ru:30333/ws
```

Storage node at startup randomly picks notification and client RPC endpoints. If endpoint is not available, then application picks another one. If all endpoints are down then application won't start.

If endpoint fails at runtime, then storage node shuts down. Later we can reinitialize app without downtime. 

Edit 1: new config looks like this now

```yaml
morph:
  rpc_endpoint:
    - http://seed1.neo.org:20333
    - http://seed2.neo.org:20333
    - http://seed3.neo.org:20333
  notification_endpoint:
    - ws://seed1.neo.org:20333/ws
    - ws://seed-go.nspcc.ru:30333/ws
  dial_timeout: 10s
```

### DIFF
--- a/cmd/neofs-node/morph.go
+++ b/cmd/neofs-node/morph.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/nspcc-dev/neo-go/pkg/util"
@@ -11,14 +12,44 @@ import (
 	"github.com/nspcc-dev/neofs-node/pkg/morph/event"
 	netmapEvent "github.com/nspcc-dev/neofs-node/pkg/morph/event/netmap"
 	"github.com/nspcc-dev/neofs-node/pkg/morph/subscriber"
+	"github.com/nspcc-dev/neofs-node/pkg/util/rand"
+	"go.uber.org/zap"
 )
 
 const newEpochNotification = "NewEpoch"
 
+var (
+	errNoRPCEndpoints = errors.New("NEO RPC endpoints are not specified in config")
+	errNoWSEndpoints  = errors.New("websocket NEO listener endpoints are not specified in config")
+)
+
 func initMorphComponents(c *cfg) {
 	var err error
 
-	c.cfgMorph.client, err = client.New(c.key, c.viper.GetString(cfgMorphRPCAddress))
+	addresses := c.viper.GetStringSlice(cfgMorphRPCAddress)
+	if len(addresses) == 0 {
+		fatalOnErr(errNoRPCEndpoints)
+	}
+
+	crand := rand.New() // math/rand with cryptographic source
+	crand.Shuffle(len(addresses), func(i, j int) {
+		addresses[i], addresses[j] = addresses[j], addresses[i]
+	})
+
+	for i := range addresses {
+		c.cfgMorph.client, err = client.New(c.key, addresses[i])
+		if err == nil {
+			c.log.Info("neo RPC connection established",
+				zap.String("endpoint", addresses[i]))
+
+			break
+		}
+
+		c.log.Info("failed to establish neo RPC connection, trying another",
+			zap.String("endpoint", addresses[i]),
+			zap.String("error", err.Error()))
+	}
+
 	fatalOnErr(err)
 
 	staticClient, err := client.NewStatic(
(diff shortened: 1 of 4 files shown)


realloc:
Is it possible to simplify config structure to make it not so deep? Something like this:
```
morph:
  rpc_endpoint:
    - http://seed1.neo.org:20333
    - http://seed2.neo.org:20333
    - http://seed3.neo.org:20333
  notification_endpoint:
    - ws://seed1.neo.org:20333/ws
    - ws://seed-go.nspcc.ru:30333/ws
```

realloc (approved):

cmd/neofs-node/morph.go
```diff
@@ -38,11 +57,33 @@ func initMorphComponents(c *cfg) {
...
+		endpoints[i], endpoints[j] = endpoints[j], endpoints[i]
 	})
+
+	for i := range endpoints {
+		subs, err = subscriber.New(c.ctx, &subscriber.Params{
+			Log:         c.log,
+			Endpoint:    endpoints[i],
+			DialTimeout: timeout,
+		})
+		if err == nil {
```
cthulhu-rider, line 79: Let's write log (may be debug) about what `err != nil`? Same in `initMorphComponents` func.

cmd/neofs-node/morph.go
```diff
@@ -38,11 +57,33 @@ func initMorphComponents(c *cfg) {
...
+
+	endpoints := c.viper.GetStringSlice(cfgMorphNotifyRPCAddress)
+	timeout := c.viper.GetDuration(cfgMorphNotifyDialTimeout)
+
+	crand := rand.New() // math/rand with cryptographic source
+	crand.Shuffle(len(endpoints), func(i, j int) {
+		endpoints[i], endpoints[j] = endpoints[j], endpoints[i]
 	})
+
+	for i := range endpoints {
```
cthulhu-rider, line 73: I think we need to add `len(endpoints) == 0` check. Same in `initMorphComponents` func.

cthulhu-rider (changes requested):

cmd/neofs-node/morph.go
alexvanin: At first I thought that last `fatalOnErr` will print error message anyway. However it would be nice to see which endpoints from the pool are down, so I put `INFO` log message in both cycles.

cmd/neofs-node/morph.go
alexvanin: Fixed.

cthulhu-rider (approved):

PR #317, opened 2021-01-14
morph/client: Return complete eACL signature from contract

alexvanin:
Related to https://github.com/nspcc-dev/neofs-contract/pull/35

### DIFF
--- a/cmd/neofs-cli/modules/container.go
+++ b/cmd/neofs-cli/modules/container.go
@@ -425,10 +425,10 @@ Container ID in EACL table will be substituted with ID from the CLI.`,
 			for i := 0; i < awaitTimeout; i++ {
 				time.Sleep(1 * time.Second)
 
-				eaclSig, err := cli.GetEACLWithSignature(ctx, id, globalCallOptions()...)
+				table, err := cli.GetEACL(ctx, id, globalCallOptions()...)
 				if err == nil {
 					// compare binary values because EACL could have been set already
-					got, err := eaclSig.EACL().Marshal()
+					got, err := table.Marshal()
 					if err != nil {
 						continue
 					}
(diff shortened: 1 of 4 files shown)


cmd/neofs-cli/modules/container.go
```diff
@@ -425,10 +425,10 @@ Container ID in EACL table will be substituted with ID from the CLI.`,
 			for i := 0; i < awaitTimeout; i++ {
 				time.Sleep(1 * time.Second)
 
-				eaclSig, err := cli.GetEACLWithSignature(ctx, id, globalCallOptions()...)
+				table, err := cli.GetEACL(ctx, id, globalCallOptions()...)
```
cthulhu-rider, line 428: I suggest to leave usage of `GetEACLWithSignature` method, but with signature check.

alexvanin: `GetEACL` method encapsulates signature check inside. I don't think explicit signature check is an important thing there.
We have `GetEACLWithSignature` in `neofs-cli get-eacl` command if you want to provide use case example for this function.

cthulhu-rider: If we use GetEACLWithSignature, we still can output the signature event if it is wrong. But this is not necessary.

cthulhu-rider (approved):

cmd/neofs-cli/modules/container.go
alexvanin: > If we use GetEACLWithSignature, we still can output the signature event if it is wrong. But this is not necessary.

That is exactly how it's done in `get-eacl` command. As for `--await` flag in `set-eacl` I don't think it is necessary there.

PR #333, opened 2021-01-20
Measure epochs in chain blocks

### LINKED ISSUE #324: Get epoch duration from NeoFS contract and calculate it in blocks
Right now NeoFS epoch duration is set in config file as astronomical time. However every NeoFS event should be tied with sidechain or mainchain blocks. 

Epoch duration should be stored in sidechain blocks, e.g. `240` is about 3600 seconds ~ 1 hour.
Value should be fetched from netmap contract, because this contracts stores network configuration parameters.
If inner ring node can't fetch this value, then use value from config (or default which should be `240`). 

/cc @realloc

cthulhu-rider:
Closes #324.

- [x] add subscription to new block events
- [x] subscribe Netmap IR processor to new blocks
- [x] change netmap processor to measure epoch in chain blocks
- [x] emit GAS in block periods


pkg/morph/event/listener.go
```diff
@@ -313,6 +345,15 @@ func (s listener) Stop() {
 	s.subscriber.Close()
 }
 
+func (s listener) RegisterBlockHandler(handler BlockHandler) {
```
alexvanin, line 348: Seems like missing block height there. And also `blockHandlers map[uint32]BlockHandler`.

cthulhu-rider: `RegisterBlockHandler` is a general purpose method. I will add the ability to subscribe to particular block height later. There will be something like this functionality:
```
// EqualHeightBlockHandler returns BlockHandler that calls bHandler only for blocks with
// Index == height.
func EqualHeightBlockHandler(height uint32, bHandler BlockHandler) BlockHandler {
	return func(b *block.Block) {
		if b.Index != height {
			return
		}

		bHandler(b)
	}
}
```

pkg/innerring/timers/block.go
```diff
@@ -0,0 +1,171 @@
...
+
+	delta := mul * t.baseDur / div
+	if delta == 0 {
+		delta = 1
+	}
+
+	t.tgt = delta
+
+	for i := range t.ps {
+		t.ps[i].reset()
```
alexvanin, line 141: If BlockTimer has delta handler (so `len(t.ps) == 1`), won't it be reset twice?
First time at this place (`Reset() -> t.resetWithBaseInterval(d) -> t.reset() -> t.ps[i].reset()`).
Second time at `Reset() -> t.ps[i].resetWithBaseInterval(d) -> t.reset()`

Is it correct?

pkg/innerring/innerring.go
```diff
@@ -121,6 +121,18 @@ func (s *Server) Start(ctx context.Context, intError chan<- error) error {
 	go s.morphListener.ListenWithError(ctx, morphErr)      // listen for neo:morph events
 	go s.mainnetListener.ListenWithError(ctx, mainnnetErr) // listen for neo:mainnet events
 
+	s.morphListener.RegisterBlockHandler(func(b *block.Block) {
+		s.log.Info("new block",
```
alexvanin, line 125: Let's use debug level for new block messages.

alexvanin (approved):
I will merge this PR after v0.14.3 release.

Looks great. I'll ask you to try update neo-go version in this PR and check if https://github.com/nspcc-dev/neo-go/pull/1687 patch works in neofs-dev-env. However there might be still issues unrelated to block parsing. Not sure if neo-go master branch is compatible with v0.92.0 chain dumps in neofs-dev-env.

pkg/morph/subscriber/subscriber.go
```diff
@@ -99,6 +103,20 @@ func (s *subscriber) Close() {
 	s.client.Close()
 }
 
+func (s *subscriber) BlockNotifications() (<-chan *block.Block, error) {
+	if err := s.client.Init(); err != nil {
```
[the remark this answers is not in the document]
cthulhu-rider: Yes, it is:
```
package client // import "github.com/nspcc-dev/neo-go/pkg/rpc/client"

func NewWS(ctx context.Context, endpoint string, opts Options) (*WSClient, error)
    NewWS returns a new WSClient ready to use (with established websocket
    connection). You need to use websocket URL for it like `ws://1.2.3.4/ws`.
    You should call Init method to initialize network magic the client is
    operating on.
```
I would be better to call `Init` inside `Subscriber` constructor, @alexvanin?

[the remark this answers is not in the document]
alexvanin: Yes, let's move it to constructor then.

pkg/morph/subscriber/subscriber.go
```diff
@@ -99,6 +103,20 @@ func (s *subscriber) Close() {
...
+		return nil, errors.Wrap(err, "could not init ws client")
+	}
+
+	if _, err := s.client.SubscribeForNewBlocks(nil); err != nil {
+		return nil, errors.Wrap(err, "could not subscribe for new block events")
+	}
+
+	s.blockChan = make(chan *block.Block)
```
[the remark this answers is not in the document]
cthulhu-rider: Agree, it will prevent possible writing to nil block channel. Fixed.

pkg/morph/event/listener.go
```diff
@@ -144,6 +153,13 @@ func (s listener) listen(ctx context.Context, intError chan<- error) error {
 }
 
 func (s listener) listenLoop(ctx context.Context, chEvent <-chan *state.NotificationEvent, intErr chan<- error) {
+	blockChan, err := s.subscriber.BlockNotifications()
+	if err != nil {
+		intErr <- errors.Wrap(err, "could not open block notifications channel")
```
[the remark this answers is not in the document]
cthulhu-rider: Fixed

pkg/innerring/timers/block.go
cthulhu-rider: `Reset` call on sub-timer won't ever happen since package does not provide the access to sub-timer's interface. So you can call `Reset` twice only on basic timer.

pkg/innerring/innerring.go
cthulhu-rider: Done.

pkg/morph/subscriber/subscriber.go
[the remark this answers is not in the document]
cthulhu-rider: Done.

PR #336, opened 2021-01-22
Use Alpine for cli instead of scratch

im-kulikov:
This is necessary in order to run the container in "limbo" (as a service) for general use, for example as service for neofs-dev-env with settings and environment variables.

Signed-off-by: Evgeniy Kulikov <kim@nspcc.ru>

### DIFF
--- a/Dockerfile.cli
+++ b/Dockerfile.cli
@@ -11,7 +11,7 @@ COPY . /src
 RUN make bin/neofs-cli
 
 # Executable image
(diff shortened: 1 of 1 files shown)


alexvanin:
Fix commit message: `[#336]` instead of `[336]`.

im-kulikov:
@alexvanin done

alexvanin (approved):

PR #341, opened 2021-01-22
Add shell in docker images

alexvanin:
It is much easier to debug and solve issues in-place with interactive command line.


realloc (approved):

im-kulikov:
`RUN apk add --update bash` incorrect way, should be `RUN apk add --no-cache bash`

because with `--update` you store `apk-cache` inside container

PR #356, opened 2021-01-28
Fix block timer behavior after reset

cthulhu-rider:
In previous implementation handler was not called more the one time after
Reset. It was caused by tick counter not being reset inside Reset method.

### DIFF
--- a/pkg/innerring/timers/block.go
+++ b/pkg/innerring/timers/block.go
@@ -136,10 +136,7 @@ func (t *BlockTimer) reset() {
 	}
 
 	t.tgt = delta
-
-	for i := range t.ps {
-		t.ps[i].reset()
-	}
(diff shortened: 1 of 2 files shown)


pkg/innerring/timers/block.go
```diff
@@ -136,10 +136,8 @@ func (t *BlockTimer) reset() {
 	}
 
 	t.tgt = delta
-
-	for i := range t.ps {
-		t.ps[i].reset()
-	}
+	t.cur = 0
+	t.tgt = delta
```
alexvanin, line 140: duplicate line 138

cthulhu-rider: fixed

alexvanin (approved):

PR #361, opened 2021-02-01
metabase: Support new match types

cthulhu-rider:
- [x] support of `STRING_NOT_EQUAL` in metabase
- [x] support of `NOT_PRESENT` in metabase
- [x] define and implement format in CLI (#362)

### DIFF
--- a/pkg/local_object_storage/metabase/select.go
+++ b/pkg/local_object_storage/metabase/select.go
@@ -3,6 +3,7 @@ package meta
 import (
 	"encoding/binary"
 	"fmt"
+	"strings"
 
 	"github.com/nspcc-dev/neofs-api-go/pkg/container"
 	"github.com/nspcc-dev/neofs-api-go/pkg/object"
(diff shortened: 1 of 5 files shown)


pkg/local_object_storage/metabase/select.go
```diff
@@ -223,10 +212,57 @@ func (db *DB) selectFastFilter(
...
+	for _, fns := range mBucketNaming {
+		for _, fn := range fns {
+			names = append(names, fn(cid))
+		}
+	}
+
+	return
+}
+
+func bucketNamesForType(cid *container.ID, mType object.SearchMatchType, typeVal string) (names [][]byte) {
```
alexvanin, line 241: Maybe `allBucketNames` can be reused in this function if we pass `object.MatchNotPresent` type. Just a thought.

pkg/local_object_storage/metabase/select.go
```diff
@@ -271,6 +307,56 @@ func (db *DB) selectFromFKBT(
...
+	fNum int, // index of filter
+) {
+	exclBktRoot := tx.Bucket(name)
+
+	mExcl := make(map[string]struct{})
+
+	bktExcl := tx.Bucket(name)
+	if bktExcl != nil {
+		_ = bktExcl.ForEach(func(k, _ []byte) error {
+			exclBktLeaf := exclBktRoot.Bucket(k)
```
alexvanin, line 328: Can't you do `exclBktLeaf := bktExcl.Bucket(k)` there?

alexvanin (approved):

pkg/local_object_storage/metabase/select.go
cthulhu-rider: Yes, fixed

pkg/local_object_storage/metabase/select.go
cthulhu-rider: I thought about that, but I decided that `allBucketNames` could have simpler logic, which is necessary when processing `MatchNotPresent`. Call `bucketNamesForType(cid, object.MatchNotPresent, "")` will not give the desired result.

Issue #316, opened 2021-01-14
Use UN/LOCODE as location identity in storage node attributes


alexvanin:
NeoFS storage nodes specify their location in node attributes. These attributes presented in the network map, so one can define placement policies with location attributes from there. However there are issues:
- attributes do not have strict formatting rules for location,
- there is more than one way you can name location, e.g. `St. Petersburg` or `Saint Petersburg`. 

To deal with these issues we can use [UN/LOCODE](https://unece.org/unlocode) as location attribute. LOCODEs are `XX XXX` ids, e.g. `CN SGH` for Shanghai. They stored in a database maintained by UN in this format:
```
CN SGH    Shanghai    Shanghai    SH    12345---    AS    1401    3114N 12129E
```

## Example

1. Storage node won't be included to the network map without LOCODE. So at first the user looks up through LOCODE database or use external service (maybe `neofs-cli` helper command?) to find proper LOCODE value.
2. User sets LOCODE in storage node config (`NEOFS_NODE_LOCODE=RU LED`) and then starts it.
3. Storage node invokes `AddPeer` method with storage node attribute `/LOCODE:RU LED`
4. Inner ring nodes makes sanity check of the storage node and converts this LOCODE to `/Country:Russia/City:Saint Petersburg`
5. Inner ring nodes approve storage node with parsed LOCODE and it gets into network map this way.

## Issues
- LOCODE encodes country, city or big transport hubs within it such as airports, railway station, etc. However it does not define regions like `Europe` or `Asia`.

alexvanin:
> LOCODE encodes country, city or big transport hubs within it such as airports, railway station, etc. However it does not define regions like Europe or Asia.

We can try to compute region based on country, however there can be inconsistency. In some databases Russia is `Asia` region, in others `Europe`. We can try to predict region more precisely based on latitude and longitude from LOCODE database, but this is not trivial.

realloc:
I looks like we could import coordinates of Polygons representing parts of the worlds' boundaries from OSM and check if the coordinates from LOCODE are in those boundaries locally, without calls to OSM or other external service.

PR #370, opened 2021-02-04
Overtake data from UN/Locode tables to local databases

cthulhu-rider:
Related #316.

### DIFF
--- a/pkg/util/locode/db/airports/calls.go
+++ b/pkg/util/locode/db/airports/calls.go
@@ -1,0 +1,186 @@
+package airportsdb
+
+import (
+	"encoding/csv"
+	"io"
+	"os"
+	"strconv"
+
+	"github.com/nspcc-dev/neofs-node/pkg/util/locode"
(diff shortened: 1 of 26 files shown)


pkg/util/locode/db/airports/calls.go
```diff
@@ -0,0 +1,186 @@
...
+	airportIATA
+	_ // ICAO
+	airportLatitude
+	airportLongitude
+	_ // Altitude
+	_ // Timezone
+	_ // DST
+	_ // Tz database time zone
+	_ // Type
+	_ // Source
```
alexvanin, line 30: This field order defined in airport database?

alexvanin (approved):

pkg/util/locode/db/airports/calls.go
cthulhu-rider: Yes, it is described [here](https://openflights.org/data.html)

PR #390, opened 2021-02-15
GC: Implement API

### LINKED ISSUE #377: GC: Implement endpoint to access GC outside of storage engine
Remove data by external invocations via endpoint -- is the simplest case for storage engine garbage collector. Endpoint should provide public method from storage engine to physically remove any existing object. Later this endpoint will be used by `Policer` component and by external monitors via `Control API`.

On delete invocation, endpoint should:
- find shard that stores the object,
- make a GCMARK in graveyard bucket.

To implement this we need:
- [ ] modify `metabase.Ihume()` to support `GCMARK` records

### LINKED ISSUE #378: GC: Run object remover job per shard
Every shard in storage engine should have several monitoring / control routines. One of these jobs: object remover.

Object remover is simple. It iterates over `Graveyard` bucket in shard's metabase and looks for records with `GCMARK` in the value. When worker sees such record, it accumulates key of this record in buffer. When the buffer is full, routine invokes `shard.Delete()` method on these objects.

To implement this we need:
- [ ] provide graveyard iterator in metabase and shard interface,
- [ ] create environment for per-shard routines in storage engine.

cthulhu-rider:
Closes #377.
Closes #378.

### DIFF
--- a/pkg/local_object_storage/metabase/iterators.go
+++ b/pkg/local_object_storage/metabase/iterators.go
@@ -1,0 +1,111 @@
+package meta
+
+import (
+	"strconv"
+
+	"github.com/nspcc-dev/neofs-api-go/pkg/container"
+	"github.com/nspcc-dev/neofs-api-go/pkg/object"
+	objectV2 "github.com/nspcc-dev/neofs-api-go/v2/object"
+	"github.com/pkg/errors"
+	"go.etcd.io/bbolt"
+)
+
+// ExpiredObject is a descriptor of expired object from DB.
+type ExpiredObject struct {
+	typ object.Type
+
+	addr *object.Address
+}
+
+// Type returns type of the expired object.
+func (e *ExpiredObject) Type() object.Type {
+	return e.typ
+}
+
+// Address returns address of the expired object.
+func (e *ExpiredObject) Address() *object.Address {
+	return e.addr
+}
+
+// ExpiredObjectHandler is an ExpiredObject handling function.
+type ExpiredObjectHandler func(*ExpiredObject) error
+
+// ErrInterruptIterator is returned by iteration handlers
+// as a "break" keyword.
+var ErrInterruptIterator = errors.New("iterator is interrupted")
+
+// IterateExpired iterates over all objects in DB which are out of date
+// relative to epoch.
+//
+// If h returns ErrInterruptIterator, nil returns immediately.
+// Returns other errors of h directly.
+func (db *DB) IterateExpired(epoch uint64, h ExpiredObjectHandler) error {
+	return db.boltDB.View(func(tx *bbolt.Tx) error {
+		return db.iterateExpired(tx, epoch, h)
+	})
+}
+
+func (db *DB) iterateExpired(tx *bbolt.Tx, epoch uint64, h ExpiredObjectHandler) error {
+	err := tx.ForEach(func(name []byte, b *bbolt.Bucket) error {
+		cidBytes := cidFromAttributeBucket(name, objectV2.SysAttributeExpEpoch)
+		if cidBytes == nil {
+			return nil
+		}
+
+		return b.ForEach(func(expKey, _ []byte) error {
+			bktExpired := b.Bucket(expKey)
+			if bktExpired == nil {
+				return nil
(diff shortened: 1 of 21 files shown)


pkg/local_object_storage/shard/gc.go
```diff
@@ -0,0 +1,312 @@
...
+		v.prevGroup.Wait()
+
+		var ctx context.Context
+		ctx, v.cancelFunc = context.WithCancel(context.Background())
+
+		v.prevGroup.Add(len(v.handlers))
+
+		for _, h := range v.handlers {
+			err := gc.workerPool.Submit(func() {
+				h(ctx, event)
```
alexvanin, line 128: `Submit` function eventually creates new go routine. Therefore it is not so good to reuse `h` literal inside. For example this code would fire go vet error.
```go
		for _, h := range v.handlers {
			go func() {
				h(ctx, event)
				v.prevGroup.Done()
			}()
		}
```

I propose to define separate variable for handler pointer:
```go
		for i := range v.handlers {
			h := v.handlers[i]
			err := gc.workerPool.Submit(func() {
				h(ctx, event)
				v.prevGroup.Done()
			})
		}
```

pkg/local_object_storage/metabase/util.go
```diff
@@ -97,6 +108,12 @@ func addressKey(addr *object.Address) []byte {
 	return []byte(addr.String())
 }
 
+// parses object address formed by addressKey.
+func addressFromKey(k []byte) (*object.Address, error) {
```
alexvanin, line 112: Would be nice to reuse this function in:
- pkg/local_object_storage/metabase/select.go:131
- pkg/local_object_storage/metabase/select.go:439
- pkg/local_object_storage/metabase/iterators.go:134
- pkg/local_object_storage/metabase/movable.go:134

pkg/local_object_storage/shard/gc.go
```diff
@@ -0,0 +1,312 @@
...
+func (s *Shard) removeGarbage() {
+	buf := make([]*object.Address, 0, s.rmBatchSize)
+
+	// iterate over metabase graveyard and accumulate
+	// objects with GC mark
+	err := s.metaBase.IterateOverGraveyard(func(g *meta.Grave) error {
+		if g.WithGCMark() {
+			buf = append(buf, g.Address())
+		}
+
```
alexvanin, line 164: Don't we need to check the length of `buf` there? If `buf` has `s.rmBarchSize` elements, then stop iterating and process buffer.

pkg/local_object_storage/shard/gc.go
cthulhu-rider: Fixed

pkg/local_object_storage/shard/gc.go
cthulhu-rider: Definitely yes, fixed

pkg/local_object_storage/metabase/iterators.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+
+		return b.ForEach(func(expKey, _ []byte) error {
+			bktExpired := b.Bucket(expKey)
+			if bktExpired == nil {
+				return nil
+			}
+
+			return bktExpired.ForEach(func(idKey, _ []byte) error {
+				expiresAt, err := strconv.ParseUint(string(expKey), 10, 64)
```
alexvanin, line 62: We don't need to calculate `expiresAt` and make an epoch check on each record in `bktEpxired` bucket. We can move it higher:

```go
			bktExpired := b.Bucket(expKey)
			if bktExpired == nil {
				return nil
			}
			expiresAt, err := strconv.ParseUint(string(expKey), 10, 64)
			...
```

pkg/local_object_storage/metabase/util.go
cthulhu-rider: Done

pkg/local_object_storage/metabase/iterators.go
cthulhu-rider: Done

alexvanin (approved):

PR #400, opened 2021-02-19
metabase: Do not add tombstone-on-tombstone records to graveyard

### LINKED ISSUE #395: Metabase graveyard should not have tombstone-to-tombstone records
## Expected Behavior
Metabase graveyard should not have records with key corresponding to tombstone address. 
## Current Behavior
Metabase does not control types of graveyard addresses and potential tombstone-to-tombstone records.
## Possible Solution

- Before recording `address1: address2` to graveyard during `Inhume` operation, make sure the `address1` is not tombstone.
- While recording `address1: address2` to graveyard remove all records with `address2` values.

## Your Environment

* Version used: **v0.15.0**

cthulhu-rider:
Closes #395.

### DIFF
--- a/pkg/local_object_storage/metabase/inhume.go
+++ b/pkg/local_object_storage/metabase/inhume.go
@@ -1,7 +1,10 @@
 package meta
 
 import (
+	"bytes"
+
 	objectSDK "github.com/nspcc-dev/neofs-api-go/pkg/object"
+	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


alexvanin (approved):
It solves issue with having `O:T` and `T:TT` records in the same graveyard bucket. However these records still can exists in two different shards. In this case it is possible to remove object's tombstone before it will remove object itself.

/cc @realloc @cthulhu-rider

PR #403, opened 2021-02-19
Remove redundant copies of the object

### LINKED ISSUE #217: Remove redundant object copies from container
version: 0.12.1 release

No deletion of redundant copies after replication and node returning with copy unavailable previously.
1. Create a container with the policy "REP 2 IN X CBF 1 SELECT 4 FROM * AS X"
2. Put an object
3. Check that we have 2 copies - on node1 and node4
4. Stop node4
5. Object has been replicated to node2
6. Return node4 
7. After that we have 3 copies of the object in the container permanently (node1, node2, node4)

cthulhu-rider:
Closes #217.

### DIFF
--- a/pkg/local_object_storage/engine/inhume.go
+++ b/pkg/local_object_storage/engine/inhume.go
@@ -11,17 +11,20 @@ import (
 
 // InhumePrm encapsulates parameters for inhume operation.
 type InhumePrm struct {
-	addr, tombstone *objectSDK.Address
+	tombstone *objectSDK.Address
+	addrs     []*objectSDK.Address
 }
 
 // InhumeRes encapsulates results of inhume operation.
 type InhumeRes struct{}
 
-// WithTarget sets object address that should be inhumed and tombstone address
+// WithTarget sets list of objects that should be inhumed and tombstone address
 // as the reason for inhume operation.
-func (p *InhumePrm) WithTarget(addr, tombstone *objectSDK.Address) *InhumePrm {
+//
+// tombstone should not be nil, addr should not be empty.
+func (p *InhumePrm) WithTarget(tombstone *objectSDK.Address, addrs ...*objectSDK.Address) *InhumePrm {
 	if p != nil {
-		p.addr = addr
+		p.addrs = addrs
 		p.tombstone = tombstone
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/local_object_storage/engine/inhume.go
```diff
@@ -73,7 +80,7 @@ func (e *StorageEngine) inhume(addr *objectSDK.Address, prm *shard.InhumePrm, ch
 				zap.String("error", err.Error()),
 			)
 		} else {
-			res = new(InhumeRes)
+			ok = true
 		}
 
 		return err == nil
```
alexvanin, line 103: I think you need `return` there instead of `return err == nil` right now.

pkg/services/policer/check.go
```diff
@@ -68,8 +69,13 @@ func (p *Policer) processNodes(ctx context.Context, addr *object.Address, nodes
 		}
 
 		if network.IsLocalAddress(p.localAddrSrc, node) {
-			shortage--
-		} else {
+			if shortage == 0 {
+				redundantLocalCopy = true
```
alexvanin, line 73: Can't we call `p.cbRedundantCopy(addr)` right there? The shortage is already `0` anyway.

pkg/local_object_storage/engine/inhume.go
cthulhu-rider: We inhume the objects in the "best" shard. so we need to stop iterator by return `err == nil`.

pkg/services/policer/check.go
cthulhu-rider: I thought that callback will be deeply nested in for-loop and if-statement, what do u think @alexvanin?

pkg/local_object_storage/engine/inhume.go
alexvanin: Oh I missed that this return statement inside shard iterator. Got it.

pkg/services/policer/check.go
alexvanin: I didn't notice any issues with callback readability inside the loop, however I don't mind to have it explicitly outside. Just wonder maybe there are other reasons to put callback outside of loop.

cthulhu-rider: I'll leave the comment that we can slightly improve the performance instead of readability, so we can return to this case later when this place of code will become some kind of "stable".

alexvanin (approved):

PR #404, opened 2021-02-20
Notary contract invocation support

### LINKED ISSUE #47: Make heuristic evaluation for extra fee
Inner ring can't calculate precise fee for contract invocation with signature collection. While most of the nodes only register signature, one node will trigger method execution, that requires much more gas to spend. Amount of extra fee depends on method complexity. Some methods depends on inner ring size, e.g. container creation triggers nep-5 transfers to every single inner ring nodes. The more inner ring nodes we have, the more gas method spend. 

So there should be some heuristic evaluation before each invocation for extra fee. 

There are also different solution approaches for this issue. For example inner ring nodes can repeat tx that not accepted in block: tx that trigger method execution lacks gas after testinvoke evaluation. But next time testinvoke evaluation will be precise. However this approach can slower all chain operations.

alexvanin:
Closes #47 because we don't need heuristic fee calculations any more.


alexvanin:
Too many changes for v0.16.0, goes to next release.

alexvanin:
Known issues:
- [ ] mint produces 8 txs at block,
- [ ] start\stop estimations goes one by one instead of being distributed in one epoch.

cthulhu-rider (approved):

PR #421, opened 2021-03-04
Prepare governance update

alexvanin:
This PR is a part of new governance model adoption. It contains morph/client methods to get chain data as list of roles from designate contract and committee list.

### DIFF
--- a/pkg/innerring/processors/governance/list.go
+++ b/pkg/innerring/processors/governance/list.go
@@ -7,7 +7,10 @@ import (
 	"github.com/pkg/errors"
 )
 
-var errNotEnoughKeys = errors.New("alphabet list in mainnet is too short")
+var (
+	errNotEnoughKeys = errors.New("alphabet list in mainnet is too short")
+	errNotEqualLen   = errors.New("old and new alphabet lists have different length")
+)
 
 // newAlphabetList returns updated list of sidechain keys with no more than 1\3
 // of new keys from mainnet list. Function returns `errNotEnoughKeys` if
@@ -59,6 +62,32 @@ func newAlphabetList(sidechain, mainnet keys.PublicKeys) (keys.PublicKeys, error
 	}
 
 	sort.Sort(result)
+
+	return result, nil
+}
+
+// updateInnerRing function removes `before` keys from `innerRing` and adds
+// `after` keys in the list. If length of `before` and `after` is not the same
+// then function returns errNotEqualLen
+func updateInnerRing(innerRing, before, after keys.PublicKeys) (keys.PublicKeys, error) {
+	lnBefore := len(before)
+	if lnBefore != len(after) {
+		return nil, errNotEqualLen
+	}
+
+	result := make(keys.PublicKeys, 0, len(innerRing))
+
+	// O(n^2) for 7 nodes is not THAT bad.
+loop:
+	for i := range innerRing {
+		for j := range before {
+			if innerRing[i].Equal(before[j]) {
+				result = append(result, after[j])
+				continue loop
+			}
+		}
+		result = append(result, innerRing[i])
+	}
 
 	return result, nil
 }
--- a/pkg/innerring/processors/governance/list.go
+++ b/pkg/innerring/processors/governance/list.go
@@ -8,16 +8,22 @@ import (
 )
 
 var (
-	errNotEnoughKeys = errors.New("alphabet list in mainnet is too short")
-	errNotEqualLen   = errors.New("old and new alphabet lists have different length")
+	errNotEnoughKeys  = errors.New("alphabet list in mainnet is too short")
+	errNotEqualLen    = errors.New("old and new alphabet lists have different length")
+	errEmptySidechain = errors.New("sidechain list is empty")
 )
 
 // newAlphabetList returns updated list of sidechain keys with no more than 1\3
-// of new keys from mainnet list. Function returns `errNotEnoughKeys` if
-// mainnet list contains less keys than sidechain list. Function returns
-// (nil, nil) if mainnet list contains all keys from sidechain list.
+// of new keys from mainnet list. Function returns `errEmptySidechain` if
+// sidechain list is empty. Function returns `errNotEnoughKeys` if mainnet
+// list contains less keys than sidechain list. Function returns (nil, nil) if
+// mainnet list contains all keys from sidechain list.
 func newAlphabetList(sidechain, mainnet keys.PublicKeys) (keys.PublicKeys, error) {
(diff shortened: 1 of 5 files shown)


pkg/morph/client/notary.go
```diff
@@ -430,11 +438,16 @@ func invocationParams(args ...interface{}) ([]sc.Parameter, error) {
...
 // mn returns M and N multi signature numbers. For NeoFS N is a length of
-// inner ring list, and M is a 2/3+1 of it (like in dBFT).
-func mn(ir []*keys.PublicKey) (m int, n int) {
+// inner ring list, and M is a 2/3+1 of it (like in dBFT). If committee is
+// true, returns M as N/2+1.
+func mn(ir []*keys.PublicKey, committee bool) (m int, n int) {
 	n = len(ir)
 	m = n*2/3 + 1
 
+	if committee {
```
cthulhu-rider, line 478: Lets do
```
if commitee {
  m = n*2/3 + 1
} else {
  m = n/2 + 1
}
```
to underline the mutual exclusive cases.

pkg/morph/client/constructor.go
```diff
@@ -100,11 +100,17 @@ func New(key *ecdsa.PrivateKey, endpoint string, opts ...Option) (*Client, error
 		return nil, err
 	}
 
+	neo, err := cli.GetNativeContractHash(nativenames.Neo)
```
cthulhu-rider, line 103: Does `cli.GetNativeContractHash` actually return static value?

pkg/morph/client/notary.go
```diff
@@ -151,6 +153,20 @@ func (c *Client) GetNotaryDeposit() (int64, error) {
 	return bigIntDeposit.Int64(), nil
 }
 
+// UpdateNotaryList updates list of notary nodes in designate contract. Requires
+// committee multi signature.
+func (c *Client) UpdateNotaryList(list keys.PublicKeys) error {
+	if c.notary == nil {
+		return errNotaryNotEnabled
```
cthulhu-rider, line 160: Is it expected to happen in app runtime? If not, I suggest to panic.

pkg/innerring/processors/governance/list.go
```diff
@@ -62,3 +65,29 @@ func newAlphabetList(sidechain, mainnet keys.PublicKeys) (keys.PublicKeys, error
...
+	}
+
+	result := make(keys.PublicKeys, 0, len(innerRing))
+
+	// O(n^2) for 7 nodes is not THAT bad.
+loop:
+	for i := range innerRing {
+		for j := range before {
+			if innerRing[i].Equal(before[j]) {
+				result = append(result, after[j])
```
cthulhu-rider, line 85: So we replace elements from `before` with the elements from `after` with the exact same indices, is this really what it takes?

pkg/innerring/processors/governance/list.go
```diff
@@ -0,0 +1,64 @@
...
+// of new keys from mainnet list. Function returns `errNotEnoughKeys` if
+// mainnet list contains less keys than sidechain list. Function returns
+// (nil, nil) if mainnet list contains all keys from sidechain list.
+func newAlphabetList(sidechain, mainnet keys.PublicKeys) (keys.PublicKeys, error) {
+	ln := len(sidechain)
+	if len(mainnet) < ln {
+		return nil, errors.Wrapf(errNotEnoughKeys, "expecting %d keys", ln)
+	}
+
+	hmap := make(map[string]bool)
```
cthulhu-rider, line 25: Preallocate `make(map[string]bool, ln)`?

pkg/innerring/processors/governance/list.go
```diff
@@ -0,0 +1,64 @@
...
+	result := make(keys.PublicKeys, 0, ln)
+
+	for _, node := range sidechain {
+		hmap[node.Address()] = false
+	}
+
+	newNodes := 0
+	newNodeLimit := ln - (ln*2/3 + 1)
```
cthulhu-rider, line 33: Can we
```
newNodeLimit := ln/3 - 1
```
?

alexvanin: We can `newNodeLimit := (ln-1)/3`, thx.

pkg/morph/client/constructor.go
alexvanin: Technically it can return different values. However it is not designed to behave like this. NEO, GAS, RoleManagement(designate) and Notary contracts are native and can be changed without script hash change.

pkg/innerring/processors/governance/list.go
alexvanin: Yeah. Indices are not important though, we just need to replace all `before` elements with `after` elements in any order. There is a guarantee that original list contains all `before` elements (if it isn't then algorithm is broken on higher levels).

pkg/morph/client/constructor.go
alexvanin: However it will be nice to get insight on this from @roman-khimov, is it legit to cache script hashes of native contracts at the application start or not.

pkg/morph/client/notary.go
alexvanin: It is not expected to happen in runtime.

This check applied in all notary related methods at `notary.go`. I don't mind to panic there but we should change it everywhere. Can you create an enhancement issue on this and then we'll fix it later?

pkg/innerring/processors/governance/list.go
alexvanin: Fixed.

pkg/morph/client/notary.go
alexvanin: Fixed.

pkg/innerring/processors/governance/list.go
cthulhu-rider: And pls double-check potential negative `newNodeLimit` values (`ln < 3`) in the function algorithm (idk can this rly happen or not)

pkg/morph/client/notary.go
cthulhu-rider: Good idea, check it out https://github.com/nspcc-dev/neofs-node/issues/445

pkg/innerring/processors/governance/list.go
alexvanin: I've checked that, and it should be okay. The only tricky part is with empty side chain list. It should not ever happen but I made length check in https://github.com/nspcc-dev/neofs-node/pull/421/commits/7153ed3186c46638b2b7a2bd22f3fbf35fb06f7a

cthulhu-rider (approved):

PR #424, opened 2021-03-13
Cache clients by address only

### LINKED ISSUE #422: Cache pkg/clients per address instead of per (address+key) tuple
Adopt https://github.com/nspcc-dev/neofs-api-go/issues/261 changes.

fyrchik:
Close #422.

Remove both private key and options from cache interface.

### DIFF
--- a/pkg/innerring/innerring.go
+++ b/pkg/innerring/innerring.go
@@ -580,7 +580,7 @@ func createListener(ctx context.Context, p *chainParams) (event.Listener, error)
 	sub, err := subscriber.New(ctx, &subscriber.Params{
 		Log:         p.log,
 		Endpoint:    p.cfg.GetString(p.name + ".endpoint.notification"),
-		DialTimeout: p.cfg.GetDuration(p.name + ".dial_timeouts"),
+		DialTimeout: p.cfg.GetDuration(p.name + ".dial_timeout"),
 	})
 	if err != nil {
 		return nil, err
@@ -603,7 +603,7 @@ func createClient(ctx context.Context, p *chainParams) (*client.Client, error) {
 		p.cfg.GetString(p.name+".endpoint.client"),
 		client.WithContext(ctx),
 		client.WithLogger(p.log),
-		client.WithDialTimeout(p.cfg.GetDuration(p.name+".dial_timeouts")),
+		client.WithDialTimeout(p.cfg.GetDuration(p.name+".dial_timeout")),
 	)
 }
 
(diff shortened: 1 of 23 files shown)


pkg/innerring/innerring.go
```diff
@@ -579,7 +580,7 @@ func createListener(ctx context.Context, p *chainParams) (event.Listener, error)
 	sub, err := subscriber.New(ctx, &subscriber.Params{
 		Log:         p.log,
 		Endpoint:    p.cfg.GetString(p.name + ".endpoint.notification"),
-		DialTimeout: p.cfg.GetDuration(p.name + ".dial_timeouts"),
+		DialTimeout: p.cfg.GetDuration(p.name + ".dial_timeout"),
```
alexvanin, line 583: Wow, good one, thanks.

pkg/innerring/rpc.go
```diff
@@ -46,8 +46,10 @@ func newClientCache(p *clientCacheParams) *ClientCache {
 	}
 }
 
-func (c *ClientCache) Get(address string, opts ...client.Option) (*client.Client, error) {
-	return c.cache.Get(c.key, address, opts...)
+func (c *ClientCache) Get(address string) (*client.Client, error) {
+	// Because cache is used by `ClientCache` exclusively,
+	// client will always have valid key.
```
alexvanin, line 51: I think you missed `c.getSG`, `c.GetHeader` and `c.GetRangeHash` functions. They use `c.Get` to get client, and then use it without new `WithKey` option.

alexvanin (commented):
It seems that key is missing in `getsvc.execCtx` and `searchsvc.execCtx`.

`getsvc.clientCacheWrapper` and `searchsvc.clientCacheWrapper` have `get` function that return `clientWrapper`. This wrapper uses client with options from `execCtx` and it did not adopt new `client.WithKey` option. Can you double check on that?

pkg/innerring/rpc.go
fyrchik: Fixed.

fyrchik:
@alexvanin Checked it again:
For both of them, `get` done in `(*execCtx).remoteClient` and corresponding `searchObjects`/`getObject` is executed with the same context (which provide the key from via `callOptions()`).

alexvanin (approved):
LGTM. I'll try to run it on dev-env to check if everything fine and wait for @cthulhu-rider.

cthulhu-rider (approved):
The changes are correct, @alexvanin pls merge after testing in dev-env.

PR #426, opened 2021-03-15
Add metrics to storage node

### LINKED ISSUE #423: Request for additional optional metrics for neofs-node
Need to get additional neofs-node metrics:
1) Counter of processed requests filtered by type
2) Sum of time spent processing requests (filtered by type)
3) Sum of time spent to store object in the storage engine for PUT operations
4) Sum of throughput (processed objects size) for all operations (filtered by type)

alexvanin:
Closes #423 

## Provided metrics

- Object service request counter per method

```
neofs_node_object_delete_req_count 0
neofs_node_object_get_req_count 0
neofs_node_object_head_req_count 0
neofs_node_object_put_req_count 0
neofs_node_object_range_hash_req_count 0
neofs_node_object_range_req_count 0
neofs_node_object_search_req_count 1
```

- Object service accumulated duration counter per method (in ns)
```
neofs_node_object_delete_req_duration 0
neofs_node_object_get_req_duration 0
neofs_node_object_head_req_duration 0
neofs_node_object_put_req_duration 0
neofs_node_object_range_hash_req_duration 0
neofs_node_object_range_req_duration 0
neofs_node_object_search_req_duration 3.310136e+06
```

- Accumulated size of payloads at `object.Put` method (in bytes)

```
neofs_node_object_put_payload 0
```

- Accumulated size of payloads at `object.Get` method (in bytes)

```
neofs_node_object_get_payload 0
```

- Storage engine accumulated durations per public method (in ns)
```
neofs_node_engine_delete_duration 0
neofs_node_engine_estimate_container_size_duration 6.11133e+06
neofs_node_engine_exists_duration 0
neofs_node_engine_get_duration 0
neofs_node_engine_head_duration 0
neofs_node_engine_inhume_duration 0
neofs_node_engine_list_containers_duration 40743
neofs_node_engine_list_objects_duration 90613
neofs_node_engine_put_duration 0
neofs_node_engine_range_duration 0
neofs_node_engine_search_duration 16023
```

## Configuration

Metrics are disabled by default. Default address is `:9090`. To configure it, use these examples

```yaml
metrics:
  enabled: true
  address: :9090
```
```
NEOFS_METRICS_ENABLED=true
NEOFS_METRICS_ADDRESS=:9090
```

### DIFF
--- a/pkg/services/object/metrics.go
+++ b/pkg/services/object/metrics.go
@@ -1,0 +1,146 @@
+package object
+
+import (
+	"context"
+
+	"github.com/nspcc-dev/neofs-api-go/v2/object"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type (
+	MetricCollector struct {
+		next ServiceServer
+	}
+)
+
+const (
+	namespace = "neofs_node"
+	subsystem = "object"
+)
+
+// Request counter metrics.
+var (
+	getCounter = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace:   namespace,
+		Subsystem:   subsystem,
+		Name:        "get_req_count",
+		Help:        "Number of get request processed",
+		ConstLabels: nil,
+	})
+
+	putCounter = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: namespace,
+		Subsystem: subsystem,
+		Name:      "put_req_count",
+		Help:      "Number of put request processed",
+	})
+
+	headCounter = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: namespace,
+		Subsystem: subsystem,
+		Name:      "head_req_count",
+		Help:      "Number of head request processed",
+	})
+
+	searchCounter = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: namespace,
+		Subsystem: subsystem,
+		Name:      "search_req_count",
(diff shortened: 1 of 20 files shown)


pkg/services/object/metrics.go
```diff
@@ -0,0 +1,159 @@
...
+		stream:       stream,
+		metrics:      m.metrics,
+	})
+}
+
+func (m MetricCollector) Put(ctx context.Context) (object.PutObjectStreamer, error) {
+	t := time.Now()
+	defer func() {
+		m.metrics.IncPutReqCounter()
+		m.metrics.AddPutReqDuration(time.Since(t))
```
cthulhu-rider, line 75: I think we should `AddPutReqDuration` in `putStreamMetric` calls too, current call does not transmit any data.

alexvanin: Probably it is better to drop duration metric in `Get` and `Put` and use it in `Send` functions in the streams. What do you think?

cthulhu-rider: The only call that doesn't transmit the data is `Put`, so this meaningfully only for it.

alexvanin: > The only call that doesn't transmit the data is Put, so this meaningfully only for it.

I see, thank you. I moved `AddPutReqDuration` to `CloseAndRecv` since it is last invocation call.

cthulhu-rider (approved):

PR #430, opened 2021-03-17
Makefile: speed up building images

fyrchik:
When building images `tidy` and `download` are
executed inside of the docker container thus
downloading all dependencies on every invoke.
This commit allows to reuse local cache by
vendoring dependencies before the actual build.

`make images` time reduced from 627 seconds to 58 seconds.


fyrchik:
I'd actually execute `go build` directly in `Dockerfile` (instead of introducing additional targets), but decided to leave it as it is.

fyrchik:
Found a bug with directory handling, fixind.

realloc:
That's behaviour by design

PR #434, opened 2021-03-18
Feature/check assets before transfers

### LINKED ISSUE #416: Check remaining assets before making transfers
Inner ring emits GAS to storage nodes at `neofs.Deposit` processing and after emission event. Node should check it's own wallet balance before sending any GAS because inner ring first priority is to operate processes in side chain.

carpawell:
Closes #416.

### DIFF
--- a/cmd/neofs-ir/defaults.go
+++ b/cmd/neofs-ir/defaults.go
@@ -94,6 +94,7 @@ func defaultConfiguration(cfg *viper.Viper) {
 	cfg.SetDefault("emit.mint.cache_size", 1000)
 	cfg.SetDefault("emit.mint.threshold", 1)
 	cfg.SetDefault("emit.mint.value", 20000000) // 0.2 Fixed8
+	cfg.SetDefault("emit.gas.balance_threshold", 0)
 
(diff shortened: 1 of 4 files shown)


cmd/neofs-ir/defaults.go
```diff
@@ -94,6 +94,7 @@ func defaultConfiguration(cfg *viper.Viper) {
 	cfg.SetDefault("emit.mint.cache_size", 1000)
 	cfg.SetDefault("emit.mint.threshold", 1)
 	cfg.SetDefault("emit.mint.value", 20000000) // 0.2 Fixed8
+	cfg.SetDefault("emit.gas.balance_threshold", 10)
```
carpawell, line 97: need help with naming and default value

alexvanin: As for naming it's fine. We can reuse it later to make a limit on all side chain GAS transfer operations, so it is good.
As for default value, let's keep it zero. We can't provide any reasonable default value there.

carpawell: ok, changed to zero

alexvanin (approved):

PR #446, opened 2021-03-26
Fetch inner ring and alphabet keys separately from chain

alexvanin:
New governance model assumes that inner ring keys and alphabet keys are stored separately on chain and not in the contract storage. To support it we should make changes in inner ring and storage nodes by using committee and role management contract getters. Also we should distinguish usage of inner ring list (only for audit right now) and alphabet list (for all voting operations).


pkg/innerring/indexer.go
```diff
@@ -0,0 +1,80 @@
...
+
+	return nil
+}
+
+func (s *innerRingIndexer) InnerRingIndex() (int32, error) {
+	if err := s.update(); err != nil {
+		return 0, errors.Wrap(err, "can't update index state")
+	}
+
+	s.RLock()
```
cthulhu-rider, line 71: As an option, we can return `index` and `size` values from method update
```
func (s *innerRingIndexer) update() (index, size int, err error)
```
 in order to reduce the number of synchronization points. Then the method body can be shortened to
```
func (s *innerRingIndexer) InnerRingIndex() (int32, error) {
	index, _, err := s.update()
	return index, err
}
```
(same for `InnerRingSize` method).

cthulhu-rider, line 71: As an option, we can return `index` and `size` values from method update
```
func (s *innerRingIndexer) update() (index, size int, err error)
```
 in order to reduce the number of synchronization points. Then the method body can be shortened to
```
func (s *innerRingIndexer) InnerRingIndex() (int32, error) {
	index, _, err := s.update()
	return index, err
}
```
(same for `InnerRingSize` method).

cmd/neofs-node/object.go
```diff
@@ -138,6 +139,24 @@ func (i *delNetInfo) TombstoneLifetime() (uint64, error) {
 	return i.tsLifetime, nil
 }
 
+type innerRingFetcher struct {
+	sidechain *morphClient.Client
+}
+
+func (n *innerRingFetcher) InnerRingKeys() ([][]byte, error) {
+	keys, err := n.sidechain.NeoFSAlphabetList()
```
cthulhu-rider, line 147: Shouldn't non-alphabet nodes also be processed in ACL sender classifier (e.g. for audit-related requests)?

alexvanin: They should and they does. Side chain stores inner ring (not alphabet) keys in `NeoFSAlphabet` role. Unfortunately we don't have special inner ring role for our side chain in NEO so we have to re-use existing ones :(

cthulhu-rider: Oh, I see now, thx!

pkg/innerring/indexer.go
```diff
@@ -52,6 +53,11 @@ func (s *innerRingIndexer) update() (err error) {
 		return err
 	}
 
+	s.alphabetIndex, err = invoke.AlphabetIndex(s.cli, s.key)
```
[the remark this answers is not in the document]
alexvanin: I am not sure what desired behavior should be. If inner ring or alphabet keys are unavailable -- it is very-very bad state of the network. It might be not safe to continue work in such network. Thus I decided to proceed only if all values are fetched correctly.

pkg/innerring/indexer.go
alexvanin: Seems reasonable. Instead of ints I decided to return structure that stores all indexes. Check it out: https://github.com/nspcc-dev/neofs-node/pull/446/commits/2052065423ddc76d64feb74bb3f74ad36f632527

cthulhu-rider: Good, but do we rly need to return a pointer? It seems like we can just `return s.ind` from `update`.

alexvanin: We don't. I was just too lazy to write `return indexes{}, err` instead of `return nil, err`. Fixed.

cthulhu-rider (approved):

PR #447, opened 2021-03-26
Sync alphabet keys from main net

alexvanin:


pkg/innerring/processors/governance/process_update.go
```diff
@@ -0,0 +1,98 @@
...
+		gp.log.Error("can't fetch inner ring list from side chain",
+			zap.String("error", err.Error()))
+	} else {
+		newInnerRing, err := updateInnerRing(innerRing, sidechainAlphabet, newAlphabet)
+		if err != nil {
+			gp.log.Error("can't create new inner ring list with new alphabet keys",
+				zap.String("error", err.Error()))
+		}
+
+		sort.Sort(newInnerRing)
```
cthulhu-rider, line 67: 1. can `newInnerRing` have the same items as `innerRing`, and if so, should we call `UpdateNeoFSAlphabetList` in this case?
2. if `err != nil` in previous statement should we perform this action or it must be placed in `else` block?

alexvanin: 1. It is not possible. In this case `mainnet` should be equal to `sidechain`, but `newAlphabetList` will return `nil`  and it will be processed:
```go
	if newAlphabet == nil {
		gp.log.Info("no governance update, alphabet list has not been changed")
		return
	}
```

2. You are right about this, it should placed in else block. Fixed.

cthulhu-rider (approved):

PR #452, opened 2021-03-31
Support updated reputation contract

alexvanin:
Related to https://github.com/nspcc-dev/neofs-contract/pull/65

- [x] Support reputation contract in morph/client
- [x] Add client wrapper
- [x] Support reputation put event 
- [x] Add reputation processor to inner ring node
- [x] Add comments
- [x] Use structure to represent reputation value instead of `[]byte`.

### DIFF
--- a/pkg/innerring/processors/reputation/process_put.go
+++ b/pkg/innerring/processors/reputation/process_put.go
@@ -3,18 +3,38 @@ package reputation
 import (
 	"encoding/hex"
 
+	"github.com/nspcc-dev/neofs-api-go/pkg/reputation"
 	"github.com/nspcc-dev/neofs-node/pkg/morph/client/reputation/wrapper"
-	"github.com/nspcc-dev/neofs-node/pkg/services/reputation"
 	"go.uber.org/zap"
 )
 
-func (rp *Processor) processPut(epoch uint64, id reputation.PeerID, value []byte) {
+func (rp *Processor) processPut(epoch uint64, id reputation.PeerID, value reputation.GlobalTrust) {
 	if !rp.alphabetState.IsAlphabet() {
 		rp.log.Info("non alphabet mode, ignore reputation put notification")
 		return
 	}
 
-	// todo: do sanity checks of value and epoch
+	// check if epoch is valid
+	currentEpoch := rp.epochState.EpochCounter()
+	if epoch >= currentEpoch {
+		rp.log.Info("ignore reputation value",
+			zap.String("reason", "invalid epoch number"),
+			zap.Uint64("trust_epoch", epoch),
+			zap.Uint64("local_epoch", currentEpoch))
+
+		return
+	}
+
+	// check signature
+	if err := value.VerifySignature(); err != nil {
+		rp.log.Info("ignore reputation value",
+			zap.String("reason", "invalid signature"),
+			zap.String("error", err.Error()))
+
+		return
+	}
+
+	// todo: do sanity checks of value
 
 	args := wrapper.PutArgs{}
 	args.SetEpoch(epoch)
@@ -24,7 +44,7 @@ func (rp *Processor) processPut(epoch uint64, id reputation.PeerID, value []byte
 	err := rp.reputationWrp.PutViaNotary(args)
 	if err != nil {
 		rp.log.Warn("can't send approval tx for reputation value",
-			zap.String("peer_id", hex.EncodeToString(id.Bytes())),
+			zap.String("peer_id", hex.EncodeToString(id.ToV2().GetValue())),
 			zap.String("error", err.Error()))
 	}
 }
--- a/pkg/innerring/processors/reputation/process_put.go
+++ b/pkg/innerring/processors/reputation/process_put.go
@@ -1,0 +1,30 @@
+package reputation
+
+import (
+	"encoding/hex"
+
+	"github.com/nspcc-dev/neofs-node/pkg/morph/client/reputation/wrapper"
+	"github.com/nspcc-dev/neofs-node/pkg/services/reputation"
+	"go.uber.org/zap"
+)
+
+func (rp *Processor) processPut(epoch uint64, id reputation.PeerID, value []byte) {
+	if !rp.alphabetState.IsAlphabet() {
+		rp.log.Info("non alphabet mode, ignore reputation put notification")
+		return
+	}
+
+	// todo: do sanity checks of value and epoch
+
+	args := wrapper.PutArgs{}
+	args.SetEpoch(epoch)
+	args.SetPeerID(id)
+	args.SetValue(value)
+
+	err := rp.reputationWrp.PutViaNotary(args)
+	if err != nil {
+		rp.log.Warn("can't send approval tx for reputation value",
+			zap.String("peer_id", hex.EncodeToString(id.Bytes())),
+			zap.String("error", err.Error()))
+	}
+}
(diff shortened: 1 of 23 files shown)


pkg/innerring/processors/reputation/process_put.go
```diff
@@ -0,0 +1,50 @@
...
+	"encoding/hex"
+
+	"github.com/nspcc-dev/neofs-api-go/pkg/reputation"
+	"github.com/nspcc-dev/neofs-node/pkg/morph/client/reputation/wrapper"
+	"go.uber.org/zap"
+)
+
+func (rp *Processor) processPut(epoch uint64, id reputation.PeerID, value reputation.GlobalTrust) {
+	if !rp.alphabetState.IsAlphabet() {
+		rp.log.Info("non alphabet mode, ignore new peer notification")
```
cthulhu-rider, line 13: I guess incorrect message

pkg/innerring/processors/reputation/process_put.go
```diff
@@ -0,0 +1,50 @@
...
+		return
+	}
+
+	// check if epoch is valid
+	currentEpoch := rp.epochState.EpochCounter()
+	if epoch >= rp.epochState.EpochCounter() {
```
cthulhu-rider, line 19:
```suggestion
	if epoch >= currentEpoch {
```

pkg/morph/client/reputation/get.go
```diff
@@ -0,0 +1,99 @@
...
+func (g *GetArgs) SetPeerID(v []byte) {
+	g.peerID = v
+}
+
+// SetID sets id of expected reputation value in reputation contract.
+func (g *GetByIDArgs) SetID(v []byte) {
+	g.id = v
+}
+
+// Reputations returns marshaled slice of reputation values.
```
cthulhu-rider, line 42:
```suggestion
// Reputations returns slice of marshaled reputation values.
```

pkg/morph/client/reputation/put.go
```diff
@@ -0,0 +1,48 @@
...
+func (p *PutArgs) SetPeerID(v []byte) {
+	p.peerID = v
+}
+
+// SetValue sets marshaled reputation value.
+func (p *PutArgs) SetValue(v []byte) {
+	p.value = v
+}
+
+// Put invokes direct call of "put reputation value" method of audit contract.
```
cthulhu-rider, line 29:
```suggestion
// Put invokes direct call of "put reputation value" method of reputation contract.
```

pkg/morph/client/reputation/put.go
```diff
@@ -0,0 +1,48 @@
...
+	return errors.Wrapf(c.client.Invoke(
+		c.putMethod,
+		int64(args.epoch),
+		args.peerID,
+		args.value,
+	), "could not invoke method (%s)", c.putMethod)
+}
+
+// PutViaNotary invokes notary call of "put reputation value" method of
+// audit contract.
```
cthulhu-rider, line 40:
```suggestion
// reputation contract.
```

pkg/morph/client/reputation/put.go
```diff
@@ -0,0 +1,48 @@
...
+func (c *Client) PutViaNotary(args PutArgs) error {
```
cthulhu-rider, line 41: Cosmetic: add boolean `notary` to `PutArgs` and leave only method `Put` (the same in `wrapper` pkg).

pkg/morph/client/reputation/wrapper/get.go
```diff
@@ -0,0 +1,95 @@
...
+		peerID reputation.PeerID
+	}
+
+	// GetByIDArgs groups the arguments of "get reputation value by
+	// reputation id" test invocation.
+	GetByIDArgs struct {
+		id ReputationID
+	}
+
+	// ListByEpochResult groups the result of "get reputation value" and
```
cthulhu-rider, line 22:
```suggestion
	// GetResult groups the result of "get reputation value" and
```

pkg/morph/client/reputation/wrapper/get.go
```diff
@@ -0,0 +1,95 @@
...
+// Reputations returns slice of reputation values.
+func (g GetResult) Reputations() []reputation.GlobalTrust {
+	return g.reputations
+}
+
+// Get invokes the call of "get reputation value" method of reputation contract.
+func (w *ClientWrapper) Get(v GetArgs) (*GetResult, error) {
+	args := reputationClient.GetArgs{}
+	args.SetEpoch(v.epoch)
+	args.SetPeerID(v.peerID.ToV2().GetValue())
```
cthulhu-rider, line 53: Why don't we use `v.peerID.Marshal()` (same in `wrapper.preparePutArgs`)?

pkg/morph/client/reputation/wrapper/put.go
```diff
@@ -0,0 +1,67 @@
...
+type (
+	// PutArgs groups the arguments of "put reputation value" invocation call.
+	PutArgs struct {
+		epoch  uint64
+		peerID reputation.PeerID
+		value  reputation.GlobalTrust
+	}
+)
+
+// SetPeerID sets epoch of reputation value.
```
cthulhu-rider, line 18:
```suggestion
// SetEpoch sets epoch of reputation value.
```

pkg/morph/client/reputation/wrapper/put.go
```diff
@@ -0,0 +1,67 @@
...
+func (p *PutArgs) SetPeerID(v reputation.PeerID) {
+	p.peerID = v
+}
+
+// SetValue sets reputation value.
+func (p *PutArgs) SetValue(v reputation.GlobalTrust) {
+	p.value = v
+}
+
+// Put invokes direct call of "put reputation value" method of audit contract.
```
cthulhu-rider, line 33:
```suggestion
// Put invokes direct call of "put reputation value" method of reputation contract.
```

pkg/morph/client/reputation/put.go
alexvanin: I would like to keep two separate functions. Notary call is an add-on for client and it should not affect functions that can be used without notary, I think.

pkg/morph/client/reputation/wrapper/get.go
alexvanin: PeerID value used as a part of the key for storage in reputation contract. Storage key has it's own size limits, so I decided to keep exact value right there without extra meta information of protobuf encoding. This should encourage developers to not change `reputation.PeerID` definition in API.

cthulhu-rider: But why contract's storage key imposes requirements to `PeerID` size? It becomes pretty version-dependent then, same as this part of the code.

alexvanin: > But why contract's storage key imposes requirements to PeerID size? 

It is NEO limitation of max storage key size (64 byte), so we can't do anything about that. 

Another thing is that bytes of marshaled `PeerID` structure is not PeerID. The same as bytes of marshaled structure of ContainerID is not ContainerID, so it seemed legit to use value in this case.

cthulhu-rider: I got it, in the background I will think about how to simultaneously adjust to the restrictions of the contract and avoid using `ToV2()` (apparently some way of getting a binary ID of a fixed size).

cthulhu-rider (approved):

PR #460, opened 2021-04-05
Feature/add tranport for local trusts

carpawell:
Implement `Router` and its necessary dependencies for local `reputation.Trust`s. Add `Router` to `reputationServer`.

Fix some logical mistakes in the reputation controller's commentaries.

### DIFF
--- a/cmd/neofs-node/reputation.go
+++ b/cmd/neofs-node/reputation.go
@@ -122,7 +122,7 @@ type localTrustLogger struct {
 }
 
 func (l *localTrustLogger) Write(t reputation.Trust) error {
-	l.log.Info("new local trust",
+	l.log.Info("received local trust",
 		zap.Uint64("epoch", l.ctx.Epoch()),
 		zap.String("peer", hex.EncodeToString(t.Peer().Bytes())),
 		zap.Stringer("value", t.Value()),
@@ -324,8 +324,11 @@ func initReputationService(c *cfg) {
 			reputationrpc.NewSignService(
 				c.key,
 				reputationrpc.NewResponseService(
-					&loggingReputationServer{
-						log: c.log,
+					&reputationServer{
+						cfg:          c,
+						log:          c.log,
+						router:       router,
+						routeBuilder: routeBuilder,
 					},
 					c.respSvc,
 				),
@@ -334,20 +337,65 @@ func initReputationService(c *cfg) {
 	)
 }
 
-type loggingReputationServer struct {
-	log *logger.Logger
+type reputationServer struct {
+	*cfg
+	log          *logger.Logger
+	router       trustcontroller.WriterProvider
+	routeBuilder reputationroute.Builder
 }
 
-func (s *loggingReputationServer) SendLocalTrust(_ context.Context, req *v2reputation.SendLocalTrustRequest) (*v2reputation.SendLocalTrustResponse, error) {
+type epochContext struct {
+	context.Context
+	epoch uint64
+}
+
+func (ctx *epochContext) Epoch() uint64 {
+	return ctx.epoch
+}
+
+type reputationOnlyKeyRemoteServerInfo struct {
+	key []byte
+}
+
+func (i *reputationOnlyKeyRemoteServerInfo) PublicKey() []byte {
+	return i.key
+}
+
+func (*reputationOnlyKeyRemoteServerInfo) Address() string {
+	return ""
+}
+
+func (s *reputationServer) SendLocalTrust(ctx context.Context, req *v2reputation.SendLocalTrustRequest) (*v2reputation.SendLocalTrustResponse, error) {
+	var passedRoute []reputationroute.ServerInfo
+
+	for hdr := req.GetVerificationHeader(); hdr != nil; hdr = hdr.GetOrigin() {
+		passedRoute = append(passedRoute, &reputationOnlyKeyRemoteServerInfo{
+			key: hdr.GetBodySignature().GetKey(),
+		})
+	}
+
+	for left, right := 0, len(passedRoute)-1; left < right; left, right = left+1, right-1 {
+		passedRoute[left], passedRoute[right] = passedRoute[right], passedRoute[left]
+	}
+
+	passedRoute = append(passedRoute, s)
+
 	body := req.GetBody()
 
-	log := s.log.With(zap.Uint64("epoch", body.GetEpoch()))
+	eCtx := &epochContext{
+		Context: ctx,
+		epoch:   body.GetEpoch(),
+	}
 
-	for _, t := range body.GetTrusts() {
-		log.Info("local trust received",
-			zap.String("peer", hex.EncodeToString(t.GetPeer().GetValue())),
-			zap.Float64("value", t.GetValue()),
-		)
+	w, err := s.router.InitWriter(reputationroute.NewRouteContext(eCtx, passedRoute))
+	if err != nil {
+		return nil, errors.Wrap(err, "could not initialize local trust writer")
+	}
+
+	for _, trust := range body.GetTrusts() {
+		if err = s.processTrust(body.GetEpoch(), apiToLocalTrust(trust), passedRoute, w); err != nil {
+			return nil, errors.Wrap(err, "could not write one of trusts")
+		}
 	}
 
 	resp := new(v2reputation.SendLocalTrustResponse)
@@ -356,10 +404,29 @@ func (s *loggingReputationServer) SendLocalTrust(_ context.Context, req *v2reput
 	return resp, nil
 }
 
-func (s *loggingReputationServer) SendIntermediateResult(_ context.Context, req *v2reputation.SendIntermediateResultRequest) (*v2reputation.SendIntermediateResultResponse, error) {
+func (s *reputationServer) SendIntermediateResult(_ context.Context, req *v2reputation.SendIntermediateResultRequest) (*v2reputation.SendIntermediateResultResponse, error) {
 	resp := new(v2reputation.SendIntermediateResultResponse)
 
 	// todo: implement me
 
 	return resp, nil
+}
+
+func apiToLocalTrust(t *v2reputation.Trust) reputation.Trust {
+	localTrust := reputation.Trust{}
+
+	localTrust.SetValue(reputation.TrustValueFromFloat64(t.GetValue()))
+	localTrust.SetPeer(reputation.PeerIDFromBytes(t.GetPeer().GetValue()))
+
+	return localTrust
+}
+
+func (s *reputationServer) processTrust(epoch uint64, t reputation.Trust,
+	passedRoute []reputationroute.ServerInfo, w trustcontroller.Writer) error {
+	err := reputationroute.CheckRoute(s.routeBuilder, epoch, reputation.PeerIDFromBytes(passedRoute[0].PublicKey()), passedRoute)
+	if err != nil {
+		return errors.Wrap(err, "wrong route of reputation trust value")
+	}
+
+	return w.Write(t)
 }
--- a/cmd/neofs-node/reputation.go
+++ b/cmd/neofs-node/reputation.go
@@ -3,17 +3,26 @@ package main
 import (
 	"bytes"
 	"context"
+	"crypto/ecdsa"
 	"encoding/hex"
 
+	"github.com/nspcc-dev/hrw"
+	apiClient "github.com/nspcc-dev/neofs-api-go/pkg/client"
+	apiNetmap "github.com/nspcc-dev/neofs-api-go/pkg/netmap"
+	reputationapi "github.com/nspcc-dev/neofs-api-go/pkg/reputation"
 	v2reputation "github.com/nspcc-dev/neofs-api-go/v2/reputation"
 	v2reputationgrpc "github.com/nspcc-dev/neofs-api-go/v2/reputation/grpc"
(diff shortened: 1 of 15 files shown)


pkg/services/reputation/local/route/managers/calls.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	peer := reputationapi.PeerID(b.peer.Bytes())
+
+	route, err := b.managerBuilder.BuildManagers(epoch, peer)
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not build managers for epoch: %d", epoch)
+	}
+
+	res := make([]reputationroute.ServerInfo, 0, len(route))
```
cthulhu-rider, line 24: Why do we need to copy `route`?

pkg/services/reputation/local/route/deps.go
```diff
@@ -18,6 +18,17 @@ type ServerInfo interface {
...
 
+// Builder groups methods to route values in the network.
+type Builder interface {
+	// NextStage must return next group of route points for passed epoch
+	// based on the passed route.
+	//
+	// Empty passed list means being at the starting point of the route.
+	//
+	// Must return empty list and no error if the endpoint of the route is reached.
+	NextStage(epoch uint64, passed []ServerInfo) ([]ServerInfo, error)
```
cthulhu-rider, line 29: The 1st argument should be type of `reputation.Trust` since route may depend not only on the epoch.

pkg/services/reputation/local/route/managers/deps.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"github.com/nspcc-dev/neofs-api-go/pkg/netmap"
+	"github.com/nspcc-dev/neofs-api-go/pkg/reputation"
+)
+
+// ManagerBuilder defines an interface for providing a list
+// of Managers for specific epoch and node.
+type ManagerBuilder interface {
+	// BuildManagers must compose list of managers,
+	// using network map of particular epoch.
```
cthulhu-rider, line 12: Lets remove `using network map` since this is an implementation detail, not a requirement for all implementations.

cmd/neofs-node/reputation.go
```diff
@@ -39,6 +47,133 @@ type localTrustIterator struct {
...
+}
+
+func (rtp *remoteLocalTrustWriter) Write(t reputation.Trust) error {
+	val := t.Value().Float64()
+	peer := t.Peer().Bytes()
+
+	apiTrust := reputationapi.NewTrust()
+
+	apiTrust.SetValue(val)
+	apiTrust.SetPeer(reputationapi.PeerIDFromV2(peer))
```
cthulhu-rider, line 87: Pls update `neofs-api-go` lib, there is a new API in `reputation` pkg there.

cmd/neofs-node/reputation.go
```diff
@@ -39,6 +47,133 @@ type localTrustIterator struct {
...
+		ctx:    ctx,
+		client: rtwp.client,
+	}, nil
+}
+
+var ErrNoServerInfoProvided = errors.New("no server info provided in InitRemove call")
+
+func (rtp *remoteLocalTrustProvider) InitRemote(srv reputationroute.ServerInfo) (trustcontroller.WriterProvider, error) {
+	if srv == nil {
+		return nil, ErrNoServerInfoProvided
```
cthulhu-rider, line 123: Interface definitions covers this case:
```
// Nil info matches the end of the route.
```

cmd/neofs-node/reputation.go
```diff
@@ -39,6 +47,133 @@ type localTrustIterator struct {
...
+	}
+
+	return &remoteLocalTrustWriterProvider{
+		client: c,
+	}, nil
+}
+
+var ErrNoCurrentNodeFoundInNetMap = errors.New("could not find current node in NetMap")
+
+// BuildManagers sorts nodes in NetMap with HRW algorithms and trivially
```
cthulhu-rider, line 150: Remove `trivially` word

pkg/services/reputation/local/route/managers/calls.go
```diff
@@ -0,0 +1,31 @@
+package managers
+
+import (
+	reputationapi "github.com/nspcc-dev/neofs-api-go/pkg/reputation"
+	reputationroute "github.com/nspcc-dev/neofs-node/pkg/services/reputation/local/route"
+	"github.com/pkg/errors"
+)
+
+// NextStage builds Manager list for node and returns it trivially.
```
cthulhu-rider, line 9: `trivially` -> `directly`

cmd/neofs-node/reputation.go
```diff
@@ -39,6 +47,133 @@ type localTrustIterator struct {
...
+// takes the next node after the current one as the only manager.
+func (mb *managerBuilder) BuildManagers(epoch uint64, id reputationapi.PeerID) (netmapapi.Nodes, error) {
+	nm, err := mb.nmSrc.GetNetMapByEpoch(epoch)
+	if err != nil {
+		return nil, err
+	}
+
+	nodes := nm.Nodes
+
+	hrw.SortSliceByValue(nodes, epoch)
```
cthulhu-rider, line 160: We need to copy `nm.Nodes` in order to avoid violation of the network map structure, or impose an additional requirement on `managerBuilder.nmSrc` for free mutation on the structure returned by the method.

cmd/neofs-node/reputation.go
```diff
@@ -99,49 +234,47 @@ func (it *localTrustIterator) Iterate(h reputation.TrustHandler) error {
...
-type localTrustLogger struct {
-	ctx trustcontroller.Context
+	managerBuilder := &managerBuilder{
+		log:   c.log,
+		nmSrc: nmSrc,
+	}
 
-	log *logger.Logger
-}
+	LocalPeerID := reputation.PeerIDFromBytes(localKey)
```
cthulhu-rider, line 247: `LocalPeerID` -> `localPeerID`

cmd/neofs-node/reputation.go
```diff
@@ -39,6 +47,133 @@ type localTrustIterator struct {
...
+
+			if managerIndex == len(nodes) {
+				managerIndex = 0
+			}
+
+			return netmapapi.Nodes{nodes[managerIndex]}, nil
+		}
+	}
+
+	return nil, ErrNoCurrentNodeFoundInNetMap
```
cthulhu-rider, line 174: We can just `return nil, nil` instead, nodes outside the network map have an empty list of managers.

cmd/neofs-node/reputation.go
```diff
@@ -39,6 +47,133 @@ type localTrustIterator struct {
...
+
+	for i, node := range nodes {
+		if bytes.Equal(node.PublicKey(), id) {
+			managerIndex := i + 1
+
+			if managerIndex == len(nodes) {
+				managerIndex = 0
+			}
+
+			return netmapapi.Nodes{nodes[managerIndex]}, nil
```
cthulhu-rider, line 170: The code is correct, cosmetic savings on variables:
```
	if i == len(nodes)-1 {
		i = 0
	}

	return netmapapi.Nodes{nodes[i]}, nil
}
```

cthulhu-rider (changes requested):

pkg/services/reputation/local/route/calls.go
```diff
@@ -0,0 +1,137 @@
...
+
+	routeMtx sync.RWMutex
+	mServers map[string]reputationcontroller.Writer
+}
+
+var ErrWrongContextProvided = errors.New("provided context does not contain required values")
+
+// InitWriter initializes and returns Writer that sends each value to its next route point.
+//
+// Ctx MUST be created by NewRouteContext function.
```
cthulhu-rider, line 46: Is such a limitation really necessary?

cmd/neofs-node/reputation.go
alexvanin: After https://github.com/nspcc-dev/neofs-node/pull/452 you can just rebase to get latest api-go version.

cmd/neofs-node/reputation.go
carpawell: i found my version more readable: `i` is too far from "managerIndex" meaning, i think. also, it is necessary to `i = -1`  in `if` statement for correct work in your suggestion which is a little confusing i guess

pkg/services/reputation/local/route/managers/deps.go
carpawell: removed

cmd/neofs-node/reputation.go
carpawell: rebased and used new api version

cmd/neofs-node/reputation.go
carpawell: deleted returning non-nil error

cmd/neofs-node/reputation.go
carpawell: removed

cmd/neofs-node/reputation.go
carpawell: added copying

cmd/neofs-node/reputation.go
carpawell: after refactor deleted that var

cmd/neofs-node/reputation.go
carpawell: deleted returning non-nil error

pkg/services/reputation/local/route/calls.go
carpawell: my bad, changed description

pkg/services/reputation/local/route/managers/calls.go
carpawell: changed

pkg/services/reputation/local/route/calls.go
```diff
@@ -0,0 +1,136 @@
...
+	"go.uber.org/zap"
+)
+
+type routeContext struct {
+	reputationcontroller.Context
+
+	passedRoute []ServerInfo
+}
+
+func (r *routeContext) Epoch() uint64 {
```
cthulhu-rider, line 17: Method is redundant, it is implemented through inheritance

cmd/neofs-node/reputation.go
```diff
@@ -334,20 +336,66 @@ func initReputationService(c *cfg) {
...
-		log.Info("local trust received",
-			zap.String("peer", hex.EncodeToString(t.GetPeer().GetValue())),
-			zap.Float64("value", t.GetValue()),
-		)
+	w, err := s.router.InitWriter(reputationroute.NewRouteContext(eCtx, passedRoute))
+	if err != nil {
+		return nil, errors.Wrap(err, "could not initialize local trust writer")
+	}
+
+	trusts := apiToLocalTrust(body.GetTrusts())
```
cthulhu-rider, line 393: We iterate over `trusts`, so to avoid additional allocation we can iterate over `body.GetTrusts()` and convert single value per iteration.

cmd/neofs-node/reputation.go
```diff
@@ -356,10 +404,33 @@ func (s *loggingReputationServer) SendLocalTrust(_ context.Context, req *v2reput
...
-func (s *loggingReputationServer) SendIntermediateResult(_ context.Context, req *v2reputation.SendIntermediateResultRequest) (*v2reputation.SendIntermediateResultResponse, error) {
+func (s *reputationServer) SendIntermediateResult(_ context.Context, req *v2reputation.SendIntermediateResultRequest) (*v2reputation.SendIntermediateResultResponse, error) {
 	resp := new(v2reputation.SendIntermediateResultResponse)
 
 	// todo: implement me
 
 	return resp, nil
 }
+
+func (s *reputationServer) Address() string {
```
cthulhu-rider, line 415: Why do we need this and `PublicKey()` methods?

cthulhu-rider (changes requested):

cmd/neofs-node/reputation.go
carpawell: i decided to pass `reputationServer` to `passedRoute` as is, not internal `reputationServer.cfg`, because i thought that `cfg` could be gone sometime in the future => hided localserver info behind method `Address()` but not field `cfg`

but maybe it was the wrong decision

cmd/neofs-node/reputation.go
carpawell: changed

pkg/services/reputation/local/route/calls.go
carpawell: forget to delete after previous changes. deleted

cmd/neofs-node/reputation.go
cthulhu-rider: I agree with your motivation, for such a need, in order to avoid an additional method at the moment, you can embed `cfg` into `reputationServer`.

cmd/neofs-node/reputation.go
cthulhu-rider: Nice, but lets return value instead of pointer since we dereference it anyway.

cmd/neofs-node/reputation.go
carpawell: ok, added

cmd/neofs-node/reputation.go
carpawell: changed

cthulhu-rider (approved):

cthulhu-rider (approved):

PR #468, opened 2021-04-08
Update neo-go to v0.94.1

alexvanin:
New neo-go version provides:
- new type for roles in `RoleManagement` contract,
- methods to get keys from `RoleManagement` contract,
- new way to sign notary transaction.


roman-khimov:
> +13 −80

Looks like we're not that bad with 0.94.1 update.

cthulhu-rider (approved):

PR #472, opened 2021-04-09
Implement write-cache

fyrchik:
1. Small objects are first accumulated in memory then saved to a cache DB.
2. Flushing to the main storage is done in the background.
3. Write-cache performs considers objects as blobs, thus there is possibly unexpected behaviour for splitted objects, which needs to be discussed (related tests are skipped).


pkg/local_object_storage/shard/put.go
```diff
@@ -41,19 +41,18 @@ func (s *Shard) Put(prm *PutPrm) (*PutRes, error) {
...
-			s.log.Debug("can't put message to writeCache, trying to blobStor")
-
-			res = nil // just in case
+		err = s.writeCache.Put(prm.obj)
+		if err == nil {
+			return nil, nil
 		}
+
+		s.log.Debug("can't put message to writeCache, trying to blobStor")
+		res = nil // just in case
```
alexvanin, line 50: You don't need this assignment anymore. Also remove comment about `// res == nil if there is no writeCache..`

pkg/local_object_storage/shard/get_test.go
```diff
@@ -68,6 +63,10 @@ func testShardGet(t *testing.T, sh *shard.Shard) {
 	})
 
 	t.Run("parent object", func(t *testing.T) {
+		if hasWriteCache {
+			t.Skip()
```
alexvanin, line 67: I think we still can pass this and `TestShard_Head/with_write_cache/virtual_object` test without skip.

The issue with these tests is that write cache do not index objects in metabase immediately. However indexing happens when data flushed to persisted storage (disk cache or even blobstor). If it is possible to wait for flush or invoke it somehow, I would do this:

```go
if hasWriteCache {
	_, err = sh.Get(getPrm)
	require.True(t, errors.Is(err, object.ErrNotFound))
	// invoke flush to persistent storage somehow, sleep maybe?
}
_, err = sh.Get(getPrm)
var expectedErr *objectSDK.SplitInfoError
require.True(t, errors.As(err, &expectedErr))
si, ok := err.(*objectSDK.SplitInfoError)
...
```

pkg/local_object_storage/metabase/put.go
```diff
@@ -349,48 +348,43 @@ func putListIndexItem(tx *bbolt.Tx, item namedBucketItem) error {
...
-	if bkt == nil {
-		// if object exists, don't have blobovniczaID and we want to update it
-		// then ignore, this should never happen
-		return ErrIncorrectBlobovniczaUpdate
-	}
-
-	objectKey := objectKey(addr.ObjectID())
-
-	if len(bkt.Get(objectKey)) == 0 {
-		return ErrIncorrectBlobovniczaUpdate
```
alexvanin, line 390: Without this check `ErrIncorrectBlobovniczaUpdate` is no longer used, so remove it.

alexvanin (approved):

cthulhu-rider (approved):

Issue #437, opened 2021-03-22
Ignore `already on chain` errors of notary contract


alexvanin:
Errors like `failed to submit notary request: Block or transaction validation failed. (-504) - invalid attribute: conflicting transaction <tx> is already on chain` in notary method invocation are totally fine and should not be marked at `WARN` level in node. It should be either ignored or moved to `DEBUG` level.

PR #475, opened 2021-04-12
bug/ignore "already on chain" errors of notary contract

carpawell:
Related to #437.

There is no custom error in the NeoGo repository that describes the error that should be ignored. Up to HTTP call. There is no even any string constant that can be used for searching: every word in the error message appears in the `fmt.Errorf("...%w..."...)` form in NeoGo code. Therefore, it seems that there is no better solution(without changing NeoGo codebase) than just searching for "already on chain" substring of error message.

### DIFF
--- a/pkg/morph/client/notary.go
+++ b/pkg/morph/client/notary.go
@@ -1,6 +1,8 @@
 package client
 
 import (
+	"strings"
+
 	"github.com/nspcc-dev/neo-go/pkg/core/native/nativenames"
 	"github.com/nspcc-dev/neo-go/pkg/core/native/noderoles"
 	"github.com/nspcc-dev/neo-go/pkg/core/transaction"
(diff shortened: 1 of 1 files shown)


roman-khimov:
A part of the problem is that you're getting this error via RPC, so it's stringified already and there is nothing you can do about it. It theoretically could be solved by error codes that are returned along with strings, but error code set we have is directly inherited from C# implementation and it defines very small set of generic errors.

pkg/morph/client/notary.go
```diff
@@ -279,7 +281,7 @@ func (c *Client) notaryInvoke(committee bool, contract util.Uint160, method stri
 		0,
 		c.notary.fallbackTime,
 		c.acc)
-	if err != nil {
+	if err != nil && !alreadyOnChainError(err) {
```
alexvanin, line 284: Add comment about the reason `already on chain` error is being ignored. For example:

```go
// Neo RPC node can return `core.ErrInvalidAttribute` error with
// `conflicting transaction <> is already on chain` message. This
// error is expected and ignored. As soon as main tx persisted on 
// chain everything is fine. This happens because notary contract 
// requires 5 out of 7 signatures to send main tx, thus last two 
// notary requests may be processed after main tx appeared on chain.
```

alexvanin (approved):

PR #476, opened 2021-04-13
Feature/add async worker notification handler

### LINKED ISSUE #440: Submit sidechain notification handlers to worker pool(s)
**Description**
Handlers of sidechain notification events are expected to be very lightweight since event `Listener` call them synchronously one-by-one. If some list of instructions isn't lightweight (e.g. contains network ops), it is expected that handler will generate go-routine for asynchronous execution of these instructions.

Right now node application registers all notification handlers regardless of their "weight".
 
**Possible solution**

1. have a notification worker pool (global or per-contract);
2. make `registerNotificationHandlers` func to set closures over worker pool: 
```
hi.SetHandler(func(ev event.Event) {
	err := c.cfgMorph.workerPool.Submit(h) // or corresponding contract pool
	if err != nil {
		c.log.Warn("could not submit notification handler to worker pool",
			zap.String("error", err.Error()),
		)
	}
})
```

**Alternatives**

We can split handlers into sync and async groups, and use worker pool for async handlers only.

**Additional context**
It would be useful to share the mechanism between applications, for example add function to event pkg:
```
func AsyncHandler(pool util.WorkerPool, handler Handler, log *logger.Logger) Handler {
	return func(ev Event) {
		err := pool.Submit(func() {
			handler(ev)
		})
		if err != nil {
			log.Warn("could not submit notification handler to worker pool",
				zap.String("error", err.Error()),
			)
		}
	}
}
```

carpawell:
Closes #440.

Added 1 notification worker pool per one contract(one in `cfgNetmap` and one in `cfgContainer` structs). Added `WorkerPoolHandler` func that creates a closure over worker pool with `event.Handler`. It is assumed that every async handler will be created using that func in the future.

In `containter` and `netmap` files of the `main` pkg added function that creates closures over handler with appropriate worker pool => to make handler execute asynchronously it is only needed to add handlers with appropriate `add...AsyncNotificationHandler()` func with the same signature.

Behavior can be configured with env vars: 
1. make all async operations sync with `NEOFS_*contract_name*_ASYNC_WORKER_ENABLED=false` (default is `true`)
2. choose the size of worker pool with `NEOFS_*contract_name*_ASYNC_WORKER_SIZE` (default is `10`)

### DIFF
--- a/cmd/neofs-node/config.go
+++ b/cmd/neofs-node/config.go
@@ -87,12 +87,16 @@ const (
 	cfgAccountingFee      = "accounting.fee"
 
 	// config keys for cfgNetmap
-	cfgNetmapContract = "netmap.scripthash"
-	cfgNetmapFee      = "netmap.fee"
+	cfgNetmapContract          = "netmap.scripthash"
+	cfgNetmapFee               = "netmap.fee"
+	cfgNetmapWorkerPoolEnabled = "netmap.async_worker.enabled"
+	cfgNetmapWorkerPoolSize    = "netmap.async_worker.size"
 
 	// config keys for cfgContainer
-	cfgContainerContract = "container.scripthash"
-	cfgContainerFee      = "container.fee"
+	cfgContainerContract          = "container.scripthash"
+	cfgContainerFee               = "container.fee"
+	cfgContainerWorkerPoolEnabled = "container.async_worker.enabled"
+	cfgContainerWorkerPoolSize    = "container.async_worker.size"
 
 	cfgGCQueueSize = "gc.queuesize"
 	cfgGCQueueTick = "gc.duration.sleep"
@@ -235,6 +239,7 @@ type cfgContainer struct {
 
 	parsers     map[event.Type]event.Parser
 	subscribers map[event.Type][]event.Handler
+	workerPool  util2.WorkerPool // pool for asynchronous handlers
 }
 
 type cfgNetmap struct {
@@ -246,6 +251,7 @@ type cfgNetmap struct {
 	parsers map[event.Type]event.Parser
 
 	subscribers map[event.Type][]event.Handler
+	workerPool  util2.WorkerPool // pool for asynchronous handlers
 
 	state *networkState
 
@@ -332,6 +338,13 @@ func initCfg(path string) *cfg {
 
 	state := newNetworkState()
 
+	// initialize async workers if it is configured so
+	containerWorkerPool, err := initContainerWorkerPool(viperCfg)
+	fatalOnErr(err)
+
+	netmapWorkerPool, err := initNetmapWorkerPool(viperCfg)
+	fatalOnErr(err)
+
 	c := &cfg{
 		ctx:         context.Background(),
 		internalErr: make(chan error),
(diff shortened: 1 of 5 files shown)


cmd/neofs-node/config.go
```diff
@@ -87,12 +87,16 @@ const (
 	cfgAccountingFee      = "accounting.fee"
 
 	// config keys for cfgNetmap
-	cfgNetmapContract = "netmap.scripthash"
-	cfgNetmapFee      = "netmap.fee"
+	cfgNetmapContract          = "netmap.scripthash"
+	cfgNetmapFee               = "netmap.fee"
+	cfgNetmapWorkerPoolEnabled = "netmap.async_worker.enabled"
```
carpawell, line 92: maybe it is better to add `notification` word here

alexvanin: As for me, it is okay as it is. Don't mind to share this pool for something else related to contract.

cmd/neofs-node/reputation.go
```diff
@@ -309,15 +309,21 @@ func initReputationService(c *cfg) {
...
-	addNewEpochNotificationHandler(c, func(ev event.Event) {
-		var reportPrm trustcontroller.ReportPrm
-
-		// report collected values from previous epoch
-		reportPrm.SetEpoch(ev.(netmap.NewEpoch).EpochNumber() - 1)
-
-		// TODO: implement and use worker pool [neofs-node#440]
-		go c.cfgReputation.localTrustCtrl.Report(reportPrm)
-	})
+	addNewEpochNotificationHandler(
```
alexvanin, line 312: 1. What about others events? I think some of them (if not all) will benefit from async execution. Consider `cmd/neofs-node/netmap.go:77` and `cmd/neofs-node/netmap.go:89` for example.

2. If some handlers are designed by a developer to be async and some are not, then I would expect to have two functions: `addNewEpochNotificationHandler` and `addNewEpochAsyncNotificationHandler` (or at least bool argument for async). This way you won't need to copy-paste `event.WorkerPoolHandler` wrapper in all handler registrations.

```go
// netmap.go
func addNewEpochNotificationHandler(c *cfg, h event.Handler) {
	addNetmapNotificationHandler(c, newEpochNotification, h)
}

func addNewEpochAsyncNotificationHandler(c *cfg, h event.Handler) {
	addNetmapNotificationHandler(c, newEpochNotification,
		event.WorkerPoolHandler(
			c.cfgNetmap.workerPool,
			h,
			c.log,
		))
}
...
// sync handler example
addNewEpochNotificationHandler(c, func(ev event.Event) {
	c.cfgNetmap.state.setCurrentEpoch(ev.(netmapEvent.NewEpoch).EpochNumber())
})

if c.cfgNetmap.reBootstrapEnabled {
	// async handler example
	addNewEpochAsyncNotificationHandler(c, func(ev event.Event) {
		n := ev.(netmapEvent.NewEpoch).EpochNumber()
		...
		}
	})
}

```

cmd/neofs-node/config.go
carpawell: ok. keep that version

cmd/neofs-node/reputation.go
carpawell: 1. did not know, what could be done async. and are there any handlers that MUST be sync or not, so decided to keep them as they are now. can change handlers that you mentioned or try to make all of them(container, object and so on) async.
2. agree, added such func to netmap and similar func to container in `main` pkg.

alexvanin: >  can change handlers that you mentioned or try to make all of them(container, object and so on) async.

I suggest you to make async handlers for all operations that use network communication, as it stated in the original issue. I found those:
- [ ] https://github.com/carpawell/neofs-node/blob/b0d023d522ffbb9d9ab262d813c7615d909b6b18/cmd/neofs-node/netmap.go#L77 (invokes `AddPeer` method at network map contract)
- [ ] https://github.com/carpawell/neofs-node/blob/b0d023d522ffbb9d9ab262d813c7615d909b6b18/cmd/neofs-node/netmap.go#L89 (`netmapLocalNodeState()` fetches network map from the contract)
- [ ] https://github.com/carpawell/neofs-node/blob/b0d023d522ffbb9d9ab262d813c7615d909b6b18/cmd/neofs-node/container.go#L109 (announce local values to other storage nodes)
- [ ] https://github.com/carpawell/neofs-node/blob/b0d023d522ffbb9d9ab262d813c7615d909b6b18/cmd/neofs-node/container.go#L116 (sends estimate result to the contract)

carpawell: added

PR #483, opened 2021-04-19
Throw panic if notary methods are not initialized

### LINKED ISSUE #445: Tight control over the initialization of the notary in the morph client
## Context
In current implementation NeoFS blockchain [Client](https://github.com/nspcc-dev/neofs-node/blob/28fb8c971c289db251275c9f1c103a88934980c0/pkg/morph/client/client.go#L26) provides some methods (`NotaryInvoke`, etc.) that should be called after successful initialization via `EnableNotarySupport` method. These group of methods returns an error if initialization wasn't performed before the call. The problem is that if this error occurs, there is nothing we can do.
## Proposal
It seems like calling described methods w/o preliminary `EnableNotarySupport` call is a programmer error. To prevent this we can:
- provide all methods that require preliminary initialization with comments about this;
- panic if the requirement is not met.

## Alternative
It seems like `EnableNotarySupport` is not called more than once, so we can perform this action in the constructor instead of the public method (with the option enabled, or create a dedicated client constructor).

alexvanin:
Closes #445

### DIFF
--- a/pkg/morph/client/notary.go
+++ b/pkg/morph/client/notary.go
@@ -47,13 +47,11 @@ const (
 	notaryBalanceOfMethod = "balanceOf"
 	setDesignateMethod    = "designateAsRole"
 
-	notaryBalanceErrMsg = "can't fetch notary balance"
+	notaryBalanceErrMsg      = "can't fetch notary balance"
+	notaryNotEnabledPanicMsg = "notary support was not enabled on this client"
 )
 
-var (
-	errNotaryNotEnabled = errors.New("notary support was not enabled on this client")
-	errUnexpectedItems  = errors.New("invalid number of NEO VM arguments on stack")
-)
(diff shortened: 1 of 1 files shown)


pkg/morph/client/notary.go
```diff
@@ -48,12 +48,10 @@ const (
 	setDesignateMethod    = "designateAsRole"
 
 	notaryBalanceErrMsg = "can't fetch notary balance"
+	notaryNotEnabledMsg = "notary support was not enabled on this client"
```
carpawell, line 51: maybe add `Err` or `Panic` word in var name like in the previous var?

carpawell (approved):

pkg/morph/client/notary.go
alexvanin: Fixed.

PR #488, opened 2021-04-21
Feature/add intermediate router

carpawell:
1. Add(implement) intermediate router for EigenTrust algorithm. 
2. Refactor `neofs-node/cmd/reputation` through dividing it onto subpkgs.
3. Implement `SendIntermediateResult` of GRPC server.
4. Add writing `GlocalTrust` results to contract directly.


pkg/services/reputation/common/deps.go
```diff
@@ -29,7 +29,7 @@ type Writer interface {
 	// Close operation.
 	//
 	// Write must not be called after Close.
-	Write(reputation.Trust) error
+	Write(Context, reputation.Trust) error
```
cthulhu-rider, line 32: Why do we need  `Context`? We have `WriterProvider` for this purpose.

pkg/services/reputation/eigentrust/storage/daughters/calls.go
```diff
@@ -52,13 +53,44 @@ func (x *Storage) DaughterTrusts(epoch uint64, daughter reputation.PeerID) (*Dau
 	return s.daughterTrusts(daughter)
 }
 
+// AllDaughterTrusts returns daughter iterator for the epoch.
+//
+// Returns false if there is no data for the epoch and daughter.
+func (x *Storage) AllDaughterTrusts(epoch uint64) (*daughterStorage, bool) {
```
cthulhu-rider: Exported method returns value of unexported type?

cmd/neofs-node/reputation/intermediate/consumers.go
```diff
@@ -0,0 +1,53 @@
...
+
+type ConsumerTrustWriter struct {
+	log     *logger.Logger
+	storage *consumerstorage.Storage
+}
+
+func (w *ConsumerTrustWriter) Write(ctx common.Context, t reputation.Trust) error {
+	eiCtx, ok := ctx.(eigencalc.Context)
+	if !ok {
+		return ErrIncorrectContext
```
cthulhu-rider, line 35: Seems not good, why do we restrict context?

pkg/services/reputation/local/storage/calls.go
```diff
@@ -40,14 +41,17 @@ type trustValue struct {
 // EpochTrustValueStorage represents storage of
 // the trust values by particular epoch.
 type EpochTrustValueStorage struct {
+	reputationcommon.ServerInfo
```
cthulhu-rider, line 44: Why do we need this?

pkg/services/reputation/eigentrust/calculator/deps.go
```diff
@@ -59,3 +59,9 @@ type IntermediateWriter interface {
 type IntermediateWriterProvider interface {
 	InitIntermediateWriter(Context) (IntermediateWriter, error)
 }
+
+// AlphaProvider must provide information about required
+// alpha parameter for eigen trust algorithm.
+type AlphaProvider interface {
+	EigenTrustAlpha() (float64, error)
```
cthulhu-rider, line 76: There are no other alphas, it can be just `Value` or `Get`.

pkg/services/reputation/common/deps.go
carpawell: since we try to use one router for `local` and `intermediate` purposes and - on the other hand - we want to use our local(mean "that repository") `Trust` struct, I decided to have some generalized "context" of the execution. if it is for intermediate usage - it must contain `Epoch()` and `I()` data, and intermediate controller must pass it

pkg/services/reputation/eigentrust/storage/daughters/calls.go
carpawell: outdated, was fixed yesterday

cmd/neofs-node/reputation/intermediate/consumers.go
carpawell: because intermediate `Trust`s have no sense without `I()` and `Epoch()` "context"

pkg/services/reputation/local/storage/calls.go
carpawell: after the addition `trusting` peer to `Trust` struct there was bug related to `Trust`s without(zero) `trusting` peer in storages

in fact, i think that it is not so necessary in that case, but to prevent potential bugs in the future, i suggest creating all local `Trust` with trusting peer(via ServerInfo in that case) always

it is used at 165 line

pkg/services/reputation/eigentrust/calculator/deps.go
carpawell: in fact i wrote it in such way, cause i was almost sure that we would be glad to store `Alpha` in global config so i named the method as it is used to be named in global config getters(e.g. `EigenTrustIterations`)

cmd/neofs-node/config.go
```diff
@@ -308,9 +314,15 @@ type cfgControlService struct {
...
 type cfgReputation struct {
+	workerPool util2.WorkerPool // pool for EigenTrust algorithm's iterations
+
 	localTrustStorage *truststorage.Storage
 
 	localTrustCtrl *trustcontroller.Controller
+
+	scriptHash util.Uint160
+
+	rptWrapper *rptwrapper.ClientWrapper
```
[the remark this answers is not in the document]
carpawell: hmm, i did it like it was done in containers, but forget to add init in `reputation` file. i think it's because i dont know should it be stored in cfg struct or not

i may add initialization in `reputation` file or just delete it. which one is better?

cmd/neofs-node/reputation/intermediate/consumers.go
cthulhu-rider: So we try to implement more detailed interface through more abstracted one, don't you think this is incorrect?

pkg/services/reputation/common/deps.go
cthulhu-rider: Initially I asked why the `Context` in the signature when it is passed through `InitWriter` in expected way?

cmd/neofs-node/reputation/intermediate/consumers.go
carpawell: found no better solution

there must be always `Epoch()` method, but for intermediate `I()` is also required. and for local `Trust` `I()` means nothing and does not exist

i look at this solution in the same way as i look at ctx with `passedRoute` unnecessary field. i may be wrong

cmd/neofs-node/config.go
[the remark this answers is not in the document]
alexvanin: I don't see usage of reputation contract wrapper outside of reputation components, so I think the better way is to remove wrapper from config.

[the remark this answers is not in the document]
carpawell: ok, deleted

pkg/services/reputation/common/deps.go
carpawell: got your question wrong.

i think that ctxs for `InitWriter` and `Write` are too different logically and in general *theoretically* may differ in practice(e.g. `passedRoute` may be same(received from one node through the same route) but received trusts are for different iterations)

there may be too much optional information if passed all through one ctx(may be just std `ctx`, may contain `passedRoute`, may have `Epoch`, may have `Iteration` and all this is optional). found it too complex

carpawell: and, by the way, think it is more traditional to pass context to func that communicates via network

cmd/neofs-node/reputation/intermediate/contract.go
```diff
@@ -0,0 +1,146 @@
...
+	apiTrust.SetValue(t.Value().Float64())
+	apiTrust.SetPeer(apiTrustedPeerID)
+
+	var managerPublicKey [33]byte
+	copy(managerPublicKey[:], fw.pubKey)
+
+	apiMangerPeerID := apireputation.NewPeerID()
+	apiMangerPeerID.SetPublicKey(managerPublicKey)
+
+	gTrust := apireputation.GlobalTrust{}
```
alexvanin, line 91: Use `apireputation.NewGlobalTrust()` there

carpawell: changed

pkg/services/reputation/local/storage/calls.go
carpawell: deleted

pkg/services/reputation/common/deps.go
carpawell: moved all ctx logic to `Init...` funcs. make `Write` methods ctx insensitive

pkg/services/reputation/common/router/calls.go
```diff
@@ -94,7 +94,7 @@ func (w *trustWriter) Write(t reputation.Trust) error {
 				continue
 			}
 
-			remoteWriter, err = provider.InitWriter(w.ctx)
+			remoteWriter, err = provider.InitWriter(w.ctx.Context)
```
cthulhu-rider, line 97: Just `w.ctx`

cthulhu-rider (approved):
General picture of the code is legitimate, I propose to merge it and edit it on the spot while debugging. @alexvanin

pkg/services/reputation/common/router/calls.go
carpawell: added commentary why this is necessary and `TODO` to think if the router can be done general without context cast

alexvanin (approved):

PR #489, opened 2021-04-21
cli: support NEP-2 and NEP-6 key formats

### LINKED ISSUE #466: Support NEP-6 wallet in neo-cli
NeoFS CLI requires private key in open WIF format to operate, like:
```
$ neofs-cli -r st1.storage.fs.neo.org:8080 -k L2j3Lqx6rJyBiyxnTUHr6TkwMZQKUTuUChsZBLmbAxYa5Rou2TqH object put --file cat.png --cid Ec1fAsQbstaVbUHboEEgtFNguCpyBKfTqhcV7NmjVSoJ --attributes img_type=cat,my_attr=cute
```
which is neither convenient nor secure. Keys are usually stored in wallets and if we're talking Neo here, NEP-6 wallets like the ones supported by neo-go.

Add support for NEP-6 wallets, `-w` parameter for wallet path specification (maybe also `-a` for address as there can be many keys stored in a wallet) and ask for password dynamically to unlock the key.

**Describe alternatives you've considered**
At the very minimum it'd be nice to support NEP-2 encrypted private keys for `-k`, but NEP-6 wallets are still preferred.

**Additional context**
https://neospcc.medium.com/neofs-n3-testnet-release-and-first-episode-of-neofs-beginners-guide-54d618e1c86b

fyrchik:
Close #466 .

NEP-2 is contains WIF encrypted with password.
NEP-6 is format used for neo-go wallets.

### DIFF
--- a/cmd/neofs-cli/modules/accounting.go
+++ b/cmd/neofs-cli/modules/accounting.go
@@ -34,17 +34,17 @@ var accountingBalanceCmd = &cobra.Command{
 			ctx = context.Background()
 		)
 
-		cli, err := getSDKClient()
+		key, err := getKey()
+		if err != nil {
+			return err
+		}
+
+		cli, err := getSDKClient(key)
 		if err != nil {
 			return err
 		}
 
 		if balanceOwner == "" {
-			key, err := getKey()
-			if err != nil {
-				return err
-			}
-
 			wallet, err := owner.NEO3WalletFromPublicKey(&key.PublicKey)
 			if err != nil {
 				return err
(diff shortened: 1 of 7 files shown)


alexvanin:
Can we offload some code to neofs-crypto/load.go? I guess other client applications like gates will benefit from it. 
/cc @realloc @roman-khimov

alexvanin (changes requested):
`getKey()`  now might require user input so it can't be called more than once per command. Unfortunately it is not the case right now. Some commands like `container list` or `accounting balance` call `getSDKClient()` with `getKey()` and then explicitly call `getKey()` again. 

Maybe it won't be that bad to store `key` as a global variable in modules, so it will be initialized once and reused in all commands then. 

Edit 1: or provide key to `getSDKClient` explicitly.

alexvanin (approved):

PR #499, opened 2021-04-29
object/search: Re-sign original requests during forwarding

cthulhu-rider:
In previous implementation node's Object Search V2 service handler created a
new request for each RPC. Now original requests are re-signed according to
API specification. Logical handler abstracts from this version-dependent
logic through `RequestForwarder` callback.

### DIFF
--- a/pkg/services/object/get/v2/util.go
+++ b/pkg/services/object/get/v2/util.go
@@ -1,13 +1,22 @@
 package getsvc
 
 import (
+	"context"
 	"crypto/sha256"
 	"hash"
+	"io"
+	"sync"
 
+	"github.com/nspcc-dev/neofs-api-go/pkg/client"
 	objectSDK "github.com/nspcc-dev/neofs-api-go/pkg/object"
 	"github.com/nspcc-dev/neofs-api-go/pkg/token"
+	rpcclient "github.com/nspcc-dev/neofs-api-go/rpc/client"
+	signature2 "github.com/nspcc-dev/neofs-api-go/util/signature"
 	objectV2 "github.com/nspcc-dev/neofs-api-go/v2/object"
 	"github.com/nspcc-dev/neofs-api-go/v2/refs"
+	"github.com/nspcc-dev/neofs-api-go/v2/rpc"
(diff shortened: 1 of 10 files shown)


pkg/services/object/get/v2/util.go
```diff
@@ -66,6 +172,77 @@ func (s *Service) toRangePrm(req *objectV2.GetRangeRequest, stream objectSvc.Get
...
+			// code below is copy-pasted from c.ObjectPayloadRangeData implementation,
+			// perhaps it is worth highlighting the utility function in neofs-api-go
+
+			// open stream
+			stream, err := rpc.GetObjectRange(c.Raw(), req, rpcclient.WithContext(stream.Context()))
+			if err != nil {
+				return nil, errors.Wrap(err, "could not create Get payload range stream")
+			}
+
+			payload := make([]byte, body.GetRange().GetLength())
```
alexvanin, line 207:
```suggestion
			payload := make([]byte, 0, body.GetRange().GetLength())
```

cthulhu-rider: Need to be fixed in api-go too

pkg/services/object/search/util.go
```diff
@@ -76,9 +76,7 @@ func (c *clientConstructorWrapper) get(addr string) (searchClient, error) {
 }
 
 func (c *clientWrapper) searchObjects(exec *execCtx) ([]*objectSDK.ID, error) {
-	return c.client.SearchObject(exec.context(),
-		exec.remotePrm(),
-		exec.callOptions()...)
```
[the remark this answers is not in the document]
cthulhu-rider: Done

alexvanin (approved):
LGTM. 

Even though we can try to adopt such behavior for `object.Put` I don't think it will be as useful as in other object requests. `object.Put` request might be resend only when object in it is complete: it has all required fields and valid signatures. 

Most of the time node gets such requests when it need to store object locally. Resending complete object is a rare case right now and session token covers it. However this might change in the future, thus we need to investigate the way session token can be dropped for complete object (#501).

As for now I suggest to merge this PR and fix possible issues (especially with ACL) as soon as test-case repo will be able to work again.

Issue #493, opened 2021-04-26
Refactor storage application


alexvanin:
Storage application now is a bit of mess right now with global state, configuration params and service implementation all in the main package. Refactor should consider:
- keeping main package small,
- moving implementations to a separate packages,
- figuring out the way state of application is stored (`cfg` structure), 
- figuring out the way configuration is read and set up with default values.

alexvanin:
After internal discussion we've decided to keep overall structure as it is for a while, because storage apps requires some bigger changes with internal balancing. So we can postpone application structure reorganization untill then. 

In this issue we want to organize all configuration values in config package and use it in app.

alexvanin:
We finished config refactor at storage node as a first step to reorganize application structure. Next changes will be described in different issues.

Closed by #553 #573, #575, #576, #577, #578, #579, #581

PR #512, opened 2021-05-11
Refactor serving of prometheus and pprof services

cthulhu-rider:
Relates to #493.


cmd/neofs-ir/main.go
```diff
@@ -57,8 +59,33 @@ func main() {
 	ctx := grace.NewGracefulContext(log)
 	intErr := make(chan error) // internal inner ring errors
 
-	pprof := profiler.NewProfiler(log, cfg)
-	prometheus := profiler.NewMetrics(log, cfg)
+	var httpServers []*httputil.Server
+
+	for _, item := range []struct {
```
alexvanin, line 64: Encapsulate filling `httpServers` in separate function down below. This will make main function a bit smaller and cleaner. Something like
```go
func initHTTPservers(cfg viper.Viper) []*http.util.Server {
 . . .
}
```

cmd/neofs-ir/defaults.go
```diff
@@ -32,13 +32,9 @@ func newConfig(path string) (*viper.Viper, error) {
 func defaultConfiguration(cfg *viper.Viper) {
 	cfg.SetDefault("logger.level", "info")
 
-	cfg.SetDefault("pprof.enabled", false)
-	cfg.SetDefault("pprof.address", ":6060")
-	cfg.SetDefault("pprof.shutdown_ttl", "30s")
+	cfg.SetDefault("profiler.shutdown_timeout", "30s")
```
alexvanin, line 35: Please add `cfg.SetDefault("profiler.address", "")` and `cfg.SetDefault("metrics.address", "")` for inner ring app.

We discussed it already in the context of alphabet contracts defaults. However in this case there are only two meaningful config records. I want to keep all available configuration values in one place as mush as it possible.

alexvanin (changes requested):

cmd/neofs-ir/main.go
cthulhu-rider: Done.

We can do the same for serve/shutdown, although their code is much smaller.

cmd/neofs-ir/defaults.go
cthulhu-rider: Done.

alexvanin (approved):

Issue #455, opened 2021-04-02
Add TLS encryption for gRPC servers


alexvanin:
Right now nodes and clients talk to each other over unsecured connection. gRPC, as the only supported transport protocol right now, supports TLS to secure messages, thus we should support it.

neofs-storage configuration should provide path to `.pem` and `.key` files and use it as TLS credentials.

alexvanin:
The issue is a bit more complex that it seems. 

## Server side

### Configuration
Storage node serves gRPC server with NeoFS API services. Node will have config options to provide TLS related files: certificate and private key.

```yaml
grpc:
  tls:
    enabled: true
    cert: cert.pem
    key: privkey.pem
```

Then, if TLS is enabled, these files will be used in gRPC server constructor.

```go
	// cmd/neofs-node/grpc.go
	var serverOpts = []grpc.ServerOption{
		grpc.MaxSendMsgSize(maxMsgSize),
	}

	if c.cfgRPC.tlsEnabled {
		creds, err := credentials.NewServerTLSFromFile(c.cfgRPC.tlsCertPath, c.cfgRPC.tlsKeyPath)
		if err != nil {
			fatalOnErr(err)
		}
		
		serverOpts = append(serverOpts, grpc.Creds(creds))
	}

	c.cfgGRPC.server = grpc.NewServer(serverOpts...)

```

### Bootstrap address

Storage node must announce TLS support in the address stored in network map. This way any node or client can look at the address and connect to node without additional protocol negotiation. We use [multiaddr](https://github.com/multiformats/multiaddr) to encode node address. The simplest way to do that is to add `/tls` protocol into address,:
```
TLS Enabled:

/ip4/77.88.55.80/tcp/8080/tls
/dns4/example.com/tcp/16512/tls

TLS Disabled:

/ip4/77.88.55.80/tcp/8080
/dns4/example.com/tcp/8080
```

Unfortunately specification does not support TLS protocol yet, even though `/tls/` included in the examples and supported by related projects ([multicodec](https://github.com/multiformats/multicodec)). To resolve this issue we can fork `go-multiaddr` repository, add TLS protocol, use it in storage node and then try to push this change into upstream.

### Summary

- [ ] Support new config options for TLS
- [ ] Enable TLS in gRPC server
- [ ] Add `TLS` protocol support to `go-multiaddr` fork
- [ ] Use `/tls` in announced multiaddr at bootstrap

## Client side

### Protocol

[...]

PR #522, opened 2021-05-17
Use host addresses as transport endpoints

alexvanin:
Related to https://github.com/nspcc-dev/neofs-node/issues/455

To enable TLS support we can't operate with IP addresses directly. Certificates are issued with host names so it is required to
pass them into RPC client. DNS resolving should be done by transport layer and not be a part of node. Therefore `IPAddrString` usage is removed from code.

Also there is fix for incorrect port-only host address parsing (`:8080`). This leads to `/dns4/tcp/8080` multiaddress which is incorrect. In this case `dns4` section should be omitted, but it breaks `manet.DialArgs`. To solve this issue we explicitly set 0.0.0.0 address.

### DIFF
--- a/pkg/network/address.go
+++ b/pkg/network/address.go
@@ -83,6 +83,13 @@ func multiaddrStringFromHostAddr(host string) (string, error) {
 		return "", err
 	}
 
+	// Empty address in host `:8080` generates `/dns4//tcp/8080` multiaddr
+	// which is invalid. It could be `/tcp/8080` but this breaks
+	// `manet.DialArgs`. The solution is to manually parse it as 0.0.0.0
+	if endpoint == "" {
+		return "/ip4/0.0.0.0/tcp/" + port, nil
+	}
+
 	var (
 		prefix = "/dns4"
 		addr   = endpoint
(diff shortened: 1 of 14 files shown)


pkg/network/address.go
```diff
@@ -83,6 +83,13 @@ func multiaddrStringFromHostAddr(host string) (string, error) {
 		return "", err
 	}
 
+	// Empty address in host :8080` generates `/dns4//tcp/8080` multiaddr
+	// which is invalid. It could be `/tcp/8080` but this breaks
+	// `manet.DialArgs`. The solution is to manually parse it as 0.0.0.0
+	if endpoint == "" {
+		endpoint = "0.0.0.0"
```
fyrchik, line 90: We can return from here immediately, without further parsing.

alexvanin: Yeah, fixed, thanks.

pkg/network/address.go
```diff
@@ -73,6 +83,13 @@ func multiaddrStringFromHostAddr(host string) (string, error) {
 		return "", err
 	}
 
+	// Empty address in host :8080` generates `/dns4//tcp/8080` multiaddr
```
carpawell, line 86: if i understand comment correctly, there is missing opening backtick before `:8080`

carpawell (approved):

PR #527, opened 2021-05-17
fix/EigenTrust-algorithm

carpawell:
Add normalizing to `localTrust`s. Add `break` in node search cycle if a node's been found.

Run `go mod tidy`.


alexvanin:
Can you do two commits, one with changes and one with `go mod tidy`?

carpawell:
divided commit onto two

alexvanin (approved):

PR #530, opened 2021-05-18
[#502] innerring: synchronize validators on mainnet alphabet update

### LINKED ISSUE #502: Syncronize side chain validators on designate notification
After https://github.com/nspcc-dev/neo-go/pull/1938 role management contract generates notifications. Inner ring node should use it as a trigger to synchronize side chain validators instead of new epoch tick.

fyrchik:
Close #502 .

### DIFF
--- a/pkg/innerring/innerring.go
+++ b/pkg/innerring/innerring.go
@@ -490,6 +490,10 @@ func New(ctx context.Context, log *zap.Logger, cfg *viper.Viper) (*Server, error
 		}
 	} else {
 		alphaSync = governanceProcessor.HandleAlphabetSync
+		err = bindMainnetProcessor(governanceProcessor, server)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	// create netmap processor
--- a/pkg/innerring/processors/governance/handlers.go
+++ b/pkg/innerring/processors/governance/handlers.go
@@ -1,12 +1,29 @@
 package governance
 
 import (
+	"github.com/nspcc-dev/neo-go/pkg/core/native"
+	"github.com/nspcc-dev/neo-go/pkg/core/native/noderoles"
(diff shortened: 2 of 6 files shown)


cthulhu-rider (approved):

alexvanin:
Will merge it after RC2.

pkg/innerring/processors/governance/events.go
```diff
@@ -9,3 +17,25 @@ func (s Sync) MorphEvent() {}
 func NewSyncEvent() Sync {
 	return Sync{}
 }
+
+// Designate represents designation event of the mainnet RoleManagement contract.
+type Designate struct {
```
alexvanin, line 22: I would recommend to move designate event to `morph/event/rolemanagement` package. Events in processor sub packages are tend to be internal, and designate event produced by external native contract.

pkg/innerring/processors/governance/events.go
```diff
@@ -9,3 +17,25 @@ func (s Sync) MorphEvent() {}
 func NewSyncEvent() Sync {
```
alexvanin, line 17: process_epoch.go still generates internal event to sync validators on each epoch. Maybe it is redundant because of RoleManagement contract notifications, what do you think? 

```go
// pkg/innerring/processors/netmap/process_epoch.go
func (np *Processor) processNewEpoch(epoch uint64) {
	...
	np.handleAlphabetSync(governance.NewSyncEvent())
}
```

alexvanin: It is not redundant, because committee changes are limited by `1/3n-1` of new keys. Thus it may require several synchronization rounds, however RoleManagement contract produces notification only once.

alexvanin (approved):

PR #541, opened 2021-05-19
[#XX] blobstor/fstree: fix bug in iterate

fyrchik:
Be able to recover address from the path. Also add tests.


alexvanin:
Use of stable marshaled address as a filename breaks original idea to distribute files in the filesystem for two reasons:
- container ID placement in the address,
- stable marshaling scheme.

Container ID takes first place in address encoding. Therefore all objects of one container will have the same prefix and will be stored in the one directory. 

Marshaling scheme makes it even worse. Any address encoded in protobuf scheme will have the same first four bytes that encode container ID filed type and size. _All_ objects will be stored in single directory. 

```
neofs.fstree.test$ tree
.
└── 0a
    └── 22
        ├── 0a202234d9ceaee6f057f20f463c9678414495e0bd96e725d9da0639a3f6b1457ef412220a201747a41871aea727311ebbc9e52b5759439a0a1d124bcfcee8286ba326786367
        ├── 0a2026e3c5a618636a84c60ed3a4a9c1bbefa4473a0db06a58396e1675336129ed2612220a203a7c469c005539d7ff6de07dade2f0a3d5542aefdf548419866de6807e16f350
        ├── 0a205b4c98eecb111820a3bdc810b470d222288b985429c0f643acd5cecaf251e80012220a2027817f24b8a99ded93b928971b515827b12444303f5cb77617a60e560e24f7cf
        ├── 0a2079a425e98456a0f89f541eccc0f782a77e8ed79282cc51c8f76abf05e4f6746612220a207013a6f639ca767afb43af09564742925f37c38415109a6eaadc79dcfa1cc624
        ├── 0a208f6f1f3707fa7d8221fdfb7169b588676d9fd90531a7879a95a686887f1e7a2b12220a204351feef51b2fc94d0e56b554db32e9c68d11e330a369e1748ea2994207816fc
        ├── 0a20a757387ecd081187edea184baa8fe94438f8b891e927c1bdacdd7411004651d612220a2077cec29e87fe2ca876daccc951ef07507b5e1c9ce10d7927cbfbe05548b96d90
        ├── 0a20b37b72874ee6c129ea60b4867cf299aafe307822eef9b7ab862ffede14f400cb12220a203b9b9d3665639b8d4dbcf33aea123a3f880728415a3d99623eb941d8fbeb8f2e
        ├── 0a20b58c06a78f2649475cb326ef82ae57c6b757121d18124f4f47d4a0ddfec267d612220a203c3890b9ab2b2bae5fefba1d61750aa75d359cd33e92a9982bea74a6efe2655f
        └── 0a20c928751ff89ebdc8f5921cd227c028e7afe29682b2064560aa98c7a4ba16540b12220a20ae6ffadf9de80dcac10db4fd5c36bd5bbe31845357c81c3abebce71190dafefd

```

I think we can merge this as a bug fix, but then we should come up with different file name encoding. 
For example we can drop protobuf encoding and use filenames with object ID encoded first.

realloc:
I'd suggest to use only `ObjectID` or start the slicing for shallow tree directory names from the tail.

alexvanin (approved):

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -40,9 +40,30 @@ const (
 var ErrFileNotFound = errors.New("file not found")
 
 func stringifyAddress(addr *objectSDK.Address) string {
-	h := sha256.Sum256([]byte(addr.String()))
+	return addr.ObjectID().String() + "." + addr.ContainerID().String()
+}
+
+func addressFromString(s string) (*objectSDK.Address, error) {
```
realloc, line 46: May it be simpler to reverse the address string? It will be less code to understand, and maybe it could work faster a little.

PR #549, opened 2021-05-21
Feature/add TLS encryption

### LINKED ISSUE #455: Add TLS encryption for gRPC servers

carpawell:
Add TLS(turns on with config(env vars)) to GRPC server and client.

In `network` pkg add implementation of `tls` protocol for `multiaddr` lib.

Make CLI support multiaddresess with `/tls` protocol.

Closes #455.

### DIFF
--- a/cmd/neofs-node/config.go
+++ b/cmd/neofs-node/config.go
@@ -354,6 +354,10 @@ func initCfg(path string) *cfg {
 		tlsKeyFile = viperCfg.GetString(cfgTLSKeyFile)
 	}
 
+	if tlsEnabled {
+		netAddr.AddTLS()
+	}
+
 	state := newNetworkState()
 
 	containerWorkerPool, err := ants.NewPool(notificationHandlerPoolSize)
--- a/cmd/neofs-node/config.go
+++ b/cmd/neofs-node/config.go
@@ -68,6 +68,9 @@ const (
 
 	// config keys for cfgGRPC
 	cfgListenAddress = "grpc.endpoint"
+	cfgTLSEnabled    = "grpc.tls.enabled"
+	cfgTLSCertFile   = "grpc.tls.certificate"
+	cfgTLSKeyFile    = "grpc.tls.key"
 
 	// config keys for API client cache
 	cfgAPIClientDialTimeout = "apiclient.dial_timeout"
@@ -206,6 +209,10 @@ type cfgGRPC struct {
 	maxChunkSize uint64
 
 	maxAddrAmount uint64
+
+	tlsEnabled  bool
+	tlsCertFile string
+	tlsKeyFile  string
 }
 
 type cfgMorph struct {
@@ -335,6 +342,18 @@ func initCfg(path string) *cfg {
 	maxChunkSize := uint64(maxMsgSize) * 3 / 4          // 25% to meta, 75% to payload
 	maxAddrAmount := uint64(maxChunkSize) / addressSize // each address is about 72 bytes
 
+	var (
+		tlsEnabled  bool
+		tlsCertFile string
+		tlsKeyFile  string
+	)
+
+	if viperCfg.GetBool(cfgTLSEnabled) {
+		tlsEnabled = true
+		tlsCertFile = viperCfg.GetString(cfgTLSCertFile)
+		tlsKeyFile = viperCfg.GetString(cfgTLSKeyFile)
+	}
+
 	state := newNetworkState()
 
 	containerWorkerPool, err := ants.NewPool(notificationHandlerPoolSize)
@@ -377,6 +396,9 @@ func initCfg(path string) *cfg {
 		cfgGRPC: cfgGRPC{
 			maxChunkSize:  maxChunkSize,
 			maxAddrAmount: maxAddrAmount,
+			tlsEnabled:    tlsEnabled,
+			tlsCertFile:   tlsCertFile,
+			tlsKeyFile:    tlsKeyFile,
 		},
 		localAddr: netAddr,
 		respSvc: response.NewService(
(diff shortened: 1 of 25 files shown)


pkg/innerring/rpc.go
```diff
@@ -136,14 +136,14 @@ func (c *ClientCache) GetHeader(task *audit.Task, node *netmap.Node, id *object.
...
-	addr, err := network.HostAddrFromMultiaddr(node.Address())
+	addr, err := network.AddressFromString(node.Address())
 	if err != nil {
 		return nil, fmt.Errorf("can't parse remote address %s: %w", node.Address(), err)
 	}
 
 	cli, err := c.Get(addr)
 	if err != nil {
-		return nil, fmt.Errorf("can't setup remote connection with %s: %w", addr, err)
+		return nil, fmt.Errorf("can't setup remote connection with %s: %w", addr.String(), err)
```
alexvanin, line 146: I don't think you need to manually call `String()` method when argument is substituted to `%s` modifier.

pkg/network/cache/client.go
```diff
@@ -27,9 +30,19 @@ func NewSDKClientCache(opts ...client.Option) *ClientCache {
 }
 
 // Get function returns existing client or creates a new one.
-func (c *ClientCache) Get(address string) (client.Client, error) {
+func (c *ClientCache) Get(netAddr *network.Address) (client.Client, error) {
+	hostAddr, err := netAddr.HostAddrString()
```
alexvanin, line 34: Since `hostAddr` is not used as key for cache map, you can move this code down near `opts := append(c.opts, client.WithAddress(hostAddr))`

pkg/network/tls.go
```diff
@@ -0,0 +1,48 @@
+package network
+
+import (
+	"fmt"
+
+	"github.com/multiformats/go-multiaddr"
+)
+
+// There is realisation of TLS protocol for
```
alexvanin, line 9:
```suggestion
// There is implementation of TLS protocol for
```

pkg/services/object/head/remote.go
```diff
@@ -71,7 +71,7 @@ func (h *RemoteHeader) Head(ctx context.Context, prm *RemoteHeadPrm) (*object.Ob
 		return nil, err
 	}
 
-	c, err := h.clientCache.Get(addr)
+	c, err := h.clientCache.Get(prm.node)
```
alexvanin, line 74: Maybe we need to reconsider do we need `addr` there at all? I see in this function

```go
	addr, err := prm.node.HostAddrString()
	if err != nil {
		return nil, err
	}
        ...
	p := new(client.ObjectHeaderParams).
		WithAddress(prm.commonHeadPrm.addr).
```

Looks a bit weird why other components need host address.

alexvanin (changes requested):

alexvanin:
Can you point out where `/tls` protocol is added to multiaddress in `NodeInfo` structure, that is announced at bootstrap?

pkg/services/object/head/remote.go
carpawell: hm, ok, it is only used in building error message, so deleted retrieving `HostAddr` string and start using `prm.node` instead

pkg/network/tls.go
carpawell: fixed

pkg/network/cache/client.go
carpawell: moved

cmd/neofs-node/container.go
```diff
@@ -229,7 +229,7 @@ func (r *remoteLoadAnnounceProvider) InitRemote(srv loadroute.ServerInfo) (loadc
 		return loadcontroller.SimpleWriterProvider(new(nopLoadWriter)), nil
 	}
 
-	hostAddr, err := network.HostAddrFromMultiaddr(addr)
+	hostAddr, err := network.AddressFromString(addr)
```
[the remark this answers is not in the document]
carpawell: agree, changed everywhere to `netAddr`

cmd/neofs-node/config.go
```diff
@@ -335,6 +342,18 @@ func initCfg(path string) *cfg {
...
+		tlsCertFile string
+		tlsKeyFile  string
+	)
+
+	if viperCfg.GetBool(cfgTLSEnabled) {
+		tlsEnabled = true
+		tlsCertFile = viperCfg.GetString(cfgTLSCertFile)
+		tlsKeyFile = viperCfg.GetString(cfgTLSKeyFile)
+	}
+
```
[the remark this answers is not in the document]
carpawell: did not want to run `viperCfg.GetString` if it's useless(when TLS is off), but can do it like you wrote, or even without `var`

carpawell:
> Can you point out where `/tls` protocol is added to multiaddress in `NodeInfo` structure, that is announced at bootstrap?

Did not get the whole idea, so have just added it now

pkg/services/object/put/remote.go
```diff
@@ -54,14 +54,14 @@ func (t *remoteTarget) Close() (*transformer.AccessIdentifiers, error) {
 		return nil, fmt.Errorf("(%T) could not receive private key: %w", t, err)
 	}
 
-	addr, err := t.addr.HostAddrString()
+	hostAddr, err := t.addr.HostAddrString()
```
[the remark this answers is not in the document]
carpawell: agree. fixed and also checked that there are no more such cases

alexvanin (approved):

PR #553, opened 2021-05-21
node: Refactor usage of configuration values

cthulhu-rider:
Relates to #493.

Package `config` with basic functionality can be reused for IR app later.

Usage of refactored config is used for logger only to start. Global transition will be made in subsequent PRs.


cmd/neofs-node/config/cast.go
```diff
@@ -0,0 +1,49 @@
...
+// Returns nil if value can not be casted.
+func StringSliceSafe(c *Config, name string) []string {
+	return cast.ToStringSlice(c.Value(name))
+}
+
+// String reads configuration value
+// from c by name and casts it to string.
+//
+// Panics if value can not be casted.
+func String(c *Config, name string) string {
```
alexvanin, line 36: I propose to:
- remove unsafe converters until we need one,
- name safe converters without `Safe` prefix.

cthulhu-rider: `<Type>` methods will be used for mandatory configurations, and `<Type>Safe` - for optional.

alexvanin: Type casting is not connected with requirements, isn't it? If some mandatory field is missing, then `<Type>` function won't panic. Did I get the idea wrong?

```
func TestString(t *testing.T) {
	configtest.ForEachFileType("test/config", func(c *config.Config) {
		require.Panics(t, func() {
			config.String(c, "mandatory_value_that_missing") // no panic there
		})
	})
}
```

cthulhu-rider: It will panic if value is not of the particular type (e.g. `String` panics if config can no be interpreted as `string`).

alexvanin (approved):

PR #561, opened 2021-05-26
Traverse request meta headers to fetch tokens in ACL checker

### LINKED ISSUE #548: Issue with operation with allow bearer token for object operations (object not found)
NeoFS version 0.19.0-42-gb587b23 (latest master)

The issue with GET operation with allow bearer token.
Detailed log: https://drive.google.com/file/d/1Zv1GoAfeieWfmmUH3rcGi8s9C5AUXv9H/view?usp=sharing
Can be opened after download as .HTML file.

alexvanin:
Fixes #548 

When storage node re-sends requests, it creates new layer in request meta header chain. New layer does not contain tokens but container link to the original request meta header. ACL service should traverse to the original request meta header to fetch relevant session and bearer tokens.

### DIFF
--- a/pkg/services/object/acl/acl_test.go
+++ b/pkg/services/object/acl/acl_test.go
@@ -1,0 +1,36 @@
+package acl
+
+import (
+	"testing"
+
+	"github.com/nspcc-dev/neofs-api-go/v2/acl"
+	acltest "github.com/nspcc-dev/neofs-api-go/v2/acl/test"
+	"github.com/nspcc-dev/neofs-api-go/v2/session"
+	sessiontest "github.com/nspcc-dev/neofs-api-go/v2/session/test"
+	"github.com/stretchr/testify/require"
+)
+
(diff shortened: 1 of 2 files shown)


pkg/services/object/acl/acl_test.go
```diff
@@ -0,0 +1,47 @@
...
+	sToken := new(session.SessionToken)
+	sToken.SetBody(sTokenBody)
+
+	owner := new(refs.OwnerID)
+	owner.SetValue([]byte("Owner ID"))
+
+	bTokenBody := new(acl.BearerTokenBody)
+	bTokenBody.SetOwnerID(owner)
+
+	bToken := new(acl.BearerToken)
```
cthulhu-rider, line 25: Now we can just use `sessiontest.GenerateBearerToken` and `acltest.GenerateBearerToken`

cthulhu-rider (approved):

PR #565, opened 2021-05-28
Finalize container sessions 

### LINKED ISSUE #525: Check container session token in container processor
After https://github.com/nspcc-dev/neofs-api-go/issues/283 and https://github.com/nspcc-dev/neofs-contract/issues/76 there will be optional session token in container and EACL bodies. Container processor must verify by checking:
- if container owner and token owner are the same,
- if token signed by the owner,
- if container or EACL signed itself by session key inside the token.

If all three checks are OK, then approve operation.

### LINKED ISSUE #560: Support container session tokens in NeoFS CLI
After #525 applications can create containers on behalf of user with session tokens. CLI can provide an example how to do that with updated api-go.

I propose to add `[--session-key <privatekey> | new ]` argument in container create command of NeoFS CLI. With this argument CLI will create container on user's behalf but sign with provided session key. If `new` keyword specified instead of the key, then it key generated in runtime.

Example:
```
$ neofs-cli -k user.key -r s01.neofs.devenv:8080 container create --policy rule.ql --session-key new --await
```

cthulhu-rider:
Closes #525.
Closes #560.

### DIFF
--- a/pkg/innerring/processors/container/common.go
+++ b/pkg/innerring/processors/container/common.go
@@ -8,10 +8,17 @@ import (
 	"fmt"
 
 	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
+	cid "github.com/nspcc-dev/neofs-api-go/pkg/container/id"
 	"github.com/nspcc-dev/neofs-api-go/pkg/owner"
 	"github.com/nspcc-dev/neofs-api-go/pkg/session"
 	"github.com/nspcc-dev/neofs-api-go/util/signature"
 	signature2 "github.com/nspcc-dev/neofs-api-go/v2/signature"
+)
+
+var (
+	errWrongSessionContext = errors.New("wrong session context")
+	errWrongSessionVerb    = errors.New("wrong token verb")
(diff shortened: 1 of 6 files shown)


pkg/innerring/processors/container/common.go
```diff
@@ -125,3 +115,37 @@ func (cp *Processor) checkSessionToken(token *session.Token) error {
...
+}
+
+func checkTokenContextWithCID(tok *session.Token, id *cid.ID, verbAssert verbAssert) error {
+	c, err := contextWithVerifiedVerb(tok, verbAssert)
+	if err != nil {
+		return err
+	}
+
+	tokCID := c.Container()
+	if tokCID != nil && !tokCID.Equal(id) {
```
alexvanin, line 146: Is it valid to have `tokCID == nil`?

alexvanin (approved):

pkg/innerring/processors/container/common.go
cthulhu-rider: Yes, it is, here `tokCID == nil` if token is signed with wildcard flag.

PR #567, opened 2021-05-31
node: Close cached clients

cthulhu-rider:


pkg/network/cache/client.go
```diff
@@ -75,3 +75,21 @@ func (c *ClientCache) Get(netAddr *network.Address) (client.Client, error) {
...
+// Ignores closing errors.
+func (c *ClientCache) CloseAll() {
+	c.mu.RLock()
+
+	{
+		for _, cl := range c.clients {
+			con := cl.Conn()
+			if con != nil {
+				_ = con.Close()
+			}
```
alexvanin, line 90: Don't you want to delete map records after connection close? I don't see how they can be reused after that.

cthulhu-rider: Lets for now just close: cache cleaning is not implemented in latest implementation, and the main point of the PR is to close clients.

PR #568, opened 2021-05-31
local_object_storage/shard/writecache: maintain size 

ZhangTao1596:

I think we should reduce memory size when objects removed from memory?

### DIFF
--- a/pkg/local_object_storage/writecache/delete.go
+++ b/pkg/local_object_storage/writecache/delete.go
@@ -19,6 +19,7 @@ func (c *cache) Delete(addr *objectSDK.Address) error {
 		if saddr == c.mem[i].addr {
 			copy(c.mem[i:], c.mem[i+1:])
 			c.mem = c.mem[:len(c.mem)-1]
+			c.curMemSize -= uint64(len(c.mem[i].data))
 			c.mtx.Unlock()
 			return nil
 		}
@@ -26,18 +27,24 @@ func (c *cache) Delete(addr *objectSDK.Address) error {
 	c.mtx.Unlock()
 
 	// Check disk cache.
-	has := false
+	var has int
 	_ = c.db.View(func(tx *bbolt.Tx) error {
 		b := tx.Bucket(defaultBucket)
-		has = b.Get([]byte(saddr)) != nil
+		has = len(b.Get([]byte(saddr)))
 		return nil
 	})
 
-	if has {
-		return c.db.Update(func(tx *bbolt.Tx) error {
+	if 0 < has {
+		err := c.db.Update(func(tx *bbolt.Tx) error {
 			b := tx.Bucket(defaultBucket)
-			return b.Delete([]byte(saddr))
+			err := b.Delete([]byte(saddr))
+			return err
 		})
+		if err != nil {
+			return err
+		}
+		c.dbSize.Sub(uint64(has))
+		return nil
 	}
 
 	err := c.fsTree.Delete(addr)
--- a/pkg/local_object_storage/writecache/persist.go
+++ b/pkg/local_object_storage/writecache/persist.go
@@ -31,19 +31,13 @@ func (c *cache) persistLoop() {
 				zap.Int("total", len(m)))
 
 			c.mtx.Lock()
+			c.curMemSize = 0
 			n := copy(c.mem, c.mem[len(m):])
 			c.mem = c.mem[:n]
 			for i := range c.mem {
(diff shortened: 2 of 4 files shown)


pkg/local_object_storage/writecache/delete.go
```diff
@@ -20,6 +20,7 @@ func (c *cache) Delete(addr *objectSDK.Address) error {
 			copy(c.mem[i:], c.mem[i+1:])
 			c.mem = c.mem[:len(c.mem)-1]
 			c.mtx.Unlock()
+			c.curMemSize -= uint64(len(c.mem[i].data))
```
fyrchik, line 23: This should be done with mutex taken (before `c.mtx.Unlock`).

pkg/local_object_storage/writecache/persist.go
```diff
@@ -26,6 +26,9 @@ func (c *cache) persistLoop() {
 
 			start := time.Now()
 			c.persistObjects(m)
+			for i := range m {
```
fyrchik, line 29: It is calculated at about 5 lines below, we need just to set it to 0 (again, with mutex).

pkg/local_object_storage/writecache/delete.go
```diff
@@ -27,17 +27,22 @@ func (c *cache) Delete(addr *objectSDK.Address) error {
 	c.mtx.Unlock()
 
 	// Check disk cache.
-	has := false
+	var has []byte
 	_ = c.db.View(func(tx *bbolt.Tx) error {
 		b := tx.Bucket(defaultBucket)
-		has = b.Get([]byte(saddr)) != nil
+		has = b.Get([]byte(saddr))
```
fyrchik, line 33: The returned value of `Get` is only valid during tx execution. We need to remember it's length instead (`var has bool` -> `var sz int`) and compare it with 0.

pkg/local_object_storage/writecache/persist.go
```diff
@@ -71,7 +64,7 @@ func (c *cache) persistToCache(objs []objectInfo) []int {
 			if err != nil {
 				return err
 			}
-
+			c.dbSize.Add(uint64(len(objs[i].data)))
```
fyrchik, line 67: It is not a good idea to do this inside `db.Update`, because if tx fails it will be rollbacked and we will end up with an incorrect counter. So we better do this in a local variable and update `dbSize` after `db.Update`.

pkg/local_object_storage/writecache/storage.go
```diff
@@ -99,9 +100,14 @@ func (c *cache) deleteFromDB(keys [][]byte) error {
...
 		b := tx.Bucket(defaultBucket)
 		for i := range keys {
+			has := b.Get(keys[i])
+			if has == nil {
+				return object.ErrNotFound
+			}
 			if err := b.Delete(keys[i]); err != nil {
 				return err
 			}
+			c.dbSize.Sub(uint64(len(has)))
```
fyrchik, line 110: Same here, better to update it outside of tx.

fyrchik (changes requested):

pkg/local_object_storage/writecache/delete.go
ZhangTao1596: Done.

pkg/local_object_storage/writecache/persist.go
ZhangTao1596: Please have review.

pkg/local_object_storage/writecache/delete.go
ZhangTao1596: Done.

pkg/local_object_storage/writecache/storage.go
ZhangTao1596: Done.

pkg/local_object_storage/writecache/persist.go
ZhangTao1596: Done.

fyrchik (approved):
LGTM, last 2 commits can be squashed.

pkg/local_object_storage/writecache/persist.go
alexvanin: Isn't this `c.dbSize.Add()` still inside `db.Update`?

alexvanin (changes requested):
Please, add issue number in commit headers. Also you can specify short path to the working dir in commit. 

```
[#568] shard/writecache: Maintain memory size when remove

```

alexvanin (approved):

PR #569, opened 2021-05-31
Remove neofs-crypto dependency

### LINKED ISSUE #562: Support NEP-6 wallets as a private key source in neofs nodes
Related to #491 

We've got wallet support in NeoFS CLI (#466) and it should be supported in inner ring and storage nodes too. This is going to be a first step to drop neofs-crypto dependency. 

Storage node config can look like that:
```yaml
node:
  wallet:
    path: /home/user/node/wallet.json
    address: NXWry4f4LJ1ynrzaPjzfhG7KSh5aBSaqWU
    password: hunter2
```

We can keep `node.key` parameter for backward compatibility, but should move key file reading logic out of neofs-crypto library (`crypto.LoadPrivateKey`). If both wallet and key parameters are present, use key parameter.

Summary:
- [ ] Add NEP-6 wallet support in Storage Node
- [ ] Add NEP-6 wallet support in Inner Ring
- [ ] Figure out what needs to be done to remove neofs-crypto dependency from node

fyrchik:
Close #562.


alexvanin (commented):
After a bunch of PRs (#573, #575, #576, #577, #578, #579, #581) we changed the way storage node processes the config. It would be great if you adopt this changes in this PR.

As the part of these changes, we want to parse public keys in config. So take a look at `initControlService()`. Right now it takes [strings from config](https://github.com/nspcc-dev/neofs-node/blob/a2547da5ae87c9d1cfe6a4f328f4844d24d75112/cmd/neofs-node/config/control/config.go#L25) but we would rather take parsed PublicKeys from there. 

Also look at `node.key` configuration value [there](https://github.com/nspcc-dev/neofs-node/blob/a2547da5ae87c9d1cfe6a4f328f4844d24d75112/cmd/neofs-node/config/node/config.go#L27). I guess it is gonna be removed and wallet subsection should be defined in this file. See examples of subsection definitions in `config/grpc` or `config/control`.

And do we really want to remove `node.key` completely?

cmd/neofs-node/config/node/config.go
```diff
@@ -24,16 +27,40 @@ var (
...
+
+// Wallet returns value of node private key from "node" section.
+func Wallet(c *config.Config) *keys.PrivateKey {
+	v := c.Sub(subsection).Sub("wallet")
+	acc, err := innerringconfig.LoadAccount(
+		config.String(v, "path"),
+		config.String(v, "address"),
+		config.String(v, "password"))
+	if err != nil {
+		return Key(c)
```
alexvanin, line 60: I am not sure this fallback should be done in config. Maybe `Wallet` can return `nil` if section is empty? 

If you want to keep fallback from one section to another, then I propose to do that in `Key` function. In the end it returns PrivateKey, thus makes more sense.

cmd/neofs-node/config/node/config.go
```diff
@@ -24,16 +27,40 @@ var (
...
+		panic(fmt.Errorf("can't read key: %w", err))
+	}
+
+	return key
+}
+
+// Wallet returns value of node private key from "node" section.
+func Wallet(c *config.Config) *keys.PrivateKey {
+	v := c.Sub(subsection).Sub("wallet")
+	acc, err := innerringconfig.LoadAccount(
```
alexvanin, line 55: This looks really odd to import something from innerring config to storage node. Can we move it somewhere in `pkg`? `pkg/util` or some sort.

cmd/neofs-node/control.go
```diff
@@ -18,24 +19,24 @@ import (
...
-	keys = append(keys, crypto.MarshalPublicKey(&c.key.PublicKey))
+	pubs = append(pubs, c.key.PublicKey().Bytes())
 
 	for i := range strKeys {
 		key, err := hex.DecodeString(strKeys[i])
 		fatalOnErr(err)
 
-		if crypto.UnmarshalPublicKey(key) == nil {
-			fatalOnErr(fmt.Errorf("invalid permitted key for Control service %s", strKeys[i]))
+		if _, err := keys.NewPublicKeyFromBytes(key, elliptic.P256()); err != nil {
```
alexvanin, line 30: I think the better way is to parse these keys in `controlconfig.AuthorizedKeys`. For the same reason we parse private key in config as well.

pkg/innerring/innerring.go
```diff
@@ -344,11 +347,11 @@ func New(ctx context.Context, log *zap.Logger, cfg *viper.Viper) (*Server, error
 		}
 	}
 
-	server.pubKey = crypto.MarshalPublicKey(&server.key.PublicKey)
+	server.pubKey = server.key.Bytes()
```
alexvanin, line 350: I guess it should be 
```suggestion
	server.pubKey = server.key.PublicKey().Bytes()
```

pkg/morph/client/constructor.go
```diff
@@ -58,22 +57,13 @@ func defaultConfig() *cfg {
 // If desired option satisfies the default value, it can be omitted.
 // If multiple options of the same config value are supplied,
 // the option with the highest index in the arguments will be used.
-func New(key *ecdsa.PrivateKey, endpoint string, opts ...Option) (*Client, error) {
+func New(key *keys.PrivateKey, endpoint string, opts ...Option) (*Client, error) {
 	if key == nil {
-		return nil, crypto.ErrEmptyPrivateKey
+		// FIXME(@fyrchik): check if panic here breaks anything
```
alexvanin, line 62: It is okay to panic there.

alexvanin (commented):
Overall LGTM. Let's fix couple issues and call it a day.

cmd/neofs-node/config/node/config.go
fyrchik: IMO this is exactly what should be done in config: there is somesome info about a key and we just want to retrieve it.
I have put this in `Wallet` for wallet to have priority over raw key. Reversed the order here.

alexvanin (approved):

PR #574, opened 2021-06-01
[#496] Use static wrappers in processors

### LINKED ISSUE #496: Use `morph/client` package in inner ring instead of `invoke`
Right now inner ring has two ways to invoke chain methods:
- by using wrapper functions in `innerring/invoke` package,
- by using wrapper structures from `morph/client` package.

(1) functions were designed to be light weight wrappers around client. (2) structures were designed with complex logic inside: taking and marshaling structures from arguments, using predefined invocation fee, etc.

(2) already used in storage node and has some adoption in inner ring for various processors. I suppose we can completely drop wrapper functions from `innerring/invoke` and use only `morph/client` wrapper instances to avoid this duality.

fyrchik:
Close #496.


pkg/morph/client/notary.go
```diff
@@ -203,11 +205,10 @@ func (c *Client) UpdateNeoFSAlphabetList(list keys.PublicKeys) error {
 // blockchain. Fallback tx is a `RET`. If Notary support is not enabled
 // it fallbacks to a simple `Invoke()`.
 //
-// This function must be invoked after `EnableNotarySupport()` otherwise it
-// throws panic.
+// This function must be invoked with notary enabled otherwise it throws panic.
 func (c *Client) NotaryInvoke(contract util.Uint160, fee fixedn.Fixed8, method string, args ...interface{}) error {
 	if c.notary == nil {
-		return c.Invoke(contract, fee, method, args...)
+		panic(notaryNotEnabledPanicMsg)
```
alexvanin, line 211: Not sure about that. Or, at least, we need to make more changes.

Inner ring, depending on config, enables or disables notary in the client. Static clients, which are wrappers, have `TryNotary` option. If option enabled, then static client uses `NotaryInvoke()`, otherwise `Invoke()`. `NotaryInvoke()`, on the other hand, can fallback to `Notary()` if notary was not enabled (you added it in https://github.com/nspcc-dev/neofs-node/commit/b5cda8cd4180db5e6d13b83fe8813923b78e020e).

Now:
1. all static clients in Inner Ring have `TryNotary` option,
2. if Inner Ring config disables notary, then `c.notary == nil`,
3. in this case Inner Ring expects to fallback to `Invoke()`.

(3) is not happening with this change. So you either need to return `Invoke()` fallback or reconsider role of `TryNotary()` flag in static client.

pkg/innerring/processors/netmap/processor.go
```diff
@@ -58,6 +59,7 @@ type (
 		alphabetState  AlphabetState
 
 		morphClient  *client.Client
```
alexvanin, line 61: Not used anymore, drop it.

alexvanin (changes requested):
All processors have `FeeProvider` but they are unused anymore. Drop them.

pkg/morph/client/neofs/wrapper/cheque.go
```diff
@@ -0,0 +1,16 @@
+package neofscontract
+
+import (
+	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
+	"github.com/nspcc-dev/neo-go/pkg/util"
+)
+
+// Cheque invokes `cheque` method of NeoFS contract.
+func (x *ClientWrapper) Cheque(id []byte, user util.Uint160, amount int64, lock util.Uint160) error {
+	return x.Cheque(id, user, amount, lock)
```
alexvanin, line 10: Linter says it should be 
```suggestion
	return (*neofscontract.Client)(x).Cheque(id, user, amount, lock)
```

pkg/morph/client/neofs/wrapper/cheque.go
```diff
@@ -0,0 +1,16 @@
...
+}
+
+// AlphabetUpdate update list of alphabet nodes.
+func (x *ClientWrapper) AlphabetUpdate(id []byte, pubs keys.PublicKeys) error {
+	return x.AlphabetUpdate(id, pubs)
```
alexvanin, line 15:
```suggestion
	return (*neofscontract.Client)(x).AlphabetUpdate(id, pubs)
```

pkg/morph/client/netmap/config.go
```diff
@@ -56,6 +56,11 @@ func (c *Client) Config(args ConfigArgs, assert func(stackitem.Item) (interface{
 	}, nil
 }
 
+func (c *Client) SetConfig(id, key []byte, value interface{}) error {
+	_, err := c.client.TestInvoke(c.configMethod, id, key, value)
```
alexvanin, line 60:
```suggestion
	return c.client.Invoke(c.configMethod, id, key, value)
```

alexvanin (changes requested):

pkg/morph/client/netmap/config.go
fyrchik: Yeah, method should also be different.

pkg/innerring/processors/neofs/process_assets.go
```diff
@@ -24,10 +20,7 @@ func (np *Processor) processDeposit(deposit *neofsEvent.Deposit) {
 	}
 
 	// send transferX to balance contract
-	err := np.morphClient.NotaryInvoke(np.balanceContract, np.feeProvider.SideChainFee(), mintMethod,
-		deposit.To().BytesBE(),
-		np.converter.ToBalancePrecision(deposit.Amount()),
-		deposit.ID())
+	err := np.balanceClient.Mint(deposit.To(), deposit.Amount(), deposit.ID())
```
alexvanin, line 23:
```suggestion
	err := np.balanceClient.Mint(deposit.To(), np.converter.ToBalancePrecision(deposit.Amount()), deposit.ID())
```

alexvanin (changes requested):
Last one.

alexvanin (approved):

PR #592, opened 2021-06-04
ir/container: Verify session token lifetime

### LINKED ISSUE #589: Validate session token lifetime
According to https://github.com/nspcc-dev/neofs-api-go/issues/298 there is no way to access session token lifetime without converting to v2 structure. It means that this field is not checked in alphabet nodes when they process `container.Put`, `container.Delete` and `container.SetEACL` notifications.

cthulhu-rider:
Closes #589.

### DIFF
--- a/pkg/innerring/processors/container/processor.go
+++ b/pkg/innerring/processors/container/processor.go
@@ -31,6 +31,7 @@ type (
 		feeProvider       *config.FeeConfig
 		cnrClient         *wrapper.Wrapper // notary must be enabled
 		idClient          *neofsid.ClientWrapper
+		netState          NetworkState
 	}
 
 	// Params of the processor constructor.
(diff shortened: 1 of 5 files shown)


pkg/innerring/processors/container/processor.go
```diff
@@ -43,9 +44,20 @@ type (
 		FeeProvider       *config.FeeConfig
 		ContainerClient   *wrapper.Wrapper
 		NeoFSIDClient     *neofsid.ClientWrapper
+		NetworkState      NetworkState
 	}
 )
 
+// NetworkState is an interface of a component
+// that provides access to network state.
+type NetworkState interface {
```
alexvanin, line 53: Define this interface together with other types, please. We have `AlphabetState` right above, for example.

alexvanin (approved):

pkg/innerring/processors/container/processor.go
cthulhu-rider: I dont think it is useful since these types are independent.

PR #593, opened 2021-06-07
object/put: Check payload size of the finished objects

### LINKED ISSUE #580: Max size of finished objects is not checked
NeoFS network configuration dictates a limit on the maximum size of the payload of objects. Storage nodes should check if incoming objects satisfy to the limit.

In current implementation storage nodes don't check payload size of "ready" objects. As a result, excessive consumption of node memory is possible, since it tries not to cache more than the maximum value in the memory of the payload within one stream of the object (`object.Put`). 

P.S. During `object.Get` this kind of check may not be legitimate, since the constraint is a dynamic configuration.

### How to reproduce
Prepare the object with payload larger than limit and stream it w/o session.

cthulhu-rider:
Closes #580.

### DIFF
--- a/pkg/services/object/put/validation.go
+++ b/pkg/services/object/put/validation.go
@@ -3,6 +3,7 @@ package putsvc
 import (
 	"bytes"
 	"crypto/sha256"
+	"errors"
 	"fmt"
 	"hash"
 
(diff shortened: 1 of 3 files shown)


pkg/services/object/put/validation.go
```diff
@@ -20,9 +21,34 @@ type validatingTarget struct {
...
+	maxPayloadSz uint64 // network config
+
+	payloadSz uint64 // payload size of the streaming object from header
+
+	writtenPayload uint64 // number of already written payload bytes
 }
 
+// errors related to invalid payload size
+var (
+	errExceedingMaxSize = errors.New("payload size is greater than the limit ")
```
alexvanin, line 34:
```suggestion
	errExceedingMaxSize = errors.New("payload size is greater than the limit")
```

pkg/services/object/put/validation.go
```diff
@@ -20,9 +21,34 @@ type validatingTarget struct {
...
+	t.payloadSz = obj.PayloadSize()
+	chunkLn := uint64(len(obj.Payload()))
+
+	// check chunk size
+	if chunkLn > t.payloadSz {
+		return errWrongPayloadSize
+	}
+
+	// check payload size limit
+	if t.payloadSz > t.maxPayloadSz {
```
alexvanin, line 48: Wouldn't it better to do this?
```suggestion
	if chunkLn > t.maxPayloadSz {
```

As far as I understand, client can send 2GiB of payload in stream with correct object size in header and this check will fail, or am I missing something?

cthulhu-rider: `chunkLn` should not be GE `payloadSz` by condition above. This condition checks that the client does not start streaming a obviously larger object. Your condition is satisfied automatically by the transitivity of the `>` operator.

pkg/services/object/put/validation.go
cthulhu-rider: Fixed thx

pkg/services/object/put/validation.go
alexvanin: `validatingTarget.WriteHeader` is not executed against unfinished objects, I missed that. Therefore my comment is invalid.

alexvanin (approved):

PR #598, opened 2021-06-09
Update inner ring config

### LINKED ISSUE #363: Move timer related config values to global configuration in netmap contract
Timer related config values might be set up through config, however if config value contains `0` or empty string, then application must look at global config in network map.

Timer options:
- [x] timers.epoch
- [x] ~~timers.emit~~

Other options:
- [x] ~~emit.storage.amount~~
- [x] eigentrust.rounds

Extra:
- [x] Add `DEBUG` option to override global values in IR

Considering options:
- timers.stop_estimation.mul
- timers.stop_estimation.div
- timers.collect_basic_income.mul
- timers.collect_basic_income.div
- timers.distribute_basic_income.mul
- timers.distribute_basic_income.div

alexvanin:
Closes #363 

This PR finishes transition to global config in inner ring node. All global config values have corresponding local config values, but they are active **only in debug build**. To make debug build try:
```
$ DEBUG=true make bin/neofs-ir
```

There are couple more changes there as well. First we came up with values for out timers in testnet and neofs-dev-env so it makes sense to make them default. Also we changed storage node config and it can't turn off or specify length of bootstrap healthchecks, so it also reflected in inner ring config.

### DIFF
--- a/pkg/innerring/config/config.go
+++ b/pkg/innerring/config/config.go
@@ -13,6 +13,7 @@ Implemented as a part of https://github.com/nspcc-dev/neofs-node/issues/363
 */
 
 import (
+	"github.com/nspcc-dev/neofs-node/misc"
 	netmapClient "github.com/nspcc-dev/neofs-node/pkg/morph/client/netmap/wrapper"
 	"github.com/spf13/viper"
 )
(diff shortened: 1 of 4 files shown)


realloc (approved):

pkg/innerring/config/config.go
```diff
@@ -31,7 +32,7 @@ func NewGlobalConfigReader(cfg *viper.Viper, nm *netmapClient.Wrapper) *GlobalCo
 
 func (c *GlobalConfig) BasicIncomeRate() (uint64, error) {
 	value := c.cfg.GetUint64("settlement.basic_income_rate")
-	if value != 0 {
+	if isDebug() && value != 0 {
```
cthulhu-rider, line 35: Code is valid, but lets wrap value reading `cfg.Get` into
```
if isDebug() {
	...
}
```
statement in order to underline that we read config only in debug build.

cthulhu-rider (approved):

PR #601, opened 2021-06-09
Implement IR Control service with health check

### LINKED ISSUE #414: Inner ring node healthcheck
After #404 inner ring node won't be in the healthy status immediately at the start. It has to deposit 1 GAS to a notary contract and wait for a block / two, so it will be able to operate with notary contract. There is an issue with neofs-dev-env because docker-compose creates containers one after another and storage nodes will bootstrap before inner ring nodes will be ready. 

To solve that we should have health status endpoint in inner ring node. Then run command to fetch if in `healthcheck` section of `docker-compose.yml` config.

Right now I see two ways to do that:
- use http server with `/healthy` handler,
- use grpc server with sort of control API service.

As for control API, we should share protobuf files and healthy format between storage node and inner ring somehow.

cthulhu-rider:
Closes #414.


cmd/neofs-cli/modules/control.go
```diff
@@ -37,6 +41,16 @@ const (
...
 )
 
+// control healthcheck flags
+const (
+	healthcheckIRFlag = "ir"
+)
+
+// control healthcheck vars
+var (
+	healthCheckIRVar bool
```
alexvanin, line 51: I would rather remove `( )` if block consists one variable or one constant.

alexvanin (approved):

cmd/neofs-cli/modules/control.go
cthulhu-rider: This is how I usually do, but I indicated in the comment the semantic grouping of names.

PR #609, opened 2021-06-15
morph/client: Check return state in invoke

alexvanin:
Test invocations are used in `Invoke` method to calculate consumed gas. We can check return code and return error if panic happened in contract. 

This is already done the same way in `TestInvoke` method.

### DIFF
--- a/pkg/morph/client/client.go
+++ b/pkg/morph/client/client.go
@@ -100,6 +100,10 @@ func (c *Client) Invoke(contract util.Uint160, fee fixedn.Fixed8, method string,
 		return err
 	}
 
+	if resp.State != HaltState {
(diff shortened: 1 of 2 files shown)


cthulhu-rider (approved):

pkg/morph/client/client.go
```diff
@@ -100,6 +100,10 @@ func (c *Client) Invoke(contract util.Uint160, fee fixedn.Fixed8, method string,
 		return err
 	}
 
+	if resp.State != HaltState {
+		return &NotHaltStateError{state: resp.State, exception: resp.FaultException}
```
cthulhu-rider, line 104: I suggest to reflect this error case in docs or make it unexported.

PR #612, opened 2021-06-15
Extend CLI key options

### LINKED ISSUE #610: NeoFS CLI input credentials
NeoFS CLI help command suggests to pass WIF, hex or wallet file to get private key.
```
$ neofs-cli --help
...
Flags:
...
  -k, --key --key new         private key in hex, WIF, NEP-2 or filepath (use --key new to generate key for request)
```

However after #562 it is not true. I suggest to keep support of WIF or hex encoded files, but we can do that in different arguments (the same way as NEOFS_KEY and NEOFS_WALLET are the same arguments in storage node config).

fyrchik:
Close #610 .

### DIFF
--- a/cmd/neofs-cli/modules/root.go
+++ b/cmd/neofs-cli/modules/root.go
@@ -24,8 +24,6 @@ import (
 const (
 	envPrefix = "NEOFS_CLI"
 
-	generateKeyConst = "new"
-
 	ttlDefaultValue = 2
 )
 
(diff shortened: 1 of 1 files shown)


cmd/neofs-cli/modules/root.go
```diff
@@ -133,25 +143,51 @@ const nep2Base58Length = 58
 
 // getKey returns private key that was provided in global arguments.
 func getKey() (*ecdsa.PrivateKey, error) {
-	privateKey := viper.GetString("key")
-	if privateKey == generateKeyConst {
+	if viper.GetBool("generate-key") {
```
cthulhu-rider, line 146: `generateKeyConst` became unused? Lets delete if so

cthulhu-rider (approved):

alexvanin (approved):

PR #613, opened 2021-06-16
Feature/add reputation sanity check in IR

### LINKED ISSUE #500: Make trust approval more reliable
Inner ring approves trust values and saves them in contract by checking signature of `GlobalTrust` structure. It is possible to improve this by checking manager correctness.

Manager of the node is calculated based on network map and epoch number in deterministic way. Inner ring node has epoch number, peer ID, manager ID and access to network map, thus it is possible to make such check. This check should be done in `pkg/innerring/reputation/process_put.go`.

carpawell:
Closes #500.

### DIFF
--- a/pkg/innerring/processors/reputation/process_put.go
+++ b/pkg/innerring/processors/reputation/process_put.go
@@ -1,14 +1,20 @@
 package reputation
 
 import (
+	"bytes"
 	"encoding/hex"
+	"errors"
+	"fmt"
 
-	"github.com/nspcc-dev/neofs-api-go/pkg/reputation"
+	apireputation "github.com/nspcc-dev/neofs-api-go/pkg/reputation"
 	"github.com/nspcc-dev/neofs-node/pkg/morph/client/reputation/wrapper"
+	"github.com/nspcc-dev/neofs-node/pkg/services/reputation"
 	"go.uber.org/zap"
 )
 
-func (rp *Processor) processPut(epoch uint64, id reputation.PeerID, value reputation.GlobalTrust) {
+var errWrongManager = errors.New("got manager that is incorrect for peer")
+
+func (rp *Processor) processPut(epoch uint64, id apireputation.PeerID, value apireputation.GlobalTrust) {
 	if !rp.alphabetState.IsAlphabet() {
 		rp.log.Info("non alphabet mode, ignore reputation put notification")
 		return
@@ -34,7 +40,14 @@ func (rp *Processor) processPut(epoch uint64, id reputation.PeerID, value reputa
 		return
 	}
 
-	// todo: do sanity checks of value
+	// check if manager is correct
+	if err := rp.checkManagers(epoch, *value.Manager(), id); err != nil {
+		rp.log.Info("ignore reputation value",
+			zap.String("reason", "wrong manager"),
+			zap.String("error", err.Error()))
+
+		return
+	}
 
 	args := wrapper.PutArgs{}
 	args.SetEpoch(epoch)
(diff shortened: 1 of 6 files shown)


pkg/innerring/processors/reputation/process_put.go
```diff
@@ -48,3 +63,18 @@ func (rp *Processor) processPut(epoch uint64, id reputation.PeerID, value reputa
 			zap.String("error", err.Error()))
 	}
 }
+
+func (rp *Processor) checkManagers(e uint64, mng apireputation.PeerID, peer apireputation.PeerID) error {
+	mm, err := rp.mngBuilder.BuildManagers(e, reputation.PeerIDFromBytes(peer.ToV2().GetPublicKey()))
+	if err != nil || len(mm) == 0 {
+		return errCouldNotBuildManagers
```
cthulhu-rider, line 70: We do not recognize what the `err` is. I suggest to return `bool` from the method, and log this error.

pkg/services/reputation/common/managers/managers.go
```diff
@@ -1,7 +1,8 @@
-package common
+package managers
```
cthulhu-rider, line 1: Why do we need this replacing?

pkg/services/reputation/common/managers/managers.go
```diff
@@ -12,6 +14,15 @@ import (
 	"go.uber.org/zap"
 )
 
+type ErrNoSuchPeerInNetmap struct {
```
alexvanin, line 17: Do we need to export that, though?

alexvanin (approved):
Overall LGTM.

pkg/services/reputation/common/managers/managers.go
carpawell: not for now, i think, but maybe it could be useful in the future

make it private?

alexvanin: Make it public when it's needed. Unlike `neofs-api-go`, it is not a library, so we are in control of this error usage.

carpawell: got it, changed

pkg/services/reputation/common/managers/managers.go
cthulhu-rider: Dont think we need a separate package for this

pkg/services/reputation/common/managers/managers.go
```diff
@@ -83,7 +93,7 @@ func (mb *managerBuilder) BuildManagers(epoch uint64, p reputation.PeerID) ([]co
 		}
 	}
 
-	return nil, nil
+	return nil, &errNoSuchPeerInNetmap{peer: p, epoch: epoch}
```
cthulhu-rider, line 96: What exactly caused this change?

pkg/innerring/processors/reputation/process_put.go
cthulhu-rider: I mean to return only `bool` from the method. Here is the only place for the error, an we log it.

We can skip check the emptiness of managers, because the code below will eventually fail the check.

pkg/innerring/processors/reputation/process_put.go
```diff
@@ -48,3 +61,18 @@ func (rp *Processor) processPut(epoch uint64, id reputation.PeerID, value reputa
...
 }
+
+func (rp *Processor) checkManagers(e uint64, mng apireputation.PeerID, peer apireputation.PeerID) error {
+	mm, err := rp.mngBuilder.BuildManagers(e, reputation.PeerIDFromBytes(peer.ToV2().GetPublicKey()))
+	if err != nil {
+		return fmt.Errorf("could not build managers: %w", err)
+	}
+
+	for _, m := range mm {
+		if bytes.Equal(mng.ToV2().GetPublicKey(), m.PublicKey()) {
```
cthulhu-rider, line 72: We don't want `ToV2`, so at least add FIXME.

pkg/innerring/processors/reputation/process_put.go
carpawell: > Here is the only place for the error, an we log it.

"manager not found" is also an error case. so one check function that returns any kind of error and logging the result of the checking after that vs logging inside after every mistake and returning bool. what should we do with that bool? one more check to do what? or just `return`?

> We can skip check the emptiness of managers, because the code below will eventually fail the check.

but what are the cons of the additional information about reasons for error? there somehow can not be the peer in the netmap at all, or the peer can have managers that are different from received one

pkg/services/reputation/common/managers/managers.go
carpawell: a deleted empty managers check in `managers`'s code and moved it here. i think that this is error case when node not in `netmap` but we are building managers for it

carpawell: reverted back

pkg/innerring/processors/reputation/process_put.go
carpawell: added `FIXME` to every `ToV2` usage in that package

cthulhu-rider (approved):

Issue #600, opened 2021-06-09
Bump dependencies


alexvanin:
We are closer and closer for major release and it will be nice to keep our dependencies updated (@realloc will definitely like it). Let's check this list down below, go through the changes and and update everything that does not break compatibility.

```
$ go list -u -m -json all | go-mod-outdated -update -direct
+--------------------------------------+------------------------------------+------------------------------------
|                MODULE                |              VERSION               |            NEW VERSION             
+--------------------------------------+------------------------------------+------------------------------------
| code.cloudfoundry.org/bytefmt        | v0.0.0-20200131002437-cf55d5288a48 | v0.0.0-20210608160410-67692ebc98de 
| github.com/golang/protobuf           | v1.4.3                             | v1.5.2                             
| github.com/google/uuid               | v1.1.1                             | v1.2.0                             
| github.com/klauspost/compress        | v1.11.3                            | v1.13.0                            
| github.com/mr-tron/base58            | v1.1.3                             | v1.2.0                             
| github.com/multiformats/go-multiaddr | v0.3.1                             | v0.3.2                             
| github.com/nspcc-dev/neofs-sdk-go    | v0.0.0-20210520210714-9dee13f0d556 | v0.0.0-20210604112451-f16d38c7b92a 
| github.com/panjf2000/ants/v2         | v2.3.0                             | v2.4.5                             
| github.com/paulmach/orb              | v0.2.1                             | v0.2.2                             
| github.com/prometheus/client_golang  | v1.6.0                             | v1.11.0                            
| github.com/spf13/cast                | v1.3.0                             | v1.3.1                             
| github.com/spf13/cobra               | v1.0.0                             | v1.1.3                             
| github.com/spf13/viper               | v1.7.0                             | v1.7.1                             
| github.com/stretchr/testify          | v1.6.1                             | v1.7.0                             
| go.etcd.io/bbolt                     | v1.3.5                             | v1.3.6                             
| go.uber.org/atomic                   | v1.5.1                             | v1.7.0                             
| go.uber.org/zap                      | v1.13.0                            | v1.17.0                            
| google.golang.org/grpc               | v1.29.1                            | v1.38.0                            
| google.golang.org/protobuf           | v1.25.0                            | v1.26.0                            
+--------------------------------------+------------------------------------+------------------------------------
```

alexvanin:
Also do https://github.com/nspcc-dev/neofs-api-go/issues/176

carpawell:
[...]

PR #615, opened 2021-06-16
blobovnicza: Simplify byte size stringer

cthulhu-rider:
Relates to #600.

```
after
creating bucket for size range	{"component": "Blobovnicza", "range": "[0:32768]"}
before
creating bucket for size range	{"component": "Blobovnicza", "range": "[0:32K]"}
```


alexvanin (approved):
Can you provide example "before & after" in the PR description?

Issue #607, opened 2021-06-11
Support multiple addresses


alexvanin:
Related https://github.com/nspcc-dev/neofs-api/issues/160

- [ ] Multiple `NODE_ADDRESS` and `GRPC_ENDOINT` addresses should be supported in config.
- [ ] Run node on all provided gRPC endpoints.
- [ ] Announce all provided node addresses.
- [ ] Iterate over sorted* list of endpoints in client cache until you establish connection.

* Sorted list is a list with all TLS endpoints before others.

cthulhu-rider:
We need to recognize `Client`'s connection failures. I can suggest two approaches.
#### 1. Dedicated `Open` or `Init` method
Before any RPC caller should call the method to establish the connection. In this approach we need to think about how `Client` behaves if connection was provided to constructor (e.g. check status). Any RPC should once do pre-opening in order to save previous behavior.
#### 2. Dedicated connection errors
Define typed `error` which can be returned from any RPC method. Check the error when we iterate over all per-address `Client`s.


As for me, 2 is clearer, but 1 also has a place to be.

alexvanin:
> 1. Dedicated Open or Init method

Looks like protocol version handshake becomes handy there.

PR #616, opened 2021-06-16
Minor refactoring of network package

cthulhu-rider:
Relates to #607.

### DIFF
--- a/cmd/neofs-cli/modules/root.go
+++ b/cmd/neofs-cli/modules/root.go
@@ -223,13 +223,8 @@ func getSDKClient(key *ecdsa.PrivateKey) (client.Client, error) {
 		return nil, err
 	}
 
-	hostAddr, err := netAddr.HostAddrString()
-	if err != nil {
-		return nil, errInvalidEndpoint
-	}
-
 	options := []client.Option{
-		client.WithAddress(hostAddr),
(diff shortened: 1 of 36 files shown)


alexvanin (commented):
If some function accept value of the `Address` instead of a pointer, then I suggest to dive a bit deeper and do not return pointers at all in network package. What do you think?

cthulhu-rider:
> If some function accept value of the `Address` instead of a pointer, then I suggest to dive a bit deeper and do not return pointers at all in network package. What do you think?

Yes, I thought about that. We can implement `Address.FromString` method instead of `AddressFromString` function. Usage:
```
var addr network.Address

if err := addr.FromString(s); err != nil {
  // handle
}
```
What do you think?

alexvanin:
> > If some function accept value of the `Address` instead of a pointer, then I suggest to dive a bit deeper and do not return pointers at all in network package. What do you think?
> 
> Yes, I thought about that. We can implement `Address.FromString` method instead of `AddressFromString` function. Usage:
> 
> ```
> var addr network.Address
> 
> if err := addr.FromString(s); err != nil {
>   // handle
> }
> ```
> 
> What do you think?

Both looks good for me.

alexvanin (approved):

PR #647, opened 2021-06-25
feature/add-`netmapCandidates`-netmap-method

carpawell:
Related to https://github.com/nspcc-dev/neofs-net-monitor/issues/12.

### DIFF
--- a/pkg/morph/client/netmap/netmap.go
+++ b/pkg/morph/client/netmap/netmap.go
@@ -36,7 +36,55 @@ type EpochSnapshotValues struct {
 	*GetNetMapValues
 }
 
-const nodeInfoFixedPrmNumber = 1
+// GetNetMapCandidatesArgs groups the arguments
+// of get network map candidates test invoke call.
+type GetNetMapCandidatesArgs struct {
+}
+
+// GetNetMapCandidatesValues groups the stack parameters
+// returned by get network map candidates test invoke.
+type GetNetMapCandidatesValues struct {
+	netmapNodes []*PeerWithState
+}
+
+func (g GetNetMapCandidatesValues) NetmapNodes() []*PeerWithState {
+	return g.netmapNodes
+}
+
+// State is an enumeration of various states of the NeoFS node.
+type State int64
+
+const (
+	// Undefined is unknown state.
+	Undefined State = iota
+
+	// Online is network unavailable state.
+	Online
+
+	// Offline is an active state in the network.
+	Offline
+)
+
+// PeerWithState groups information about peer
+// and its state in network map.
+type PeerWithState struct {
+	peer  []byte
+	state State
+}
+
+func (ps PeerWithState) State() State {
+	return ps.state
+}
+
+func (ps PeerWithState) Peer() []byte {
+	return ps.peer
+}
+
+const (
+	nodeInfoFixedPrmNumber = 1
+
+	peerWithStateFixedPrmNumber = 2
+)
 
 // SetDiff sets argument for snapshot method of
 // netmap contract.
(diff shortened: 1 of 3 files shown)


pkg/morph/client/netmap/netmap.go
```diff
@@ -36,7 +36,45 @@ type EpochSnapshotValues struct {
...
-const nodeInfoFixedPrmNumber = 1
+// GetNetMapCandidatesArgs groups the arguments
+// of get network map candidates test invoke call.
+type GetNetMapCandidatesArgs struct {
+}
+
+// GetNetMapCandidatesValues groups the stack parameters
+// returned by get network map candidates test invoke.
+type GetNetMapCandidatesValues struct {
+	NetmapNodes []*PeerWithState
```
cthulhu-rider, line 47: Pls make it unexported, or it will be hard to keep code and logic compatibility.

pkg/morph/client/netmap/netmap.go
```diff
@@ -36,7 +36,45 @@ type EpochSnapshotValues struct {
...
+}
+
+type State int64
```
cthulhu-rider, line 50: Pls add docs to all public elements.

pkg/morph/client/netmap/netmap.go
```diff
@@ -107,6 +145,103 @@ func (c *Client) EpochSnapshot(args EpochSnapshotArgs) (*EpochSnapshotValues, er
...
+		}
+
+		res.NetmapNodes = append(res.NetmapNodes, node)
+	}
+
+	return res, nil
+}
+
+func peerWithStateFromStackItem(prm stackitem.Item) (*PeerWithState, error) {
+	if prm.Type() != stackitem.StructT {
```
cthulhu-rider, line 191: Why do we need explicit `stackitem.StructT` which is supported by `client.ArrayFromStackItem`?

pkg/morph/client/netmap/netmap.go
```diff
@@ -107,6 +145,103 @@ func (c *Client) EpochSnapshot(args EpochSnapshotArgs) (*EpochSnapshotValues, er
...
+			"unexpected stack item count (PeerWithState): expected %d, has %d",
+			peerWithStateFixedPrmNumber,
+			ln,
+		)
+	}
+
+	var res PeerWithState
+
+	// peer
+	if prms[0].Type() != stackitem.StructT {
```
cthulhu-rider, line 209: Same as for `stackitem.StructT` and below

pkg/morph/client/netmap/netmap.go
```diff
@@ -107,6 +145,103 @@ func (c *Client) EpochSnapshot(args EpochSnapshotArgs) (*EpochSnapshotValues, er
...
+	if err != nil {
+		return nil, fmt.Errorf("could not parse contract response: %w", err)
+	}
+
+	return candVals, nil
+}
+
+func peersWithStateFromStackItems(stack []stackitem.Item, method string) (*GetNetMapCandidatesValues, error) {
+	if ln := len(stack); ln != 1 {
+		return nil, fmt.Errorf("unexpected stack intem count (%s): %d", method, ln)
```
cthulhu-rider, line 166:
```suggestion
		return nil, fmt.Errorf("unexpected stack item count (%s): %d", method, ln)
```
and re-check if it caused by copy-paste.

cthulhu-rider (changes requested):

pkg/morph/client/netmap/netmap.go
carpawell: made unexported and add getter to it

pkg/morph/client/netmap/netmap.go
carpawell: added

pkg/morph/client/netmap/netmap.go
carpawell: an additional layer of checks since there are too deep nesting of structures(there was no nested structs parsing before that PR)

pkg/morph/client/netmap/netmap.go
carpawell: fixed and checked

pkg/morph/client/netmap/netmap.go
```diff
@@ -107,6 +155,92 @@ func (c *Client) EpochSnapshot(args EpochSnapshotArgs) (*EpochSnapshotValues, er
...
+
+	var res PeerWithState
+
+	// peer
+	if res.peer, err = peerInfoFromStackItem(prms[0]); err != nil {
+		return nil, fmt.Errorf("could not get bytes from 'node' field of PeerWithState: %w", err)
+	}
+
+	// state
+	if prms[1].Type() != stackitem.IntegerT {
```
alexvanin, line 220: I think you already discussed that this check is redundant. Big integers can be returned as `ByteArrayT` from the RPC. This is not the case for client status, but in general this can happen with integers and `IntFromStackItem` already covers this case.

alexvanin (changes requested):

pkg/morph/client/netmap/netmap.go
carpawell: ok, deleted that check too

alexvanin (approved):

PR #653, opened 2021-06-29
Shard: Fix unexpected return when remove garbage

ZhangTao1596:
@alexvanin


ZhangTao1596:
https://github.com/ZhangTao1596/neofs-node/blob/64a46bb464f33bd9537e5f74601f673393d855ff/pkg/local_object_storage/metabase/graveyard.go#L61-L63

PR #680, opened 2021-07-09
Add Docker images for neofs-adm

realloc:
%subj%


alexvanin (approved):
I think it is time to think about separate `docker` dir for docker files.

PR #690, opened 2021-07-13
Fallback to wallet section if storage node's key is not set

alexvanin:
Some users want to specify only wallet section in the SN. It is not possible if `Key` throws panic on empty value. Instead it should
fallback to wallet section. Panic is suitable if node's key is provided but invalid.


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ Changelog for NeoFS Node
 
 ### Fixed
 - Do not require MainNet attributes in "Without MainNet" mode ([#663](https://github.com/nspcc-dev/neofs-node/issues/663)).
+- User can specify only wallet section without node key ([#xxx](xxx))
```
carpawell, line 8: is PR num and link going to be actualized?

carpawell (approved):

CHANGELOG.md
alexvanin: Yep. It is, already.

PR #694, opened 2021-07-13
Reduce amount of open connections in storage node

### LINKED ISSUE #692: Connection overflow
It seems that neofs-node v0.22.2 does not reuse existing connections and creates new one on each request. At one time there is no enough ulimit to open new connection and we see errors like in https://github.com/nspcc-dev/neofs-node/issues/676.

Here the example from one of our storage nodes couple hours after restart.

```
# ls -l /proc/9758/fd | grep socket  | wc -l
568
# lsof -i -a -p 9758 | grep 199.247.0.222 | wc -l
199
```

I guess we need to look at client cache. Probably it is related to group address support.

alexvanin:
Closes #692 

First commit fixes issue with connection leak.
Second commit reduces amount of total connections by reusing the same client cache in all storage node components.

### DIFF
--- a/pkg/network/cache/multi.go
+++ b/pkg/network/cache/multi.go
@@ -324,9 +324,14 @@ func (x *multiClient) Close() error {
 func (x *multiClient) RawForAddress(addr network.Address) *rawclient.Client {
 	x.mtx.Lock()
 
-	c := x.createForAddress(addr).Raw()
+	strAddr := addr.String()
+
+	c, cached := x.clients[strAddr]
+	if !cached {
+		c = x.createForAddress(addr)
+	}
 
 	x.mtx.Unlock()
 
-	return c
+	return c.Raw()
 }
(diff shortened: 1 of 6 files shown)


carpawell (approved):

pkg/network/cache/multi.go
```diff
@@ -324,9 +324,14 @@ func (x *multiClient) Close() error {
 func (x *multiClient) RawForAddress(addr network.Address) *rawclient.Client {
 	x.mtx.Lock()
 
-	c := x.createForAddress(addr).Raw()
+	strAddr := addr.String()
+
+	c, cached := x.clients[strAddr]
+	if !cached {
```
cthulhu-rider, line 330: It is sensible to place this logic inside the `createForAddress`

cthulhu-rider (approved):

pkg/network/cache/multi.go
alexvanin: We can do small refactor there. Check https://github.com/nspcc-dev/neofs-node/pull/694/commits/ab78fa1830b935cc877e152b13ef0d6310abb360

I would like to keep `createForAddress` only for creation of new clients. To get or create we can use `x.client(addr)`.

PR #708, opened 2021-07-21
Sync first epoch duration at application startup

### LINKED ISSUE #679: Internal timer syncronization
In N3 Testnet RC3 we've got an issue with epoch 939. The epoch was frozen for some time, because internal epoch timers of Alphabet nodes were not synchronized. Notary transactions from one nodes failed before other nodes send them. In 240 blocks we could not collect more than 4 valid notary transaction (need at least 5). Even though Alphabet nodes synchronize timers after new epoch notification, some nodes may be restarted between epochs, which can lead to this. It will work the same way in notary disabled environment, because `votes` are valid only for [20 blocks](https://github.com/nspcc-dev/neofs-contract/blob/4d648431d6faabb1edd3c6ceb1f73aa8be955918/common/vote.go#L25). 

To solve this issue we need either:
- increase valid time of notary transaction (and more blocks for `votes` in contract) and make it **at least** amount of blocks per epoch (240 for testnet)
- store information about new epoch block in netmap contract so restarted node can synchronize it's clock.

I think first solution will be easier to implement, but it might work not that great in disabled notary environment. 
Second solution requires to modify netmap contract and tweak inner ring event scheduler.

alexvanin:
Closes #679 

At the startup, Inner Ring application calculates block duration before next epoch tick and processes new epoch event at this block.

### DIFF
--- a/pkg/innerring/innerring.go
+++ b/pkg/innerring/innerring.go
@@ -25,6 +25,7 @@ import (
 	"github.com/nspcc-dev/neofs-node/pkg/innerring/processors/reputation"
 	"github.com/nspcc-dev/neofs-node/pkg/innerring/processors/settlement"
 	auditSettlement "github.com/nspcc-dev/neofs-node/pkg/innerring/processors/settlement/audit"
+	timerEvent "github.com/nspcc-dev/neofs-node/pkg/innerring/timers"
 	"github.com/nspcc-dev/neofs-node/pkg/morph/client"
 	auditWrapper "github.com/nspcc-dev/neofs-node/pkg/morph/client/audit/wrapper"
 	balanceWrapper "github.com/nspcc-dev/neofs-node/pkg/morph/client/balance/wrapper"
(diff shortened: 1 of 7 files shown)


alexvanin:
To process single event in the future, I added `NewOneTickTimer` constructor in `morph/timer` (https://github.com/nspcc-dev/neofs-node/pull/708/commits/4ab91aedeb544482b6ac2967efc11101d412c25a). Maybe there are better solution for this.

carpawell (approved):
it took some time for me to get the idea of what `EpochBlock` returns, i had to go to read it in contract code
wouldn't it be better to add smth like "last" or "start" to the method name?

alexvanin:
> wouldn't it be better to add smth like "last" or "start" or both to the method name?

I thought about that. `LastEpochBlock()` in the contract and in the code will be better, what do you think?

carpawell:
> what do you think?

yes, it sounds better for me

PR #711, opened 2021-07-22
Docs/Make config examples usable

### LINKED ISSUE #699: Make example config usable for apllication start
Example config provides some unusable values right now, because it was designed as a config format example, not a real config. However people tend to use it without modifying, so it's better to make it cleaner. We can get some values from neofs-dev-env storage configuration.

Related to #646

carpawell:
Closes #699.

Permissions are set on the assumption that default executable dirs will be added soon.


config/example/node.env
```diff
@@ -1,10 +1,10 @@
 NEOFS_LOGGER_LEVEL=debug
 
 NEOFS_PROFILER_ADDRESS=127.0.0.1:6060
-NEOFS_PROFILER_SHUTDOWN_TIMEOUT=15s
+NEOFS_PROFILER_SHUTDOWN_TIMEOUT=30s
 
 NEOFS_METRICS_ADDRESS=127.0.0.1:9090
-NEOFS_METRICS_SHUTDOWN_TIMEOUT=15s
+NEOFS_METRICS_SHUTDOWN_TIMEOUT=30s
```
alexvanin, line 7: What's wrong with 15s timeouts?

carpawell: set default value

alexvanin: It's better to *avoid* default values in this config, but keep values sane in the same time.
Configuration test fetches some *non default* values and compare them correctly.
To check default values configuration test uses empty config.

carpawell: forget that there are tests for default values

rolled back and changed other default values

alexvanin (approved):

PR #720, opened 2021-07-23
Feature/Delete notary related envs

### LINKED ISSUE #709: Delete `WITHOUT_NOTARY` env option
It is possible to deduct if a `notary` is enabled or not on startup without explicitly reading this from config: check if native `notary` is presented in chain or not.

Delete that config env and start checking notary availability automatically on startup.

carpawell:
Closes #709.

### DIFF
--- a/pkg/innerring/innerring.go
+++ b/pkg/innerring/innerring.go
@@ -303,7 +303,6 @@ func New(ctx context.Context, log *zap.Logger, cfg *viper.Viper) (*Server, error
 
 	// parse notary support
 	server.feeConfig = config.NewFeeConfig(cfg)
-	server.mainNotaryConfig, server.sideNotaryConfig = parseNotaryConfigs(cfg)
 
 	// prepare inner ring node private key
 	acc, err := utilConfig.LoadAccount(
@@ -315,25 +314,6 @@ func New(ctx context.Context, log *zap.Logger, cfg *viper.Viper) (*Server, error
 	}
 
 	server.key = acc.PrivateKey()
-
-	withoutMainNet := cfg.GetBool("without_mainnet")
-
-	// get all script hashes of contracts
-	server.contracts, err = parseContracts(
-		cfg,
-		withoutMainNet,
-		server.mainNotaryConfig.disabled,
-		server.sideNotaryConfig.disabled,
-	)
-	if err != nil {
-		return nil, err
-	}
-
-	// parse default validators
-	server.predefinedValidators, err = parsePredefinedValidators(cfg)
-	if err != nil {
-		return nil, fmt.Errorf("ir: can't parse predefined validators list: %w", err)
-	}
 
 	morphChain := &chainParams{
 		log:  log,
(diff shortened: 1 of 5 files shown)


pkg/morph/client/constructor.go
```diff
@@ -105,12 +106,6 @@ func New(key *keys.PrivateKey, endpoint string, opts ...Option) (*Client, error)
 		waitInterval: cfg.waitInterval,
 	}
 
-	if len(cfg.notaryOpts) != 0 {
```
alexvanin, line 108: Should we remove `cfg.notaryOpts` then?

pkg/innerring/innerring.go
```diff
@@ -376,21 +352,57 @@ func New(ctx context.Context, log *zap.Logger, cfg *viper.Viper) (*Server, error
...
+		server.mainnetClient, err = createClient(ctx, mainnetChain)
 		if err != nil {
 			return nil, err
 		}
 	}
 
+	server.mainNotaryConfig, server.sideNotaryConfig = parseNotaryConfigs(
+		cfg,
+		server.morphClient.ProbeNotary(),
+		!withoutMainNet && server.mainnetClient.ProbeNotary(),
```
alexvanin, line 365: Isn't it will be the same thing in the end?
```suggestion
		server.mainnetClient.ProbeNotary(),
```

alexvanin (commented):
LGTM

pkg/morph/client/constructor.go
carpawell: looks like it is useless then

deleted

pkg/innerring/innerring.go
carpawell: if i remember correctly what i wanted to do:

if `withoutMainNet == true` and notary is enabled, then `server.mainnetClient` is `server.morphClient` in fact, then `ProbeNotary()` is called on sidechain in fact, then notary enabled in sidechain causes IR to require notary main contracts. it is not correct

alexvanin: > then notary enabled in sidechain causes IR to require notary main contracts. it is not correct

You mean `EnableNotarySupport()` will be called twice on morph chain client? If so, we can make `EnableNotarySupport()` to ignore if it was already enabled.

However I am fine with this solution if you add some comment about that. Looks a bit confusing at first glance.

carpawell: no-no, i mean, that the line that you suggest to change is arg to `parseNotaryConfigs` func

that arg defines if notary is presented in main and if it is necessary to parse additional envs

if use your variant in "`withoutMain==true`, notary in sidechain enabled" case, that arg is true since `server.mainnetClient.ProbeNotary()` is sidechain in fact => `server.mainNotaryConfig.disabled` is false => IR try to parse additional contracts hashes because of the line `372` and try to enable it because of the line `389`

carpawell: added commentary to that argument

alexvanin (approved):

PR #721, opened 2021-07-28
neofs-adm: generate storage wallets

fyrchik:
Close #685 .

PR #724, opened 2021-07-28
neofs-adm: implement `morph force-new-epoch`

### LINKED ISSUE #686: New epoch update command in neofs-adm
This command should look up to NNS contract (deployed with ID:0) and find address of network map contract. Then it should fetch current epoch number and prepare multi signed transaction of `netmap.NewEpoch()` with incremented argument. Look at [tick.sh](https://github.com/nspcc-dev/neofs-dev-env/blob/master/bin/tick.sh) in neofs-dev-env that does the same.

Passwords can be read from config file.

```
$ neofs-adm morph force-new-epoch [-c config.yaml] \
  -r https://neofs.private.corp.com:30333
  --alphabet-wallets ./alphabet-wallets
> Password for AZ wallet: 
> Password for BUKY wallet:
...
```

fyrchik:
Close #686 .
Allow to force epoch change. This is independent from #721 .

### DIFF
--- a/cmd/neofs-adm/internal/modules/morph/epoch.go
+++ b/cmd/neofs-adm/internal/modules/morph/epoch.go
@@ -1,13 +1,75 @@
 package morph
 
 import (
+	"errors"
+	"fmt"
+
+	nns "github.com/nspcc-dev/neo-go/examples/nft-nd-nns"
+	"github.com/nspcc-dev/neo-go/pkg/core/transaction"
+	"github.com/nspcc-dev/neo-go/pkg/io"
+	"github.com/nspcc-dev/neo-go/pkg/smartcontract"
+	"github.com/nspcc-dev/neo-go/pkg/smartcontract/callflag"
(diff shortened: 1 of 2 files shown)


cmd/neofs-adm/internal/modules/morph/root.go
```diff
@@ -87,4 +87,5 @@ func init() {
 	RootCmd.AddCommand(forceNewEpoch)
 	forceNewEpoch.Flags().String(alphabetWalletsFlag, "", "path to alphabet wallets dir")
 	forceNewEpoch.Flags().StringP(endpointFlag, "r", "", "N3 RPC node endpoint")
+	forceNewEpoch.Flags().String(contractsInitFlag, "", "path to archive with compiled NeoFS contracts")
```
alexvanin, line 90: I suggest to reconsider this flag. NNS should be enough to find contract script hashes, isn't it?

alexvanin (approved):

PR #727, opened 2021-07-28
Feature/Allow non executable permissions in config

### LINKED ISSUE #646: data file are executable
```
rei # l /srv/neofs/data0/blobovnicza/0/0 
-rwxr-x--- 1 neofs-storage neofs-storage 10489856 Jun 24 04:40 /srv/neofs/data0/blobovnicza/0/0
```

carpawell:
Closes #646.

### DIFF
--- a/cmd/neofs-node/config.go
+++ b/cmd/neofs-node/config.go
@@ -3,13 +3,12 @@ package main
 import (
 	"context"
 	"net"
-	"os"
 	"path"
 	"sync"
 	"time"
 
 	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
-	"github.com/nspcc-dev/neo-go/pkg/util"
+	neogoutil "github.com/nspcc-dev/neo-go/pkg/util"
 	"github.com/nspcc-dev/neofs-api-go/pkg"
 	apiclient "github.com/nspcc-dev/neofs-api-go/pkg/client"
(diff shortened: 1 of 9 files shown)


alexvanin:
@anatoly-bogatyrev @532910 @realloc Can you provide some feedback on what you would expect as a NeoFS administrator from  node when it have non-existent dir in config path to blob stor?

Now node creats new dir with the same rights as blob stor files. Unlike blob stor files, dirs should be executable. This PR adds `--x--x--x` rights for new dirs, so you can specify 644 for blob stor files and dir will be 755. *However* 600 will turn into 711 dir. Looks odd. 

Maybe  `--x------` for dirs will solve the issue?

532910:
1. Yes, dirs must be executable
2. Permissions should follow umask

carpawell:
according to https://github.com/nspcc-dev/neofs-node/issues/646#issuecomment-880055575, made `0660` as default(if `perm` is empty in config) permissions that would be also affected by `umask`. for dirs `--x--x---` is added(`0660` becomes `0770`)

if custom perms are presented in config, `--x--x---` is added for dirs and it is also affected by `umask`

is that expected behavior now?

alexvanin:
> is that expected behavior now?

Let's stick with that. In case of different proposals, we will consider them in different issues.

PR #730, opened 2021-07-29
Feature/add disable morph cache response option

### LINKED ISSUE #704: Toggle contract timeout in Storage Nodes
While testing NeoFS eACL,  I'm always frustrated when test has to wait about 30 seconds until Storage Node places next request to blockchain and fetches updated eACL. 

I'd like to reduce this timeout to, for instance, 1 second, or disable the cache totally. Another solution is to trigger the contract data fetch.

carpawell:
Closes #704.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ Changelog for NeoFS Node
 
 ## [Unreleased]
 
+### Upgrading from v0.23.0
+Added `NEOFS_MORPH_DISABLE_CACHE` env. If `true`, none of
+the `eACL`/`netmap`/`container` RPC responses cached.
+
 ## [0.23.0] - 2021-07-23 - Wando (완도, 莞島)
 
 Improved stability for notary disabled environment.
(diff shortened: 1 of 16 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,10 @@ Changelog for NeoFS Node
 
 ## [Unreleased]
 
+### Upgrading from v0.23.0
+Added `NEOFS_MORPH_DISABLE_CACHE` env for test purposes. If `true`,
```
alexvanin, line 7: Omit `test purpose`.

pkg/services/object/acl/eacl/types.go
```diff
@@ -6,16 +6,16 @@ import (
 	bearer "github.com/nspcc-dev/neofs-api-go/v2/acl"
 )
 
-// Storage is the interface that wraps
-// basic methods of extended ACL table storage.
-type Storage interface {
-	// GetEACL reads the table from the storage by identifier.
+// Source is the interface that wraps
+// basic methods of extended ACL table source.
+type Source interface {
```
alexvanin, line 11: Unnecessary change. Previous naming is completely fine.

carpawell: it sounds like implementation *must* store tables. so i found it strange to name it "storage" if it may be used not in cache context(does not store)

also, i wanted to rename similar interface in `netmap` but... it is already named as "Source"

roll back?

CHANGELOG.md
carpawell: deleted

alexvanin (approved):

PR #735, opened 2021-07-30
Refactor/Delete multiaddr TLS implementation

### LINKED ISSUE #661: Delete implementation of `TLS` protocol for `multiaddr` library
There has been added implementation of `TLS` protocol in #549 in https://github.com/nspcc-dev/neofs-node/pull/549/commits/6d475a95023806fe2aa186a29a95e2ba2ec3043c.

But now `TLS` was added to the original library, therefore this part of the code is useless and should be deleted. (Possibly it should be done after the next release in the `multiaddr` library).

carpawell:
Closes #661.

### DIFF
--- a/pkg/network/tls.go
+++ b/pkg/network/tls.go
@@ -1,53 +1,20 @@
 package network
 
 import (
-	"fmt"
-
 	"github.com/multiformats/go-multiaddr"
 )
 
-// There is implementation of TLS protocol for
-// github.com/multiformats/go-multiaddr library in this file.
-//
-// After addition TLS protocol via `multiaddr.AddProtocol` function
-// the library is ready to parse "tls" protocol with empty body, e.g.:
-//
-//		"/dns4/localhost/tcp/8080/tls"
-
 const (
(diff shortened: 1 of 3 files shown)


pkg/network/tls.go
```diff
@@ -1,48 +1,16 @@
...
-// the library is ready to parse "tls" protocol with empty body, e.g.:
-//
-//		"/dns4/localhost/tcp/8080/tls"
-
 const (
 	tlsProtocolName = "tls"
-
-	// tlsProtocolCode is chosen according to its draft version's code in
-	// original multiaddr repository: https://github.com/multiformats/multicodec.
 	tlsProtocolCode = 0x01c0
```
alexvanin, line 9: Unused?

carpawell: was used for searching for `tls` protocol in `TLSEnabled` method

but i missed that it is exported const in library, so deleted that and changed `TLSEnabled` method

alexvanin (approved):

PR #737, opened 2021-08-03
[#737] network/cache: handle request cancelled in multiclient

ZhangTao1596:
@alexvanin  Please have a look.


alexvanin:
Can you provide description of the use cases when this change affects node execution?

ZhangTao1596:
I just think we  shouldn't try another address when the error is caused by cancellation.

alexvanin (approved):

PR #764, opened 2021-08-24
adm: Implement command to refill storage node's GAS

### LINKED ISSUE #758: neofs-adm generate-storage-wallet - --initial-gas for existing wallet
It would be convenient to be able to transfer gas (--initial-gas) not only to a new wallet via generate-storage-wallet, but also to an existing one.
It can be implemented due to not change the wallet if it already exists and only makes a transfer.

cthulhu-rider:
Closes #758.

Add `refill-gas` sub-command to `morph` command which provides the ability
to refill storage node's GAS.

Command performs some actions from `generate-storage-wallet` runner, so
common code is moved to separate functions.

### DIFF
--- a/cmd/neofs-adm/internal/modules/morph/generate.go
+++ b/cmd/neofs-adm/internal/modules/morph/generate.go
@@ -117,7 +117,11 @@ func addMultisigAccount(w *wallet.Wallet, m int, name, password string, pubs key
 	return nil
 }
 
-func generateStorageCreds(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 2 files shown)


alexvanin (changes requested):
Do not optimize any cobra \ viper related bindings. They are explicit as possible in the whole neofs-adm utility. All cobra commands defined in `root.go`, use `init()` from there. I suggest you to follow pattern for now.

alexvanin (approved):

PR #765, opened 2021-08-24
services/policer: Select pseudo-random list of objects to check

### LINKED ISSUE #715: Select random objects for replication
Upon replication cycle, Storage Node lists stored objects, cuts first N ones and tries to replicate them. The issue is that `List` operation is consistent and selects objects sorted uniformly for every iteration. If Storage Node stores thousands of objects, and N is significantly less that object amount, this Node would perform replication for the same object every time. 

Shuffle listed objects and then select N for replication.

cthulhu-rider:
Closes #715.

In previous implementation of Policer's job queue the same list of objects for
processing was selected at each iteration. This was caused by consistent
return of `engine.List` function.

Use `rand.Shuffle` function to compose pseudo-random list of all objects in
order to approximately evenly distribute objects to work.

### DIFF
--- a/pkg/services/policer/queue.go
+++ b/pkg/services/policer/queue.go
@@ -3,6 +3,7 @@ package policer
 import (
 	"github.com/nspcc-dev/neofs-api-go/pkg/object"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/engine"
+	"github.com/nspcc-dev/neofs-node/pkg/util/rand"
 )
 
 type jobQueue struct {
@@ -18,6 +19,10 @@ func (q *jobQueue) Select(limit int) ([]*object.Address, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	rand.New().Shuffle(len(res), func(i, j int) {
+		res[i], res[j] = res[j], res[i]
+	})
 
 	if len(res) < limit {
 		return res, nil


pkg/services/policer/queue.go
```diff
@@ -1,6 +1,8 @@
 package policer
 
 import (
+	"math/rand"
```
alexvanin, line 4: We have an implementation of math/rand with cryptographic source, let's use it.
Examples: 
https://github.com/nspcc-dev/neofs-node/blob/8eadf18e015981dd8a11e4ce661e6fc2b5388a2e/pkg/services/audit/auditor/por.go#L70-L72
https://github.com/nspcc-dev/neofs-node/blob/0bcc5056828d8cc00bc921953ca5f8a682e22183/pkg/innerring/processors/audit/process.go#L76-L81

pkg/services/policer/queue.go
```diff
@@ -19,9 +21,16 @@ func (q *jobQueue) Select(limit int) ([]*object.Address, error) {
...
-		return res, nil
+	num := len(res)
+
+	if num > limit {
+		num = limit
+		res = res[:limit]
 	}
 
-	return res[:limit], nil
+	rand.Shuffle(num, func(i, j int) {
```
alexvanin, line 31: Shuffling before cutting provides more entropy, isn't it? 

If node stores 1 000 objects and it is able to replicate 100 of them, then it is going to replicate the same 100 objects but in different order. Not sure this is expected solution.

alexvanin (approved):

Issue #746, opened 2021-08-04
Storage nodes do not switch between RPC endpoints


alexvanin:
Storage node config provides a list of endpoints for Neo RPC client. Node creates client with random RPC endpoint. Unfortunately it did not switch to other endpoint even if RPC node is dead. 

In case of Listener component, it produces internal error and application is restarted with new connection. 
In case of the client, Storage node will just return errors on incoming NeoFS API requests that trigger side chain invocations.

While it is hard to maintain listener component switch, we can easily define multiclient component for RPC node (the same way we have multiclient of NeoFS API client).

Additionally, we can adopt this mulitclient in Inner Ring node application.

/cc @carpawell

cthulhu-rider:
Several approaches can be considered:
- forget about the node on failure 4ever;
- give it the lowest priority for the next time;
- do exactly like API multi-client: every time iterate over fixed list of endpoints.

@alexvanin

alexvanin:
> do exactly like API multi-client: every time iterate over fixed list of endpoints.

:+1:

alexvanin:
Closed in #769

PR #769, opened 2021-08-25
Implement and use morph multi-client

cthulhu-rider:
Relates to #746.

### DIFF
--- a/pkg/morph/client/client.go
+++ b/pkg/morph/client/client.go
@@ -79,6 +79,29 @@ var errEmptyInvocationScript = errors.New("got empty invocation script from neo 
 
 var errScriptDecode = errors.New("could not decode invocation script from neo node")
 
+// implementation of error interface for NeoFS-specific errors.
+type neofsError struct {
+	err error
+}
+
+func (e neofsError) Error() string {
+	return fmt.Sprintf("neofs error: %v", e.err)
+}
+
+// wraps NeoFS-specific error into neofsError. Arg must not be nil.
+func wrapNeoFSError(err error) error {
+	return neofsError{err}
+}
+
+// unwraps NeoFS-specific error if err is type of neofsError. Otherwise, returns nil.
+func unwrapNeoFSError(err error) error {
+	if e := new(neofsError); errors.As(err, e) {
+		return e.err
+	}
+
+	return nil
+}
+
 // Invoke invokes contract method by sending transaction into blockchain.
 // Supported args types: int64, string, util.Uint160, []byte and bool.
 func (c *Client) Invoke(contract util.Uint160, fee fixedn.Fixed8, method string, args ...interface{}) error {
@@ -121,11 +144,11 @@ func (c *Client) Invoke(contract util.Uint160, fee fixedn.Fixed8, method string,
 	}
 
 	if resp.State != HaltState {
-		return &notHaltStateError{state: resp.State, exception: resp.FaultException}
+		return wrapNeoFSError(&notHaltStateError{state: resp.State, exception: resp.FaultException})
 	}
 
 	if len(resp.Script) == 0 {
-		return errEmptyInvocationScript
+		return wrapNeoFSError(errEmptyInvocationScript)
 	}
 
 	script := resp.Script
@@ -178,7 +201,7 @@ func (c *Client) TestInvoke(contract util.Uint160, method string, args ...interf
 	}
 
 	if val.State != HaltState {
-		return nil, &notHaltStateError{state: val.State, exception: val.FaultException}
+		return nil, wrapNeoFSError(&notHaltStateError{state: val.State, exception: val.FaultException})
 	}
 
 	return val.Stack, nil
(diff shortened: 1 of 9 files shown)


pkg/morph/client/notary.go
```diff
@@ -213,15 +258,26 @@ func (c *Client) UpdateNeoFSAlphabetList(list keys.PublicKeys) error {
...
 		return c.Invoke(contract, fee, method, args...)
 	}
 
 	return c.notaryInvoke(false, contract, method, args...)
 }
 
-func (c *Client) notaryInvokeAsCommittee(contract util.Uint160, method string, args ...interface{}) error {
-	return c.notaryInvoke(true, contract, method, args...)
+func (c *Client) notaryInvokeAsCommittee(method string, args ...interface{}) error {
+	designate, err := c.GetDesignateHash()
```
alexvanin, line 275: Keep contract address in the `notaryInvokeAsCommittee` signature. The only usage of role management (designate) contract is just a case, not general rule.

alexvanin:
TLDR: can we try to distinguish client related errors and NeoFS related errors in iterator? If the error is NeoFS related, then do not iterate over clients and return error immediately.

`morph/client` returns various errors from the neo-go client and it's own as well. `Invoke()` has great example:
```go
	resp, err := c.client.InvokeFunction(contract, method, params, cosigner)
	if err != nil {
		return err // <- native client error
	}

	if resp.State != HaltState {
		return &notHaltStateError{state: resp.State, exception: resp.FaultException} // <- NeoFS related business logic error
	}

	if len(resp.Script) == 0 {
		return errEmptyInvocationScript // <- NeoFS related business logic error
	}
```

As for the business logic errors, we don't need to retry the request in all available neo-go clients, because it is not related to the remote node health or connection. As for neo-go client calls, we expect to have `err == nil` while they are healthy and connection is established. So, I propose to define `NeoRPCError` type (not final name), wrap all neo-go client errors and check `errors.As()` in the iterator. What do you think?

pkg/morph/client/notary.go
```diff
@@ -93,17 +102,27 @@ func (c *Client) EnableNotarySupport(opts ...NotaryOption) error {
 		alphabetSource: cfg.alphabetSource,
 	}
 
-	return nil
-}
+	// update client cache
+	for _, cached := range c.clients {
+		cached.notary = c.sharedNotary
+	}
```
alexvanin, line 108: This loop sets `sharedNotary` structure for already created clients. I suppose new clients will miss it?

cthulhu-rider:
> TLDR: can we try to distinguish client related errors and NeoFS related errors in iterator? If the error is NeoFS related, then do not iterate over clients and return error immediately. What do you think?

I think it would be very useful. Does it need to be done in separate PR?

alexvanin:
> I think it would be very useful. Does it need to be done in separate PR?

Either way is fine.

pkg/morph/client/notary.go
cthulhu-rider: New clients has same setter of the field in `createForAddress` method. Since `EnableNotarySupport` can be called at any time, I added this loop to ensure all clients work in the same mode. I did it as an alternative of pre-clearing the cache before turning on the notary.

pkg/morph/client/notary.go
cthulhu-rider: Ok, I'll return previous signature. I'd still implement the method with the proposed signature since two places of code to the same instructions. What would be the best name for it?

pkg/morph/client/client.go
```diff
@@ -37,8 +49,6 @@ type Client struct {
 
 	gas util.Uint160 // native gas script-hash
 
-	designate util.Uint160 // native designate script-hash
```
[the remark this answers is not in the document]
cthulhu-rider: Caching is implemented on neo-go client side, so I decided to remove the additional level.

pkg/morph/client/client.go
```diff
@@ -308,20 +406,36 @@ func toStackParameter(value interface{}) (sc.Parameter, error) {
 			result.Value = int64(0)
 		}
 	default:
-		return result, fmt.Errorf("chain/client: unsupported parameter %v", value)
+		return result, neofsError{fmt.Errorf("chain/client: unsupported parameter %v", value)}
```
carpawell, line 409: is there any logic when you choose b/w `noefsError{}` and `wrapNeoFSError()` or its a mistake?

pkg/morph/client/client.go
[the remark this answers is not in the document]
alexvanin: In fact, all native contracts cached in the neo-go client, including `gas` above, are cached. If you want to delete additional level, maybe we delete this level for all contracts? But I would prefer to see it in a different PR.

pkg/morph/client/notary.go
alexvanin: I see, all new clients created by `createForAddress` will already have shared notary, missed that
```go
	c := &Client{
		singleClient: &singleClient{
			notary:       x.sharedNotary,
		},
	}
```

pkg/morph/client/client.go
cthulhu-rider: Mistake, fixed

pkg/morph/client/client.go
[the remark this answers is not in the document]
cthulhu-rider: Done

alexvanin (approved):

carpawell (approved):

PR #783, opened 2021-08-31
Read object headers from write cache

### LINKED ISSUE #762: ObjectExists error when we try to put an object via S3-GW.
When we try to put an exactly non-existing object via S3 GW into NeoFS, we receive an error:
```
An error occurred (InternalError) when calling the PutObject operation (reached max retries: 4): object exists
```
But the object is created.

## Expected Behavior

Success response.

## Possible Solution

It happens when write-cache is on, we tried to turn off cache, and putting an object become successful. Also it seems like  S3 GW retries to send PutObject after success with the first PutObject.

http://realloc.spb.ru/tmp/lolcatz/

cthulhu-rider:
Closes #762.

Easiest to implement working solution is presented. I left some notes for future optimizations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Changelog for NeoFS Node
 
 ### Fixed
 - Work of a storage node with one Neo RPC endpoint instead of a list (#746).
+- Lack of support for HEAD operation on the object write cache (#762).
 
 ## [0.24.0] - 2021-08-30 Anmyeondo (안면도, 安眠島)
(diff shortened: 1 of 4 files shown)


alexvanin (approved):
Looks OK for the initial solution. Let's see if this fix anything in @masterSplinter01 bench.

carpawell (approved):

masterSplinter01:
Launched benchmark locally multiple times, all passed successfully :)

PR #785, opened 2021-08-31
Feature/Improve release process

carpawell:
Release binaries are not built in the clean reproducible environment. It would be safer to build binaries in container, the same way we do for docker images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,6 +78,16 @@ images: image-storage image-ir image-cli image-adm
 # Build dirty local Docker images
 dirty-images: image-dirty-storage image-dirty-ir image-dirty-cli image-dirty-adm
 
+# Run `make %` in Golang container
+docker/%:
+	docker run --rm -it \
+	-v `pwd`:/src \
+	-w /src \
+	-u "$$(id -u):$$(id -g)" \
+	--env HOME=/src \
+	golang:$(GO_VERSION) make $*
+
+
 # Run all code formatters
 fmts: fmt imports
 
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,7 @@ image-%:
 		-t $(HUB_IMAGE)-$*:$(HUB_TAG) .
 
 # Build all Docker images
-images: image-storage image-ir image-cli image-adm
+images: image-storage image-ir image-cli image-adm image-storage-testnet
 
 # Build dirty local Docker images
 dirty-images: image-dirty-storage image-dirty-ir image-dirty-cli image-dirty-adm
--- a/Makefile
+++ b/Makefile
@@ -10,12 +10,11 @@ HUB_IMAGE ?= nspccdev/neofs
 HUB_TAG ?= "$(shell echo ${VERSION} | sed 's/^v//')"
 
 GO_VERSION ?= 1.16
+ARCH = amd64
 
 BIN = bin
-DIRS = $(BIN)
-
 RELEASE = release
-ARCH = amd64
+DIRS = $(BIN) $(RELEASE)
 
 # List of binaries to build.
 CMDS = $(notdir $(basename $(wildcard cmd/*)))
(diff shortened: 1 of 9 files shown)


alexvanin (commented):
Works great. The only thing I am not sure: do we really need to build images in `make prepare-release`? Taking into account that image deploy will be automated by @sami-nspcc soon, I suppose we can leave only binary \ tar generation routine.

carpawell:
@alexvanin, sure

deleted all the thing about docker

realloc:
I'd suggest doing another thing instead:
- Add clean build in docker environment, the same way it's done for [neofs-spec](https://github.com/nspcc-dev/neofs-spec/blob/master/Makefile#L133)
- Leave release binaries creation to @sami-nspcc
- Let @sami-nspcc build and attach binaries for each new release page she will notice

carpawell:
@realloc I made a mistake: it is golang ALPINE image that does not contain `make` util. So I updated Makefile with `docker/%` target since, I think, it is OK to build binaries not with `alpine` image.

@alexvanin, @cthulhu-rider, please, take a look at https://github.com/nspcc-dev/neofs-node/pull/785/commits/13a5952723a776f92b1de460e35794b21cf9d6ed: I added updating go instruction and there is one more place that contains go version

docs/release-instruction.md
```diff
@@ -26,33 +28,36 @@ Use `vX.Y.Z` tag for releases and `vX.Y.Z-rc.N` for release candidates
...
-organization in dockerhub)
+Create images for `neofs-storage`, `neofs-ir` and `neofs-cli`, `neofs-adm` applications
+and push them into Docker Hub (so that releasing requires privileges in nspccdev
+organization in Docker Hub)
 
-    $ make images
-    $ docker push nspccdev/neofs-storage:0.12.0
-    $ docker push nspccdev/neofs-ir:0.12.0
-    $ docker push nspccdev/neofs-cli:0.12.0
+    $ make images && make image-storage-testnet
```
realloc, line 45: Can we add `image-storage-testnet` into the `images` target deps?

Makefile
```diff
@@ -9,14 +9,21 @@ DEBUG ?= false
...
-DIRS= $(BIN)
+DIRS = $(BIN)
+
+RELEASE = release
+ARCHS = amd64
 
 # List of binaries to build.
 CMDS = $(notdir $(basename $(wildcard cmd/*)))
 BINS = $(addprefix $(BIN)/, $(CMDS))
+RELEASE_BINS = $(foreach ARCH,$(ARCHS),$(addsuffix -$(ARCH), $(addprefix $(RELEASE)/, $(CMDS))))
```
realloc, line 23: Why it's a separate list? Shuldn't the regular `bin` and release `bin` be the same?

Makefile
```diff
@@ -37,6 +44,28 @@ $(DIRS):
...
 
+# Directory for release files
+$(RELEASE):
+	@echo "⇒ Ensure dir: $@"
+	@mkdir -p $@
+
+# Prepare binaries and archives for release
+prepare-$(RELEASE): $(RELEASE) $(RELEASE_BINS)
+
+$(RELEASE)/%: dep
```
realloc, line 55: This looks the same way as regular `dep`, apart from arch support. I suggest adding different arch support as a separate change when we need it. Currently, we support only amd64.

docs/release-instruction.md
carpawell: sure

Makefile
carpawell: because i think of `make all` as "I want to have neofs binaries in my local env for using it/testing it", so it shouldn't be with arch suffixes and I don't want to have all arch variations to be build

on the other hand, `make prepare-release` is something that shouldn't be used by devs every day, so I think it should be used separately

Makefile
carpawell: why can't it be like already done? do not see any problems with that

at least it is different because of different targets that do not intersect(`BINS` and `RELEASE_BINS`)

carpawell:
@realloc, @alexvanin, simplified release target

alexvanin (approved):

PR #787, opened 2021-08-31
Fix attribute parser issues

alexvanin:
This PR fixes:
1. inconsistent order of the returned attributes,
2. multiple attribute parent support.

As for (1), attributes does not require any special order, because they are linked together with `Parents` field. However for the tests it would be better to have consistent parsing function, so we won't see any 
```
             	Error Trace:	parser_test.go:67
            	Error:      	Not equal: 
            	            	expected: "Ke/y2"
            	            	actual  : "K:ey1"
```
errors.

As for (2), previously we didn't reuse existing attributes so the `Parents` list will have at most one record. 

It is better to review per commit than whole PR.

### DIFF
--- a/pkg/util/attributes/parser.go
+++ b/pkg/util/attributes/parser.go
@@ -48,7 +48,8 @@ func ParseV2Attributes(attrs []string, excl []string) ([]*netmap.NodeAttribute, 
 			key := pair[0]
 			value := pair[1]
 
-			if at, ok := cache[key]; ok && at.Value() != value {
+			attribute, present := cache[key]
+			if present && attribute.Value() != value {
 				return nil, errNonUniqueBucket
 			}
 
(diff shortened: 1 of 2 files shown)


pkg/util/attributes/parser.go
```diff
@@ -48,28 +48,35 @@ func ParseV2Attributes(attrs []string, excl []string) ([]*netmap.NodeAttribute,
...
+				value = replaceEscaping(value, true)
 
-			attribute := netmap.NewNodeAttribute()
-			attribute.SetKey(key)
-			attribute.SetValue(value)
+				attribute = netmap.NewNodeAttribute()
+				attribute.SetKey(key)
+				attribute.SetValue(value)
+
+				cache[key] = attribute
```
cthulhu-rider, line 70: Can `key` be changed by `replaceEscaping` above? If so, we have different key formats when read and write the cache, is it ok?

cthulhu-rider:
There is a typo in [commit](https://github.com/nspcc-dev/neofs-node/pull/787/commits/169695838221e783ffd72b5ac1b66745d3bc24d4) header.

pkg/util/attributes/parser.go
alexvanin: It is definitely not okay. I will add a test for that and will fix it.

alexvanin: Done in https://github.com/nspcc-dev/neofs-node/pull/787/commits/45f40bea1140819d90d9ef12e0339e3af4659c40

carpawell (approved):

cthulhu-rider (approved):

PR #798, opened 2021-09-06
Do not listen RPC nodes in transition state

### LINKED ISSUE #795: Do not listen events from unhealthy Neo RPC nodes
Consider Neo RPC node unhealthy if it is restoring the chain. In this case Inner Ring and Storage nodes may receive events from the past to handle. Since Inner Ring application mostly stateless, it can't filter out such events. The simplest way to fix it --- wait for the chain to be restored or choose other Neo RPC endpoint.

## How to check Neo RPC node state
Unfortunately, there is no RPC to check node status. Possible solution is to monitor block height for short period of time and classify the state based on the speed of height change, e.g.: healthy if there are less than 2 new blocks per second. 

/cc @AnnaShaleva @roman-khimov @fyrchik

alexvanin:
This PR fixes #795 and introduces several changes:

* Subscriber component can wait specified time duration before RPC node return block height equal or higher than specified minimal height.
* Added bbolt.DB wrapper for persistent thread-safe state storage.
* Added config record for both Storage and Inner Ring applications with the path to state storage.
```yaml
node:
  persistent_state:
    path: /srv/node/.state
```
* Storage and Inner Ring applications will use persistent state to save latest processed block number for all available chains.
* At the initialization stage, Storage and Inner Ring applications will wait up to 10 seconds for RPC node to return block height equal or higher than previously saved latest processed block number.

--- 

RPC node transition state may affect not only listener but client as well. However this PR does not solves this issue, because client (or mulitclient) does not initialized straight away. It also may switch between connections so the block height awaiting routine might become a bit more complex for client than for the listener.


cmd/neofs-node/morph.go
```diff
@@ -99,11 +99,19 @@ func listenMorphNotifications(c *cfg) {
...
+
 	for i := range endpoints {
 		subs, err = subscriber.New(c.ctx, &subscriber.Params{
-			Log:         c.log,
-			Endpoint:    endpoints[i],
-			DialTimeout: timeout,
+			Log:            c.log,
+			Endpoint:       endpoints[i],
+			DialTimeout:    timeout,
+			RPCInitTimeout: 10 * time.Second,
```
carpawell, line 113: what's the point in using 10 seconds timeout? can neo-go finish syncing in that time?

pkg/morph/subscriber/subscriber.go
```diff
@@ -188,3 +199,29 @@ func New(ctx context.Context, p *Params) (Subscriber, error) {
...
+			return fmt.Errorf("could not init ws client: didn't reach expected height %d", startFrom)
+		default:
+		}
+		height, err := wsClient.GetBlockCount()
+		if err != nil {
+			return fmt.Errorf("could not get block height: %w", err)
+		} else if height >= startFrom {
+			return nil
+		}
+		time.Sleep(100 * time.Millisecond)
```
carpawell, line 225: isn't it too frequent?

carpawell (approved):

cmd/neofs-node/morph.go
alexvanin: Removed in https://github.com/nspcc-dev/neofs-node/pull/798/commits/b52d3fcbf6d8800b0952b784a1abba50134ac2a7

pkg/morph/subscriber/subscriber.go
alexvanin: Removed in https://github.com/nspcc-dev/neofs-node/pull/798/commits/b52d3fcbf6d8800b0952b784a1abba50134ac2a7

alexvanin:
After consulting with @532910 , https://github.com/nspcc-dev/neofs-node/pull/798/commits/b52d3fcbf6d8800b0952b784a1abba50134ac2a7 removes RPC Init timeout. In the case of remote node failure, NeoFS app usually fails straight away. There is no point in changing that for subscriber component. It also removes any issues with default timeout duration.

PR #799, opened 2021-09-06
Unify storage engine logs

### LINKED ISSUE #790: Unify logs of data storage component updates
Every component that stores data should log all writing operation related to this data (put, delete). It is better to unify the log format for these operations.

Storage сomponents:
- [x] blobstor blobovniczas
- [x] blobstor fstree
- [x] writecache bbolt
- [x] writecache fstree
- [x] writecache in-mem

It is important to log both internal operations and external operations. E.g. writecache in-memory can drop the object because of the `Delete` invocation and because of persist loop. Both cases must be logged.

cthulhu-rider:
Closes #790.

Unified message head is hard-coded to `local object storage operation`. Log level is hard-coded to `Info`.

### DIFF
--- a/pkg/local_object_storage/blobstor/blobovnicza.go
+++ b/pkg/local_object_storage/blobstor/blobovnicza.go
@@ -12,6 +12,7 @@ import (
 	objectSDK "github.com/nspcc-dev/neofs-api-go/pkg/object"
 	"github.com/nspcc-dev/neofs-node/pkg/core/object"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobovnicza"
+	storagelog "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/internal/log"
 	"go.uber.org/zap"
 )
 
@@ -181,10 +182,7 @@ func (b *blobovniczas) put(addr *objectSDK.Address, data []byte) (*blobovnicza.I
 
 		id = blobovnicza.NewIDFromBytes([]byte(p))
 
-		b.log.Debug("object successfully saved in active blobovnicza",
-			zap.String("path", p),
-			zap.Stringer("addr", addr),
-		)
+		storagelog.Write(b.log, storagelog.AddressField(addr), storagelog.OpField("blobovniczas PUT"))
 
 		return true, nil
 	}
@@ -262,7 +260,7 @@ func (b *blobovniczas) delete(prm *DeleteSmallPrm) (res *DeleteSmallRes, err err
 			return nil, err
 		}
 
-		return b.deleteObject(blz, bPrm)
+		return b.deleteObject(blz, bPrm, prm)
 	}
 
 	activeCache := make(map[string]struct{})
@@ -273,7 +271,7 @@ func (b *blobovniczas) delete(prm *DeleteSmallPrm) (res *DeleteSmallRes, err err
 		// don't process active blobovnicza of the level twice
 		_, ok := activeCache[dirPath]
 
-		res, err = b.deleteObjectFromLevel(bPrm, p, !ok)
+		res, err = b.deleteObjectFromLevel(bPrm, p, !ok, prm)
 		if err != nil {
 			if !errors.Is(err, object.ErrNotFound) {
 				b.log.Debug("could not remove object from level",
@@ -351,7 +349,7 @@ func (b *blobovniczas) getRange(prm *GetRangeSmallPrm) (res *GetRangeSmallRes, e
 // tries to delete object from particular blobovnicza.
 //
 // returns no error if object was removed from some blobovnicza of the same level.
-func (b *blobovniczas) deleteObjectFromLevel(prm *blobovnicza.DeletePrm, blzPath string, tryActive bool) (*DeleteSmallRes, error) {
+func (b *blobovniczas) deleteObjectFromLevel(prm *blobovnicza.DeletePrm, blzPath string, tryActive bool, dp *DeleteSmallPrm) (*DeleteSmallRes, error) {
 	lvlPath := path.Dir(blzPath)
 
 	log := b.log.With(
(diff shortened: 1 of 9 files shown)


pkg/local_object_storage/internal/log/log.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/nspcc-dev/neofs-node/pkg/util/logger"
+	"go.uber.org/zap"
+)
+
+// a distinctive part of all messages
+const headMsg = "local object storage operation"
+
+// Write writes message about storage engine's operation to logger.
+//
+// Address should be type of *object.Address or string.
```
carpawell, line 13: how do logs of `*object.Address` look?

wouldn't it be better to use type switch?

cthulhu-rider: Type switch is done in `zap.Any`, `*object.Address` is processed as `fmt.Stringer`. I don't think we need to do explicit cast on our side.

carpawell: ok

pkg/local_object_storage/blobstor/blobovnicza.go
```diff
@@ -262,7 +260,7 @@ func (b *blobovniczas) delete(prm *DeleteSmallPrm) (res *DeleteSmallRes, err err
 			return nil, err
 		}
 
-		return b.deleteObject(blz, bPrm)
+		return b.deleteObject(blz, bPrm, prm.addr)
```
carpawell, line 263: why not just add getter to delete parameter `bPrm`?

as I see you put the same address twice: one pure and one wrapped in `bPrm`

carpawell (approved):

pkg/local_object_storage/blobstor/blobovnicza.go
cthulhu-rider: Because we have access to address on this level so pull it from the lower is redundant

carpawell: still do not get the idea of wrapping and passing something and just passing it at the same time

but ok

pkg/local_object_storage/writecache/persist.go
```diff
@@ -64,6 +65,7 @@ func (c *cache) persistToCache(objs []objectInfo) []int {
 			}
 			sz += uint64(len(objs[i].data))
 			doneMem = append(doneMem, i)
+			storagelog.Write(c.log, objs[i].addr, "db PUT")
```
alexvanin, line 68: Aren't those objects are also deleted from memory?

pkg/local_object_storage/blobstor/blobovnicza.go
```diff
@@ -544,12 +542,14 @@ func (b *blobovniczas) getRangeFromLevel(prm *GetRangeSmallPrm, blzPath string,
...
 
 // removes object from blobovnicza and returns DeleteSmallRes.
-func (b *blobovniczas) deleteObject(blz *blobovnicza.Blobovnicza, prm *blobovnicza.DeletePrm) (*DeleteSmallRes, error) {
+func (b *blobovniczas) deleteObject(blz *blobovnicza.Blobovnicza, prm *blobovnicza.DeletePrm, addr *objectSDK.Address) (*DeleteSmallRes, error) {
 	_, err := blz.Delete(prm)
 	if err != nil {
 		return nil, err
 	}
 
+	storagelog.Write(b.log, addr, "blobovniczas DELETE")
```
alexvanin, line 551: Can we specify blobovnicza id there?

alexvanin (commented):
Log examples in the PR description will be appreciated.

pkg/local_object_storage/blobstor/blobovnicza.go
cthulhu-rider: Sure

pkg/local_object_storage/internal/log/log.go
```diff
@@ -0,0 +1,19 @@
...
+func Write(logger *logger.Logger, addr interface{}, op string) {
```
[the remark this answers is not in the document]
cthulhu-rider: I thought about it. We have `component` field on each storage component' log, e.g. `{..., "component": "BlobStor", ...}`. Do we need to add one more field?

pkg/local_object_storage/writecache/persist.go
cthulhu-rider: Only if the method is called from `persistLoop`. I'll add log there.

pkg/local_object_storage/blobstor/blobovnicza.go
```diff
@@ -544,12 +542,18 @@ func (b *blobovniczas) getRangeFromLevel(prm *GetRangeSmallPrm, blzPath string,
 }
 
 // removes object from blobovnicza and returns DeleteSmallRes.
-func (b *blobovniczas) deleteObject(blz *blobovnicza.Blobovnicza, prm *blobovnicza.DeletePrm) (*DeleteSmallRes, error) {
+func (b *blobovniczas) deleteObject(blz *blobovnicza.Blobovnicza, prm *blobovnicza.DeletePrm, dp *DeleteSmallPrm) (*DeleteSmallRes, error) {
```
alexvanin, line 545: `blobovnicza.DeletePrm` looks a bit useless after this change, but if you feel ok about it, I don't mind to keep it.

alexvanin (approved):

pkg/local_object_storage/blobstor/blobovnicza.go
cthulhu-rider: It is accepted as parameter because we call `deleteObject` several times with the same args. The best way would be to keep all parameters in a single context structure, and pass it, changing it if necessary. But now it's easier to just pass an additional argument.

PR #801, opened 2021-09-07
Fix/Add options to netmap wrapper constructor

### LINKED ISSUE #793: `Netmap` Morph wrapper does not have `TryNotary` option
Add `TryNotary` option to the 4-th morph client layer (wrapper) and use it instead of 3-the client's option in IR constructor.

carpawell:
Closes #793.

Also fixes `errcheck` linter warning and adds `AsAlphabet` option.

### DIFF
--- a/pkg/morph/client/netmap/wrapper/wrapper.go
+++ b/pkg/morph/client/netmap/wrapper/wrapper.go
@@ -22,9 +22,25 @@ type Wrapper struct {
 	client *netmap.Client
 }
 
+// Option allows to set an optional
+// parameter of Wrapper.
+type Option func(*opts)
+
+type opts []client.StaticClientOption
+
+func defaultOpts() *opts {
+	return new(opts)
+}
+
 // NewFromMorph returns the wrapper instance from the raw morph client.
-func NewFromMorph(cli *client.Client, contract util.Uint160, fee fixedn.Fixed8, opts ...client.StaticClientOption) (*Wrapper, error) {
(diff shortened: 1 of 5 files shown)


pkg/morph/client/netmap/wrapper/wrapper.go
```diff
@@ -22,9 +22,25 @@ type Wrapper struct {
 	client *netmap.Client
 }
 
+// Option allows to set an optional
+// parameter of ClientWrapper.
```
cthulhu-rider, line 26:
```suggestion
// parameter of Wrapper.
```

cthulhu-rider, line 26:
```suggestion
// parameter of Wrapper.
```

cthulhu-rider (approved):

pkg/morph/client/netmap/wrapper/wrapper.go
carpawell: do you mean to change it in every wrapper?

i copy-pasted that from other contract wrappers

cthulhu-rider: Type is named `Wrapper`, so the doc should follow it.

alexvanin (approved):

pkg/morph/client/netmap/wrapper/wrapper.go
carpawell: ok, rewrited commentary and add commit that fixes the same for other wrappers

PR #807, opened 2021-09-08
Feat/Add notary notification for container contract

### LINKED ISSUE #731: Failed to set-eacl if its size exceeds 1024 bytes
If size of eacl table exceeds 1024 bytes it cannot be set. Such table, for example, may contain:
* 7 allowed records for specific public key
* 7 allowed records for specific public key with filter by `FileName` attribute
* 5 (other than `put` and `delete`) allowed records for  `others` role with filter by `FileName` attribute
* 7 denied records for `others` role

## Expected Behavior
EACL has been persisted on sidechain

## Current Behavior
```
rpc error: transport error: rpc error: code = Unknown desc = could not handle request: could not handle request: could not execute SetEACL request: could not invoke method (setEACL): chain/client: contract execution finished with state FAULT; exception: error encountered at instruction 3045 (SYSCALL): failed to invoke syscall 1634664853: notification size shouldn't exceed 1024
```

## Steps to Reproduce

1. Try to set-eacl that exceeds 1024 bytes: ` ./bin/neofs-cli -w wallet.json --address NbUgTSFvPmsRxmGeWpuuGeJUoRoi6PErcM  -r s01.neofs.devenv:8080 --cid 6Dk9CTMhRY5M8UXfay3KauKgK9i9khWCRDo8hwDT9Bhv container set-eacl --table eacl.json --await
`

## Your Environment
* Version used: v0.22.1
* Operating System and version (`uname -a`): Linux 5.8.0-59-generic # 66~20.04.1-Ubuntu SMP Thu Jun 17 11:14:10 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

carpawell:
Closes #731.

Also: 
- merges `ContractProcessor` and `NotaryContractProcessor` interfaces;
- adds notary notification support for `delete` container operation;
- improves logging in notary listener;
- refactors binding processors and parsers/handlers.

Merge after #814.

### DIFF
--- a/pkg/morph/event/listener.go
+++ b/pkg/morph/event/listener.go
@@ -368,6 +368,11 @@ func (l listener) parseAndHandleNotary(nr *response.NotaryRequestEvent) {
 		return
 	}
 
+	log := l.log.With(
+		zap.String("contract", notaryEvent.ScriptHash().StringLE()),
+		zap.Stringer("method", notaryEvent.Type()),
+	)
+
 	notaryKey := notaryRequestTypes{}
 	notaryKey.SetMempoolType(nr.Type)
 	notaryKey.SetRequestType(notaryEvent.Type())
@@ -379,7 +384,7 @@ func (l listener) parseAndHandleNotary(nr *response.NotaryRequestEvent) {
 	l.mtx.RUnlock()
 
 	if !ok {
-		l.log.Debug("notary parser not set")
+		log.Debug("notary parser not set")
 
 		return
 	}
@@ -387,7 +392,7 @@ func (l listener) parseAndHandleNotary(nr *response.NotaryRequestEvent) {
 	// parse the notary event
 	event, err := parser(notaryEvent)
 	if err != nil {
-		l.log.Warn("could not parse notary event",
+		log.Warn("could not parse notary event",
 			zap.String("error", err.Error()),
 		)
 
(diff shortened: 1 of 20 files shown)


alexvanin:
Check if container size announcements works correctly. I suppose they are not, because Storage node use container wrapper with `TryNotary()` option to invoke `putContainerSize` method. This method is not supposed to be invoked with notary service.

carpawell:
@alexvanin, yes, it seems that `putContainerSize` is the only one method that SN invokes and that does not work with notary, fixed that in https://github.com/nspcc-dev/neofs-node/pull/807/commits/a38dfd3d3c318811f71a783de661c998d7f670ea

also, please, take a look at https://github.com/nspcc-dev/neofs-node/pull/807/commits/704788e06df05f71f722c018cd71f939c326b40e, I added unexpected opcodes check to `Put` parser and reuse that approach in `setEACL` parser

also improved a little logging in https://github.com/nspcc-dev/neofs-node/pull/807/commits/8862d5e361402a3276b1e07440d31d24ed2cbb0d

pkg/morph/event/listener.go
```diff
@@ -368,6 +368,11 @@ func (l listener) parseAndHandleNotary(nr *response.NotaryRequestEvent) {
 		return
 	}
 
+	log := l.log.With(
+		zap.String("contract_LE", notaryEvent.ScriptHash().StringLE()),
```
alexvanin, line 372: I suggest to assume that all string values are `LE`, because it is human readable format by default for Neo related things. Or you really want to emphasize encoding?

alexvanin (approved):

pkg/morph/event/listener.go
carpawell: no, i do not want

just saw "script hash LE"

deleted "LE"

PR #809, opened 2021-09-08
Limit write-cache size

### LINKED ISSUE #776: Improve write cache size counting
`dbSize` was initially introduced when only db was used (no fs-tree). There were another problem with it: estimating database size based on size of keys is approximate and is less than the real size. If there is an error writing object in writecache db (e.g. no space), we just write in blobstore.
I think the best way to handle this is to separately count number of objects stored in a db and in a file-system and perform a rough estimation of how much space is currently taken. This info should be used in `Put` and persist-related routines.

_Originally posted by @fyrchik in https://github.com/nspcc-dev/neofs-node/issues/734#issuecomment-906407121_

cthulhu-rider:
Closes #776.

I'd prefer to remove DB size counter in a separate PR.


pkg/local_object_storage/writecache/persist.go
```diff
@@ -79,7 +92,33 @@ func (c *cache) persistToCache(objs []objectInfo) []int {
...
+		)
+
+		if failMem[i] < 0 {
+			// actually, since the overflow was detected in DB tx, the required space could well have been freed,
+			// but it is easier to consider the entire method atomic
+			overflowInd = -failMem[i]
+		} else {
+			// check if object will overflow write-cache size limit
+			if updCacheSz = c.incSizeFS(cacheSz); updCacheSz > c.maxCacheSize {
+				overflowInd = i
```
alexvanin, line 110: It looks a bit suspicious when `overflowInd` sets the index of `failMem` slice there, but it used as the index of `objs` slice down there. Is it correct?

pkg/local_object_storage/writecache/state.go
```diff
@@ -0,0 +1,103 @@
...
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("could not read write-cache DB counter: %w", err)
+	}
+
+	x.cDB.Store(inDB)
+
+	// FIXME: calculate the actual value in FSTree (new method?).
+	//  For now we can think that db/fs = 50/50.
```
alexvanin, line 95: Issue for that?

alexvanin (commented):
Feedback from @fyrchik will be appreciated.

pkg/local_object_storage/writecache/persist.go
cthulhu-rider: It is correct here: `i` is not an index of `failMem` slice but the value. But the remark helped to find the error above - in `if` block. Renamed to `objInd` for readability and fixed.

pkg/local_object_storage/writecache/state.go
cthulhu-rider: https://github.com/nspcc-dev/neofs-node/issues/821

pkg/local_object_storage/writecache/persist.go
```diff
@@ -46,24 +46,37 @@ func (c *cache) persistLoop() {
...
 			if uint64(len(objs[i].data)) >= c.smallObjectSize {
 				failMem = append(failMem, i)
 				continue
 			}
 
+			// check if object will overflow write-cache size limit
+			updCacheSz := c.incSizeDB(cacheSz)
+			if updCacheSz > c.maxCacheSize {
+				// set negative index
+				failMem = append(failMem, -i)
```
alexvanin, line 66: Are there possible issues with `-0` index?

pkg/local_object_storage/writecache/persist.go
```diff
@@ -79,15 +92,45 @@ func (c *cache) persistToCache(objs []objectInfo) []int {
...
+			overflowInd = -objInd
+		} else {
+			// check if object will overflow write-cache size limit
+			if updCacheSz = c.incSizeFS(cacheSz); updCacheSz > c.maxCacheSize {
+				overflowInd = objInd
+			}
+		}
+
+		if overflowInd >= 0 {
+			for j := range objs[overflowInd:] {
```
alexvanin, line 115: If I understand correctly, `objs` will contain both `doneMem` and `failMem` objects at this point of execution. In this case with `range objs[overflowInd:]`, you can mark some objects from `doneMem` slice. 

Example: consider `objs [0, 1, 2, 3]` where `1 2` --  big objects, `0 3` -- small objects. bbolt.DB is exhausted. Then `failMem` contains [-1,-2]. Therefore objs[1:] will be added to `failDisk` including 3, which might be incorrect.

pkg/local_object_storage/writecache/persist.go
cthulhu-rider: Sure, we should handle zero case.

pkg/local_object_storage/writecache/persist.go
cthulhu-rider: Yes, we need to exclude indices from `doneMem`, fixed.

alexvanin (approved):

PR #813, opened 2021-09-08
Return the composite error of PUT operation

### LINKED ISSUE #778: Expand `incomplete object put` error
If all container nodes refuse to store the object, node throws `incomplete object put` error. While we are lacking of proper error handling, wrapping latest error message in the chain of put request is better than doing absolutely nothing.

cthulhu-rider:
Closes #778.

### DIFF
--- a/pkg/services/object/put/distributed.go
+++ b/pkg/services/object/put/distributed.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"sync/atomic"
 
 	"github.com/nspcc-dev/neofs-node/pkg/core/object"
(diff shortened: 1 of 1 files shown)


pkg/services/object/put/distributed.go
```diff
@@ -128,6 +132,10 @@ loop:
 	}
 
 	if !traverser.Success() {
+		if err, ok := resErr.Load().(error); ok {
+			return nil, err
```
alexvanin, line 136: Maybe wrap it around `errIncompletePut`? This is useful error, node tried all placements and haven't succeed, but requires a bit more details from `resErr` if they are available.

alexvanin (approved):

PR #824, opened 2021-09-13
Metabase restoration

### LINKED ISSUE #789: Support metabase reinitalization
Storage data or metabase might be corrupted because of external reasons. In this case Storage node may provide inconsistent responses of object service request, e.g. successful HEAD, but failed GET.  We should have an option to reinitialize and sync metabases within available shards. 

Proposal: 
- define metabase restore function in the shard, that goes over all available blobovnizcas, fstrees, writecache files, reads all the objects and translate them over metabase data,
- define storage engine function, that goes through selected shards and calls metabase restore function.

To initialize metabase restore we can:
1. support restore operation in the Storage node start if it is configured this way,
2. add separate `neofs-node` command to restore the data,
3. add separate `neofs-cli` command to restore the data.

I would go with (1), but I am a bit worried that it might took too long for the storage node to startup. Maybe startup and restore should be different independent operations.

Questions:
- Should we rebuild database from the start or try to sync available data? As for me, syncing and merging seems quite complex than empty start, so I prefer complete rebuild.
- How object parsing order going to affect the resulting metabase?

Related to https://github.com/nspcc-dev/neofs-node/issues/394 but implying repairer is a runtime worker and metabase reinitialization is done before Store node start.

cthulhu-rider:
Closes #789.


alexvanin (approved):

pkg/local_object_storage/blobstor/iterate.go
```diff
@@ -27,7 +48,78 @@ func (i *IteratePrm) SetIterationHandler(h IterationHandler) {
...
+	return err
+}
+
+// IterateObjects is a helper function which iterates over BlobStor and passes decoded objects to f.
+func IterateObjects(blz *BlobStor, f func(obj *object.Object, blzID *blobovnicza.ID) error) error {
+	var obj *object.Object
+
+	return IterateBinaryObjects(blz, func(data []byte, blzID *blobovnicza.ID) error {
+		if obj == nil {
+			obj = object.New()
```
carpawell, line 116: why not just create obj above and do not do such check once per obj?

carpawell (approved):

pkg/local_object_storage/blobstor/iterate.go
cthulhu-rider: Because handler can never be called.

PR #828, opened 2021-09-15
IR: Support multiple notification endpoints

### LINKED ISSUE #792: Support multilistener component for Neo events
We must check if listeners are rotated during initial start. 

Additionally we can try to rotate listeners in runtime, however we decided not to do that in initial connection pool task: https://github.com/nspcc-dev/neofs-node/issues/157

cthulhu-rider:
Closes #792.

### DIFF
--- a/pkg/innerring/innerring.go
+++ b/pkg/innerring/innerring.go
@@ -879,14 +879,34 @@ func New(ctx context.Context, log *zap.Logger, cfg *viper.Viper) (*Server, error
 }
 
 func createListener(ctx context.Context, p *chainParams) (event.Listener, error) {
-	sub, err := subscriber.New(ctx, &subscriber.Params{
-		Log:            p.log,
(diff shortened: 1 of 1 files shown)


pkg/innerring/innerring.go
```diff
@@ -879,14 +879,34 @@ func New(ctx context.Context, log *zap.Logger, cfg *viper.Viper) (*Server, error
...
+			StartFromBlock: p.from,
+		})
+		if err == nil {
+			break
+		}
+
+		p.log.Info("failed to establish websocket neo event listener, trying another",
+			zap.String("endpoint", endpoints[i]),
+			zap.String("error", err.Error()),
+		)
```
alexvanin, line 909: If none of the endpoints are healthy, do we fail at `event.NewListener()`?

cthulhu-rider: Yes, constructor returns an error on nil `Subscriber`.

alexvanin (approved):

PR #833, opened 2021-09-20
Support NetworkInfo extension

cthulhu-rider:


alexvanin:
Is this feature requires update on the server side neo-go RPC node? I want to merge it, but if it does not work with neo-go v0.97.2, then we can't use it until the release.

/cc @cthulhu-rider @fyrchik

PR #846, opened 2021-09-24
Fix/Add random nonce to SN's notary request

### LINKED ISSUE #843: Use random nonce at non-alphabet notary invocations
Notary transaction data may be the same when storage node repeats the same contract method invocation (for example set empty extneded ACL list). Constant nonce produces `<tx> already on chain` errors in this case.

- [ ] Randomize `nonce` when notary transaction created at storage node
- [ ] Inherit `nonce` from storage node `AddPeer` transaction in netmap processor
- [ ] Use current epoch in `UpdatePeerState` at `processNetmapCleanupTick()`

carpawell:
Fixes #843.

### DIFF
--- a/pkg/morph/client/notary.go
+++ b/pkg/morph/client/notary.go
@@ -16,6 +16,7 @@ import (
 	"github.com/nspcc-dev/neo-go/pkg/util"
 	"github.com/nspcc-dev/neo-go/pkg/vm/opcode"
 	"github.com/nspcc-dev/neo-go/pkg/wallet"
+	"github.com/nspcc-dev/neofs-node/pkg/util/rand"
 	"go.uber.org/zap"
 )
 
@@ -230,6 +231,7 @@ func (c *Client) UpdateNotaryList(list keys.PublicKeys) error {
 
 	return c.notaryInvokeAsCommittee(
 		setDesignateMethod,
+		1, // FIXME: do not use constant nonce for alphabet NR: #844
 		noderoles.P2PNotary,
 		list,
 	)
@@ -253,6 +255,7 @@ func (c *Client) UpdateNeoFSAlphabetList(list keys.PublicKeys) error {
 
 	return c.notaryInvokeAsCommittee(
 		setDesignateMethod,
+		1, // FIXME: do not use constant nonce for alphabet NR: #844
 		noderoles.NeoFSAlphabet,
 		list,
 	)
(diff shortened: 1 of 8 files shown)


pkg/morph/client/notary.go
```diff
@@ -263,18 +266,18 @@ func (c *Client) UpdateNeoFSAlphabetList(list keys.PublicKeys) error {
 // it fallbacks to a simple `Invoke()`.
 //
 // This function must be invoked with notary enabled otherwise it throws panic.
-func (c *Client) NotaryInvoke(contract util.Uint160, fee fixedn.Fixed8, method string, args ...interface{}) error {
+func (c *Client) NotaryInvoke(contract util.Uint160, fee fixedn.Fixed8, nonce uint32, method string, args ...interface{}) error {
```
cthulhu-rider, line 269: Isn't it time to structure the parameters in `NotaryInvokePrm`?

cthulhu-rider (approved):

pkg/morph/client/notary.go
carpawell: i think it is a good idea, but also I think that it would be good to keep `Invoke` and `NotaryInvoke` funcs in sync so add `Prm` structs for both of them or even it should be the same struct

/cc @alexvanin

cthulhu-rider: Then you can define `InvokePrm`, and inherit it in `NotaryInvokePrm`. I think single type is not a good idea.

pkg/morph/client/notary.go
```diff
@@ -3,6 +3,7 @@ package client
 import (
 	"errors"
 	"fmt"
+	"math/rand"
```
alexvanin, line 6: Use math rand with cryptographic source from util package there. I don't like any use of default `math/rand` where random values matter.

```go
import "github.com/nspcc-dev/neofs-node/pkg/util/rand"
rand.New().Uint32()

carpawell: ok, is it ok to call `New` every time I need rand number or it can be global var in `client` pgk?

alexvanin: Sure it is better to be a global singleton.

carpawell: changed dependency

alexvanin (approved):

PR #850, opened 2021-09-24
Fix erroneous deletion marks

### LINKED ISSUE #840: Object removed when extended ACL restricts delete
There is a heisenbug when object considered removed even though delete operation was restricted by extended ACL

## Expected Behavior
1. Create public container with `REP 2 IN X CBF 1 SELECT 4 FROM * AS X` policy for neofs-dev-env
2. Upload two objects that has to be split
3. Delete one object using generated key successfully
4. Set extended ACL that restricts DELETE operation for all users
5. Try to delete another object and see `access to operation DELETE is denied by extended ACL check`
6. Get another object successfully

## Current Behavior
On the step (6) now we see `could not handle request: object already removed` error.
This does not happen always, so I suppose it may depend on object placement.

## Possible Solution
Even though DELETE operation was restricted, storage node probably still send tombstones that produce `object already removed` error. We should look into that.

## Your Environment
latest neofs-dev-env with max_object_size=1000
neofs-node from master branch

### LINKED ISSUE #849: policer: Log redunant copy occasion
It'd be useful to write log message when storage node detects redundant local object copy.

cthulhu-rider:
Closes #840.
Closes #849.

### DIFF
--- a/pkg/local_object_storage/metabase/get.go
+++ b/pkg/local_object_storage/metabase/get.go
@@ -85,8 +85,13 @@ func (db *DB) get(tx *bbolt.Tx, addr *objectSDK.Address, checkGraveyard, raw boo
 	key := objectKey(addr.ObjectID())
 	cid := addr.ContainerID()
 
-	if checkGraveyard && inGraveyard(tx, addr) {
-		return nil, object.ErrAlreadyRemoved
+	if checkGraveyard {
+		switch inGraveyard(tx, addr) {
+		case 1:
+			return nil, object.ErrNotFound
+		case 2:
+			return nil, object.ErrAlreadyRemoved
+		}
 	}
 
 	// check in primary index
--- a/pkg/local_object_storage/shard/list.go
+++ b/pkg/local_object_storage/shard/list.go
@@ -19,6 +19,7 @@ func (r *ListContainersRes) Containers() []*cid.ID {
 	return r.containers
 }
 
+// List returns all objects physically stored in the Shard.
 func (s *Shard) List() (*SelectRes, error) {
 	lst, err := s.metaBase.Containers()
 	if err != nil {
@@ -27,6 +28,7 @@ func (s *Shard) List() (*SelectRes, error) {
 
 	res := new(SelectRes)
 	filters := object.NewSearchFilters()
+	filters.AddPhyFilter()
 
 	for i := range lst {
 		ids, err := meta.Select(s.metaBase, lst[i], filters) // consider making List in metabase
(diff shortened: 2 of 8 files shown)


pkg/local_object_storage/metabase/get.go
```diff
@@ -85,8 +85,13 @@ func (db *DB) get(tx *bbolt.Tx, addr *objectSDK.Address, checkGraveyard, raw boo
...
 
-	if checkGraveyard && inGraveyard(tx, addr) {
-		return nil, object.ErrAlreadyRemoved
+	if checkGraveyard {
+		switch inGraveyard(tx, addr) {
+		case 1:
+			return nil, object.ErrNotFound
+		case 2:
+			return nil, object.ErrAlreadyRemoved
+		}
```
alexvanin, line 94: Add a small test for that in `get_test.go`. For example
```go
	t.Run("get removed object", func(t *testing.T) {
		obj := generateAddress()
		ts := generateAddress()

		require.NoError(t, meta.Inhume(db, obj, ts))
		_, err := meta.Get(db, obj)
		require.ErrorIs(t, err, object.ErrAlreadyRemoved)

		obj = generateAddress()
		require.NoError(t, meta.Inhume(db, obj, nil))
		_, err = meta.Get(db, obj)
		require.ErrorIs(t, err, object.ErrNotFound)
	})
```

alexvanin (approved):

pkg/local_object_storage/shard/list.go
```diff
@@ -27,6 +28,7 @@ func (s *Shard) List() (*SelectRes, error) {
 
 	res := new(SelectRes)
 	filters := object.NewSearchFilters()
+	filters.AddPhyFilter()
```
alexvanin, line 31: Change `List()` test to check this condition. I suggest to add parent for all objects in the test. They should be ignored in `List()` output. 

https://github.com/nspcc-dev/neofs-node/blob/master/pkg/local_object_storage/shard/list_test.go#L36-L50

```go
		for j := 0; j < N; j++ {
			obj := generateRawObjectWithCID(t, cid)
			addPayload(obj, 1<<2)

			// add parent as virtual object, it must be ignored in List()
			parent := generateRawObjectWithCID(t, cid)
			obj.SetParentID(parent.Object().ID())
			obj.SetParent(parent.Object().SDK())

			objs[obj.Object().Address().String()] = 0

			putPrm.WithObject(obj.Object())

			_, err := sh.Put(putPrm)
			require.NoError(t, err)
		}
	}
```

carpawell (approved):

PR #854, opened 2021-09-27
Fix/Do not use global flags

### LINKED ISSUE #810: Senseless keys in `neofs-cli accounting balance`
There are keys available for `neofs-cli accounting balance` command:
```
$ neofs-cli accounting --help
Operations with accounts and balances

Usage:
  neofs-cli accounting [command]

Available Commands:
  balance     Get internal balance of NeoFS account

Flags:
  -h, --help   help for accounting

Global Flags:
      --address string        address of wallet account
      --binary-key string     path to the raw private key file
  -c, --config string         config file (default is $HOME/.config/neofs-cli/config.yaml)
      --generate-key          generate new private key
  -r, --rpc-endpoint string   remote node address (as 'multiaddr' or '<host>:<port>')
      --ttl uint32            TTL value in request meta header (default 2)
  -v, --verbose               verbose output
  -w, --wallet string         path to the wallet
      --wif string            WIF or NEP-2
  -x, --xhdr strings          Request X-Headers in form of Key=Value
```
The global flags `--xhdr` and `--ttl` aren't applicable to the balance request: we cannot specify X-Headers for accounts, and setting of TTL to any value will not affect the response.
I believe these flags confuses a user, and my proposal is to get rid of them for `accounting balance`. 
What do you think?

carpawell:
Closes #810.

Also delete `ttl`, `--generate-key` and `xhdr` flags from
`accounting balance` command and refactor
command initialization.

### DIFF
--- a/cmd/neofs-cli/modules/root.go
+++ b/cmd/neofs-cli/modules/root.go
@@ -26,8 +26,6 @@ import (
 
 const (
 	envPrefix = "NEOFS_CLI"
-
-	ttlDefaultValue = 2
 )
 
 const xHeadersFlag = "xhdr"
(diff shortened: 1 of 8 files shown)


cmd/neofs-cli/modules/root.go
```diff
@@ -37,9 +35,58 @@ var xHeaders []string
...
+	ttlShorthand = ""
+	ttlDefault   = 2
+	ttlUsage     = "TTL value in request meta header"
+
+	xHeadersKey       = "xhdr"
+	xHeadersShorthand = "x"
+	xHeadersUsage     = "Request X-Headers in form of Key=Value"
+)
+
+var xHeadersDefault []string = nil
```
cthulhu-rider, line 88:
```suggestion
var xHeadersDefault []string
```

cthulhu-rider (approved):

cmd/neofs-cli/modules/root.go
carpawell: deleted

PR #855, opened 2021-09-28
Check public key of RPC responses

### LINKED ISSUE #645: Always proceed public keys during intra-container communication
NeoFS does not prohibit the announcement of the same network addresses to different nodes.

In current implementation storage node works with network addresses only during the work (placement traversing). Examples:

- to check if node info is a descriptor of the local node we compare addresses from the network map;
- we don't check if response was signed with the expected key.

TODO:

1. define remote node info interface similar to the one from reputation or container service;
2. make placement traverser to return list of these interfaces instead of `[]network.Address`;
3. extend the interface with method to get public key;
4. replace `network.IsLocalAddress` function with `IsLocalServer(ServerInfo)` and use it where it is needed;
5. add the ability to receive response pubic key from the client (API Go library or SDK);
6. check keys from response signatures in Object service.

cthulhu-rider:
Closes #645.

### DIFF
--- a/pkg/core/client/errors.go
+++ b/pkg/core/client/errors.go
@@ -1,0 +1,9 @@
+package client
+
+import (
+	"errors"
+)
+
+// ErrWrongPublicKey is returned when the client's response is signed with a key different
+// from the one declared in the network map.
+var ErrWrongPublicKey = errors.New("public key is different from the key in the network map")
--- a/pkg/network/cache/client.go
+++ b/pkg/network/cache/client.go
@@ -4,6 +4,7 @@ import (
 	"sync"
 
 	"github.com/nspcc-dev/neofs-api-go/pkg/client"
+	clientcore "github.com/nspcc-dev/neofs-node/pkg/core/client"
 	"github.com/nspcc-dev/neofs-node/pkg/network"
 )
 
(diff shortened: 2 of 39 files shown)


pkg/core/client/errors.go
```diff
@@ -0,0 +1,9 @@
+package client
+
+import (
+	"github.com/pkg/errors"
```
carpawell, line 4: was it added by mistake?

cthulhu-rider: Yes, fixed

refugee-burma:
good

pkg/network/cache/client.go
```diff
@@ -37,10 +41,10 @@ func (c *ClientCache) Get(netAddr network.AddressGroup) (client.Client, error) {
 	//  but network.StringifyGroup is order-dependent.
 	//  This works until the same mixed group is transmitted
 	//  (for a network map, it seems to be true).
-	mAddr := network.StringifyGroup(netAddr)
+	cacheKey := hex.EncodeToString(info.PublicKey()) + network.StringifyGroup(netAddr)
```
alexvanin, line 44: I guess we can drop `netAddr` part from `cacheKey` completely, because public keys are good enough for node identification.

alexvanin (approved):

pkg/network/cache/client.go
cthulhu-rider: For node identification - agree. But if storage node changes the network endpoints in network map, cache will still return cached clients with old endpoints. I guess we need to think it over again.

PR #862, opened 2021-09-29
Try pebble for writecache

fyrchik:
TODO write more extensive benchmark.


fyrchik:
We decided to continue using bolt for now.

PR #867, opened 2021-09-29
governance: Add more logs of the alphabet and inner ring lists

alexvanin:


pkg/innerring/processors/governance/process_update.go
```diff
@@ -103,3 +114,15 @@ func (gp *Processor) processAlphabetSync() {
...
+func prettyKeys(keys keys.PublicKeys) string {
+	sb := strings.Builder{}
+	for _, key := range keys {
+		sb.WriteString(hex.EncodeToString(key.Bytes()))
+		sb.WriteString(",")
+	}
+
+	ln := sb.Len()
+
+	return sb.String()[:ln-1]
```
cthulhu-rider, line 127: We'll panic if `keys` is empty. I'd check this or add function comment `keys must not be empty`. 

Also double-check is the condition `if newAlphabet == nil` enough.

alexvanin: Fixed panic.

cthulhu-rider (approved):

PR #872, opened 2021-09-30
cmd/neofs-node: Don't lose local container size estimations

alexvanin:
Containers with one storage node do not produce size estimations. The reason is that load controller uses `nopLoadWriter` which loses the data. It does not accumulated in storage and therefore does not announced to the container contract, which is incorrect.

### DIFF
--- a/cmd/neofs-node/container.go
+++ b/cmd/neofs-node/container.go
@@ -228,16 +228,6 @@ func (*morphLoadWriter) Close() error {
 	return nil
 }
 
-type nopLoadWriter struct{}
-
-func (nopLoadWriter) Put(containerSDK.UsedSpaceAnnouncement) error {
-	return nil
-}
-
-func (nopLoadWriter) Close() error {
-	return nil
-}
-
 type remoteLoadAnnounceProvider struct {
 	key *ecdsa.PrivateKey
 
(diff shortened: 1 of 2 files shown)


cthulhu-rider:
@alexvanin your solution fixes the problem, but the reason is on the upper level. If you remember we use `Router` to control the route of the values inside the container. `Router` implementation uses  route `Builder` entity, which plots a route during the value transmission. The starting point of the route is always local node. In described case with single-node container route builder should immediately return the end of the route, which means that there is no need to transfer the value between the nodes of the container. In the program, this will result in hitting condition
```
if srv == nil {
	return r.deadEndProvider, nil
}
```

alexvanin:
> @alexvanin your solution fixes the problem, but the reason is on the upper level. If you remember we use `Router` to control the route of the values inside the container. `Router` implementation uses route `Builder` entity, which plots a route during the value transmission. The starting point of the route is always local node. In described case with single-node container route builder should immediately return the end of the route, which means that there is no need to transfer the value between the nodes of the container. In the program, this will result in hitting condition
> 
> ```
> if srv == nil {
> 	return r.deadEndProvider, nil
> }
> ```

Check latest changes with solution in higher level entity. `NextStage` of the router ignored case with single passed value.

alexvanin:
Maybe after this fix, we can remove 
```go
if len(route) == 0 {
	route = []ServerInfo{nil}
}
```
and no-op writer for container load announcement. What do you think?

cthulhu-rider (approved):

cthulhu-rider:
> What do you think?
Only no-op writer, the condition is legit

PR #873, opened 2021-09-30
Fix/make notary deposits depend on balances

carpawell:


carpawell:
this is how balances look with that fix:

image

carpawell:
this is how graphs look in the last dev-env(1000 GAS once in 50 blocks `IR` -> `SN`; 0.00000001 GAS or `SN`'s balance/2 once in epoch `SN` -> `Notary Service`) but epoch duration is 20 blocks.
image

PR #874, opened 2021-10-04
docs: Add comments for config values

### LINKED ISSUE #858: describe config parameters
My notes after talking with @cthulhu-rider :

```YAML

# All sizes are in bytes, multiplicative suffixes are not supported yet.

refill_metabase: false          # sync metabase with blobstore on start, expensive, leave false untill complete understanding

writecache:
  path: tmp/0/cache             # writecache root directory
  mem_size: 2147483648          # coarse RAM usage limit for small objects
  small_size: 16384             # small objects goes to boltdb (leave default)
  max_size: 134217728           # objects bigger that <max_size> bypassing writecache and goes to storage directly
  workers_number: 30            # thread number of writecache flushers 
  size_limit: 3221225472        # coarse writecache total size (5% of metabase part is a good start)

metabase:
  path: tmp/0/meta              #
  perm: 0644                    #

blobstor:
  path: tmp/0/blob              #
  perm: 0644                    #
  compress: true                # zstd, level 3, unconfigurable right now
  shallow_depth: 5              # directory structure depth
  small_size_limit: 102400      # bigger objects goes to filesystem, smaller -- into blobovnicza

blobovnicza:
  size: 4194304                 # coarse one db size limit, total size: size*width^(depth+1), 5% of blobstor is a good start)
  shallow_depth: 1              # directory structure depth
  shallow_width: 4              # and width
  opened_cache_size: 50         #  maximum number of opened database files

gc:                             # full object removal
  remover_batch_size: 150       # number of objects to remove 
  remover_sleep_interval: 2m    # frequency of invocation
```

This should be added to [config/example/node.yaml](https://github.com/nspcc-dev/neofs-node/blob/master/config/example/node.yaml) or documented somewhere else.

carpawell:
Closes #858.


config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
 
 node:
-  key: ./wallet.key
+  key: ./wallet.key  # path to a binary private key
   wallet:
-    path: "./wallet.json"
-    address: "NcpJzXcSDrh5CCizf4K9Ro6w4t59J5LKzz"
-    password: "password"
-  addresses:
+    path: "./wallet.json"  # path to a NEO wallet; ignored if key is presented
```
532910, line 15: а не наоборот: ключ будет проигнорирован, при наличии кошелька?

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
   reputation: 441995f631c1da2b133462b71859494a5cd45e90
   proxy: ad7c6b55b737b696e5c82c85445040964a03e97f
 
 morph:
-  dial_timeout: 30s
-  disable_cache: true
-  rpc_endpoint:
+  dial_timeout: 30s  # timeout for side chain NEO RPC client connection
+  disable_cache: true  # do not use TTL cache for side chain GET operations
+  rpc_endpoint:  # side chain NEO RPC endpoints; are shuffled and used one by one until the first success
```
532910, line 60: shuffled or round-robin?

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
-    deposit_amount: 100000001
-    deposit_duration: 1001
+    deposit_amount: 100000001  # notary deposit GAS amount
+    deposit_duration: 1001  # notary deposit expiration in blocks
 
 mainchain:
-  dial_timeout: 30s
-  rpc_endpoint:
+  dial_timeout: 30s  # timeout for main chain NEO RPC client connection
+  rpc_endpoint:  # main chain NEO RPC endpoints; are shuffled and used one by one until the first success
```
532910, line 72: same here

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
 
 replicator:
-  put_timeout: 15s
+  put_timeout: 15s  # timeout for the Replicator PUT remote operation
 
 object:
   put:
-    pool_size_remote: 100
-    pool_size_local: 101
+    pool_size_remote: 100  # numbers of async workers for remote PUT operations
```
532910, line 87: number

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
+    pool_size_local: 101  # numbers of async workers for local PUT operations
```
532910, line 88: same here

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
 
 storage:
-  shard_num: 2
+  shard_num: 2  # total shards amount
```
532910, line 91: not sure if "amount" is suitable word here

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
-        path: tmp/0/cache
-        mem_size: 2147483648
-        small_size: 16384
-        max_size: 134217728
-        workers_number: 30
-        size_limit: 3221225472
+        path: tmp/0/cache  # write-cache root directory
+        mem_size: 2147483648  # coarse RAM usage limit for "small" objects
+        small_size: 16384 # size threshold for objects that are considered to be "small"
+        max_size: 134217728  # objects that are bigger than <max_size> bypass write-cache and go to the storage directly
```
532910, line 102: looks like it should be "and goes"

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
+        workers_number: 30  # thread number of write-cache flushers
```
532910, line 103: number of  write-cache flusher threads

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
+        size_limit: 3221225472  # coarse write-cache total size
```
532910, line 104: units (bytes) should be specified

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
+        compress: true  # zstd, level 3, unconfigurable right now
+        shallow_depth: 5 # directory structure depth
+        small_size_limit: 102400  # size threshold for objects that are considered to be "small"
 
         blobovnicza:
-          size: 4194304
-          shallow_depth: 1
-          shallow_width: 4
-          opened_cache_size: 50
+          size: 4194304  # coarse one db size limit, total size: size*width^(depth+1)
```
532910, line 118: total size will be: size*width^(depth+1)

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
+          shallow_depth: 1  # directory structure depth
+          shallow_width: 4   # directory structure and width
```
532910, line 120: why "and"?

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
 
       gc:
-        remover_batch_size: 150
-        remover_sleep_interval: 2m
+        remover_batch_size: 150  # number of objects to be removed by the garbage collector
+        remover_sleep_interval: 2m  # frequency of the garbage collector invocation
 
     1:
-      refill_metabase: true
+      refill_metabase: true  # sync metabase with blobstore on start, expensive, leave false until complete understanding
```
532910, line 128: вроде выглядит избыточным, хватит описание одного шарда

532910 (commented):
Стоит послушать @EliChin

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
+  addresses:  # addresses for bootstrapping in the Network map
     - s01.neofs.devenv:8080
     - /dns4/s02.neofs.devenv/tcp/8081
     - grpc://127.0.0.1:8082
     - grpcs://localhost:8083
   attribute_0: "Price:11"
   attribute_1: UN-LOCODE:RU MSK
-  relay: true
-  persistent_state:
+  relay: true  # (not)register Node in the Network map on startup
```
EliChin, line 25: Если здесь имеется в виду "(не) регистрировать ноду" -- "(do not) register Node", если "(не)зарегистрированная нода" -- "(non)registered Node"

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
-      enabled: false
+      enabled: false  # use TLS for a gRPC connection
 
 control:
-  authorized_keys:
+  authorized_keys:  # list of raw hex public keys that have rights to use the Control Service
     - 035839e45d472a3b7769a2a1bd7d54c4ccd4943c3b40f547870e83a8fcbfb3ce11
     - 028f42cfcb74499d7b15b35d9bff260a1c8d27de4f446a627406a382d8961486d6
-  grpc:
+  grpc:  # endpoint that are listened by the Control Service
```
EliChin, line 47: "endpoint that **is**"

cthulhu-rider:
I'd also mention in `README.md` the fact that explanations are located in `yaml`.

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
 logger:
-  level: debug
+  level: debug  # logger level
```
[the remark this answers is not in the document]
carpawell: added possible values

config/example/node.yaml
carpawell: this is how it is done now: if key != "" wallet config is read and only in that case

config/example/node.yaml
carpawell: fixed

config/example/node.yaml
carpawell: misclick: wanted it to be "endpoint**s** that are listened by the Control Service"

fixed

config/example/node.yaml
carpawell: they are shuffled and 2nd(in the shuffled list) is not going to be used(even connection is not established) until the first one returns error and so on

config/example/node.yaml
carpawell: hmmm

in fact, node does not use a connection to the main chain at all after the last(`v0.25.0`) release, so marked that section as deprecated

config/example/node.yaml
carpawell: fixed

config/example/node.yaml
carpawell: fixed

config/example/node.yaml
carpawell: changed to "number"

config/example/node.yaml
carpawell: hm, why?

"objects bypass and go..."

config/example/node.yaml
532910: or just "Control Service endpoint"

config/example/node.yaml
carpawell: changed

config/example/node.yaml
carpawell: changed

config/example/node.yaml
carpawell: misclick

deleted

config/example/node.yaml
carpawell: theoretically, they can differ and it is already done so I suggest to keep it

config/example/node.yaml
532910: OK!

carpawell:
@cthulhu-rider, updated README.md

config/example/README.md
```diff
@@ -1,7 +1,8 @@
 # Examples of correct configuration file structures 
 
 Here are files in all supported formats and with all possible configuration values
-of NeoFS applications.
+of NeoFS applications. `node.yaml` file contains commentaries to the configuration
```
532910, line 4: could you make it a link?

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
 logger:
-  level: debug
+  level: debug  # logger level: one of "info"(default), "debug", "warn", "error", "dpanic", "panic", "fatal"
```
532910, line 2: what is the difference between panic and dpanic?

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
 
 profiler:
-  address: 127.0.0.1:6060
-  shutdown_timeout: 15s
+  address: 127.0.0.1:6060  # endpoint for Node profiling
+  shutdown_timeout: 15s  # timeout for graceful profiling HTTP server shutdown
```
532910, line 6: graceful means it will not kick me out if I connect?

532910 (commented):
see comments

config/example/README.md
carpawell: sure

config/example/node.yaml
carpawell: it means that "graceful shutdown"(closing connection and so on) process has its own timeout before it is gonna be killed

I think I chosen wrong word order, made it as "# timeout for profiling HTTP server graceful shutdown"

532910: graceful shutdown timeout for HTTP profiling
but it still not clear, why it is "graceful"

config/example/node.yaml
carpawell: lib that we use documents it as following:
```
// DPanicLevel logs are particularly important errors. In development the
// logger panics after writing the message.
    DPanicLevel
// PanicLevel logs a message, then panics.
    PanicLevel
```

DPanic means "development panic", our logger is going to panic only in the development mode

config/example/node.yaml
carpawell: added ", bytes" suffix in every mem-size-config values

config/example/node.yaml
carpawell: what exactly don't you like? term or its usage?

config/example/node.yaml
532910: > DPanic means "development panic", our logger is going to panic only in the development mode

This is a good note!

config/example/node.yaml
532910: What does "graceful" mean? 
For example nginx defines it as follow:
> Old worker processes close listen sockets and continue to service old clients. After all clients are serviced, old worker processes are shut down.

Is this true in our case?

config/example/README.md
```diff
@@ -1,7 +1,8 @@
 # Examples of correct configuration file structures 
 
 Here are files in all supported formats and with all possible configuration values
-of NeoFS applications.
+of NeoFS applications. [node.yaml](node.yaml) file contains commentaries to the configuration
```
532910, line 4: contains configuration comments

532910: See [node.yaml](node.yaml) for configuration comments.

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
 object:
   put:
-    pool_size_remote: 100
-    pool_size_local: 101
+    pool_size_remote: 100  # number of async workers for remote PUT operations
+    pool_size_local: 101  # number of async workers for local PUT operations
 
 storage:
-  shard_num: 2
+  shard_num: 2  # total shards number
```
EliChin, line 91: I think "total number of shards" would sound better

config/example/node.yaml
carpawell: i used this term just because golang std lib use it:
```
// Shutdown gracefully shuts down the server without interrupting any
// active connections. Shutdown works by first closing all open
// listeners, then closing all idle connections, and then waiting
// indefinitely for connections to return to idle and then shut down.
// If the provided context expires before the shutdown is complete,
// Shutdown returns the context's error, otherwise it returns any
// error returned from closing the Server's underlying Listener(s).
```

this is exactly what `shutdown_timeout` is used for

config/example/node.yaml
carpawell: changed

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
+        compress: false  # zstd, level 3, unconfigurable right now
+        shallow_depth: 5  # directory structure depth
+        small_size_limit: 102400  # size threshold for objects that are considered to be "small", bytes
 
         blobovnicza:
-          size: 4194304
-          shallow_depth: 1
-          shallow_width: 4
-          opened_cache_size: 50
+          size: 4194304  # coarse one db size limit, total size will be: size*width^(depth+1), bytes
```
alexvanin, line 152: I am not sure `coarse` is a suitable word there. Probably direct translation of "грубый", but I haven't seen usage of it in this context.

What about  
```suggestion
          size: 4194304  # approximate size limit of single blobovnicza instance, total size will be: size*width^(depth+1), bytes
```

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
 
       blobstor:
-        path: tmp/1/blob
-        perm: 0644
-        compress: false
-        shallow_depth: 5
-        small_size_limit: 102400
+        path: tmp/1/blob  # blobstor path
+        perm: 0644  # permissions for blobstor files(directories: +x for current user and group)
+        compress: false  # zstd, level 3, unconfigurable right now
```
alexvanin, line 147: I don't understand anything from this comment besides that it uses zstd lib.

```suggestion
        compress: false  # turn on/off zstd compression of stored objects
```

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
-        path: tmp/1/cache
-        mem_size: 2147483648
-        small_size: 16384
-        max_size: 134217728
-        workers_number: 30
-        size_limit: 4294967296
+        path: tmp/1/cache  # write-cache root directory
+        mem_size: 2147483648  # coarse RAM usage limit for "small" objects, bytes
+        small_size: 16384  # size threshold for objects that are considered to be "small", bytes
+        max_size: 134217728  # objects that are bigger than <max_size> bypass write-cache and go to the storage directly, bytes
```
alexvanin, line 136: Use the same wording from `small_size`, because these values are related
```suggestion
        max_size: 134217728  # size threshold for big objects which bypass write-cache and go to the storage directly, bytes
```

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
       writecache:
-        path: tmp/1/cache
-        mem_size: 2147483648
-        small_size: 16384
-        max_size: 134217728
-        workers_number: 30
-        size_limit: 4294967296
+        path: tmp/1/cache  # write-cache root directory
+        mem_size: 2147483648  # coarse RAM usage limit for "small" objects, bytes
+        small_size: 16384  # size threshold for objects that are considered to be "small", bytes
```
alexvanin, line 135: Big objects bypass write cache but what happens with small ones?

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
     - https://rpc2.morph.fs.neo.org:40341
-  notification_endpoint:
+  notification_endpoint:  # side chain NEO RPC notification endpoints
     - wss://rpc1.morph.fs.neo.org:40341/ws
     - wss://rpc2.morph.fs.neo.org:40341/ws
   notary:
-    deposit_amount: 100000001
-    deposit_duration: 1001
+    deposit_amount: 100000001  # notary deposit GAS amount
+    deposit_duration: 1001  # notary deposit expiration in blocks
```
alexvanin, line 68: If I remember correctly, it is not an expiration value, it is more like `repeat notary deposit after this number of blocks`

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
-  dial_timeout: 30s
-  disable_cache: true
-  rpc_endpoint:
+  dial_timeout: 30s  # timeout for side chain NEO RPC client connection
+  disable_cache: true  # do not use TTL cache for side chain GET operations
+  rpc_endpoint:  # side chain NEO RPC endpoints; are shuffled and used one by one until the first success
     - https://rpc1.morph.fs.neo.org:40341
     - https://rpc2.morph.fs.neo.org:40341
-  notification_endpoint:
+  notification_endpoint:  # side chain NEO RPC notification endpoints
```
alexvanin, line 63: Mention shuffling there too, if you mention it with other endpoint lists.

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
-  authorized_keys:
+  authorized_keys:  # list of raw hex public keys that have rights to use the Control Service
     - 035839e45d472a3b7769a2a1bd7d54c4ccd4943c3b40f547870e83a8fcbfb3ce11
     - 028f42cfcb74499d7b15b35d9bff260a1c8d27de4f446a627406a382d8961486d6
-  grpc:
+  grpc:  # endpoints that are listened by the Control Service
     endpoint: 127.0.0.1:8090
 
-contracts:
+contracts:  # side chain NEOFS contract script hashes
```
alexvanin, line 50: Specify, that these are optional and they override values from NNS contract.

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
   1:
-    endpoint: s02.neofs.devenv:8080
+    endpoint: s02.neofs.devenv:8080  # endpoint for gRPC server
     tls:
-      enabled: false
+      enabled: false  # use TLS for a gRPC connection
 
 control:
-  authorized_keys:
+  authorized_keys:  # list of raw hex public keys that have rights to use the Control Service
```
alexvanin, line 44:
```suggestion
  authorized_keys:  # list of hex-encoded public keys that have rights to use the Control Service
```

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
     - s01.neofs.devenv:8080
     - /dns4/s02.neofs.devenv/tcp/8081
     - grpc://127.0.0.1:8082
     - grpcs://localhost:8083
   attribute_0: "Price:11"
   attribute_1: UN-LOCODE:RU MSK
-  relay: true
-  persistent_state:
+  relay: true  # (do not)register Node in the Network map on startup
+  persistent_state:  # path to store Node state persistently
```
alexvanin, line 26:
```suggestion
  persistent_state:  # path to persistent state file of Storage node
```

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
+  key: ./wallet.key  # path to a binary private key
   wallet:
-    path: "./wallet.json"
-    address: "NcpJzXcSDrh5CCizf4K9Ro6w4t59J5LKzz"
-    password: "password"
-  addresses:
+    path: "./wallet.json"  # path to a NEO wallet; ignored if key is presented
+    address: "NcpJzXcSDrh5CCizf4K9Ro6w4t59J5LKzz"  # address of a NEO account in the wallet; ignored if key is presented
+    password: "password"  # password for a NEO account in the wallet; ignored if key is presented
+  addresses:  # addresses for bootstrapping in the Network map
```
alexvanin, line 18:
```suggestion
  addresses:  # list of addresses announced by Storage node in the Network map
```

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
     - s01.neofs.devenv:8080
     - /dns4/s02.neofs.devenv/tcp/8081
     - grpc://127.0.0.1:8082
     - grpcs://localhost:8083
   attribute_0: "Price:11"
   attribute_1: UN-LOCODE:RU MSK
-  relay: true
-  persistent_state:
+  relay: true  # (do not)register Node in the Network map on startup
```
alexvanin, line 25:
```suggestion
  relay: true  # start Storage node in relay mode without bootstrapping into the Network map
```

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
-      enabled: false
+      enabled: false  # use TLS for a gRPC connection
 
 control:
-  authorized_keys:
+  authorized_keys:  # list of raw hex public keys that have rights to use the Control Service
     - 035839e45d472a3b7769a2a1bd7d54c4ccd4943c3b40f547870e83a8fcbfb3ce11
     - 028f42cfcb74499d7b15b35d9bff260a1c8d27de4f446a627406a382d8961486d6
-  grpc:
+  grpc:  # endpoints that are listened by the Control Service
```
alexvanin, line 47: Is this a list of endpoint**s**? I thought it is possible to specify only one endpoint for control service.

alexvanin (changes requested):
Do not use `coarse`. I propose to use `approximate` instead.

config/example/node.yaml
carpawell: made a mistake

changed this to "notary deposit period in blocks"

config/example/node.yaml
carpawell: how about "# size threshold for "small" objects which are cached in key-value DB, not on FS, bytes"?

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
     - https://rpc1.morph.fs.neo.org:40341
     - https://rpc2.morph.fs.neo.org:40341
-  notification_endpoint:
+  notification_endpoint:  # side chain NEO RPC notification endpoints; are shuffled and used only the first non-error one
     - wss://rpc1.morph.fs.neo.org:40341/ws
     - wss://rpc2.morph.fs.neo.org:40341/ws
   notary:
-    deposit_amount: 100000001
-    deposit_duration: 1001
+    deposit_amount: 100000001  # notary deposit GAS amount
```
aprasolova, line 67: These comments bring no information for a person who isn't familiar with NeoFS internal machinery. Would you please provide some hints on what value should one put here? What is the range for this amount? How the node behaviour will change it I put `deposit_amount` to `1`; to `0`; to `1000000000000000001`?

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
+        small_size: 16384  # size threshold for "small" objects which are cached in key-value DB, not on FS, bytes
+        max_size: 134217728  # size threshold for "big" objects which bypass write-cache and go to the storage directly, bytes
+        workers_number: 30  # number of write-cache flusher threads
+        size_limit: 3221225472  # approximate write-cache total size, bytes
 
       metabase:
-        path: tmp/0/meta
-        perm: 0644
+        path: tmp/0/meta  # metabase path
+        perm: 0644  # permissions for metabase files(directories: +x for current user and group)
```
aprasolova, line 108: Should this value ever be configurable? How the node behaviour would change if I put `0777`; `0600`? 

The same question for blobstor files permissions.

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
       blobstor:
-        path: tmp/0/blob
-        perm: 0644
-        compress: true
-        shallow_depth: 5
-        small_size_limit: 102400
+        path: tmp/0/blob  # blobstor path
+        perm: 0644  # permissions for blobstor files(directories: +x for current user and group)
+        compress: true  # turn on/off zstd(level 3) compression of stored objects
+        shallow_depth: 5  # directory structure depth
```
aprasolova, line 114: How does this value affects Storage Node performance? Please provide some rule of thumb on how a user should select this value.

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
+        small_size_limit: 102400  # size threshold for objects that are considered to be "small", bytes
```
aprasolova, line 115: How small size variation would affect storage engine performance?

config/example/node.yaml
```diff
@@ -1,159 +1,159 @@
...
 
         blobovnicza:
-          size: 4194304
-          shallow_depth: 1
-          shallow_width: 4
-          opened_cache_size: 50
+          size: 4194304  # approximate size limit of single blobovnicza instance, total size will be: size*width^(depth+1), bytes
```
aprasolova, line 118: As a random user, I have no idea what size of blobovnicza with all its dimensions should I have on my machine. Is `shallow_depth: 1; shallow_width: 4` and `shallow_depth: 0; shallow_width: 8` the same or not?

I wonder if this section ever should be configurable by a user. Maybe, we should not expose that much internal parameters. Instead, we might somehow provide pre-configurable blobovniczy with description for every configuration and when it is most applicable. What do you think?

aprasolova (changes requested):

config/example/node.yaml
alexvanin: In short: we don't know yet. In the search of "golden ratio", we should be able to change these parameters, so here we go.

Depth is related to amount of blobovniczas that are exist in the shard. More blobovniczas -> more fast k:v storage for small objects -> faster responses for those objects. But you can increase width value and do the same but differently. :smiley:

config/example/node.yaml
alexvanin: The bigger `small_size` -> the more objects stored in the fast k:v storage -> the faster this limit goes off and performance drops due to cache skip.

config/example/node.yaml
532910: Moreover, I believe default permissions must not be all readable.

532910: Example file is for humans, not for testing propose. Moreover it contains configuration notes.
It is perceived as a good starting point and permissions with non zeros at the end looks strange.

532910: Moreover, good example configs describe default settings as commented out.

PR #875, opened 2021-10-04
Replace golint by revive

masterSplinter01:


alexvanin (changes requested):
Specify latest golangci lint version in github workflow. Current version does not support revive.

PR #893, opened 2021-10-07
Support size suffixes in config

### LINKED ISSUE #857: size multiplicative suffixes support
Please add support for multiplicative suffixes (like kb, mb, gb, tb) in config files.

fyrchik:
Fix #857 .


fyrchik:
See also https://github.com/spf13/cast/pull/129 .

fyrchik:
@532910 floats are not supported, at least for now. I will add TB suffixes.

PR #904, opened 2021-10-11
log levels should go in a verbosity order

532910:
Signed-off-by: Sergio Nemirowski <sergio@nspcc.ru>


alexvanin:
> Agree. Fixed. 

It seems that commit message still has previous header.
```
log levels should go in a verbosity order

config/example: Set log levels in verbosity order

Signed-off-by: Sergio Nemirowski <sergio@nspcc.ru>
```


> is a self id, should it be really added?

Self id is totally fine. When there is no issue to link, we use self id. Without the link it is hard to group commits when merge commits are disabled.

In your case, you can link with https://github.com/nspcc-dev/neofs-node/issues/858

PR #920, opened 2021-10-15
Fix/Rename RPC flag for control service

### LINKED ISSUE #879: --rpc-endpoint should have a different name for control command
control command uses control port, not rpc, so key should have a different name

```
% ./neofs-cli-amd64 --rpc-endpoint localhost:8090 -w storage_wallet.json control netmap-snapshot
```

The accepted solution is rename `rpc-endpoint` to something common like `address` for example.

carpawell:
Closes #879.


carpawell:
@alexvanin, @cthulhu-rider, @532910 need help with [naming](https://github.com/nspcc-dev/neofs-node/pull/920/files#diff-60e51ab03f90c61aa7b430055bf99bc23b5103885338e77530e305bd2af8f88dR56)

carpawell:
> Also 2 commits don't seem logically separated, I would squash them.

@fyrchik, the first one contains fix in fact: there was `initCommonFlagsWithoutRPC` that DID init RPC flag

PR #947, opened 2021-10-27
writecache: refactor object persisting

### LINKED ISSUE #866: Need to verify: writecache max_size is not working
Version: v0.24.1

```
 writecache:
        path: /meta3/write_cache3
        mem_size: 1073741824
        db_size: 0
        small_size: 32768
        max_size: 67108864

```

As a result of the neofs-node work over load generator we observe:
```
/dev/nvme1n1p1  219G  219G     0 100% /meta0
/dev/nvme1n1p3  219G  219G     0 100% /meta2
/dev/nvme1n1p2  219G  219G     0 100% /meta1
/dev/nvme1n1p4  221G  221G     0 100% /meta3
```

fyrchik:
a1696a8 introduced some logic which in some situations prevented big objects
to be persisted in FSTree. In this commit a refactoring is done with the
goal of simplifying the code and also checking #866 issue.

1. Split a monstrous function into multiple simple ones: memory objects
   can only be small and for writing through the cache we can do a dispatch
   in `Put` itself.
2. Determine objects to be put in database before the actual update
   as setting up a transaction has non-zero overhead.

Fix #866.
Testing done:
1. Restrict max cache size.
2. Put both small and big objects.
3. Check that writecache size doesn't exceeds the max size, but has objects put to FSTree and database.

### DIFF
--- a/pkg/local_object_storage/writecache/persist.go
+++ b/pkg/local_object_storage/writecache/persist.go
@@ -26,7 +26,7 @@ func (c *cache) persistLoop() {
 			sort.Slice(m, func(i, j int) bool { return m[i].addr < m[j].addr })
 
 			start := time.Now()
-			c.persistObjects(m)
+			c.persistSmallObjects(m)
 			c.log.Debug("persisted items to disk",
 				zap.Duration("took", time.Since(start)),
 				zap.Int("total", len(m)))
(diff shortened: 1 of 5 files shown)


pkg/local_object_storage/writecache/persist.go
```diff
@@ -26,7 +26,7 @@ func (c *cache) persistLoop() {
 			sort.Slice(m, func(i, j int) bool { return m[i].addr < m[j].addr })
 
 			start := time.Now()
-			c.persistObjects(m)
+			c.persistSmallObjects(m)
```
alexvanin, line 29: Is it okay to persist only small objects there?

alexvanin (commented):
Add issue prefix to commit header.

pkg/local_object_storage/writecache/persist.go
fyrchik: Yes, becaues of https://github.com/nspcc-dev/neofs-node/blob/master/pkg/local_object_storage/writecache/put.go#L33

alexvanin (approved):

PR #950, opened 2021-11-01
cli: Refactor usage of NeoFS API client

cthulhu-rider:
Similar to https://github.com/nspcc-dev/neofs-node/pull/946


realloc:
I'd suggest minimizing wrappers and reconsider this particular refactoring change.

cthulhu-rider:
@realloc, there are no new wrappers or interfaces, only helper functions over `Client` with docs. They will help you more easily handle upcoming breaking changes in the client interface and behavior. Such functionality will make it easier to pull up new client features and respond to changes in behavior.

Issue #910, opened 2021-10-12
Too small `till` value in notary deposit


carpawell:
There could be a situation when epoch duration changes(decreases significantly) and `IR`, `SN` try to make new notary deposits with `*epochDuration* + *someConstant*` `till` deposit value. If a previous epoch duration was too big(comparing to a current) then `SN` and `IR` could not make new notary deposits due to the fact that their `till` value is lower than the previous, so they gеt `till shouldn't be less then the previous value <>` error.

### Possible solutions:
1. Just ignore such errors(already done). It leads to a theoretical situation when the notary balance is out and it is impossible to make a new one.
2. Make "smart" deposit logic: do deposits with `till = max(*epochDuration* + *someConstant*, *currentTill*+1)`. It leads to strange long deposits but seems that there is nothing bad with that. Also, it requires more RPC calls, that is not good.
3. ...

AnnaShaleva:
I vote for "smart" deposits, it shouldn't be hard to implement and maintain. Several remarks:
- you're *not* obliged to increase deposit's `till` value with each new deposit. There are three constraints to the new `till` value:
  - new `till` must be *not less* than the current chain's height
  - new `till` must be *not less* than the previously set `till` value (if so)
  - new `till` will only be set if the transfer sender *is the owner of the deposit*, otherwise default `till` value will be used (the previously set `till` value if so, or the current height + 5760 otherwise)
So in the described situation, you may use `till = max(*epochDuration* + *someConstant*, *currentTill*)` in order not to produce  "strange long deposits".
- `currentTill` value can be defined free by using `invokefunction` RPC call with the following arguments: notary contract hash + "expirationOf" method name + uint160 address of the deposit owner.
- `IR` and `SN` can also monitor their notary balances via `balanceOf` notary contract method and do not deposit more funds when the balance is enough (just use `lockDepositUntil` to prolong the deposit if needed).


>  it requires more RPC calls, that is not good

It requires only one RPC call (per epoch?) to define the current `till` value. If it's bad, then each `SN` and `IR` node may maintain their `till` value locally and update it by themselves without RPC call after each successful deposit. Remember, that only the deposit owner is able to change deposit `till`, so this approach requires only one `expirationOf` RPC call at the node bootstrap to define initial deposit value.

PR #953, opened 2021-11-01
Make notary deposit on notification instead of timer

alexvanin:
Related to #910 

During v0.26.1 pre release testing, I saw this warning in IR logs:
```
warn    innerring/blocktimer.go:106     can't stop epoch estimation     {"epoch": 17, "error": "could not invoke method (stopContainerEstimation): failed to submit notary request: Block or transaction validation failed. (-504) - fallback transaction is valid after deposit is unlocked: V
alidUntilBlock is 3300, deposit lock expires at 3245"}
```

Turns out, that notary deposit did not trigger by new epoch event. Instead deposits were triggered by epoch timer. It is not correct, because timer can never fire up, due to desynchronization or external epoch changes.

### DIFF
--- a/pkg/innerring/processors/netmap/process_epoch.go
+++ b/pkg/innerring/processors/netmap/process_epoch.go
@@ -4,12 +4,15 @@ import (
 	"github.com/nspcc-dev/neofs-node/pkg/innerring/processors/audit"
 	"github.com/nspcc-dev/neofs-node/pkg/innerring/processors/governance"
 	"github.com/nspcc-dev/neofs-node/pkg/innerring/processors/settlement"
+	netmapEvent "github.com/nspcc-dev/neofs-node/pkg/morph/event/netmap"
 	"go.uber.org/zap"
 )
 
 // Process new epoch notification by setting global epoch value and resetting
 // local epoch timer.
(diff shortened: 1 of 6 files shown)


fyrchik (approved):

pkg/innerring/processors/netmap/process_epoch.go
```diff
@@ -48,6 +48,7 @@ func (np *Processor) processNewEpoch(epoch uint64) {
 	np.handleNewAudit(audit.NewAuditStartEvent(epoch))
 	np.handleAuditSettlements(settlement.NewAuditEvent(epoch))
 	np.handleAlphabetSync(governance.NewSyncEvent())
+	np.handleNotaryDeposit(nil)
```
cthulhu-rider, line 51: I suggest do one of:

- make `NotaryDepositHandler` of type `func()`
- pass causal event to it (new epoch event?).
This will avoid potential panics.

cthulhu-rider (approved):

pkg/innerring/processors/netmap/process_epoch.go
alexvanin: new epoch event seems ok.

PR #961, opened 2021-11-06
Support NeoFS API status returns

cthulhu-rider:
Based on and blocked by https://github.com/nspcc-dev/neofs-api-go/pull/350.

To reduce the number of changes in server interfaces, all work with statuses is concentrated in the signature service. In the future, it should be done differently: get rid of the `error` from  the signatures of the RPC handlers and work only with statuses.


pkg/services/util/sign.go
```diff
@@ -52,79 +62,192 @@ func NewUnarySignService(key *ecdsa.PrivateKey) *SignService {
 }
 
 func (s *RequestMessageStreamer) Send(req interface{}) error {
+	// req argument should be strengthen with type RequestMessage
+	s.statusSupported = isStatusSupported(req.(RequestMessage)) // panic is OK here for now
```
carpawell, line 66: what panic? NPE?

carpawell (approved):

pkg/services/util/sign.go
cthulhu-rider: Unsafe cast.

PR #963, opened 2021-11-08
ir/netmap: Call AddPeer method if existing candidate was updated

### LINKED ISSUE #938: Resign `AddPeer` request if it contains new NodeInfo
Storage nodes invoke `netmap.AddPeer` every epoch. Alphabet nodes use these invocations as a heartbeat to keep node in the network map, so they do not resign such invocations. 

Storage node might want to update its `NodeInfo` (e.g. change some attributes without changing the key). In this case Alphabet notes should not ignore such invocation. If Alphabet nodes resign such invocations, netmap contract will update contract storage (possible after https://github.com/nspcc-dev/neofs-contract/issues/146).

cthulhu-rider:
Closes #938.

### DIFF
--- a/pkg/innerring/processors/netmap/cleanup_table.go
+++ b/pkg/innerring/processors/netmap/cleanup_table.go
@@ -1,7 +1,9 @@
 package netmap
 
 import (
+	"bytes"
 	"encoding/hex"
+	"fmt"
 	"sync"
 
 	"github.com/nspcc-dev/neofs-api-go/pkg/netmap"
@@ -12,12 +14,18 @@ type (
 		*sync.RWMutex
 		enabled    bool
 		threshold  uint64
-		lastAccess map[string]epochStamp
+		lastAccess map[string]epochStampWithNodeInfo
 	}
 
 	epochStamp struct {
 		epoch      uint64
 		removeFlag bool
+	}
+
+	epochStampWithNodeInfo struct {
+		epochStamp
+
+		binNodeInfo []byte
 	}
 )
 
@@ -26,7 +34,7 @@ func newCleanupTable(enabled bool, threshold uint64) cleanupTable {
 		RWMutex:    new(sync.RWMutex),
 		enabled:    enabled,
 		threshold:  threshold,
-		lastAccess: make(map[string]epochStamp),
+		lastAccess: make(map[string]epochStampWithNodeInfo),
 	}
 }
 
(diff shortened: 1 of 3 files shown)


pkg/innerring/processors/netmap/cleanup_table.go
```diff
@@ -36,33 +44,50 @@ func (c *cleanupTable) update(snapshot *netmap.Netmap, now uint64) {
...
+// Returns true if node hasn't been accessed yet, or:
+//  * remove flag is unset, and
+//  * binary node info is changed.
+func (c *cleanupTable) touch(keyString string, now uint64, binNodeInfo []byte) bool {
 	c.Lock()
 	defer c.Unlock()
 
 	access, ok := c.lastAccess[keyString]
-	result := !access.removeFlag && ok
+	result := !ok || !access.removeFlag && !bytes.Equal(access.binNodeInfo, binNodeInfo)
```
alexvanin, line 82:
```suggestion
	result := !ok || access.removeFlag || !bytes.Equal(access.binNodeInfo, binNodeInfo)
```

If `removeFlag` is set, then `touch()` always has to return `true` to resign transaction.
If `binNodeInfo` has been changed, then `touch()` always has to return true to resign transaction.

pkg/innerring/processors/netmap/cleanup_table_test.go
```diff
@@ -108,7 +116,7 @@ func TestCleanupTable(t *testing.T) {
 			cnt := 0
 			key := hex.EncodeToString(infos[1].PublicKey())
 
-			require.False(t, c.touch(key, 4)) // one node was updated
+			require.False(t, c.touch(key, 4, mapInfos[key])) // one node was updated
```
alexvanin, line 119: I suppose this should be `require.True` because `touch()` return value was inverted. Maybe it is related to comment above.

pkg/innerring/processors/netmap/cleanup_table.go
cthulhu-rider: Fixed

pkg/innerring/processors/netmap/cleanup_table_test.go
cthulhu-rider: Changed along with https://github.com/nspcc-dev/neofs-node/pull/963#discussion_r746622222

alexvanin (approved):

PR #965, opened 2021-11-10
Continuous replication

alexvanin:
Blocked by #948

### DIFF
--- a/cmd/neofs-node/config.go
+++ b/cmd/neofs-node/config.go
@@ -197,6 +197,10 @@ type cfgLocalStorage struct {
 
 type cfgObjectRoutines struct {
 	putRemote *ants.Pool
+
+	putRemoteCapacity int
+
+	replication *ants.Pool
 }
 
 type cfgControlService struct {
@@ -449,10 +453,13 @@ func initObjectPool(cfg *config.Config) (pool cfgObjectRoutines) {
 
 	optNonBlocking := ants.WithNonblocking(true)
 
-	pool.putRemote, err = ants.NewPool(objectconfig.Put(cfg).PoolSizeRemote(), optNonBlocking)
-	if err != nil {
-		fatalOnErr(err)
-	}
+	pool.putRemoteCapacity = objectconfig.Put(cfg).PoolSizeRemote()
+
+	pool.putRemote, err = ants.NewPool(pool.putRemoteCapacity, optNonBlocking)
+	fatalOnErr(err)
+
+	pool.replication, err = ants.NewPool(pool.putRemoteCapacity)
+	fatalOnErr(err)
 
 	return pool
 }
@@ -485,4 +492,11 @@ func (c *cfg) bootstrap() error {
 // needBootstrap checks if local node should be registered in network on bootup.
 func (c *cfg) needBootstrap() bool {
 	return c.cfgNetmap.needBootstrap
+}
+
+// ObjectServiceLoad implements system loader interface for policer component.
+// It is calculated as size/capacity ratio of "remote object put" worker.
+// Returns float value between 0.0 and 1.0.
+func (c *cfg) ObjectServiceLoad() float64 {
+	return float64(c.cfgObject.pool.putRemote.Running()) / float64(c.cfgObject.pool.putRemoteCapacity)
 }
--- a/pkg/services/policer/policer.go
+++ b/pkg/services/policer/policer.go
@@ -1,9 +1,9 @@
 package policer
 
 import (
-	"sync"
 	"time"
 
+	lru "github.com/hashicorp/golang-lru"
 	"github.com/nspcc-dev/neofs-node/pkg/core/container"
 	"github.com/nspcc-dev/neofs-node/pkg/core/netmap"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/engine"
@@ -12,15 +12,22 @@ import (
 	"github.com/nspcc-dev/neofs-node/pkg/services/replicator"
 	"github.com/nspcc-dev/neofs-node/pkg/util/logger"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
+	"github.com/panjf2000/ants/v2"
 	"go.uber.org/zap"
 )
+
+// NodeLoader provides application load statistics.
+type nodeLoader interface {
+	// ObjectServiceLoad returns object service load value in [0:1] range.
+	ObjectServiceLoad() float64
+}
 
 // Policer represents the utility that verifies
 // compliance with the object storage policy.
 type Policer struct {
 	*cfg
 
-	prevTask prevTask
+	cache *lru.Cache
 }
 
 // Option is an option for Policer constructor.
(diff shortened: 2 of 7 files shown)


cmd/neofs-node/config.go
```diff
@@ -449,7 +453,14 @@ func initObjectPool(cfg *config.Config) (pool cfgObjectRoutines) {
 
 	optNonBlocking := ants.WithNonblocking(true)
 
-	pool.putRemote, err = ants.NewPool(objectconfig.Put(cfg).PoolSizeRemote(), optNonBlocking)
+	pool.putRemoteCapacity = objectconfig.Put(cfg).PoolSizeRemote()
+
+	pool.putRemote, err = ants.NewPool(pool.putRemoteCapacity, optNonBlocking)
+	if err != nil {
+		fatalOnErr(err)
```
cthulhu-rider, line 460: `if` statement can be omitted.

pkg/services/policer/policer.go
```diff
@@ -12,15 +12,22 @@ import (
 	"github.com/nspcc-dev/neofs-node/pkg/services/replicator"
 	"github.com/nspcc-dev/neofs-node/pkg/util/logger"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
+	"github.com/panjf2000/ants/v2"
 	"go.uber.org/zap"
 )
 
+// NodeLoader provides functions to fetch application load statistic.
+type nodeLoader interface {
```
cthulhu-rider, line 20:
```suggestion
type NodeLoader interface {
```

pkg/services/policer/policer.go
```diff
@@ -12,15 +12,22 @@ import (
 	"github.com/nspcc-dev/neofs-node/pkg/services/replicator"
 	"github.com/nspcc-dev/neofs-node/pkg/util/logger"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
+	"github.com/panjf2000/ants/v2"
 	"go.uber.org/zap"
 )
 
+// NodeLoader provides functions to fetch application load statistic.
```
cthulhu-rider, line 19:
```suggestion
// NodeLoader provides application load statistics.
```

pkg/services/policer/process.go
```diff
@@ -2,113 +2,80 @@ package policer
...
-}
-
-type workScope struct {
-	val int
-
-	expRate int // in %
-}
-
 func (p *Policer) Run(ctx context.Context) {
-	defer func() {
```
cthulhu-rider, line 28: Were these logs so useless?

pkg/services/policer/process.go
```diff
@@ -2,113 +2,80 @@ package policer
...
-			)
-
-			return
-		case task, ok := <-p.trigger:
-			if !ok {
-				p.log.Warn("trigger channel is closed")
+		addrs, cursor, err = p.jobQueue.Select(cursor, p.batchSize)
+		if err != nil {
+			if errors.Is(err, engine.ErrEndOfListing) {
+				time.Sleep(1 * time.Second) // finished whole cycle, sleep a bit
```
cthulhu-rider, line 29:
```suggestion
				time.Sleep(time.Second) // finished whole cycle, sleep a bit
```

pkg/services/policer/process.go
```diff
@@ -2,113 +2,80 @@ package policer
...
+				continue
+			}
+			p.log.Warn("failure at object select for replication", zap.Error(err))
```
cthulhu-rider, line 32:
```suggestion
			p.log.Warn("failure at object select for replication", zap.String("error", err.Error()))
```
and `\n` matters.

pkg/services/policer/process.go
```diff
@@ -2,113 +2,80 @@ package policer
...
+					v, ok := p.cache.Get(addrStr)
+					if ok && time.Since(v.(time.Time)) < p.evictDuration {
+						return
+					}
+
+					p.processObject(ctx, addr)
+					p.cache.Add(addrStr, time.Now())
+				})
+				if err != nil {
+					p.log.Warn("pool submission", zap.Error(err))
```
cthulhu-rider, line 52: `zap.String`

pkg/services/policer/policer.go
```diff
@@ -12,15 +12,22 @@ import (
...
+type nodeLoader interface {
+	// ObjectServiceLoad returns object service load value.
```
cthulhu-rider, line 21:
```suggestion
	// ObjectServiceLoad returns object service load value in [0:1] range.
```

pkg/services/policer/process.go
```diff
@@ -2,113 +2,80 @@ package policer
...
-		p.processObject(ctx, addrs[i])
+		case <-ticker.C:
+			neofsSysLoad := p.loader.ObjectServiceLoad()
+			newCapacity := int((1.0 - neofsSysLoad) * float64(p.maxCapacity))
+			if newCapacity == 0 {
+				newCapacity++
+			}
 
-		p.prevTask.undone--
+			if p.taskPool.Cap() != newCapacity {
```
cthulhu-rider, line 73: Required check? Thought maybe `pool.Tune` does nothing if same capacity is passed.

cthulhu-rider (approved):

cthulhu-rider:
We can keep control of pool capacity value on the app-side, and on replicator's side only control the frequency of updates. In theory, you can even leave the periodicity on the side of the application.

pkg/services/policer/process.go
alexvanin: Actually, `\n` are not produced anymore, even in wrapped errors! I think we simplified logger structure a while ago and it is not a case to worry now. 

```go
a := errors.New("deep error")
b := fmt.Errorf("included error: %w", a)
d := fmt.Errorf("included error two: %w", b)
c.log.Info("error", zap.Error(d)) // logger with config from neofs-dev-env

// 2021-11-24T09:50:09.841Z        info    neofs-node/main.go:59   error   {"error": "included error two: included error: deep error"} 
```

pkg/services/policer/policer.go
alexvanin: Should it be public? I am not sure it needs to be exported somewhere out of policer package.

pkg/services/policer/process.go
alexvanin: This check is mostly for logger, to avoid log messages every second when capacity didn't change.

pkg/services/policer/process.go
cthulhu-rider: Sorry, I meant `\n` is source code.

pkg/services/policer/process.go
alexvanin: Nope, accidentally removed it. Put it back.

alexvanin:
> We can keep control of pool capacity value on the app-side, and on replicator's side only control the frequency of updates. In theory, you can even leave the periodicity on the side of the application.

Yes, both are viable options. I can't say what are pros to stick with one or another option, but they are definitely will work fine.

pkg/services/policer/process.go
alexvanin: Buy anyway, with current logger configuration, `zap.Error` is totally legit, though it produced some ugly output before and we stopped using it.

PR #968, opened 2021-11-11
config: support floating-point sizes

fyrchik:
Rounding is done using the suffix boundary (i.e. floating point value
with `gb` suffix will return size which is an integer number of
megabytes).


532910 (approved):

alexvanin (approved):

cmd/neofs-node/config/cast.go
```diff
@@ -164,12 +165,28 @@ func SizeInBytesSafe(c *Config, name string) uint64 {
 // with minor corrections (allow to use both `k` and `kb` forms.
 // Seems like viper allows to convert sizes but corresponding parser in `cast` package
 // is missing.
-func safeMul(a, b uint64) uint64 {
-	c := a * b
-	if a > 1 && b > 1 && c/b != a {
+
+// safeMul returns size*multiplier, rounding down to the
+// multiplier/1024 number of bytes.
+// Returns 0 on if overflow is detected.
```
cthulhu-rider, line 171: is `on if` correct?

cthulhu-rider (approved):
@fyrchik add issue tag to commit header.

carpawell (approved):

PR #971, opened 2021-11-14
Feat/Generate nonce and vub based on notifications information

carpawell:
This PR refactors morph wrappers to group their call parameters. `InvokePrmOptional` is public in all level parameters to save us from too much useless copy-paste of `TxHash()` method. It will be considered to be private after moving to `neofs-sdk-go` and reducing the number of wrapper levels. 

`InvokePrmOptional` was also planned to store optional `transaction.Transaction` that is going to be signed by Alphabet. So after that PR, there will be another that deletes `NotarySignAndInvokeTX`and its usage.

Closes #844.

PR #975, opened 2021-11-16
Combined notary witnesses

alexvanin:
Add invoker witness as the third witness. Required for netmap methods checks (https://github.com/nspcc-dev/neofs-contract/issues/154). 

Blocked by neo-go release, and https://github.com/nspcc-dev/neofs-node/pull/971

### DIFF
--- a/pkg/morph/client/notary.go
+++ b/pkg/morph/client/notary.go
@@ -441,7 +441,11 @@ func (c *Client) notaryInvoke(committee, invokedByAlpha bool, contract util.Uint
 	_, n := mn(alphabetList, committee)
 	u8n := uint8(n)
 
-	cosigners, err := c.notaryCosigners(alphabetList, committee)
+	if !invokedByAlpha {
+		u8n++
+	}
+
+	cosigners, err := c.notaryCosigners(invokedByAlpha, alphabetList, committee)
 	if err != nil {
 		return err
 	}
@@ -510,7 +514,7 @@ func (c *Client) notaryInvoke(committee, invokedByAlpha bool, contract util.Uint
 	err = c.client.AddNetworkFee(
 		mainTx,
 		notaryFee,
-		c.notaryAccounts(multiaddrAccount)...,
+		c.notaryAccounts(invokedByAlpha, multiaddrAccount)...,
 	)
 	if err != nil {
 		return err
@@ -538,8 +542,8 @@ func (c *Client) notaryInvoke(committee, invokedByAlpha bool, contract util.Uint
 	return nil
 }
 
-func (c *Client) notaryCosigners(ir []*keys.PublicKey, committee bool) ([]transaction.Signer, error) {
-	s := make([]transaction.Signer, 0, 3)
+func (c *Client) notaryCosigners(invokedByAlpha bool, ir []*keys.PublicKey, committee bool) ([]transaction.Signer, error) {
+	s := make([]transaction.Signer, 0, 4)
 
 	// first we have proxy contract signature, as it will pay for the execution
 	s = append(s, transaction.Signer{
@@ -563,6 +567,16 @@ func (c *Client) notaryCosigners(ir []*keys.PublicKey, committee bool) ([]transa
 		AllowedGroups:    c.signer.AllowedGroups,
 	})
 
+	if !invokedByAlpha {
+		// then we have invoker signature
+		s = append(s, transaction.Signer{
+			Account:          hash.Hash160(c.acc.GetVerificationScript()),
+			Scopes:           c.signer.Scopes,
+			AllowedContracts: c.signer.AllowedContracts,
+			AllowedGroups:    c.signer.AllowedGroups,
+		})
+	}
+
 	// last one is a placeholder for notary contract signature
 	s = append(s, transaction.Signer{
 		Account: c.notary.notary,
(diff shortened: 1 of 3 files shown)


pkg/morph/client/notary.go
```diff
@@ -510,6 +510,14 @@ func (c *Client) notaryCosigners(ir []*keys.PublicKey, committee bool) ([]transa
 		AllowedGroups:    c.signer.AllowedGroups,
 	})
 
+	// then we have invoker signature
+	s = append(s, transaction.Signer{
+		Account:          hash.Hash160(c.acc.GetVerificationScript()),
```
AnnaShaleva, line 515: Please, use `c.acc.ScriptHash()` for that.

pkg/morph/client/notary.go
```diff
@@ -586,6 +597,17 @@ func (c *Client) notaryWitnesses(invokedByAlpha bool, multiaddr *wallet.Account,
 		VerificationScript: multiaddr.GetVerificationScript(),
 	})
 
+	// then we have invoker witness
+	invokeScript = append(
```
AnnaShaleva, line 601: Use `c.acc.SignTx(...)` instead of subsequent 10 lines of code. It does exactly the same thing.

pkg/morph/event/notary_preparator.go
```diff
@@ -338,9 +339,14 @@ func (p Preparator) validateWitnesses(w []transaction.Witness, alphaKeys keys.Pu
 		return errIncorrectAlphabet
 	}
 
-	// the third one must be a placeholder for notary
+	// the third one must be a invoker witness
```
AnnaShaleva, line 342: s/a/an

AnnaShaleva (changes requested):
I have several questions:
1. The code assumes that third witness (the invoker's one) is *always* present, which means that even when `invokedByAlpha` is true, the invoker signature wil also be attached to the transaction. Question: who is the invoker in case of `invokeByAlpha` is set to true?
2. Initial transaction is signed by the invoker and that's OK for cases when alphabet nodes do not change the transaction hashable part (script, nonce, VUB etc.). But there's the case when alphabet nodes change the initial transaction script, and after that invoker's signature becomes invalid. Is it possible? And if so, then do you handle it somehow?

alexvanin:
> The code assumes that third witness (the invoker's one) is always present, which means that even when invokedByAlpha is true, the invoker signature wil also be attached to the transaction. Question: who is the invoker in case of invokeByAlpha is set to true?

Invoker is going to be an alphabet node, therefore 3rd witness is a regular signature of alphabet node (not multisignature like 2nd witness). I had a choice:
- parse notary requests with variable amount of witnesses (3 for `invokeByAlpha = true` and 4 for `invokeByAlpha = false`),
- add mandatory invoker wintess even if it unnecessary for `invokeByAlpha = true`. 

I thought it will be more consistent to have the one scheme with 4 witnesses for all NeoFS related notary requests. It should not break anything but it is redundant, I agree. Is it bad?

> Initial transaction is signed by the invoker and that's OK for cases when alphabet nodes do not change the transaction hashable part (script, nonce, VUB etc.). But there's the case when alphabet nodes change the initial transaction script, and after that invoker's signature becomes invalid. Is it possible? And if so, then do you handle it somehow?

It is possible. We have one case with `netmap.AddPeer` method (https://github.com/nspcc-dev/neofs-contract/issues/154). To solve it we are going to split this method into two separate methods. _First_ method, invoked by Storage node, only checks witness of invoker. Alphabet node creates new notary request, that invokes _second_ method, which only checks witness of Alphabet.

AnnaShaleva:
> I thought it will be more consistent to have one scheme with 4 witnesses for all NeoFS related notary requests. It should not break anything but it is redundant, I agree. Is it bad?

I also thought about its redundancy, just needed to clarify. I think it's not critical and we may safely use the common scheme with 4 witnesses for all cases.

> It is possible. We have one case with netmap.AddPeer method (nspcc-dev/neofs-contract#154). To solve it we are going to split this method into two separate methods. First method, invoked by Storage node, only checks witness of invoker. Alphabet node creates new notary request, that invokes second method, which only checks witness of Alphabet.

I firstly have a thought that except two different `netmap.AddPeer` methods we also need two different methods for client to construct proper notary signers/witnesses for two different cases. But after the second review, I realised that we don't need to split them on the client side, so everything is OK with the current code.

AnnaShaleva:
>  But after the second review, I realised that we don't need to split them on the client side, so everything is OK with the current code.

Or maybe we still need. Consider the case: storage node constructs notary request with `netmap.AddPeer` call and sends it to the network. Alphabet nodes receive the request and construct their own, it contains 4 signers: proxy + alphabet + invoker + notary. And `invoker` here is an alphabet node (you wrote the same):

> Invoker is going to be an alphabet node, therefore 3rd witness is a regular signature of alphabet node (not multisignature like 2nd witness).

Each alphabet node should construct same main transaction with same set of signers. And here's the problem: alphabet node `A` will put herself as the 3rd signer (because it's a standard signature signer, not multisig); alphabet node `B` will put herself as the 3rd signer and etc. As a result, main transactions generated by alphabet nodes will be different (3rd signer is unique).

So we need to change our signing scheme, otherwise even standard notary requests won't be collected.

alexvanin:
> Each alphabet node should construct same main transaction with same set of signers. And here's the problem: alphabet node A will put herself as the 3rd signer (because it's a standard signature signer, not multisig); alphabet node B will put herself as the 3rd signer and etc. As a result, main transactions generated by alphabet nodes will be different (3rd signer is unique).

Oh, you are right. I guess we have no other choice other than use variable amount of witnesses in notary requests: 4 for `invokeByAlpha = false` and 3 for `invokeByAlpha = true`. But that's okay.

pkg/morph/client/notary.go
alexvanin: Can you give a reference for that in latest neo-go? Couldn't find `ScriptHash()` implementation on account structure.

AnnaShaleva: Oh, sorry, that's my bad, there's no such API defined on `Account`. There's only `acc.PrivateKey().GetScriptHash()`, but it's better to keep ` hash.Hash160(c.acc.GetVerificationScript())` then.

cthulhu-rider (approved):

alexvanin:
Merged to unblock subnetwork implementation. Any other witness related changes may be addressed in separate issues and PRs.

PR #978, opened 2021-11-16
Ignore non-HALT netmap.AddPeer notary invocations 

### LINKED ISSUE #976: Add test invoke check for `AddPeer` notification parser
Notary subsystem can provide `main tx` with invalid witnesses which will fail test invocation. As for `netmap.AddPeer` we have to check test invocation on alphabet nodes manually, because instead of resigning, alphabet nodes create new notary transaction. So it should be double checked that original `main tx` executes correctly.

alexvanin:
Closes #976

### DIFF
--- a/pkg/innerring/processors/netmap/process_peers.go
+++ b/pkg/innerring/processors/netmap/process_peers.go
@@ -18,6 +18,19 @@ func (np *Processor) processAddPeer(ev netmapEvent.AddPeer) {
 		return
 	}
 
+	// check if notary transaction is valid, see #976
+	if originalRequest := ev.NotaryRequest(); originalRequest != nil {
+		tx := originalRequest.MainTransaction
+		ok, err := np.netmapClient.Morph().IsValidScript(tx.Script, tx.Signers)
+		if err != nil || !ok {
+			np.log.Warn("non-halt notary transaction",
+				zap.String("method", "netmap.AddPeer"),
+				zap.String("hash", tx.Hash().StringLE()),
+				zap.Error(err))
+			return
+		}
+	}
+
 	// unmarshal node info
 	nodeInfo := netmap.NewNodeInfo()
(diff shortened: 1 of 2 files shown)


pkg/innerring/processors/netmap/process_peers.go
```diff
@@ -18,6 +18,19 @@ func (np *Processor) processAddPeer(ev netmapEvent.AddPeer) {
 		return
 	}
 
+	// check if notary transaction is valid, see #976
+	if originalRequest := ev.NotaryRequest(); originalRequest != nil {
+		tx := originalRequest.MainTransaction
+		ok, err := np.netmapClient.Morph().IsValidScript(tx.Script, tx.Signers)
```
AnnaShaleva, line 24: IIUC, currently `tx.Signers` still include alphabet multisigner. And `addPeer` still checks for this alphabet signer (https://github.com/nspcc-dev/neofs-contract/blob/8553320e1c611d0462339f48fbd5b36358f33b83/netmap/netmap_contract.go#L213). As a result, witness check for new public key will be skipped, isn't it?

We need to refactor `addPeer` code along with these changes.

AnnaShaleva (commented):
Otherwise LGTM.

pkg/innerring/processors/netmap/process_peers.go
alexvanin: Sure, `addPeer` will be ([most probably](https://github.com/nspcc-dev/neofs-contract/issues/154#issuecomment-970341435)) split into two methods and it will contain mandatory node witness check.
```go
		publicKey := nodeInfo[2:35] // offset:2, len:33
		if !runtime.CheckWitness(publicKey) {
		}
```

AnnaShaleva (approved):

alexvanin:
This PR does not affect implementation for current version of node and contracts, so I merge it. It is going to be useful after https://github.com/nspcc-dev/neofs-contract/issues/154

PR #986, opened 2021-11-26
IR: Listen and process `Subnet.Put/Delete` notifications

### LINKED ISSUE #973: IR: Listen and process `Subnet.Put/Delete` notifications
* Blocked by https://github.com/nspcc-dev/neofs-sdk-go/issues/79
* Related to https://github.com/nspcc-dev/neofs-contract/issues/122

## Summary
`Put` method of `Subnet` contract throws notifications about new subnetworks (carries IDs and binary information). Alphabet nodes should verify all infos and approve only valid ones. Approved subnets will be eventually  stored in `Subnet` contract.

## Task

- [x] Define format of `Put` notification event. Implement notification parser.
- [x] Create IR subnet processor and make it to listen for `Put` notifications from sidechain.
- [x] Verify the event data: it must have non-empty valid ID and owner ID.
- [x] Implement morph client's method to call `Put` method of `Subnet` contract.
- [x] Call `Put` method on notification event approval.

### UPD @cthulhu-rider 24.11.21
`Delete` notifications should be processed also.

cthulhu-rider:
* Closes #973


pkg/innerring/subnet.go
```diff
@@ -0,0 +1,332 @@
...
+	"github.com/nspcc-dev/neofs-node/pkg/util"
+	"github.com/nspcc-dev/neofs-sdk-go/owner"
+	"github.com/nspcc-dev/neofs-sdk-go/subnet"
+	subnetid "github.com/nspcc-dev/neofs-sdk-go/subnet/id"
+	"github.com/panjf2000/ants/v2"
+	"go.uber.org/zap"
+)
+
+// IR server's component to handle Subnet contract notifications.
+type subnetHandler struct {
```
alexvanin, line 24: It looks like `subnetHandler` is literally subnet processor: a structure that contains worker pool, external dependencies such as client or validators. Is there a reason to define it such way?

pkg/morph/event/subnet/put.go
```diff
@@ -0,0 +1,144 @@
...
+		return nil, fmt.Errorf("id item: %w", err)
+	}
+
+	// parse owner
+	put.ownerID, err = client.BytesFromStackItem(items[1])
+	if err != nil {
+		return nil, fmt.Errorf("owner item: %w", err)
+	}
+
+	// parse public key
```
carpawell, line 95:
```suggestion
	// parse subnet information
```

pkg/innerring/subnet.go
```diff
@@ -0,0 +1,332 @@
...
+		handlerInfo.SetRequestType(notifyTyp)
+
+		s.morphListener.SetNotaryParser(parserInfo)
+		s.morphListener.RegisterNotaryHandler(handlerInfo)
+	}
+
+	// subnet creation
+	listenEvent(subnetCreateEvName, subnetevents.ParseNotaryPut, s.onlyAlphabetEventHandler(s.catchSubnetCreation))
+	// subnet removal
+	listenEvent(subnetRemoveEvName, subnetevents.ParseNotaryDelete, s.onlyAlphabetEventHandler(s.catchSubnetRemoval))
```
carpawell, line 139: in notary enabled environmental you need to listen to just `Delete` notifications

carpawell (changes requested):

pkg/innerring/subnet.go
cthulhu-rider: I've decided to have all components related to event listening in a single layer of the app. It doesn't make sense to pass contract address, worker pool and so on to processor and get parser/handlers back. If IR server controls event listener, then it should do all this work on its side.

pkg/morph/event/subnet/delete.go
```diff
@@ -0,0 +1,109 @@
...
+
+// TxHash returns hash of the transaction which thrown the notification event.
+// Makes sense only in non-notary environments (see NotaryMainTx).
+func (x Delete) TxHash() util.Uint256 {
+	return x.txHash
+}
+
+// NotaryMainTx returns main transaction of the request in the Notary service.
+// Returns nil in non-notary environments.
+func (x Delete) NotaryMainTx() *transaction.Transaction {
```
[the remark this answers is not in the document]
cthulhu-rider: All they do is pull out main tx. Since types like `Delete` don't need to follow some interface, it's enough to provide main tx reader.

carpawell (approved):

pkg/innerring/subnet.go
alexvanin: I think it does make sense to group all operations related to the contract in separate entites, so the server is mostly application constructor and storage of the application state. All event handlings are done in processor, so it is quite easy to find, lookup and debug what application does when event appears. E.g. you can simply find out all actions IR is doing after new epoch event in corresponding processor:
https://github.com/nspcc-dev/neofs-node/blob/011d0f605b8149d3dc42b660f214b7f39fb4a116/pkg/innerring/processors/netmap/process_epoch.go#L55-L60

Anyway, like it or not, all other events processed by separate processor entities, so it is kinda architecture decision. It will be more consistent, if subnet events will be processed the same way. Let's rework that in https://github.com/nspcc-dev/neofs-node/issues/991

PR #989, opened 2021-11-28
node: Configure subnets

### LINKED ISSUE #983: Node: Configure the bootstrap subnet list
* Blocked by https://github.com/nspcc-dev/neofs-sdk-go/issues/82

## Summary
NeoFS storage nodes are registered in system subnets (at least one). We should provide the ability to configure the subnets that the node will enter during bootstrap.

## Task

- [ ] Add `subnet` sub-section to `node` section of storage node config. Accept list of subnet IDs in string format.  Add separate configuration value to toggle entry to zero subnet (`bool`).
- [ ] Parse subnet IDs in `neofs-node` app and write them into `NodeInfo` used for bootstap.

## Note
In current NeoFS API version subnet IDs are set through node attributes. We already configure attributes of storage node in `neofs-node` app, therefore a collision is possible. I propose to implement the prevalence of the `subnet` section over the manual setting of the attributes. This fact is worth mentioning in the configuration examples `config/example`.

cthulhu-rider:
* Closes #983


cmd/neofs-node/netmap.go
```diff
@@ -165,6 +169,29 @@ func initNetmapService(c *cfg) {
...
+	var (
+		id  subnetid.ID
+		err error
+	)
+
+	subnetCfg.IterateSubnets(func(idTxt string) {
+		err = id.UnmarshalText([]byte(idTxt))
+		fatalOnErrDetails("parse subnet entry", err)
+
+		c.cfgNodeInfo.localInfo.EnterSubnet(id)
```
carpawell, line 186: what if `exit_zero: true` and `entries` contains `0`?

carpawell (approved):

cmd/neofs-node/netmap.go
cthulhu-rider: `exit_zero` overrides attributes and `entries` (see comment in example config).

carpawell: i just meant that it looks confusing. error instead of overriding

PR #992, opened 2021-11-28
ADM: Add subnet-related commands

### LINKED ISSUE #979: Add administation module for subnetworks
* Blocked by https://github.com/nspcc-dev/neofs-contract/issues/122

## Summary
`Subnet` contract of NeoFS sidechain provides the functionality to create, manage and remove NeoFS subnetworks. We need to provide convenient functionality which allows to do it from admin utility.

## Task

- [ ] Add `subnet` module (`cobra.Command`).
- [ ] Add `create` sub-command which creates new subnetwork using a private key. ID is generated randomly. It'd nice to provide same await approach as `neofs-cli container create`.
- [ ] Add `remove` sub-command which removes subnetwork by ID using a private key. It'd nice to provide same await approach as `neofs-cli container delete`.
- [ ] Provide access to methods for managing lists of keys (clients and nodes) with commands of the same name. Add/remove takes a list of keys. To manage client keys group ID should be provided or generated randomly.

cthulhu-rider:
* Closes #979


cmd/neofs-adm/internal/modules/morph/initialize_deploy.go
```diff
@@ -217,7 +221,12 @@ func (c *initializeContext) deployContracts(method string) error {
 		if method == updateMethodName {
 			ctrHash, err = nnsResolveHash(c.Client, nnsHash, ctrName+".neofs")
 			if err != nil {
-				return fmt.Errorf("can't resolve hash for contract update: %w", err)
+				if errors.Is(err, errMissingNNSRecord) {
+					// if contract not found we deploy it instead of update
+					method = deployMethodName
```
fyrchik, line 226: We should add `method := method` to the start of this loop, otherwise `method` will be changed for the following iterations. `subnet` is last contract but we shouldn't depend on this.

cmd/neofs-adm/internal/modules/morph/subnet.go
```diff
@@ -0,0 +1,870 @@
...
+		return nil
+	},
+}
+
+// cmdSubnetAdmin subnet flags.
+const (
+	// subnet ID to be managed
+	flagSubnetAdminSubnet = flagSubnet
+	// admin public key
+	flagSubnetAdminID = "id"
```
carpawell: suggest to use `admin` instead of `id` to make it more clear for user

cmd/neofs-adm/internal/modules/morph/subnet.go
```diff
@@ -0,0 +1,870 @@
...
+	Short: "Remove client of the NeoFS subnet.",
+	RunE: func(cmd *cobra.Command, _ []string) error {
+		return manageSubnetClients(cmd, true)
+	},
+}
+
+// cmdSubnetNode flags.
+const (
+	// node ID
+	flagSubnetNode = "id"
```
carpawell, line 658: suggest using `node` instead of `id` to make it more clear for user

carpawell (approved):

Issue #749, opened 2021-08-05
Use CustomGroup scope for side chain contract invocations


alexvanin:
Global scope is not that good even if it is used in side chain. Update neofs-contract repo with custom group scope support in manifest and specify this group scope in side chain invocation of morph client.

roman-khimov:
See `contract manifest add-group` command in neo-go to sign manifests (nspcc-dev/neo-go#2100).

alexvanin:
See main chain scope fix in #751

fyrchik:
It can be easier with https://github.com/nspcc-dev/neo-go/pull/2251

alexvanin:
Closed in https://github.com/nspcc-dev/neofs-node/pull/1124

PR #993, opened 2021-11-29
neofs-adm: include neofs contracts in a custom group

fyrchik:
Related #749 .

Morph client will be changed in another PR.
Testing is done on this branch https://github.com/fyrchik/neofs-adm-tests/tree/force-new-epoch

### DIFF
--- a/cmd/neofs-adm/internal/modules/morph/epoch.go
+++ b/cmd/neofs-adm/internal/modules/morph/epoch.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/nspcc-dev/neo-go/pkg/core/transaction"
+	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
 	"github.com/nspcc-dev/neo-go/pkg/io"
 	"github.com/nspcc-dev/neo-go/pkg/smartcontract"
 	"github.com/nspcc-dev/neo-go/pkg/smartcontract/callflag"
@@ -30,10 +31,20 @@ func forceNewEpochCmd(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("can't get netmap contract hash: %w", err)
 	}
 
-	res, err := wCtx.Client.InvokeFunction(nmHash, "epoch", []smartcontract.Parameter{}, []transaction.Signer{{
+	signer := transaction.Signer{
 		Account: wCtx.CommitteeAcc.Contract.ScriptHash(),
 		Scopes:  transaction.Global, // Scope is important, as we have nested call to container contract.
-	}})
+	}
+
+	groupKey, err := nnsResolveKey(wCtx.Client, cs.Hash, groupKeyDomain)
+	if err != nil {
+		cmd.Println("Can't resolve neofs contract group key, using Global scope")
+	} else {
+		signer.Scopes = transaction.CustomGroups
+		signer.AllowedGroups = keys.PublicKeys{groupKey}
+	}
+
+	res, err := wCtx.Client.InvokeFunction(nmHash, "epoch", []smartcontract.Parameter{}, nil)
 	if err != nil || res.State != vm.HaltState.String() || len(res.Stack) == 0 {
 		return errors.New("can't fetch current epoch from the netmap contract")
 	}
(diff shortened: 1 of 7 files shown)


cmd/neofs-adm/internal/modules/morph/epoch.go
```diff
@@ -30,10 +31,21 @@ func forceNewEpochCmd(cmd *cobra.Command, args []string) error {
...
+	groupKey, err := nnsResolveKey(wCtx.Client, cs.Hash, groupKeyDomain)
+	if err != nil {
+		cmd.Println("Can't resolve neofs contract group key, using Global scope")
+	} else {
+		signer.Scopes = transaction.CustomGroups
+		signer.AllowedGroups = keys.PublicKeys{groupKey}
+	}
+
+	res, err := wCtx.Client.InvokeFunction(nmHash, "epoch", []smartcontract.Parameter{},
+		[]transaction.Signer{signer})
```
alexvanin, line 48: Is it required to specify scope for read-only operations like fetching epoch number? I would rather keep them as simple as possible, what do you think?

cmd/neofs-adm/internal/modules/morph/initialize.go
```diff
@@ -28,6 +28,8 @@ type initializeContext struct {
 	// ConsensusAcc is used for retrieving committee address and verification script.
 	ConsensusAcc *wallet.Account
 	Wallets      []*wallet.Wallet
+	// ContractWallet is a wallet for providing contract group signature.
+	ContractWallet *wallet.Wallet
```
alexvanin, line 32: It called `manifest wallet` in code above.

cmd/neofs-adm/internal/modules/morph/generate.go
```diff
@@ -42,6 +42,11 @@ func generateAlphabetCreds(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	err = initializeManifestWallet(walletDir)
+	if err != nil {
+		return err
+	}
```
alexvanin, line 48: Correct me if I am wrong: in the case when we already have all alphabet wallets, we call `generate-alphabet` to generate new wallets but we will take only new manifest wallet from that.

alexvanin (approved):

cmd/neofs-adm/internal/modules/morph/epoch.go
fyrchik: Fixed.

cmd/neofs-adm/internal/modules/morph/generate.go
fyrchik: I think we can initialize it during deploy if it is missing.

cmd/neofs-adm/internal/modules/morph/initialize.go
fyrchik: Fixed.

cmd/neofs-adm/internal/modules/morph/generate.go
alexvanin: So `generate-alphabet` will not rewrite existing wallets now, right? Or it needs to be implemented separately?

cmd/neofs-adm/internal/modules/morph/epoch.go
alexvanin: I still see this snippet below, maybe there is some misunderstanding. Is it necessary to specify group in read-only method invocations? I don't mind to have it, but it is a bit useless in this case, IMO.
```go
	} else {
		signer.Scopes = transaction.CustomGroups
		signer.AllowedGroups = keys.PublicKeys{groupKey}
	}
```

fyrchik: No, it is not necessary.

cmd/neofs-adm/internal/modules/morph/generate.go
fyrchik: Current behaviour:
1. `generate-alphabet` creates alphabet wallets and a contract wallet, overwrites all existing files.
2. `morph init` opens all wallets from disk and creates a contract wallet if it is missing.

PR #994, opened 2021-11-30
[#878] neofs-node: default to secure TLS settings

### LINKED ISSUE #878: TLS vulnerabilities
The same as https://github.com/nspcc-dev/neo-go/issues/2016.

```
TLS 1      offered (deprecated)
TLS 1.1    offered (deprecated)
Triple DES Ciphers / IDEA                     offered
Obsolete CBC ciphers (AES, ARIA etc.)         offered
Has server cipher order?     no (NOT ok)
SWEET32 (CVE-2016-2183, CVE-2016-6329)    VULNERABLE, uses 64 bit block ciphers
BEAST (CVE-2011-3389)                     TLS1: ECDHE-RSA-AES256-SHA AES256-SHA ECDHE-RSA-AES128-SHA AES128-SHA ECDHE-RSA-DES-CBC3-SHA DES-CBC3-SHA 
                                           VULNERABLE -- but also supports higher protocols  TLSv1.1 TLSv1.2 (likely mitigated)
LUCKY13 (CVE-2013-0169), experimental     potentially VULNERABLE, uses cipher block chaining (CBC) ciphers with TLS. Check patches
```

Thought the neo-go issue is closed I still believe this should be fixed:
1. Min TLS should be 1.2 and vulnerable mechanisms must be disabled.
2. There should be configuration settings to define cipher suites.

fyrchik:
Close #878 .

Support TLS >=1.2 only and strong cipher suites.
I have also checked NIST recommendations (
https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-52r2.pdf), but they are less strict and include ciphers considered insecure by go stdlib.

I don't think allowing user to precisely specify cipher suites is worth it.

### DIFF
--- a/cmd/neofs-node/config/grpc/config.go
+++ b/cmd/neofs-node/config/grpc/config.go
@@ -78,6 +78,11 @@ func (tls TLSConfig) CertificateFile() string {
 	return v
 }
 
+// UseInsecureCrypto returns true if TLS 1.2 cipher suite should not be restricted.
+func (tls TLSConfig) UseInsecureCrypto() bool {
+	return config.BoolSafe(tls.cfg, "use_insecure_crypto")
+}
+
 // IterateEndpoints iterates over subsections ["0":"N") (N - "num" value)
 // of "grpc" section of c, wrap them into Config and passes to f.
 //
--- a/cmd/neofs-node/config/grpc/config_test.go
+++ b/cmd/neofs-node/config/grpc/config_test.go
@@ -31,12 +31,17 @@ func TestGRPCSection(t *testing.T) {
 			case 0:
 				require.Equal(t, "s01.neofs.devenv:8080", sc.Endpoint())
 
+				require.NotNil(t, tls)
 				require.Equal(t, "/path/to/cert", tls.CertificateFile())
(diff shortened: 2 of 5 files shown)


cmd/neofs-node/config/grpc/config.go
```diff
@@ -78,6 +78,11 @@ func (tls TLSConfig) CertificateFile() string {
 	return v
 }
 
+// UseInsecureCrypto returns true if TLS 1.2 cipher suite should not be restricted.
```
cthulhu-rider, line 81: Follow the same pattern of docs like other methods.

cthulhu-rider (approved):
LGTM

532910:
If this is going to be fixed, should nspcc-dev/neo-go#2016 be fixed as well?

> I don't think allowing user to precisely specify cipher suites is worth it.

Disagree:
1. It should not be difficult, you don't need to parse or validate user setting, just bypass it to upstream ssl library.
2. go is static, you can't `LD_PRELOAD` [TLS Interposer](https://netfuture.ch/tools/tls-interposer/)

alexvanin (approved):

carpawell (approved):

PR #996, opened 2021-11-30
neofs-adm: implement `deposit-notary` command

### LINKED ISSUE #995: ADM: Add `deposit-notary` command
We need to deposit funds for some operations using Notary service. It'd be convinent to have dedicated command in admin utility.

fyrchik:
Close #995 .

### DIFF
--- a/cmd/neofs-adm/internal/modules/morph/notary.go
+++ b/cmd/neofs-adm/internal/modules/morph/notary.go
@@ -1,1 +1,132 @@
 package morph
+
+import (
+	"fmt"
+	"strconv"
+
+	"github.com/nspcc-dev/neo-go/cli/input"
+	"github.com/nspcc-dev/neo-go/pkg/core/native/nativenames"
+	"github.com/nspcc-dev/neo-go/pkg/core/transaction"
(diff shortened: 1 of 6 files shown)


cmd/neofs-adm/internal/modules/morph/notary.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	height, err := c.GetBlockCount()
+	if err != nil {
+		return fmt.Errorf("can't get current height: %v", err)
+	}
+
+	bw := io.NewBufBinWriter()
+	emit.AppCall(bw.BinWriter, gasHash, "transfer", callflag.All,
+		accHash, notaryHash.BytesBE(), int64(gasAmount), []interface{}{nil, int64(height + 5760)})
```
alexvanin, line 92: Is 5760 has special meaning?
Maybe it can be default value for new `--duration` flag.

fyrchik: It is the default value in notary contract:
https://github.com/nspcc-dev/neo-go/blob/master/pkg/core/native/notary.go#L48

alexvanin (approved):

PR #1012, opened 2021-12-07
Call Register method on peer approval

### LINKED ISSUE #1009: Call `netmap.Register` method for storage node approval in notary enabled environment
Adopt changes from https://github.com/nspcc-dev/neofs-contract/pull/189

In notary enabled environment, call `Register` method instead of `AddPeer`. We can't call `AddPeer` because of https://github.com/nspcc-dev/neofs-contract/issues/154#issuecomment-970341435

cthulhu-rider:
* closes #1009

### DIFF
--- a/pkg/innerring/processors/netmap/process_peers.go
+++ b/pkg/innerring/processors/netmap/process_peers.go
@@ -88,6 +88,11 @@ func (np *Processor) processAddPeer(ev netmapEvent.AddPeer) {
 		prm := netmapclient.AddPeerPrm{}
 		prm.SetNodeInfo(nodeInfo)
 
+		// In notary environments we call Register method instead of AddPeer.
+		// It differs from AddPeer only by name, so we can do this in the same form.
+		// See https://github.com/nspcc-dev/neofs-contract/issues/154.
+		const methodAddPeerNotary = "register"
+
 		if nr := ev.NotaryRequest(); nr != nil {
 			// create new notary request with the original nonce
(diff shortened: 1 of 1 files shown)


pkg/morph/client/netmap/wrapper/wrapper.go
```diff
@@ -29,26 +29,43 @@ type Wrapper struct {
 // parameter of Wrapper.
 type Option func(*opts)
 
-type opts []client.StaticClientOption
+type opts struct {
+	alphabet bool
```
carpawell, line 33: why have you connected all logic with `alphabet`? seems like it is gonna work but it's too implicit IMO

alexvanin (changes requested):
The fix should be a bit different. 
`Register` method should be invoked in notary enabled environment, because there Alphabet nodes do not resign tx, but creates new notary tx. 

Therefore the only change should be performed there: use `register` instead of `addPeer` in https://github.com/nspcc-dev/neofs-node/blob/89eca449b61a918aabc816f8f40bdcc6131931e3/pkg/innerring/processors/netmap/process_peers.go#L98

alexvanin (approved):

pkg/morph/client/netmap/wrapper/wrapper.go
cthulhu-rider: No longer relevant

carpawell (approved):

PR #1022, opened 2021-12-08
writecache: Fix db counter decrement

ZhangTao1596:

@alexvanin

### DIFF
--- a/pkg/local_object_storage/writecache/storage.go
+++ b/pkg/local_object_storage/writecache/storage.go
@@ -98,7 +98,6 @@ func (c *cache) deleteFromDB(keys [][]byte) error {
 	if len(keys) == 0 {
 		return nil
 	}
-	var sz uint64
 	err := c.db.Update(func(tx *bbolt.Tx) error {
 		b := tx.Bucket(defaultBucket)
 		for i := range keys {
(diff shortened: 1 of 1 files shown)


alexvanin:
Will be included in the patch for v0.27.0 release.

alexvanin:
Rebase it after https://github.com/nspcc-dev/neofs-node/pull/1028

pkg/local_object_storage/writecache/storage.go
```diff
@@ -109,15 +108,16 @@ func (c *cache) deleteFromDB(keys [][]byte) error {
...
-			sz += uint64(len(has))
 			storagelog.Write(c.log, storagelog.AddressField(string(keys[i])), storagelog.OpField("db DELETE"))
 		}
 		return nil
 	})
 	if err != nil {
 		return err
 	}
-	c.objCounters.DecDB()
+	for range keys {
```
cthulhu-rider, line 118: It is not always possible to delete all objects by keys. I propose, taking into account the lightness of calling `c.objCounters.DecDB()`, to do it right in the transaction above after successful `b.Delete(keys[i])`. Or we can use counter and iterate until its value here.

ZhangTao1596: If an error occurs, will the deleted object not be recovered?

cthulhu-rider: Oh, yes, ur right, I missed that we only change the internal storage of the BoltDB. According to `Update` docs `the entire transaction is rolled back`, so your code is correct.

cthulhu-rider (approved):
@ZhangTao1596 could you please squash your commits into a single one (with issue num)?

ZhangTao1596:
> @ZhangTao1596 could you please squash your commits into a single one (with issue num)?

👌

alexvanin (approved):

PR #1031, opened 2021-12-09
Feat/Limit `neo-go` connections per host

### LINKED ISSUE #780: Use MaxConnsPerHost in neo-go client
If https://github.com/nspcc-dev/neo-go/issues/2149 is going to be implemented, we should use it.

carpawell:
Closes #780.

### DIFF
--- a/cmd/neofs-adm/internal/modules/morph/n3client.go
+++ b/cmd/neofs-adm/internal/modules/morph/n3client.go
@@ -20,12 +20,16 @@ type clientContext struct {
 }
 
 func getN3Client(v *viper.Viper) (*client.Client, error) {
+	// number of opened connections
+	// by neo-go client per one host
+	const maxConnsPerHost = 10
+
 	ctx := context.Background() // FIXME(@fyrchik): timeout context
 	endpoint := v.GetString(endpointFlag)
 	if endpoint == "" {
 		return nil, errors.New("missing endpoint")
 	}
-	c, err := client.New(ctx, endpoint, client.Options{})
+	c, err := client.New(ctx, endpoint, client.Options{MaxConnsPerHost: maxConnsPerHost})
 	if err != nil {
 		return nil, err
 	}
--- a/cmd/neofs-node/config/morph/config.go
+++ b/cmd/neofs-node/config/morph/config.go
@@ -19,6 +19,9 @@ const (
 
 	// NotaryDepositDurationDefault is a default deposit duration.
 	NotaryDepositDurationDefault uint32 = 1000
+
+	// MaxConnPerHostDefault is a default maximum of connections per host of the morph client.
+	MaxConnPerHostDefault = 10
 )
 
 // RPCEndpoint returns list of values of "rpc_endpoint" config parameter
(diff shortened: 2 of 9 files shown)


cmd/neofs-adm/internal/modules/morph/n3client.go
```diff
@@ -20,12 +20,14 @@ type clientContext struct {
 }
 
 func getN3Client(v *viper.Viper) (*client.Client, error) {
+	const maxConnsPerHost = 1
```
cthulhu-rider, line 23: Pls add a brief meaning, otherwise const is not useful.

cthulhu-rider (approved):

cmd/neofs-adm/internal/modules/morph/n3client.go
carpawell: @alexvanin, @cthulhu-rider, let's discuss what limit should we use and should it be configurable or not

alexvanin: TLDR:
1. `maxConnsPerHost = 1` is too small, let's use `maxConnsPerHost = 10`.
2. For NeoFS installations in controlled environment, we can improve performance by increasing `maxConnsPerHost`, so maybe it makes sense to make it configurable.

I did some benchmarks with the client, which fetches 1000 blocks with tx application logs info from the testnet side chain (about 2250 requests in total). 
Here are the results:

| MaxConnsPerHost  | Total duration with 100 routines, sec. | Total duration with 10 routines, sec. |
| ------------- | ------------- | ------------- |
| 0|4.089|20.369
1|120.244|111.339
3|56.372|53.042 
5|37.894|36.364 
10|17.215|24.188
15|9.829|19.737 
20|7.789|20.07

Two things:
- with `MaxConnsPerHost = 0` number of connections equals the number of workers, that's why there is no effect for 10 and 15 `MaxConnsPerHost` with 10 routines,
- `MaxConnsPerHost = 1` affects performance drastically. 

In the storage nodes I expect from 10 to 100 routines call neo-go client to fetch the data (e.g. up to 100 simultaneous `object.Get` request with different container ID). In the private installation we can configure RPC nodes and balance system load so every Storage node can fire up 20, 50 or even 100 connections to RPC node. 

In public networks, I am afraid we can't predict system load and RPC availability. Therefore it makes sense to limit amount of connections. Let's say RPC node processes about 1000 incoming connections. With `maxConnsPerHost = 10` it will process 100 Storage nodes, which is more than enough. 

/cc @realloc @roman-khimov

cmd/neofs-adm/internal/modules/morph/n3client.go
```diff
@@ -20,12 +20,16 @@ type clientContext struct {
...
+	// by neo-go client per one host
+	const maxConnsPerHost = 10
+
 	ctx := context.Background() // FIXME(@fyrchik): timeout context
 	endpoint := v.GetString(endpointFlag)
 	if endpoint == "" {
 		return nil, errors.New("missing endpoint")
 	}
-	c, err := client.New(ctx, endpoint, client.Options{})
+	c, err := client.New(ctx, endpoint, client.Options{MaxConnsPerHost: maxConnsPerHost})
```
alexvanin, line 32: Actually, `neofs-adm` might ignore this limit.

The limit is important for running applications that reuse neo-go connection over time. `neofs-adm` runs in a single thread and exits on the finish, so it can live with default 0. I don't mind to set it though, do it as you wish.

alexvanin (approved):

cmd/neofs-adm/internal/modules/morph/n3client.go
carpawell: thought about it, but decided that to have a limit and do not use it is better than not to have

Issue #980, opened 2021-11-18
CLI: Output of NeoFS status codes


cthulhu-rider:
After 
* https://github.com/nspcc-dev/neofs-api/pull/176
* https://github.com/nspcc-dev/neofs-node/pull/961

NeoFS CLI is ready to adjust its behavior depending on the NeoFS return code.

Examples:

- write output `Code: XXX`
- map to process status code and exit with it.

cthulhu-rider:
Note that we don't have enough OS exit codes for one-to-one correspondence:
* gostd lib says `For portability, the status code should be in the range [0, 125].` https://pkg.go.dev/os#Exit
* bash-scripting guide confirms https://tldp.org/LDP/abs/html/exitcodes.html

In practice, we should have enough 126 values per command. Upon reaching the value of 125, it will be possible to come up with a workaround,

PR #1039, opened 2021-12-13
Feat/Make return codes depend on errors

carpawell:
Related to #980.

This PR makes `cli` try to cast client's errors to one of the well-known `sdk` errors and return the corresponding exit code. It does not make sense to try to localize global status code since it is version-dependent and requires more casing logic.

When more general errors are added in the `sdk` library, `switch` has to be expanded with them.


alexvanin (approved):
Seems ok for me

cthulhu-rider (approved):

PR #1040, opened 2021-12-14
Support autocomplete in neofs-adm

### LINKED ISSUE #1013: shell autocomplete generator
It would be nice if `neofs-cli` could generate autocomplete file for different shells.

Something like https://gohugo.io/commands/hugo_gen_autocomplete/

alexvanin:
Closes #1013

### DIFF
--- a/cmd/neofs-adm/internal/modules/autocomplete.go
+++ b/cmd/neofs-adm/internal/modules/autocomplete.go
@@ -1,0 +1,58 @@
+package modules
+
+import (
+	"os"
+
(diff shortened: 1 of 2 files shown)


carpawell (approved):

cthulhu-rider (approved):
Candidate for sharing between utilities? In CLI I see a similar code.

alexvanin:
> Candidate for sharing between utilities? In CLI I see a similar code.

Yes, we can refactor it a bit to share this code. There should be help message generator that substitutes application name.

PR #1048, opened 2021-12-20
Feat/Add shard listing in `CLI`

### LINKED ISSUE #1043: Support shard listing in CLI via Control API
**:related to shard dump/restore:**

To operate with shards, admin should be able at least list them and see details about them. There should be service in Control API to provide list of shards and additional information about them. It might be two separate RPC that CLI can combine. Shard info should contain at least:
- db path
- mode (read-only / read-write)

Possible implementation:
```
$ neofs-cli control shards list
id:XXX
path:/etc/neofs/shard0
mode: read-only

id:YYY
path:/etc/neofs/shard1
mode: read-write
```

carpawell:
Closes #1043. Also, adds filling `shard`'s info with its components' info (`metabase`, `blobstore`, `writecache`). It seems that `Shard`'s `DumpInfo()` worked not as it was expected since it returned information that was never filled.


pkg/services/control/types.proto
```diff
@@ -119,3 +119,33 @@ enum HealthStatus {
...
+    bytes shard_ID = 1 [json_name = "shardID"];
+
+    // Path to shard's metabase.
+    string metabase_path = 2 [json_name = "metabasePath"];
+
+    // Path to shard's blobstore.
+    string blobstore_path = 3 [json_name = "Path"];
+
+    // Path to shard's write-cache, empty if disabled.
+    string writecache_path = 4 [json_name = "Path"];
```
alexvanin, line 135: Same JSON names for blobstor and writecache.

pkg/services/control/types.proto
```diff
@@ -119,3 +119,33 @@ enum HealthStatus {
...
+// Shard description.
+message ShardInfo {
+    // ID of the shard.
+    bytes shard_ID = 1 [json_name = "shardID"];
+
+    // Path to shard's metabase.
+    string metabase_path = 2 [json_name = "metabasePath"];
+
+    // Path to shard's blobstore.
+    string blobstore_path = 3 [json_name = "Path"];
```
alexvanin, line 132: It is `blobstor`, not `blobstore`.

pkg/services/control/types.proto
carpawell: added prefix

pkg/services/control/types.proto
carpawell: renamed

alexvanin (approved):

PR #1055, opened 2021-12-27
neofs-adm: update contracts in a single transaction

### LINKED ISSUE #1035: Dry-run for contract updates
neofs-adm updates contracts one by one. If one contract fails, then previous contracts won't be roll backed. This makes updating a bit more complicated if one contract fails to update. Then we should prepare custom version of the contracts, where `update` is ignored for already updated contracts. 

Maybe we can do test invokes for all contracts before sending actual transactions to the network. `--dry-run` flag will be very appreciated.

fyrchik:
Close #1035.

Same thing is desireable for deploy, but alphabet contracts have to be deployed from different senders to have different hash. Also contrary to `update` network starts to function only after all contracts were deployed, so there we won't end in an inconsistent state.

### DIFF
--- a/cmd/neofs-adm/internal/modules/morph/initialize_deploy.go
+++ b/cmd/neofs-adm/internal/modules/morph/initialize_deploy.go
@@ -12,9 +12,11 @@ import (
 	"path"
 	"strings"
 
+	"github.com/nspcc-dev/neo-go/pkg/core/native"
 	"github.com/nspcc-dev/neo-go/pkg/core/native/nativenames"
 	"github.com/nspcc-dev/neo-go/pkg/core/state"
 	"github.com/nspcc-dev/neo-go/pkg/core/transaction"
+	io2 "github.com/nspcc-dev/neo-go/pkg/io"
 	"github.com/nspcc-dev/neo-go/pkg/rpc/client"
 	"github.com/nspcc-dev/neo-go/pkg/smartcontract"
 	"github.com/nspcc-dev/neo-go/pkg/smartcontract/manifest"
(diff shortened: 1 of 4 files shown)


fyrchik:
I have tested update from master to master.
GAS cost is 47.9157516, script length is 47794, which is close to the maximum allowed script length (65535). However tests are done with a single-node committee. Alphabet contract is currently 4445 in size, so if we have 7 nodes, this will fail. But because alphabet contracts are the same, we can reuse NEF and manifest for performing multiple updates.

cmd/neofs-adm/internal/modules/morph/initialize_deploy.go
```diff
@@ -147,27 +152,137 @@ func (c *initializeContext) deployContracts(method string) error {
 	}
 	nnsHash := nnsCs.Hash
 
+	totalGasCost := int64(0)
+	w := io2.NewBufBinWriter()
+
 	var keysParam []smartcontract.Parameter
 
+	// Update script size for a single-node committee is close to the maximum allowed size of 65535.
```
alexvanin, line 160: If update script size is close to tx limit, maybe we can split alphabet contract deployment and other contract deployment?

alexvanin:
Also check [lint errors](https://github.com/nspcc-dev/neofs-node/runs/4641866362?check_suite_focus=true)

cmd/neofs-adm/internal/modules/morph/initialize_deploy.go
fyrchik: It is now 48k which is 17k less than 65k. I think we can safely do everything in one transaction unless we add another contract.

alexvanin (approved):

Issue #1044, opened 2021-12-16
Change shard mode in CLI via Control API


alexvanin:
**:related to shard dump/restore:**
Blocked by #1043 

Implement Control API method to update shard mode. There should be two shard modes:
- read-write (default)
- read-only

Make sure, that storage engine works with read only shards correctly (chooses another shard to store data on).

```
$ neofs-cli control shards set-status --id X --status read-only
```

PR #1057, opened 2021-12-27
Feat/Add read-only shard mode

carpawell:
Related to #1044. Deleted unused shard modes and suggest adding them when needed. Do not know if shard should be able to tick its `GC` (and remove objects) and accept `Inhume()` calls when it's in `read-only` mode or not.

### DIFF
--- a/cmd/neofs-node/config/engine/shard/config.go
+++ b/cmd/neofs-node/config/engine/shard/config.go
@@ -1,11 +1,14 @@
 package shardconfig
 
 import (
+	"fmt"
+
 	"github.com/nspcc-dev/neofs-node/cmd/neofs-node/config"
 	blobstorconfig "github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/engine/shard/blobstor"
 	gcconfig "github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/engine/shard/gc"
 	metabaseconfig "github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/engine/shard/metabase"
 	writecacheconfig "github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/engine/shard/writecache"
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/shard"
 )
 
 // Config is a wrapper over the config section
@@ -67,4 +70,26 @@ func (x *Config) RefillMetabase() bool {
 		(*config.Config)(x),
 		"resync_metabase",
 	)
+}
+
+// Mode return value of "mode" config parameter.
+//
+// Panics if read value is not one of predefined
+// shard modes.
+func (x *Config) Mode() (m shard.Mode) {
+	s := config.StringSafe(
+		(*config.Config)(x),
+		"mode",
+	)
+
+	switch s {
+	case "read-write", "":
+		m = shard.ModeReadWrite
+	case "read-only":
+		m = shard.ModeReadOnly
+	default:
+		panic(fmt.Sprintf("unknown shard mode: %s", s))
+	}
+
+	return
 }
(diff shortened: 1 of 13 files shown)


alexvanin (commented):
`shard.Delete`, `shard.ToMoveIt` should also check shard mode?

alexvanin:
>  Do not know if shard should be able to tick its GC (and remove objects) and accept Inhume() calls when it's in read-only mode or not.

Restrict any changes in metabase and blobstor in `read-only` mode, so `Inhume` must not be executed.

We can stop GC ticks, but it requires to change more internal logic (e.g. removing `sync.Once`), maybe we can optimize it later.

cmd/neofs-node/config/engine/shard/config.go
```diff
@@ -68,3 +71,25 @@ func (x *Config) RefillMetabase() bool {
...
+	s := config.StringSafe(
+		(*config.Config)(x),
+		"mode",
+	)
+
+	switch s {
+	case "read-write", "":
+		m = shard.ModeReadWrite
+	case "read-only":
+		m = shard.ModeReadOnly
```
fyrchik, line 89: Do we plan to support any more modes? IMO having `readonly: true` is somewhat more common.

pkg/local_object_storage/shard/gc.go
```diff
@@ -171,7 +171,12 @@ func (gc *gc) stop() {
 
 // iterates over metabase graveyard and deletes objects
 // with GC-marked graves.
+// Does nothing if shard is in "read-only" mode.
 func (s *Shard) removeGarbage() {
+	if s.getMode() == ModeReadOnly {
+		return
```
fyrchik, line 177: Shouldn't we panic here?

pkg/local_object_storage/shard/mode.go
```diff
@@ -1,50 +1,43 @@
...
 	_ Mode = iota
 
-	// ModeActive is a Mode value for active shard.
-	ModeActive
-
-	// ModeInactive is a Mode value for inactive shard.
-	ModeInactive
+	// ModeReadWrite is a Mode value for shard that is available
+	// for read and write operations. Default shard mode.
+	ModeReadWrite
```
fyrchik, line 19: May be it's time to have `ReadWrite` as default mode (0)?

fyrchik:
What about opening all bolt-db databases with `Readonly: true` in `bbolt.Options` and with 0444 permission?

cmd/neofs-node/config/engine/shard/config.go
alexvanin: I think @realloc has many more mode example for the neofs-next: degraded mode, fault, evacuation.

pkg/local_object_storage/shard/gc.go
alexvanin: This PR doesn't stop GC worker, thus it is a valid case to call `removeGarbage` on read-only shard. We can't easily stop and start GC worker, it will require a bit of change in internal components (remove some `sync.Once` code), so we can postpone it for separate PR.

cmd/neofs-node/config/engine/shard/config.go
realloc: In the future we'd like to have at least:
- read-write // Normal
- read-only // Both MB and BS are working, but in R/O.
- read-only-purge // Can only delete data. When MB or BS are temporary full
- fault // Both Mb and BS are not functional
- degraded-read-write // MB is lost, operations limited, but can slowly accept writes
- degraded-read-only // MB is lost, R/O with limited operations
- degraded-read-only-meta // BS lost, only MB in R/O

carpawell:
> shard.Delete, shard.ToMoveIt should also check shard mode?

@alexvanin, thought that `Delete` should not be called in a normal situation (except by `GC`), but agree that it is better to add the same return statement explicitly

added to `Delete` and `ToMoveIt`

pkg/local_object_storage/shard/mode.go
carpawell: have nothing against it

made it default

alexvanin (approved):

PR #1071, opened 2022-01-10
Fix missing response status at ACL problems

### LINKED ISSUE #1062: Status can't cover errors during acl check
I think this is because acl checker is outside of sign service.
## Expected Behavior
Status in response can carry errors during acl check. For example, `container not found` error.

## Current Behavior
Still return error like before

`Get` a object whoes container id is inexistent.

BTW, Why we need status in response? We can throw exception with [StatusCode](https://grpc.github.io/grpc/csharp/api/Grpc.Core.StatusCode.html) in c# and return error in golang as we do before.
Now we need to check both response and exception(error).

cthulhu-rider:
* closes #1062

### DIFF
--- a/cmd/neofs-node/object.go
+++ b/cmd/neofs-node/object.go
@@ -352,16 +352,6 @@ func initObjectService(c *cfg) {
 		},
 	)
 
-	respSvc := objectService.NewResponseService(
-		splitSvc,
-		c.respSvc,
-	)
-
-	signSvc := objectService.NewSignService(
-		&c.key.PrivateKey,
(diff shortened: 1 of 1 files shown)


cthulhu-rider:
@ZhangTao1596 could you please check if the problem is fixed?

ZhangTao1596:
> @ZhangTao1596 could you please check if the problem is fixed?

Tested. It works well.

carpawell (approved):

alexvanin (approved):

PR #1073, opened 2022-01-10
blobstor: allow to disable compression based on content-type

### LINKED ISSUE #1060: Compress objects depending on their type
There is not much sense in compressing some data types, like mpeg4/5 video or jpeg pictures. Maybe, NeoFS node should have an option with the list of types to skip payload compression on object save in blob store.

fyrchik:
Close #1060.

For some data compression makes little sense, as it is already compressed.
This commit allows to leave such data unchanged based on `Content-Type`
attribute. Currently exact, prefix and suffix matching are supported.


cmd/neofs-node/config/engine/shard/blobstor/config.go
```diff
@@ -88,6 +88,15 @@ func (x *Config) Compress() bool {
 	)
 }
 
+// UncompressableContentTypes returns value of "compress_skip_content_types" config parameter.
```
cthulhu-rider, line 91:
```suggestion
// UncompressableContentTypes returns value of "disable_compression_content_types" config parameter.
```

pkg/local_object_storage/writecache/flush.go
```diff
@@ -125,7 +125,7 @@ func (c *cache) flushBigObjects() {
 					return nil
 				}
 
-				if _, err := c.blobstor.PutRaw(addr, data); err != nil {
+				if _, err := c.blobstor.PutRaw(addr, data, true); err != nil {
```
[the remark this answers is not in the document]
fyrchik: I haven't yet found the right way to make a decision here.
So we either
1. Unmarshal object just to determine if it needs to be compressed. This is slow.
2. Put compressed objects in writecache. This needs some refactoring, I think I will implement it right here.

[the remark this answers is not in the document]
cthulhu-rider: or since we put objects to writecache in "unmarshaled" format we can
3. Calc compression flag in the same way and keep its value in writecache. Later we can pass the flag during `flush`.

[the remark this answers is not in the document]
fyrchik: Yes, 3 should be faster. Fixed.

cmd/neofs-node/config/engine/shard/blobstor/config.go
fyrchik: `compression_exclude_content_types` ? @532910 @realloc

cthulhu-rider (approved):

cmd/neofs-node/config/engine/shard/blobstor/config.go
532910: exclude is better than disable
`no_compressed_content_types`
`nocompression_for_content_types`
`uncompressed_content_types`
`don't_compress_content_types`

realloc: `compression_exclude_content_types`?

alexvanin (approved):

PR #1076, opened 2022-01-11
Read IR list from role management contract w/o mainchain

### LINKED ISSUE #1025: IsActive() status when main chain disabled
During the tests I had a situation when main chain was disabled, but alphabet node still consideres non active (IsActive() returns false). Not sure if this is correct behavior. Unfortunately it is quite hard to keep track of such things when you have so many running modes.

cthulhu-rider:
* closes #1025

### DIFF
--- a/pkg/innerring/innerring.go
+++ b/pkg/innerring/innerring.go
@@ -536,7 +536,9 @@ func New(ctx context.Context, log *zap.Logger, cfg *viper.Viper) (*Server, error
 
 	var irf irFetcher
 
-	if !server.mainNotaryConfig.disabled {
+	if server.withoutMainNet || !server.mainNotaryConfig.disabled {
+		// if mainchain is disabled we should use NeoFSAlphabetList client method according to its docs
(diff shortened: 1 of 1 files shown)


pkg/innerring/innerring.go
```diff
@@ -536,7 +536,9 @@ func New(ctx context.Context, log *zap.Logger, cfg *viper.Viper) (*Server, error
 
 	var irf irFetcher
 
-	if !server.mainNotaryConfig.disabled {
+	if server.withoutMainNet || !server.mainNotaryConfig.disabled {
+		// if mainchain is disabled we should use NeoFSAlphabetList client method according to its docs
+		// (naming `...WithNotary` will not always be correct)
```
carpawell, line 541: should we rename it then?

carpawell (approved):

pkg/innerring/innerring.go
alexvanin: As soon as we drop notary disabled mode from the code base.

alexvanin (approved):

PR #1078, opened 2022-01-11
Factor out autocomplete command generation

### LINKED ISSUE #1041: Share autocomplete code between apps
> Candidate for sharing between utilities? In CLI I see a similar code.

Yes, we can refactor it a bit to share this code. There should be help message generator that substitutes application name.

_Originally posted by @alexvanin in https://github.com/nspcc-dev/neofs-node/issues/1040#issuecomment-993600546_

fyrchik:
Close #1041.


pkg/util/autocomplete/autocomplete.go
```diff
@@ -0,0 +1,60 @@
...
+  To load completions for each session, execute once:
+  $ %s completion fish > ~/.config/fish/completions/%s.fish
+`
+
+// Command returns cobra command structure for autocomplete routine.
+func Command(name string) *cobra.Command {
+	return &cobra.Command{
+		Use:   "completion [bash|zsh|fish|powershell]",
+		Short: "Generate completion script",
+		Long: fmt.Sprintf(longHelpTemplate,
```
carpawell, line 42: wouldn't it be more accurate to replace it somehow in the copy of the template than "`name, name, name...`"

carpawell (approved):

pkg/util/autocomplete/autocomplete.go
fyrchik: The only other way I see is to use `text/template` but it is more heavy and cumbersome to setup. Linter will catch any errors regarding parameter count anyway.

alexvanin (approved):

alexvanin (approved):

PR #1087, opened 2022-01-13
Upd/Update `neofs-sdk-go`

carpawell:
Update includes:
1. New simple client mode that parses erroneous status codes as error and
returns them from the calls of the client methods. (for `cli` only).
2. Client is struct now, not an interface. This leads to the new interfaces inside `pkg/client` package.

### DIFF
--- a/pkg/core/client/client.go
+++ b/pkg/core/client/client.go
@@ -1,15 +1,58 @@
 package client
 
 import (
+	"context"
+	"io"
+
 	rawclient "github.com/nspcc-dev/neofs-api-go/v2/rpc/client"
 	"github.com/nspcc-dev/neofs-node/pkg/network"
 	"github.com/nspcc-dev/neofs-sdk-go/client"
+	"github.com/nspcc-dev/neofs-sdk-go/container"
+	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
+	"github.com/nspcc-dev/neofs-sdk-go/eacl"
+	"github.com/nspcc-dev/neofs-sdk-go/owner"
 )
 
 // Client is an interface of NeoFS storage
 // node's client.
 type Client interface {
-	client.Client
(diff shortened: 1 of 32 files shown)


fyrchik (commented):
Splitting this in 3 commits doesn't make sense to me, as tests fail between them, so a whole PR can be considered atomic.

pkg/core/client/client.go
```diff
@@ -1,15 +1,58 @@
...
+	CreateSession(context.Context, uint64, ...client.CallOption) (*client.CreateSessionRes, error)
+
+	Raw() *rawclient.Client
+
+	Conn() io.Closer
+}
+
+// MultiAddressClient is an interface of the
+// Client that supports multihost work.
+type MultiAddressClient interface {
```
cthulhu-rider, line 54: Lets try to avoid this type. If some package needs such an interface, let it define it itself.

carpawell: so you mean to reorganize interfaces? should it be done in that PR?

i ask because there are no new interfaces: this PR's `Client` is ex `Client` from SDK and this PR's `MultiAddressClient` is ex `Client` from the same package of the node repo, so just renaming/moving, no new interfaces in fact

carpawell:
> Splitting this in 3 commits doesn't make sense to me, as tests fail between them, so a whole PR can be considered atomic.

I did this because every commit's body has some additional information that is important (IMO). I believe that every PR should be atomic but I treat `[#xxxx] ...` at the beginning of the commit message as a way to show its atomicity. But every single commit could break something.

But if you insist I have nothing against it.

pkg/core/client/client.go
cthulhu-rider: I got it, then do not rename and abstract from the fact that the client is multiaddress.

cthulhu-rider: We've decided with @carpawell to reconsider client interfaces later.

cthulhu-rider (approved):

PR #1088, opened 2022-01-14
[#1088] writecache: Fix small flushed mark

ZhangTao1596:

### DIFF
--- a/pkg/local_object_storage/writecache/persist.go
+++ b/pkg/local_object_storage/writecache/persist.go
@@ -79,12 +79,14 @@ func (c *cache) persistSmallObjects(objs []objectInfo) {
 	if err != nil {
 		overflowIndex = 0
 	} else {
-		c.evictObjects(overflowIndex)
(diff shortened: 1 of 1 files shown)


fyrchik (commented):
Nice catch! We need only to fix `c.evictObjects` argument before to remove right number of elements.
Otherwise, LGTM

ZhangTao1596:
> Nice catch! We need only to fix `c.evictObjects` argument before to remove right number of elements. Otherwise, LGTM

Yeah. evictObjects argument needs fix too. Should I copy https://github.com/nspcc-dev/neofs-node/pull/1089/commits/f1a6262508a33b094432414596dce01795933d04 here?

fyrchik:
@ZhangTao1596 yes, that works.

ZhangTao1596:
> @ZhangTao1596 yes, that works.

Done.

fyrchik (approved):

PR #1089, opened 2022-01-18
Allow to dump data from shard

### LINKED ISSUE #1085: Dump / restore objects in engine and shard
Implement dump / restore functions in engine package. Before that, define format of the dumped data. Possible options:
- tar stream of marshaled objects
- protobuf encoded slice of objects
- something else?
---
image
image

fyrchik:
Close #1085 . Related #1057 . Depends on #1088 .

Proto-marshaled list is not an option because it requires (let's be realistic) having the slice in memory.
I decided to not use TAR archives because of 512-byte header overhead for each object. It is noticeable only when lot's of small objects without attributes are present, though.

### DIFF
--- a/pkg/local_object_storage/shard/evacuate.go
+++ b/pkg/local_object_storage/shard/evacuate.go
@@ -7,6 +7,7 @@ import (
 	"os"
 
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor"
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/writecache"
 )
 
 var dumpMagic = []byte("NEOF")
@@ -72,7 +73,7 @@ func (s *Shard) Evacuate(prm *EvacuatePrm) (*EvacuateRes, error) {
 	var count int
 
 	if s.hasWriteCache() {
-		err := s.writeCache.Iterate(func(data []byte) error {
+		err := s.writeCache.Iterate(new(writecache.IterationPrm).WithHandler(func(data []byte) error {
 			var size [4]byte
 			binary.LittleEndian.PutUint32(size[:], uint32(len(data)))
 			if _, err := w.Write(size[:]); err != nil {
@@ -85,7 +86,7 @@ func (s *Shard) Evacuate(prm *EvacuatePrm) (*EvacuateRes, error) {
 
 			count++
 			return nil
-		})
+		}))
 		if err != nil {
 			return nil, err
 		}
--- a/pkg/local_object_storage/shard/evacuate.go
+++ b/pkg/local_object_storage/shard/evacuate.go
@@ -72,7 +72,23 @@ func (s *Shard) Evacuate(prm *EvacuatePrm) (*EvacuateRes, error) {
 	var count int
 
 	if s.hasWriteCache() {
-		// TODO evacuate objects from write cache
+		err := s.writeCache.Iterate(func(data []byte) error {
+			var size [4]byte
+			binary.LittleEndian.PutUint32(size[:], uint32(len(data)))
+			if _, err := w.Write(size[:]); err != nil {
+				return err
+			}
+
+			if _, err := w.Write(data); err != nil {
+				return err
+			}
+
+			count++
+			return nil
+		})
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	var pi blobstor.IteratePrm
--- a/pkg/local_object_storage/shard/evacuate.go
+++ b/pkg/local_object_storage/shard/evacuate.go
@@ -14,8 +14,9 @@ var dumpMagic = []byte("NEOF")
 
 // EvacuatePrm groups the parameters of Evacuate operation.
 type EvacuatePrm struct {
-	path   string
-	stream io.Writer
+	path         string
+	stream       io.Writer
+	ignoreErrors bool
 }
 
 // WithPath is an Evacuate option to set the destination path.
@@ -28,6 +29,13 @@ func (p *EvacuatePrm) WithPath(path string) *EvacuatePrm {
 // It takes priority over `path` option.
 func (p *EvacuatePrm) WithStream(r io.Writer) *EvacuatePrm {
 	p.stream = r
+	return p
+}
+
+// WithIgnoreErrors is an Evacuate option to allow ignore all errors during iteration.
+// This includes invalid blobovniczas as well as corrupted objects.
+func (p *EvacuatePrm) WithIgnoreErrors(ignore bool) *EvacuatePrm {
+	p.ignoreErrors = ignore
 	return p
 }
 
@@ -86,7 +94,7 @@ func (s *Shard) Evacuate(prm *EvacuatePrm) (*EvacuateRes, error) {
 
 			count++
 			return nil
-		}))
+		}).WithIgnoreErrors(prm.ignoreErrors))
 		if err != nil {
 			return nil, err
 		}
@@ -94,6 +102,9 @@ func (s *Shard) Evacuate(prm *EvacuatePrm) (*EvacuateRes, error) {
 
 	var pi blobstor.IteratePrm
 
+	if prm.ignoreErrors {
+		pi.IgnoreErrors()
+	}
 	pi.SetIterationHandler(func(elem blobstor.IterationElement) error {
 		data := elem.ObjectData()
 
--- a/pkg/local_object_storage/shard/evacuate.go
+++ b/pkg/local_object_storage/shard/evacuate.go
@@ -1,0 +1,102 @@
+package shard
+
+import (
+	"encoding/binary"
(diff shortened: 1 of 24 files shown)


pkg/local_object_storage/writecache/mode.go
```diff
@@ -0,0 +1,52 @@
...
+	// 1. Persist objects already in memory on disk.
+	c.persistMemoryCache()
+
+	// 2. Suspend producers to ensure there are channel send operations in fly.
+	// metaCh and directCh can be populated either during Put or in background memory persist thread.
+	// Former possibility is eliminated by taking `modeMtx` mutex and
+	// latter by explicit persist in the previous step.
+	// flushCh is populated by `flush` with `modeMtx` is also taken.
+	// Thus all producers are shutdown and we only need to wait until all channels are empty.
+	for len(c.metaCh) != 0 || len(c.directCh) != 0 || len(c.flushCh) != 0 {
```
carpawell, line 48: did not get the idea, these channels are not buffered, is it possible to get non-zero `len` here?

pkg/local_object_storage/shard/mode.go
```diff
@@ -40,6 +44,10 @@ func (s *Shard) SetMode(m Mode) error {
 	s.m.Lock()
 	defer s.m.Unlock()
 
+	if s.hasWriteCache() {
+		s.writeCache.SetMode(writecache.Mode(m))
```
carpawell, line 48: `write-cache` and `shard` mode enums have to be in sync then, maybe add explicit `switch` there to prevent bugs? or use the same mode for both (if `write-cache` will always has the same modes as `shard`)

pkg/local_object_storage/shard/restore.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/nspcc-dev/neofs-node/pkg/core/object"
+)
+
+// ErrInvalidMagic is returned when dump format is invalid.
+var ErrInvalidMagic = errors.New("invalid magic")
+
+// RestorePrm groups the parameters of Restore operation.
+type RestorePrm struct {
+	path         string
+	ignoreErrors bool
```
carpawell, line 19: should the `Evacuate` operation have the same `ignoreErrors`? i think evacuating "all I can" is even more important than restoring

pkg/local_object_storage/shard/evacuate.go
```diff
@@ -0,0 +1,106 @@
...
+// Returns any error encountered.
+func (s *Shard) Evacuate(prm *EvacuatePrm) (*EvacuateRes, error) {
+	s.m.RLock()
+	defer s.m.RUnlock()
+
+	if s.info.Mode != ModeReadOnly {
+		return nil, ErrMustBeReadOnly
+	}
+
+	f, err := os.Create(prm.path)
```
carpawell, line 49: do we need the dump file to have any specific permissions? @realloc @alexvanin

realloc: `0640` would be enough.
There should be an option to stream the dump, let  us don't forget to implement it later.

pkg/local_object_storage/shard/evacuate.go
```diff
@@ -0,0 +1,106 @@
...
+
+import (
+	"encoding/binary"
+	"errors"
+	"os"
+
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor"
+)
+
+var dumpMagic = []byte("NEOF")
```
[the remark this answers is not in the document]
fyrchik: I believe it is more common https://wikiless.org/wiki/List_of_file_signatures?lang=en . No technical reason, just an arbitrary choice.

pkg/local_object_storage/writecache/mode.go
fyrchik: Yes, they are not, but they _could_ be, so I put this here to avoid possible future bugs. Their capacity isn't reflected in their type.

pkg/local_object_storage/shard/restore.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// WithPath is an Restore option to set the destination path.
```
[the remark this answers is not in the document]
fyrchik: fixed.

pkg/local_object_storage/shard/restore.go
```diff
@@ -0,0 +1,123 @@
...
+//
+// Option is required.
+func (p *RestorePrm) WithPath(path string) *RestorePrm {
+	p.path = path
+	return p
+}
+
+// WithIgnoreErrors is n Restore option which allows to ignore errors encountered during restore.
```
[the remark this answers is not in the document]
fyrchik: fixed.

pkg/local_object_storage/shard/restore.go
```diff
@@ -0,0 +1,123 @@
...
+
+	f, err := os.OpenFile(prm.path, os.O_RDONLY, os.ModeExclusive)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	var m [4]byte
+	_, _ = io.ReadFull(f, m[:])
+	if !bytes.Equal(m[:], []byte{0x4e, 0x45, 0x4f, 0x46}) {
```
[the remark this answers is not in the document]
fyrchik: fixed.

pkg/local_object_storage/shard/evacuate.go
fyrchik: Changed mode, added streams for evacuate/restore.

pkg/local_object_storage/shard/mode.go
fyrchik: I have thought about a test for comparison, but we also can move mode to a separate package so that all components can reuse it. Will update PR.

fyrchik: Fixed.

pkg/local_object_storage/shard/restore.go
fyrchik: Fixed.

alexvanin (approved):
What you think about restoring on the engine level? 

I thought about the case when admin wants to reorganize storage structure from single shard to multiple (say four) shards. With the data restore on the engine level, admin can 
1. dump data into TAR, 
2. reconfigure and start storage node with four shards,
3. restore TAR on the engine level and distribute objects among new shards uniformly.

/cc @carpawell @cthulhu-rider @fyrchik @realloc

realloc:
The presented scenario is possible, but for the simple shard reshape there should be no need to do dump. Shard can be set inactive an some way and the data would be moved or replicated.

Dump/Restore is mainly for fast data evacuation from the malfunctioning device or node. In the storage system there will be no free space to store the dump. All storage devices are used for actual storage and the system partitions are read-only, in most cases. That's why the initial dump/restore design describes the case of streaming the data out. The stream can be processed by CLI and saved to a file and then streamed from the file back, but that's up to the operator to decide.

At the stage we are now, I agree that we need to make it work with files first, but later in reality there would almost be no use-cases for the engine to work with files directly. I suggest to make the most simple dump/restore implementation to try it out and rewrite using streaming via gRPC API.

pkg/local_object_storage/shard/evacuate.go
```diff
@@ -0,0 +1,102 @@
...
+	s.m.RLock()
+	defer s.m.RUnlock()
+
+	if s.info.Mode != ModeReadOnly {
+		return nil, ErrMustBeReadOnly
+	}
+
+	w := prm.stream
+	if w == nil {
+		f, err := os.OpenFile(prm.path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0640)
```
cthulhu-rider: Do we actually need read flag?

pkg/local_object_storage/shard/evacuate.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// Count return amount of object written.
+func (r *EvacuateRes) Count() int {
+	return r.count
+}
+
+var ErrMustBeReadOnly = errors.New("shard must be in read-only mode")
+
+// Evacuate dumps all objects from the shard to a file.
```
cthulhu-rider: `...or a provided target stream.`

pkg/local_object_storage/shard/evacuate.go
```diff
@@ -72,7 +72,23 @@ func (s *Shard) Evacuate(prm *EvacuatePrm) (*EvacuateRes, error) {
 	var count int
 
 	if s.hasWriteCache() {
-		// TODO evacuate objects from write cache
+		err := s.writeCache.Iterate(func(data []byte) error {
```
cthulhu-rider: Maybe it's worth to unify the handler with the one below.

cthulhu-rider (approved):

PR #1091, opened 2022-01-18
Do not cut `data` args in `NNS`

carpawell:
Same as previously reverted #1038. Is actual for `v0.14.0` contract version.


alexvanin:
I will merge it as soon as public networks will be updated to v0.14.0.

alexvanin (approved):

Issue #1083, opened 2022-01-12
Confusing `neofs-cli` config file usage


carpawell:
1. Add some warning words when using flags from config file (in verbose mode or even in general). It may lead to mistakes when a user uses flags that he did not mention in the command line, especially when config is read from the default path and has not been placed with `-c` flag.
2. Do something with a confusing situation when config file says to use `binary-key` and user uses `-w` flag (config file's key would be used since it is parsed firstly in `neofs-cli` and no warning would be printed).

carpawell:
IMO trying to read config file from some _default_ path should be dropped.

alexvanin:
Let's try to use single flag for everything:
```
--wallet [wallet-path]
--wallet [WIF]
--wallet [binary-key]
```

This will fix issue (2). Verbose flag still may print what wallet (or wallet address) has been used.

fyrchik:
Related #610.

carpawell:
is `wallet` a good name for such a flag?

alexvanin:
I think yes, because  NEP-6 compatible JSON Wallet is a preferred way to specify credentials, unless user knows what he is doing with his WIF or binary private key.

carpawell:
How about `1.` of the issue?

alexvanin:
> How about 1. of the issue?

The core of the problem is gone with single credential key. But I don't mind to mention if the default config file is used in verbose mode. Can you do that @carpawell ? If so, reopen the issue.

PR #1093, opened 2022-01-19
[#1083] neofs-cli: use single flag for key and wallet

fyrchik:
Currently have static priority of what key is used irregardless of
whether a flag was provided via CLI or in config. This makes it
impossible to override some of the config settings. While we could try
to check if the key is provided by CLI by binding CLI flag under to
viper under a different name the same problem would occur for config/environment
variables. Fixing all of this with current set of keys is too complicate.
In this commit we revert changes from #610 and use a single flag for all types of keys.

### DIFF
--- a/cmd/neofs-cli/modules/root.go
+++ b/cmd/neofs-cli/modules/root.go
@@ -44,20 +44,10 @@ const (
 	generateKeyDefault   = false
 	generateKeyUsage     = "generate new private key"
 
-	binaryKey          = "binary-key"
-	binaryKeyShorthand = ""
-	binaryKeyDefault   = ""
-	binaryKeyUsage     = "path to the raw private key file"
-
 	walletPath          = "wallet"
 	walletPathShorthand = "w"
 	walletPathDefault   = ""
-	walletPathUsage     = "path to the wallet"
-
-	wif          = "wif"
-	wifShorthand = ""
-	wifDefault   = ""
-	wifUsage     = "WIF or NEP-2"
+	walletPathUsage     = "WIF (NEP-2) string or path to the wallet or binary key"
 
 	address          = "address"
 	addressShorthand = ""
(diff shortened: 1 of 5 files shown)


cmd/neofs-cli/modules/root.go
```diff
@@ -183,29 +173,40 @@ func getKey() (*ecdsa.PrivateKey, error) {
...
 	}
 
-	priv, err := keys.NewPrivateKeyFromWIF(wif)
-	if err != nil {
-		return nil, fmt.Errorf("%w: %v", errInvalidKey, err)
+	if len(keyDesc) == nep2Base58Length {
+		// Check base58-encoding to avoid returning invalid password error when key is in fact incorrect.
+		for _, c := range keyDesc {
+			if !strings.ContainsRune("123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz", c) {
+				return nil, errInvalidKey
```
carpawell, line 203: wouldn't it be better to have a special error for that case? "provided key is incorrect" may confuse I think especially if I use WIF with `wallet` flag

cmd/neofs-cli/modules/root.go
```diff
@@ -183,29 +173,40 @@ func getKey() (*ecdsa.PrivateKey, error) {
...
+	if err == nil {
+		return getKeyFromWallet(w, viper.GetString(address))
 	}
 
-	priv, err := keys.NewPrivateKeyFromWIF(wif)
-	if err != nil {
-		return nil, fmt.Errorf("%w: %v", errInvalidKey, err)
+	if len(keyDesc) == nep2Base58Length {
+		// Check base58-encoding to avoid returning invalid password error when key is in fact incorrect.
+		for _, c := range keyDesc {
```
carpawell, line 201: wouldn't trying just to decode it be more readable?

cmd/neofs-cli/modules/root.go
fyrchik: I have removed this part. Checking full NEP-2 format seems an overkill, and checking length should be enough for most cases.

carpawell (approved):

alexvanin (approved):

PR #1095, opened 2022-01-19
Feat/Delete `v2` package in eacl service

carpawell:
Valid only after https://github.com/nspcc-dev/neofs-sdk-go/pull/124


cthulhu-rider:
Reworked in https://github.com/nspcc-dev/neofs-node/pull/1095

Issue #1074, opened 2022-01-10
NeoFS CLI refactor proposal


alexvanin:
This is the draft of NeoFS CLI refactor proposal. I believe that refactor will simplify introduction of new sub sections and sub commands into the application, such as `util eacl-builder`. Right now we have flat structure with one package for whole application.

```
└── modules
    ├── accounting.go
    ├── acl.go
    ├── completion.go
    ├── container.go
    ├── control.go
    ├── netmap.go
    ├── object.go
    ├── root.go
    ├── storagegroup.go
    └── util.go
```

Single package is great for sharing helper functions between sections. We share private key obtaining function, flag initialization, etc. However these section files slowly grow larger. `util` section already includes multiple sub sections and sub commands, and this is becoming hard to maintain. 

I want to address this issue with two changes.

## Define sections and sub sections in packages
```
$ neofs-cli util locode generate ...

cmd/neofs-cli/modules/util/locode/
├── generate.go
├── info.go
└── root.go
```

In the example above, `util` section is defined in `modules/util` package and locode sub section is stored in `modules/util/locode` package. `generate` command is defined in corresponding `generate.go` file there.

Every section will contain `root.go` file that includes definition of the section. Files that describe commands define command and it's implementation.

```go
// structure of root.go files
var Cmd = &cobra.Command{
	Use:   "locode",
	Short: "Working with NeoFS UN/LOCODE database",
}

func init() {
	Cmd.AddCommand(locodeInfoCmd, locodeGenerateCmd)
}
```

```go
// structure of command implementation files
var locodeGenerateCmd = &cobra.Command{
	Use:   "generate",
	Short: "generate UN/LOCODE database for NeoFS",
	Run:   locodeGenerate,
}

func init() {
	flags := locodeGenerateCmd.Flags()
	// process flags for the command
}

func locodeGenerate(cmd *cobra.Command, _ []string) {
	...
}
```

## Share code in internal packages
There are plenty of code that shared among these sections. It can be stored in internal package the same way as client, even though it [should be removed](https://github.com/nspcc-dev/neofs-node/pull/950#pullrequestreview-796681663).

I tried such structure
```
cmd/neofs-cli/internal
├── client
│   ├── client.go
│   ├── doc.go
│   └── prm.go
├── errors
│   └── errors.go
├── flags
│   └── flags.go
└── io
    ├── eacl.go
    ├── key.go
    ├── prettyprint.go
    ├── token.go
    └── verbose.go
```

[...]

PR #1097, opened 2022-01-20
neofs-cli: Add 'acl extended create' command

alexvanin:
Follows neofs-cli refactor scheme from #1074

```
$ neofs-cli acl extended create \
  --cid EutHBsdT1YCzHxjCfQHnLPL1vFrkSyLSio4vkphfnEk \
  -r 'allow get obj:Key=Value others'  \
  -r 'deny put others'
{
  "version": {
    "major": 2,
    "minor": 11
  },
  "containerID": {
    "value": "A5BYxvLAy0ksUzsKTRTvd8wPeKvMztUofYShogEc+4E="
  },
  "records": [
    {
      "operation": "GET",
      "action": "ALLOW",
      "filters": [
        {
          "headerType": "OBJECT",
          "matchType": "STRING_EQUAL",
          "key": "Key",
          "value": "Value"
        }
      ],
      "targets": [
        {
          "role": "OTHERS",
          "keys": []
        }
      ]
    },
    {
      "operation": "PUT",
      "action": "DENY",
      "filters": [],
      "targets": [
        {
          "role": "OTHERS",
          "keys": []
        }
      ]
    }
  ]
}
```

```
Rule consist of these blocks: <action> <operation> [<filter1> ...] [<target1> ...]

Action is 'allow' or 'deny'.

Operation is a object service verb: 'get', 'head', 'put', 'search', 'delete', 'getrange', or 'getrangehash'.

Filters consist of <typ>:<key><match><value>
  Typ is 'obj' for object applied filter or 'req' for request applied filter. 
  Key is a valid unicode string corresponding to object or request header key.
  Match is '==' for matching and '!=' for non-matching filter.
  Value is valid unicode string corresponding to object or request header value.

Target is 
  'user' for container owner, 
  'system' for Storage nodes in container and Inner Ring nodes,
  'others' for all other request senders, 
  'pubkey:<key1>,<key2>,...' for exact request sender, where <key> is a hex-encoded 33-byte public key.
```


alexvanin:
@EliChin Please, synchronize your doc paper with this version of eACL constructor command.

EliChin (approved):

fyrchik (approved):

cmd/neofs-cli/modules/acl/extended/create.go
```diff
@@ -0,0 +1,265 @@
...
+  Match is '==' for matching and '!=' for non-matching filter.
+  Value is valid unicode string corresponding to object or request header value.
+
+Target is 
+  'user' for container owner, 
+  'system' for Storage nodes in container and Inner Ring nodes,
+  'others' for all other request senders, 
+  'pubkey:<key1>,<key2>,...' for exact request sender, where <key> is a hex-encoded 33-byte public key.
+
+When both '--rule' and '--in' arguments are uses, '--rule' records will be placed higher in resulting extended ACL table.
```
carpawell, line 42:
```suggestion
When both '--rule' and '--in' arguments are used, '--rule' records will be placed higher in resulting extended ACL table.
```

cmd/neofs-cli/modules/acl/extended/create.go
```diff
@@ -0,0 +1,265 @@
...
+`,
+	Example: `neofs-cli acl extended create --cid EutHBsdT1YCzHxjCfQHnLPL1vFrkSyLSio4vkphfnEk -f rules.txt --out table.json
+neofs-cli acl extended create --cid EutHBsdT1YCzHxjCfQHnLPL1vFrkSyLSio4vkphfnEk -r 'allow get obj:Key=Value others' -r 'deny put others'`,
+	RunE: createEACL,
```
carpawell, line 46: in other commands, we do not use `RunE` to prevent `cli` from printing help message on every error

does such change relate to refactoring of `cli`?

cmd/neofs-cli/modules/acl/extended/create.go
```diff
@@ -0,0 +1,265 @@
...
+
+	return strings.Split(strings.TrimSpace(string(data)), "\n"), nil
+}
+
+// parseTable parses eACL table from the following form:
+// <action> <operation> [<filter1> ...] [<target1> ...]
+//
+// Examples:
+// allow get req:X-Header=123 obj:Attr=value others:0xkey1,key2 system:key3 user:key4
+func parseTable(tb *eacl.Table, args []string) error {
```
carpawell, line 133: maybe add some unit test for such utils?

cmd/neofs-cli/modules/acl/extended/create.go
alexvanin: I decided not to use `exitOnErr` like functions because they are not available without refactor. But you are right, I still can use `Run` and print error message manually.

PR #1106, opened 2022-01-24
Balance precisions

### LINKED ISSUE #1084: Reconsider balance precision decrease to Fixed8 from #122
Let the CLI decided output format. Storage node should return Fixed12 precision value. CLI can produce Fixed8 output for compatibility with NEP-17 assets.

cthulhu-rider:
* closes #1084

### DIFF
--- a/cmd/neofs-cli/modules/accounting.go
+++ b/cmd/neofs-cli/modules/accounting.go
@@ -1,10 +1,11 @@
 package cmd
 
 import (
-	"fmt"
-	"math"
+	"math/big"
 
+	"github.com/nspcc-dev/neo-go/pkg/encoding/fixedn"
 	internalclient "github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/client"
+	"github.com/nspcc-dev/neofs-node/pkg/util/precision"
 	"github.com/nspcc-dev/neofs-sdk-go/accounting"
 	"github.com/nspcc-dev/neofs-sdk-go/owner"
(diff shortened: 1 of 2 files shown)


cmd/neofs-cli/modules/accounting.go
```diff
@@ -97,12 +98,8 @@ func prettyPrintDecimal(cmd *cobra.Command, decimal *accounting.Decimal) {
...
 	} else {
-		// divider = 10^{precision}; v:365, p:2 => 365 / 10^2 = 3.65
-		divider := math.Pow(10, float64(decimal.Precision()))
+		amountF8 := precision.Convert(decimal.Precision(), 8, big.NewInt(decimal.Value()))
 
-		// %0.8f\n for precision 8
-		format := fmt.Sprintf("%%0.%df\n", decimal.Precision())
-
-		cmd.Printf(format, float64(decimal.Value())/divider)
+		cmd.Println(fixedn.ToString(amountF8, 8))
```
fyrchik, line 103: Why don't we just print `decimal.Value()` with original precision using `fixedn.ToString`?

cthulhu-rider: It's the point of the issue, isn't it @alexvanin?

fyrchik: Sorry, I probably misunderstood the issue.

fyrchik (approved):

Issue #596, opened 2021-06-09
Remove ToV2 casts in service/acl


alexvanin:
ToV2 casts are bad and we want to have our node version-independent as much as possible.

Remove `pkg/services/object/acl/eacl/v2` pkg from the repo.

PR #1111, opened 2022-01-26
Refactor/Reorganize `object/acl`

carpawell:
Alternative to #1095 solution of the #596.

Since it was decided to keep logic that works with requests in the repo (and not move it to `SDK`) it is impossible to remove `v2` package in `object/acl` service since something has to work with v2 requests anyway (implement [interface](https://github.com/nspcc-dev/neofs-node/blob/v0.27.4/pkg/services/object/server.go#L36) in object pipeline in other words). So that PR reorganizes the structure of that directory to be similar to other packages in `service/object`: there are `v2` packages in `acl` and `acl/eacl` dirs now. They parse all v2-versioned structs into `neofs-sdk-go` abstractions and pass them to the corresponding `acl`/`eacl` packages. `v2` packages are the only packages that do import `neofs-api-go` library. `eacl` and `acl` provide public functions that only accepts `sdk` structures.

Valid only after https://github.com/nspcc-dev/neofs-sdk-go/pull/128 and rebasing.


pkg/services/object/acl/v2/types.go
```diff
@@ -0,0 +1,29 @@
...
+	eaclSDK "github.com/nspcc-dev/neofs-sdk-go/eacl"
+	"github.com/nspcc-dev/neofs-sdk-go/owner"
+)
+
+// ACLChecker is an interface that must provide
+// ACL related checks.
+type ACLChecker interface {
+	// BasicACLCheck must return true only if request
+	// passes basic ACL validation.
+	BasicACLCheck(RequestInfo) bool
```
cthulhu-rider, line 15: I'd start the names from verb 
```suggestion
	CheckBasicACL(RequestInfo) bool
```

pkg/services/object/acl/v2/types.go
```diff
@@ -0,0 +1,29 @@
...
+	// StickyBitCheck must return true only if sticky bit
+	// is disabled or enabled but request contains correct
+	// owner field.
+	StickyBitCheck(RequestInfo, *owner.ID) bool
+}
+
+// Classifier is an interface that must classify sender
+// via passed meta information, container and its ID.
+type Classifier interface {
+	Classify(MetaWithToken, *cidSDK.ID, *container.Container) (role eaclSDK.Role, isIR bool, ownerKey []byte, err error)
```
cthulhu-rider, line 28: Oh, too many return values, maybe structure them?

cthulhu-rider: Btw do we really need separate `Classifier` component? As I see all `ACLChecker` methods accept `RequestInfo`.

carpawell: made that private and deleted interface

structured return result

cthulhu-rider (approved):

alexvanin (approved):

Issue #1094, opened 2022-01-19
Fix TODOs or link them to issues


fyrchik:
I have walked through some of the TODOs and found out that we have more and more of them with each release.
```
> for i in (git tag); echo -n "$i: "; git checkout $i 2>/dev/null; rg -i todo | wc -l; end
v0.12.0: 57
...
v0.15.0: 68
...
v0.20.0: 82
...
v0.26.0: 89
...
v0.27.4: 91
```

While some of them are reasonable, Many of them can be fixed immediately or have a link to the corresponding issue.

alexvanin:
As @fyrchik suggested, we can try to solve todos from object storage components at least. They are pretty much the same behavior issues so we need to decide what behavior is suitable and implement it.

E.g.:
https://github.com/nspcc-dev/neofs-node/blob/ac82899e850dc5bfe52be73307c72117e5592e0f/pkg/local_object_storage/engine/delete.go#L52

fyrchik:
@alexvanin yes, and the second part would be to link remaining TODOs to tasks, possibly create some.

fyrchik:
Much better now:
```
$ rg -i 'todo|fixme' | wc -l
44
```

PR #1115, opened 2022-01-27
Fix some TODOs

fyrchik:
Related #1094 .

### DIFF
--- a/pkg/local_object_storage/blobovnicza/control.go
+++ b/pkg/local_object_storage/blobovnicza/control.go
@@ -1,7 +1,6 @@
 package blobovnicza
 
 import (
-	"errors"
 	"fmt"
 	"path"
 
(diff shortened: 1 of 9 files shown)


pkg/services/audit/auditor/pdp.go
```diff
@@ -108,7 +108,9 @@ func (c *Context) splitPayload(id *object.ID) []uint64 {
 
 func (c *Context) collectHashes(p *gamePair) {
 	fn := func(n *netmap.Node, rngs []*object.Range, hashWriter func([]byte)) {
-		// TODO: add order randomization
+		rand.Shuffle(len(rngs), func(i, j int) {
+			rngs[i], rngs[j] = rngs[j], rngs[i]
+		})
```
alexvanin, line 113: It seems to me that this change is not quite correct. @cthulhu-rider, can you double check? I think original TODO meant to randomize order of network calls (`c.cnrCom.GetRangeHash()`). But those calls are strictly tied with `hashWriter()` and it makes the logic incorrect.

Hashes of provided `rngs` are appended into a `hh1` and `hh2` lists. 
```go
	p.hh1 = make([][]byte, 0, len(p.rn1))

	fn(p.n1, p.rn1, func(h []byte) {
		p.hh1 = append(p.hh1, h)
	})

	p.hh2 = make([][]byte, 0, len(p.rn2))

	fn(p.n2, p.rn2, func(h []byte) {
		p.hh2 = append(p.hh2, h)
	})
```

The order of hashes in this list is important, because they are matched to each other after homomorphic concatenation. 

```
|         hh1[0]         |   hh1[1]  |   hh1[2]  |
|------------|-----------|-----------|-----------| : object ranges
|   hh2[0]   |   hh2[1]  |         hh2[2]        |
```

pkg/local_object_storage/blobovnicza/control.go
```diff
@@ -49,32 +48,35 @@ func (b *Blobovnicza) Init() error {
...
 
+			buck, err := tx.CreateBucketIfNotExists(key)
 			if err != nil {
-				return false, fmt.Errorf("(%T) could not create bucket for bounds [%d:%d]: %w",
-					b, lower, upper, err)
+				return false, fmt.Errorf("(%T) could not create bucket for bounds %s: %w",
+					b, rangeStr, err)
 			}
+
+			b.log.Debug("bucket already exists, calculating size")
```
alexvanin, line 72: This message will be produced even when bucket was not exist before. I would rather remove that message then.

alexvanin (changes requested):

pkg/services/audit/auditor/pdp.go
fyrchik: Yeah, removed this commmit.

pkg/local_object_storage/blobovnicza/control.go
fyrchik: Fixed.

alexvanin (approved):

PR #1117, opened 2022-01-27
config: replace `use_write_cache` with `writecache.enabled`

fyrchik:
This is the way things are done with `grpc.tls` and in neo-go.


532910 (approved):
lgtm if removing sanity check is OK

cmd/neofs-node/config/engine/shard/writecache/config.go
```diff
@@ -31,6 +31,13 @@ func From(c *config.Config) *Config {
 	return (*Config)(c)
 }
 
+// Enabled returns true if write-cache is enabled and false otherwise.
+//
+// Panics if values is not boolean.
```
cthulhu-rider, line 36:
```suggestion
// Panics if value is not boolean.
```

cthulhu-rider (approved):

carpawell (approved):

Issue #1053, opened 2021-12-24
Use the same endpoint for notifications and RPC requests


alexvanin:
NeoFS nodes have two separate pools of endpoint addresses:
- to listen side chain events,
- to send side chain RPC requests.
Node might listen events from one endpoint but send RPC request to another. There is an issue with that approach.

Sidechain blocks are not distributed uniformly among RPC nodes. There is a case when NeoFS node receives notification from block which is not processed by other RPC node, where NeoFS sends requests.

Example: https://github.com/nspcc-dev/neofs-node/blob/96efe0f2941244c298f43053c52adf040cce1f54/cmd/neofs-node/netmap.go#L144-L157

In this example NeoFS node tries to fetch new network map on `NewEpoch` event. If RPC node did not process block with new epoch yet, then fetch invocation will fail and new network map will not be processed by the node.
```
error   neofs-node/netmap.go:156        could not update node state on new epoch        {"epoch": 2529, "error": "could not perform test invocation (snapshotByEpoch): chain/client: contract execution finished with state FAULT; exception: at instruction 3593 (THROW): unhandled exception: \"incorrect diff\""}
```
---

To avoid such cases I propose to use single endpoint for listening and sending events. In case of errors, switch to another endpoint both in listener and in the client.

1. Merge endpoint pools into single pool
```yaml
morph:
  endpoint:
    - wss://rpc1.morph.fs.neo.org:40341/ws
    - wss://rpc2.morph.fs.neo.org:40341/ws
```

2. Use web socket neo-go client to send requests

[WSClient](https://github.com/nspcc-dev/neo-go/blob/501ca0dedba41bb23de24987e16f1890b47fc914/pkg/rpc/client/wsclient.go#L23) encapsulates ordinary client that we use in morph package.

3. Drop multiclient component from RPC

4. Provide seamless switch from one endpoint to another in runtime

In case of network errors during RPC requests or closed channel in subscriber, switch to another endpoint.

alexvanin:
I propose to do it in two steps.

1. Use single pool of endpoints in config to fetch one web socket address for _one_ client and _one_ listener. In case of failures, do not anything. This kinda reverts https://github.com/nspcc-dev/neofs-node/issues/746

2. Define components that will implement both `Listener` interface and necessary neo-go client functions. This component will use monitor routine to check health of nodes and switch between endpoints in case of failures like https://github.com/nspcc-dev/neofs-node/issues/746.

PR #1122, opened 2022-01-29
neofs-adm: add interactive configurator for storage node

### LINKED ISSUE #1090: Storage node configurator for N3 networks
Storage node configuration for public networks might not be an easy task which requires deposit on main chain, correct LOCODE input, correct announcement address (see https://github.com/nspcc-dev/neofs-node/issues/923), etc.

While hardcore engineers can tune configuration manually by reading [config example](https://github.com/nspcc-dev/neofs-node/blob/master/config/example/node.yaml) and find correct contract address for deposit in [docs](https://testcdn.fs.neo.org/doc/integrations/endpoints/), I believe there should be simple way to configure minimal viable config for public networks.

We can improve UX with storage node configurator: an application, that will check side chain balance for bootstrap and ask simple question to prepare storage node config.

Something like this:

```
$ neofs-storage-configurator --wallet [...] --address [...] ./path-to-config.yaml
> Choose the network [mainnet/testnet]:
> Your account has no side chain GAS, you need to make initial deposit, continue? [y/n]:
> Amount of GAS to deposit [1.0]:
> Publicly announced address:
> Listening port [8080]:
> TLS Certificate (optional): 
> UN-LOCODE attribute in [XX-YYY]: 
> Path to the storage dir (all available storage will be used): 
```

Pool of RPC endpoints for main chain and side chain is hadrcoded inside the application.
Application will validate all inputs (will not accept local addresses as announced addresses, will check UN-LOCODE format).

It can be:
- subcommand of neofs-adm,
- new application,
- new application in new repo.

fyrchik:
Close #1090 .

### DIFF
--- a/cmd/neofs-adm/internal/modules/storagecfg/config.go
+++ b/cmd/neofs-adm/internal/modules/storagecfg/config.go
@@ -1,0 +1,143 @@
+package storagecfg
+
+const configTemplate = `logger:
+  level: info  # logger level: one of "debug", "info" (default), "warn", "error", "dpanic", "panic", "fatal"
+
+node:
+  wallet:
+    path: {{ .Wallet.Path }}  # path to a NEO wallet; ignored if key is presented
+    address: {{ .Wallet.Account }}  # address of a NEO account in the wallet; ignored if key is presented
+    password: {{ .Wallet.Password }}  # password for a NEO account in the wallet; ignored if key is presented
+  addresses:  # list of addresses announced by Storage node in the Network map
+    - {{ .AnnouncedAddress }}
+  attribute_0: UN-LOCODE:{{ .Attribute.Locode }}
+  relay: {{ .Relay }}  # start Storage node in relay mode without bootstrapping into the Network map
+  subnet:
+    exit_zero: false # toggle entrance to zero subnet (overrides corresponding attribute and occurrence in entries)
+    entries: [] # list of IDs of subnets to enter in a text format of NeoFS API protocol (overrides corresponding attributes)
+
+grpc:
+  num: 1  # total number of listener endpoints
+  0:
+    endpoint: {{ .Endpoint }}  # endpoint for gRPC server
+    tls:{{if .TLSCert}}
+      enabled: true  # enable TLS for a gRPC connection (min version is TLS 1.2)
+      certificate: {{ .TLSCert }}  # path to TLS certificate
+      key: {{ .TLSKey }}  # path to TLS key
+    {{- else }}
+      enabled: false # disable TLS for a gRPC connection
+    {{- end}}
+
+control:
+  authorized_keys:  # list of hex-encoded public keys that have rights to use the Control Service
+    {{- range .AuthorizedKeys }}
+    - {{.}}{{end}}
+  grpc:
+    endpoint: {{.ControlEndpoint}}  # endpoint that is listened by the Control Service
+
+morph:
+  dial_timeout: 20s  # timeout for side chain NEO RPC client connection
+  disable_cache: false  # use TTL cache for side chain GET operations
+  rpc_endpoint:  # side chain N3 RPC endpoints
+    {{- range .MorphRPC }}
+    - https://{{.}}{{end}}
+  notification_endpoint:  # side chain N3 RPC notification endpoints
+    {{- range .MorphRPC }}
+    - wss://{{.}}/ws{{end}}
+{{if not .Relay }}
+storage:
+  shard_pool_size: 15  # size of per-shard worker pools used for PUT operations
+  shard_num: 1  # total number of shards
+
+  default: # section with the default shard parameters
+    metabase:
+      perm: 0644  # permissions for metabase files(directories: +x for current user and group)
+
+    blobstor:
+      perm: 0644  # permissions for blobstor files(directories: +x for current user and group)
+      depth: 2  # max depth of object tree storage in FS
+      small_object_size: 102400  # 100KiB, size threshold for "small" objects which are stored in key-value DB, not in FS, bytes
+      compress: true  # turn on/off Zstandard compression (level 3) of stored objects
+      compression_exclude_content_types:
+        - audio/*
+        - video/*
+
+      blobovnicza:
+        size: 1073741824  # approximate size limit of single blobovnicza instance, total size will be: size*width^(depth+1), bytes
+        depth: 1  # max depth of object tree storage in key-value DB
+        width: 4   # max width of object tree storage in key-value DB
+        opened_cache_capacity: 50  # maximum number of opened database files
+
+    gc:
+      remover_batch_size: 200  # number of objects to be removed by the garbage collector
+      remover_sleep_interval: 5m  # frequency of the garbage collector invocation
+
+  shard:
+    0:
+      mode: "read-write"  # mode of the shard, must be one of the: "read-write" (default), "read-only"
+
+      metabase:
+        path: {{ .MetabasePath }}  # path to the metabase
+
+      blobstor:
+        path: {{ .BlobstorPath }}  # path to the blobstor
+{{end}}`
+
+const (
+	neofsMainnetAddress   = "2cafa46838e8b564468ebd868dcafdd99dce6221"
+	balanceMainnetAddress = "dc1ec98d9d0c5f9dfade16144defe08cffc5ca55"
+	neofsTestnetAddress   = "b65d8243ac63983206d17e5221af0653a7266fa1"
+	balanceTestnetAddress = "e0420c216003747626670d1424569c17c79015bf"
+)
+
+var n3config = map[string]struct {
+	MorphRPC        []string
+	RPC             []string
+	NeoFSContract   string
+	BalanceContract string
+}{
+	"testnet": {
+		MorphRPC: []string{
+			"rpc01.morph.testnet.fs.neo.org:51331",
+			"rpc02.morph.testnet.fs.neo.org:51331",
+			"rpc03.morph.testnet.fs.neo.org:51331",
+			"rpc04.morph.testnet.fs.neo.org:51331",
+			"rpc05.morph.testnet.fs.neo.org:51331",
+			"rpc06.morph.testnet.fs.neo.org:51331",
+			"rpc07.morph.testnet.fs.neo.org:51331",
+		},
+		RPC: []string{
+			"rpc01.testnet.n3.nspcc.ru:21331",
+			"rpc02.testnet.n3.nspcc.ru:21331",
+			"rpc03.testnet.n3.nspcc.ru:21331",
+			"rpc04.testnet.n3.nspcc.ru:21331",
+			"rpc05.testnet.n3.nspcc.ru:21331",
+			"rpc06.testnet.n3.nspcc.ru:21331",
+			"rpc07.testnet.n3.nspcc.ru:21331",
+		},
+		NeoFSContract:   neofsTestnetAddress,
+		BalanceContract: balanceTestnetAddress,
+	},
+	"mainnet": {
+		MorphRPC: []string{
+			"rpc1.morph.fs.neo.org:40341",
+			"rpc2.morph.fs.neo.org:40341",
+			"rpc3.morph.fs.neo.org:40341",
+			"rpc4.morph.fs.neo.org:40341",
+			"rpc5.morph.fs.neo.org:40341",
+			"rpc6.morph.fs.neo.org:40341",
+			"rpc7.morph.fs.neo.org:40341",
+		},
+		RPC: []string{
+			"rpc1.n3.nspcc.ru:10331",
+			"rpc2.n3.nspcc.ru:10331",
+			"rpc3.n3.nspcc.ru:10331",
+			"rpc4.n3.nspcc.ru:10331",
+			"rpc5.n3.nspcc.ru:10331",
+			"rpc6.n3.nspcc.ru:10331",
+			"rpc7.n3.nspcc.ru:10331",
+		},
+		NeoFSContract:   neofsMainnetAddress,
+		BalanceContract: balanceMainnetAddress,
+	},
+}
--- a/cmd/neofs-adm/internal/modules/storagecfg/root.go
+++ b/cmd/neofs-adm/internal/modules/storagecfg/root.go
@@ -89,6 +89,9 @@ func storageConfig(cmd *cobra.Command, args []string) {
 		}
 	}
 
+	historyPath := filepath.Join(os.TempDir(), "neofs-adm.history")
+	readline.SetHistoryPath(historyPath)
+
 	var c config
 
 	c.Wallet.Path, _ = cmd.Flags().GetString(walletFlag)
@@ -217,8 +220,11 @@ func getWalletAccount(w *wallet.Wallet, prompt string) string {
 }
 
 func getString(prompt string) string {
-	s, err := input.ReadLine(prompt)
+	s, err := readline.Line(prompt)
 	fatalOnErr(err)
+	if s != "" {
+		_ = readline.AddHistory(s)
+	}
 	return s
 }
 
@@ -258,6 +264,8 @@ func getPath(prompt string) string {
 	if p == "" {
 		return p
 	}
+
+	_ = readline.AddHistory(p)
 
 	abs, err := filepath.Abs(p)
 	if err != nil {
--- a/cmd/neofs-adm/internal/modules/storagecfg/root.go
+++ b/cmd/neofs-adm/internal/modules/storagecfg/root.go
@@ -1,0 +1,405 @@
+package storagecfg
+
+import (
+	"bytes"
+	"context"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"math/big"
+	"math/rand"
+	"net"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"text/template"
+	"time"
+
+	"github.com/chzyer/readline"
+	"github.com/nspcc-dev/neo-go/cli/flags"
+	"github.com/nspcc-dev/neo-go/cli/input"
+	"github.com/nspcc-dev/neo-go/pkg/core/native/nativenames"
+	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
+	"github.com/nspcc-dev/neo-go/pkg/encoding/address"
+	"github.com/nspcc-dev/neo-go/pkg/encoding/fixedn"
+	"github.com/nspcc-dev/neo-go/pkg/rpc/client"
+	"github.com/nspcc-dev/neo-go/pkg/smartcontract"
+	"github.com/nspcc-dev/neo-go/pkg/smartcontract/trigger"
+	"github.com/nspcc-dev/neo-go/pkg/util"
+	"github.com/nspcc-dev/neo-go/pkg/vm"
+	"github.com/nspcc-dev/neo-go/pkg/wallet"
+	"github.com/spf13/cobra"
+)
+
+const (
+	walletFlag  = "wallet"
+	accountFlag = "account"
+)
+
+const (
+	defaultControlEndpoint = "127.0.0.1:8090"
+	defaultDataEndpoint    = "127.0.0.1"
+)
+
+// RootCmd is a root command of config section.
+var RootCmd = &cobra.Command{
+	Use:   "storage-config [-w wallet] [-a acccount] [<path-to-config>]",
+	Short: "Section for storage node configuration commands.",
+	Run:   storageConfig,
+}
+
+func init() {
+	fs := RootCmd.Flags()
+
+	fs.StringP(walletFlag, "w", "", "path to wallet")
+	fs.StringP(accountFlag, "a", "", "wallet account")
+}
+
+type config struct {
+	AnnouncedAddress string
+	AuthorizedKeys   []string
+	ControlEndpoint  string
+	Endpoint         string
+	TLSCert          string
+	TLSKey           string
+	MorphRPC         []string
+	Attribute        struct {
+		Locode string
+	}
+	Wallet struct {
+		Path     string
+		Account  string
+		Password string
+	}
+	Relay        bool
+	BlobstorPath string
+	MetabasePath string
+}
+
+func storageConfig(cmd *cobra.Command, args []string) {
+	var outPath string
+	if len(args) != 0 {
+		outPath = args[0]
+	} else {
+		outPath = getPath("File to write config at [./config.yml]: ")
+		if outPath == "" {
+			outPath = "./config.yml"
+		}
+	}
+
+	var c config
+
+	c.Wallet.Path, _ = cmd.Flags().GetString(walletFlag)
+	if c.Wallet.Path == "" {
+		c.Wallet.Path = getPath("Path to the storage node wallet: ")
+	}
+
+	w, err := wallet.NewWalletFromFile(c.Wallet.Path)
+	fatalOnErr(err)
+
+	c.Wallet.Account, _ = cmd.Flags().GetString(accountFlag)
+	if c.Wallet.Account == "" {
+		addr := address.Uint160ToString(w.GetChangeAddress())
+		c.Wallet.Account = getWalletAccount(w, fmt.Sprintf("Wallet account [%s]: ", addr))
+		if c.Wallet.Account == "" {
+			c.Wallet.Account = addr
+		}
+	}
+
+	accH, err := flags.ParseAddress(c.Wallet.Account)
+	fatalOnErr(err)
+
+	acc := w.GetAccount(accH)
+	if acc == nil {
+		fatalOnErr(errors.New("can't find account in wallet"))
+	}
+
+	c.Wallet.Password, err = input.ReadPassword(fmt.Sprintf("Account password for %s: ", c.Wallet.Account))
+	fatalOnErr(err)
+
+	err = acc.Decrypt(c.Wallet.Password, keys.NEP2ScryptParams())
+	fatalOnErr(err)
+
+	c.AuthorizedKeys = append(c.AuthorizedKeys, hex.EncodeToString(acc.PrivateKey().PublicKey().Bytes()))
+
+	var network string
+	for {
+		network = getString("Choose network [mainnet]/testnet: ")
+		switch network {
+		case "":
+			network = "mainnet"
+		case "testnet", "mainnet":
+		default:
+			cmd.Println(`Network must be either "mainnet" or "testnet"`)
+			continue
+		}
+		break
+	}
+
+	c.MorphRPC = n3config[network].MorphRPC
+
+	depositGas(cmd, acc, network)
+
+	c.Attribute.Locode = getString("UN-LOCODE attribute in [XX YYY] format: ")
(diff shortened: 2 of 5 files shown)


cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -0,0 +1,304 @@
...
+			continue
+		}
+		if err = c.Init(); err != nil {
+			continue
+		}
+		return c
+	}
+
+	fatalOnErr(fmt.Errorf("can't create N3 client: %w", err))
+	panic("unreachable")
```
cthulhu-rider, line 303: Really needed?

cmd/neofs-adm/internal/modules/storagecfg/config.go
```diff
@@ -0,0 +1,148 @@
...
+
+const (
+	neofsMainnetAddress   = "2cafa46838e8b564468ebd868dcafdd99dce6221"
+	balanceMainnetAddress = "dc1ec98d9d0c5f9dfade16144defe08cffc5ca55"
+	neofsTestnetAddress   = "b65d8243ac63983206d17e5221af0653a7266fa1"
+	balanceTestnetAddress = "e0420c216003747626670d1424569c17c79015bf"
+)
+
+var n3config = map[string]struct {
+	MorphRPC        []string
```
cthulhu-rider, line 99: Visibility of the fields varies for a reason?

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -0,0 +1,304 @@
...
+		switch strings.ToLower(s) {
+		case "y", "yes":
+			return true
+		case "n", "no":
+			return false
+		default:
+			if len(s) == 0 {
+				return def
+			}
+			continue
```
cthulhu-rider, line 212: Maybe it's worth to print what are we waiting for. Otherwise `continue` is redundant

cthulhu-rider (approved):

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -0,0 +1,304 @@
...
+var RootCmd = &cobra.Command{
+	Use:   "storage-config",
+	Short: "Section for storage node configuration commands.",
+	Run:   storageConfig,
+}
+
+func init() {
+	fs := RootCmd.Flags()
+
+	fs.String("rpc-endpoint", "", "N3 endpoint")
```
alexvanin, line 55: Unused. We can keep it, though, but it should be optional.

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -0,0 +1,304 @@
...
+		Account  string
+		Password string
+	}
+	Relay        bool
+	BlobstorPath string
+	MetabasePath string
+}
+
+func storageConfig(cmd *cobra.Command, args []string) {
+	if len(args) == 0 {
```
alexvanin, line 83: It is not clear that we expect config file path there. I guess arguments like `--out` will be better, or at least mention config file path in usage examples.

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -0,0 +1,304 @@
...
+		fatalOnErr(cmd.Usage())
+		return
+	}
+
+	var c config
+
+	c.Wallet.Path = viper.GetString(walletFlag)
```
alexvanin, line 90: Cobra flags were not bind with viper, thus app asks for wallet path anyway:
```go
$ ./neofs-adm storage-config -w ../keys/storage_wallet.json config.yaml
Path to the storage node wallet:  
```

See binding examples in neofs-cli.

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -0,0 +1,304 @@
...
+
+	w, err := wallet.NewWalletFromFile(c.Wallet.Path)
+	fatalOnErr(err)
+
+	c.Wallet.Account = viper.GetString(accountFlag)
+	if c.Wallet.Account == "" {
+		c.Wallet.Account = getString("Wallet account: ")
+	}
+
+	accH, err := flags.ParseAddress(c.Wallet.Account)
```
alexvanin, line 103: Maybe use default exchange wallet if wallet address is not set? With some warning output, maybe. Now it breaks on empty input.

```
$  ./neofs-adm storage-config config.yaml
Path to the storage node wallet: ../keys/storage_wallet.json
Wallet account: 
Error: zero length string
```

cmd/neofs-adm/internal/modules/storagecfg/config.go
```diff
@@ -0,0 +1,148 @@
...
+}{
+	"testnet": {
+		MorphRPC: []string{
+			"rpc01.morph.testnet.fs.neo.org:51331",
+			"rpc02.morph.testnet.fs.neo.org:51331",
+			"rpc03.morph.testnet.fs.neo.org:51331",
+			"rpc04.morph.testnet.fs.neo.org:51331",
+			"rpc05.morph.testnet.fs.neo.org:51331",
+			"rpc06.morph.testnet.fs.neo.org:51331",
+			"rpc07.morph.testnet.fs.neo.org:51331",
```
alexvanin, line 112: All these RPC addresses miss `https` scheme and neo-go client library fails at init.

```
failed to get network magic: Post "rpc07.morph.testnet.fs.neo.org:51331": unsupported protocol scheme "rpc07.morph.testnet.fs.neo.org"
```

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -0,0 +1,304 @@
...
+	network := getString("Choose network [mainnet]/testnet: ")
+	switch network {
+	case "":
+		network = "mainnet"
+	case "testnet", "mainnet":
+	default:
+		fatalOnErr(errors.New("invalid network"))
+	}
+
+	c.MorphRPC = n3config[network].MorphRPC
```
alexvanin, line 128: `c.MorphRPC` seems unused besides this assignment?

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -0,0 +1,304 @@
...
+
+	depositGas(acc, network)
+
+	c.Attribute.Locode = getString("UN-LOCODE attribute in [XX YYY] format: ")
+	c.AnnouncedAddress = getString("Publicly announced address: ")
```
alexvanin, line 133: Can we make some sort of sanity checks for announnced address? For instance it is not `0.0.0.0/32` at least. We can add more checks later.

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+
+	mainClient := initClient(n3config[network].RPC)
+	neofsHash, _ := util.Uint160DecodeStringLE(n3config[network].neofsContract)
+
+	gasHash, err := mainClient.GetNativeContractHash(nativenames.Gas)
+	fatalOnErr(err)
+
+	w := io.NewBufBinWriter()
+	emit.AppCall(w.BinWriter, gasHash, "transfer", callflag.All,
```
alexvanin, line 276: You definitely don't like wrapper function in neo-go client for NEP-17 trransfer, do you? :smiley:

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -0,0 +1,304 @@
...
+
+	_, err = mainClient.SignAndPushInvocationTx(w.Bytes(), acc, -1, 0, nil)
+	fatalOnErr(err)
+}
+
+func initClient(rpc []string) *client.Client {
+	var c *client.Client
+	var err error
+
+	for _, endpoint := range rpc {
```
alexvanin, line 288: Shuffle those RPC before iterating to distributed workload among them.

alexvanin (changes requested):

cmd/neofs-adm/internal/modules/storagecfg/root.go
alexvanin: One more thing: maybe it is worth to wait for transaction to appear on chain before going to the next step. What do you think?

cmd/neofs-adm/internal/modules/storagecfg/root.go
alexvanin: Nah, don't keep it. We need at least two endpoints for main and side chain then, too complicated.

cmd/neofs-adm/internal/modules/storagecfg/config.go
fyrchik: Fixed.

cmd/neofs-adm/internal/modules/storagecfg/root.go
fyrchik: We can return `nil`, though `panic` better communicates our intention.

cmd/neofs-adm/internal/modules/storagecfg/root.go
fyrchik: It is printed in `readline.Line` above, so the workflow is like this:
```
Use node as a relay? yes/[no]: kek
Use node as a relay? yes/[no]: please
Use node as a relay? yes/[no]: yes
...
```

cmd/neofs-adm/internal/modules/storagecfg/root.go
fyrchik: Fixed.

cmd/neofs-adm/internal/modules/storagecfg/root.go
alexvanin: > We can return `nil`, though `panic` better communicates our intention.

Yeah, but it looks like you fighting go compiler with this line. Maybe just this, then?

```diff
- 	fatalOnErr(fmt.Errorf("can't create N3 client: %w", err))
-	panic("unreachable")
+	panic(fmt.Errorf("can't create N3 client: %w", err))
```

cmd/neofs-adm/internal/modules/storagecfg/root.go
fyrchik: Fixed.

cmd/neofs-adm/internal/modules/storagecfg/config.go
fyrchik: This is done to reuse the same data for `https://` and `ws://` in config template. Fixed client initialization.

cmd/neofs-adm/internal/modules/storagecfg/root.go
fyrchik: It is used in a template generation.

cmd/neofs-adm/internal/modules/storagecfg/root.go
fyrchik: It is an interactive CLI application, what do we gain by shuffling from the user POV?

fyrchik: Ok, distributing load on our RPC in case this utility has a lot of active users.

fyrchik: Fixed.

cmd/neofs-adm/internal/modules/storagecfg/config.go
alexvanin: Got it. I hope after https://github.com/nspcc-dev/neofs-node/issues/1053 we will have only `ws://` endpoints. Right now it is not reliable to work with two separate lists for incoming events and outcoming invocations.

cmd/neofs-adm/internal/modules/storagecfg/root.go
fyrchik: Fixed help message. I think it is easier to use `storage-config out.json` withoult a flag in the most simple case. I think we can even ask for a file if it wasn't provided.

config/testnet/config.yml
```diff
@@ -21,6 +21,7 @@ morph:
   dial_timeout: 20s
 
 contracts:
+  neofs: b65d8243ac63983206d17e5221af0653a7266fa1
```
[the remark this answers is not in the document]
fyrchik: Ok, dropped this.

cmd/neofs-adm/internal/modules/storagecfg/root.go
fyrchik: Fixed, fixed.

cmd/neofs-adm/internal/modules/storagecfg/root.go
fyrchik: Fixed.

cmd/neofs-adm/internal/modules/storagecfg/root.go
fyrchik: Removed `continue`.

cmd/neofs-adm/internal/modules/storagecfg/root.go
fyrchik: It would be strange to have a stack trace for this error, but not the others.

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -0,0 +1,360 @@
...
+
+	c.AuthorizedKeys = append(c.AuthorizedKeys, hex.EncodeToString(acc.PrivateKey().PublicKey().Bytes()))
+
+	network := getString("Choose network [mainnet]/testnet: ")
+	switch network {
+	case "":
+		network = "mainnet"
+	case "testnet", "mainnet":
+	default:
+		fatalOnErr(errors.New("invalid network"))
```
carpawell, line 131: maybe repeat the question here instead of exit?

typo in network-word may lead to losing all the previous progress but is not critical

carpawell (approved):

cmd/neofs-adm/internal/modules/storagecfg/root.go
alexvanin: Not working as expected yet:

```
$ ./neofs-adm storage-config
File to write config at [./config.yml]: ./testnet.yaml
Path to the storage node wallet: ../keys/storage_wallet.json
Wallet account [NRVRcypmJyXL985fF9CDx1yiJ8hwGNKsge]: 
Error: zero length string
```

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -0,0 +1,360 @@
...
+		}
+	}
+}
+
+func initClient(rpc []string) *client.Client {
+	var c *client.Client
+	var err error
+
+	shuffled := make([]string, len(rpc))
+	copy(rpc, shuffled)
```
alexvanin, line 341:
```suggestion
	copy(shuffled, rpc)
```

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -0,0 +1,360 @@
...
+	var balance *big.Int
+	if len(res.Stack) != 0 {
+		balance, _ = res.Stack[0].TryInteger()
+	}
+
+	if balance == nil {
+		fatalOnErr(errors.New("invalid response from balance contract"))
+	}
+
+	ok := getConfirmation(false, fmt.Sprintf("Current balance is %s, make a deposit? y/[n]: ",
```
alexvanin, line 284:
```suggestion
	ok := getConfirmation(false, fmt.Sprintf("Current NeoFS balance is %s, make a deposit? y/[n]: ",
```

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -0,0 +1,360 @@
...
+		fixedn.ToString(balance, 8)))
```
alexvanin, line 285:
```suggestion
		fixedn.ToString(balance, 12)))
```

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -0,0 +1,360 @@
...
+	for {
+		c.AnnouncedAddress = getString("Publicly announced address: ")
+		addr, port, err := net.SplitHostPort(c.AnnouncedAddress)
+		if err != nil {
+			cmd.Println("Address must have form A.B.C.D:PORT")
+			continue
+		}
+
+		ip := net.ParseIP(addr)
+		if !ip.IsGlobalUnicast() || !ip.IsPrivate() {
```
alexvanin, line 148: I guess it should be opposite, announce address most **not** be global unicast or private address.
```suggestion
		if ip.IsGlobalUnicast() || ip.IsPrivate() {
```

Otherwise:
```
Publicly announced address: ssh.neofs.alexvan.in:5051
IP must be either global unicast or private.
Publicly announced address: 199.247.0.222:5051
IP must be either global unicast or private.
```

Also it seems that go1.16 does not support `IsPrivate`.

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -0,0 +1,360 @@
...
+			cmd.Println("Port must be an integer.")
+			continue
+		}
+
+		break
+	}
+
+	c.Endpoint = getString(fmt.Sprintf("Listening address [%s]: ", defaultDataEndpoint))
+	if c.Endpoint == "" {
+		c.Endpoint = defaultDataEndpoint
```
alexvanin, line 164: I suggest to build default data endpoint with the port from public address.

```
Publicly announced address: hostname.com:5051
Listening address [127.0.0.1:5051]: 
```

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -0,0 +1,360 @@
...
+	defer timer.Stop()
+
+	at := trigger.Application
+
+loop:
+	for {
+		select {
+		case <-tick.C:
+			_, err := mainClient.GetApplicationLog(txHash, &at)
+			if err == nil {
```
alexvanin, line 322: Shouldn't it be `err != nil` there to keep waiting and `else { break loop; }`?

cmd/neofs-adm/internal/modules/storagecfg/config.go
```diff
@@ -0,0 +1,148 @@
...
+
+node:
+  wallet:
+    path: {{ .Wallet.Path }}  # path to a NEO wallet; ignored if key is presented
+    address: {{ .Wallet.Account }}  # address of a NEO account in the wallet; ignored if key is presented
+    password: {{ .Wallet.Password }}  # password for a NEO account in the wallet; ignored if key is presented
+  addresses:  # list of addresses announced by Storage node in the Network map
+    - {{ .AnnouncedAddress }}
+  attribute_0: UN-LOCODE:{{ .Attribute.Locode }}
+  attribute_1: Price:{{ .Attribute.Price }}
```
alexvanin, line 14: neofs-adm didn't ask for `Price` attribute in the dialog. Maybe it is better to avoid price for now, because it may mislead users on guarantee profit, however this `Price` attribute only viable for storage group payments, which are quite rare. 

Or specify default price of `100000` as we do in testnet config.

cmd/neofs-adm/internal/modules/storagecfg/root.go
alexvanin: Oh, I thought `fatalOnErr` produces panic, but it just prints to stderr and calls `os.Exit(1)`.

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -0,0 +1,360 @@
...
+		}
+	}
+	if pos != 0 {
+		return newLine, pos - len([]rune(dir))
+	}
+	return newLine, 0
+}
+
+func getPath(prompt string) string {
+	readline.SetAutoComplete(filenameCompleter{})
```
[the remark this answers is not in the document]
fyrchik: This autocomplete shouldn't influence other commands. Added a defer to reset it.

cmd/neofs-adm/internal/modules/storagecfg/root.go
fyrchik: Removed `IsPrivate`.

cmd/neofs-adm/internal/modules/storagecfg/root.go
fyrchik: Fixed

cmd/neofs-adm/internal/modules/storagecfg/root.go
fyrchik: Yes, fixed logic.

cmd/neofs-adm/internal/modules/storagecfg/config.go
fyrchik: Ok, removed price.

cmd/neofs-adm/internal/modules/storagecfg/root.go
fyrchik: Yes, asking again is good, fixed.

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -0,0 +1,386 @@
...
+	fatalOnErr(err)
+
+	err = acc.Decrypt(c.Wallet.Password, keys.NEP2ScryptParams())
+	fatalOnErr(err)
+
+	c.AuthorizedKeys = append(c.AuthorizedKeys, hex.EncodeToString(acc.PrivateKey().PublicKey().Bytes()))
+
+	var network string
+	for {
+		network := getString("Choose network [mainnet]/testnet: ")
```
carpawell, line 129:
```suggestion
		network = getString("Choose network [mainnet]/testnet: ")
```

carpawell (changes requested):

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -0,0 +1,386 @@
...
+
+	at := trigger.Application
+
+loop:
+	for {
+		select {
+		case <-tick.C:
+			_, err := mainClient.GetApplicationLog(txHash, &at)
+			if err == nil {
+				break loop
```
alexvanin, line 349: Print `\n` before break.

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -0,0 +1,386 @@
...
+			newLine = append(newLine, tail)
+		}
+	}
+	if pos != 0 {
+		return newLine, pos - len([]rune(dir))
+	}
+	return newLine, 0
+}
+
+func getPath(prompt string) string {
```
alexvanin, line 243: Can we provide absolute paths in config? I think it will be more consistent, because storage node can be started from different working dir and relative paths will be invalid.

cmd/neofs-adm/internal/modules/storagecfg/root.go
alexvanin: I think it is a bit broken now
```
Publicly announced address: 199.247.0.222:5051
Listening address [[127.0.0.1:8080]:5051]:
```

cmd/neofs-adm/internal/modules/storagecfg/config.go
```diff
@@ -0,0 +1,147 @@
...
+control:
+  authorized_keys:  # list of hex-encoded public keys that have rights to use the Control Service
+    {{- range .AuthorizedKeys }}
+    - {{.}}{{end}}
+  grpc:
+    endpoint: {{.ControlEndpoint}}  # endpoint that is listened by the Control Service
+
+morph:
+  dial_timeout: 20s  # timeout for side chain NEO RPC client connection
+  disable_cache: true  # do not use TTL cache for side chain GET operations
```
alexvanin, line 40: This one is `false` by default, I think better keep it this way.

cmd/neofs-adm/internal/modules/storagecfg/config.go
```diff
@@ -0,0 +1,147 @@
...
+      perm: 0644  # permissions for blobstor files(directories: +x for current user and group)
+      depth: 2  # max depth of object tree storage in FS
+      small_object_size: 102400  # 100KiB, size threshold for "small" objects which are stored in key-value DB, not in FS, bytes
+      compress: true  # turn on/off Zstandard compression (level 3) of stored objects
+      compression_exclude_content_types:
+        - audio/*
+        - video/*
+
+      blobovnicza:
+        size: 4194304  # approximate size limit of single blobovnicza instance, total size will be: size*width^(depth+1), bytes
```
alexvanin, line 66: 4 MiB blobovniczas are small. Default is 1<<30 which is 1 GiB.

cmd/neofs-adm/internal/modules/storagecfg/config.go
```diff
@@ -0,0 +1,147 @@
...
+
+      metabase:
+        path: {{ .MetabasePath }}  # path to the metabase
+
+      blobstor:
+        path: {{ .BlobstorPath }}  # path to the blobstor
+
+      gc:
+        remover_batch_size: 150  # number of objects to be removed by the garbage collector
+        remover_sleep_interval: 2m  # frequency of the garbage collector invocation
```
alexvanin, line 87: `gc` section is in defaults, so we can omit that.

cmd/neofs-adm/internal/modules/storagecfg/root.go
fyrchik: Fixed.

cmd/neofs-adm/internal/modules/storagecfg/root.go
fyrchik: fixed.

cmd/neofs-adm/internal/modules/storagecfg/config.go
fyrchik: Fixed.

cmd/neofs-adm/internal/modules/storagecfg/config.go
fyrchik: Fixed.

cmd/neofs-adm/internal/modules/storagecfg/config.go
fyrchik: Fixed.

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -0,0 +1,406 @@
...
+	var addr, port string
+	for {
+		c.AnnouncedAddress = getString("Publicly announced address: ")
+		addr, port, err = net.SplitHostPort(c.AnnouncedAddress)
+		if err != nil {
+			cmd.Println("Address must have form A.B.C.D:PORT")
+			continue
+		}
+
+		ip, err := net.ResolveIPAddr("ip", addr)
```
carpawell, line 158: err check?

fyrchik: Fixed.

alexvanin (approved):

carpawell (approved):

PR #1123, opened 2022-01-29
.golangci.yml: add `unused` linter

fyrchik:
Also fix critical bug regarding subnet removal.


pkg/innerring/subnet.go
```diff
@@ -263,7 +263,7 @@ func (s *Server) handleSubnetCreation(e event.Event) {
 // catchSubnetRemoval catches event of subnet removal from listener and queues the processing.
 func (s *Server) catchSubnetRemoval(e event.Event) {
 	err := s.subnetHandler.workerPool.Submit(func() {
-		s.handleSubnetCreation(e)
+		s.handleSubnetRemoval(e)
```
alexvanin, line 266: ouch

.golangci.yml
```diff
@@ -51,4 +52,4 @@ issues:
   exclude-rules:
     - path: policy/grammar\.go # ignore structtag issues there
       linters:
-        - govet
\ No newline at end of file
+        - govet
```
alexvanin, line 55: By the way, we can drop this exclusion rule. Code of the placement policy parser has been moved to SDK a while ago.

alexvanin (approved):

alexvanin:
Also it would be nice to have `[#1123]` in all commit headers so all these changes can be visually grouped in commit history (cause we don't use merge commits).

.golangci.yml
fyrchik: Fixed.

PR #1127, opened 2022-01-31
Set shard mode to read-only if there are too many errors

### LINKED ISSUE #1118: Make shard read-only on errors
As described in https://github.com/nspcc-dev/neofs-node/issues/1094#issuecomment-1018259445, there are many places where unexpected error is encountered during shard operation. We can make shard read-only in this situation, carefully checking that this is indeed a reasonable action for all possible errors returned.

Related #1115 .

fyrchik:
Close #1118 .

Error counter can be cleared with an additional field via `SetShardMode` and is reported in `ListShards` information.


config/example/node.yaml
```diff
@@ -99,6 +99,7 @@ storage:
   # note: shard configuration can be omitted for relay node (see `node.relay`)
   shard_pool_size: 15 # size of per-shard worker pools used for PUT operations
   shard_num: 2  # total number of shards
+  shard_ro_error_threshold: 100
```
cthulhu-rider, line 102: Name is quite telling, but let's still add a verbal explanation (in particular, that with a zero/NaN value, the restriction is not set)

fyrchik: Fixed.

carpawell (approved):
the first commit message body contains the typo in the `during` word

also, small style notes: we try to write commit message in the following form:
`[#xxxx] zone: Capitalized_word`

pkg/local_object_storage/engine/shards.go
```diff
@@ -91,23 +93,23 @@ func (e *StorageEngine) unsortedShards() []hashedShard {
 	shards := make([]hashedShard, 0, len(e.shards))
 
 	for _, sh := range e.shards {
-		shards = append(shards, hashedShard{sh})
+		shards = append(shards, hashedShard{errorCount: sh.errorCount, Shard: sh.Shard})
```
carpawell, line 96: linter

```suggestion
		shards = append(shards, hashedShard(sh))
```

pkg/local_object_storage/engine/engine.go
```diff
@@ -26,12 +27,49 @@ type StorageEngine struct {
...
+	*shard.Shard
+}
+
+// reportShardError checks that amount of errors doesn't exceed configured threshold.
+// If it does, shard is set to read-only mode.
+func (e *StorageEngine) reportShardError(
+	sh hashedShard,
+	msg string,
+	err error,
+	fields ...zap.Field) {
```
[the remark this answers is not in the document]
fyrchik: The line would be too long.

pkg/local_object_storage/engine/shards.go
fyrchik: Fixed.

pkg/local_object_storage/engine/engine.go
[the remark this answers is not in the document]
carpawell: just to know: what length is suitable for you? i use 120 char threshold that is why this comment appeared

cmd/neofs-node/config/engine/config.go
```diff
@@ -58,3 +58,10 @@ func ShardPoolSize(c *config.Config) uint32 {
 
 	return ShardPoolSizeDefault
 }
+
+// ShardErrorThreshold returns value of "shard_ro_error_threshold" config parameter from "storage" section.
+//
+// Returns 0 if the value is missing.
```
alexvanin, line 64: I think it is not good to be fault intolerant by default, but I don't have any good default value for that, 10, 100, who knows...

pkg/local_object_storage/engine/engine.go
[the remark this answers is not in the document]
fyrchik: Lately I try to wrap lines at 100.

alexvanin (approved):
LGTM. While global error counter is okay, I also thought more about sampler.

Sampler resets error counter every `t` seconds and stores up to `n` recent samples. If sum of errors in `n` samples > threshold: enable read-only mode.

cmd/neofs-node/config/engine/config.go
fyrchik: I think this non-zero default is good, but it should be provided in config, shard should not define any policy by itself IMO.

alexvanin: `neofs-node/config` package was designed for default app configuration, so it fits there. But still don't know what would be a good value for that.

PR #1129, opened 2022-02-01
Remove intermediate wrapper from morph clients

### LINKED ISSUE #625: Reduce amount of wrappers in `pkg/morph/client`
Currently we have hierarchy of clients talking to our contracts:
1. Raw neo-go client.
2. Static client talking to a specific contract (see also nspcc-dev/neo-go#1967)
3. Low-level wrapper in `pkg/morph/client/balance`
4. High-level wrapper in `pkg/morph/client/balance/wrapper`.

Do we need to separate 3 and 4? I suggest removing one of the levels as it feels artificial and unnecessary verbose.
With additional struct it is also easier to forget to add argument on a higher level. With direct invocation compiler ensures completeness for us.

As @cthulhu-rider noticed, we can have `func Method(args MethogArgs)` in contract and use the struct imported from `neofs-contract` repo directly (this will need client support from neo-go though).

I am OK with any variant as long as it reduces the amount of code needed to write and maintain while increasing readability.

fyrchik:
Depends on #1125 .
Close #625 .

So after this it is hard to argue for having a tower of wrappers, as the interface is the same, but we have 2k LOC less.
It is also faster, because there are no intermediate conversions (including array allocations just to convert element types).


cthulhu-rider (approved):

pkg/innerring/processors/container/process_eacl.go
```diff
@@ -42,12 +42,11 @@ func (cp *Processor) checkSetEACL(e container.SetEACL) error {
 	binTable := e.Table()
 	tableHash := sha256.Sum256(binTable)
 
+	// verify the identity of the container owner
```
cthulhu-rider, line 45: Previous position was correct, `key.Verify` just check the signature as said above.

fyrchik: Fixed

pkg/morph/client/netmap/new_epoch.go
```diff
@@ -6,27 +6,12 @@ import (
...
-func (c *Client) NewEpoch(args NewEpochArgs) error {
+// NewEpoch updates NeoFS epoch number through
+// Netmap contract call.
+func (c *Client) NewEpoch(epoch uint64) error {
 	prm := client.InvokePrm{}
-
 	prm.SetMethod(newEpochMethod)
-	prm.SetArgs(args.number)
-	prm.InvokePrmOptional = args.InvokePrmOptional
+	prm.SetArgs(epoch)
```
alexvanin, line 14: You should either convert it into `int64` or update code in `pkg/morph/client/client.go#384`. 

```diff
	case int64: // TODO: add other numerical types
		result.Type = sc.IntegerType
+	case uint64:
+		result.Type = sc.IntegerType
+		result.Value = int64(v)
```

I didn't check all the code, maybe there is more cases when we pass `uint64` into args in this PR, so maybe change above is more preferred.

alexvanin (changes requested):

pkg/morph/client/netmap/new_epoch.go
fyrchik: The problem is that this is `smartcontract.Parameter` which has a specific logic when marshaling in JSON https://github.com/nspcc-dev/neo-go/blob/master/pkg/smartcontract/parameter.go#L67 . Similar thing happens during unmarshaling on the server side https://github.com/nspcc-dev/neo-go/blob/01d15ff473f9984023e9c1e04be107981acec80e/pkg/rpc/request/txBuilder.go#L65 .

fyrchik: nspcc-dev/neo-go#2350

pkg/morph/client/neofs/cheque.go
```diff
@@ -6,7 +6,7 @@ import (
 	"github.com/nspcc-dev/neofs-node/pkg/morph/client"
 )
 
-// ChequePrm groups the arguments of Cheque operation.
+// ChequePrm groups parameters of AlphabetUpdate operation.
```
carpawell, line 9: misclick?

pkg/morph/client/neofs/bind.go
```diff
@@ -58,11 +49,15 @@ func (x *Client) BindKeys(args BindKeysArgs) error {
 	return nil
 }
 
+// UnbindKeysPrm groups parameters of BindKeys operation.
```
carpawell, line 52:
```suggestion
// UnbindKeysPrm groups parameters of UnbindKeys operation.
```

pkg/morph/client/netmap/config.go
```diff
@@ -82,42 +175,145 @@ func (s *SetConfigPrm) SetValue(value interface{}) {
...
-	rs []stackitem.Item
+// SetKey sets binary key to configuration parameter.
+func (c *ConfigPrm) SetKey(v []byte) {
+	c.key = v
 }
 
-// IterateRecords iterates over all config records and passes them to f.
+// Config performs the test invoke of get config value
+// method of NeoFS Netmap contract.
+func (c *Client) Config(p ConfigPrm, assert func(stackitem.Item) (interface{}, error)) (interface{}, error) {
```
carpawell, line 282: do we need it public if we have getters for all (all, yes?) the config values?

pkg/morph/client/audit/client.go
```diff
@@ -25,7 +27,12 @@ const (
...
 )
 
-// New creates, initializes and returns the Client instance.
-func New(c *client.StaticClient) *Client {
-	return &Client{client: c}
+// NewFromMorph returns the wrapper instance from the raw morph client.
+func NewFromMorph(cli *client.Client, contract util.Uint160, fee fixedn.Fixed8, opts ...client.StaticClientOption) (*Client, error) {
+	sc, err := client.NewStatic(cli, contract, fee, opts...)
+	if err != nil {
+		return nil, err
```
carpawell, line 34:
```suggestion
		return nil, fmt.Errorf("could not create static client of audit contract: %w", err)
```

to sync with the other constructors

carpawell (commented):
what rule did you use when deciding if the parameters of the call should be kept in a group or separated?

pkg/morph/client/neofs/cheque.go
fyrchik: Fixed

pkg/morph/client/neofs/bind.go
fyrchik: Fixde

pkg/morph/client/audit/client.go
fyrchik: Fixed.

pkg/morph/client/netmap/config.go
fyrchik: No, fixed.

fyrchik:
@carpawell 0-1 parameter methods are used without a structure, 3+ parameter methods are used with structure. Did you notice some inconsistencies?

alexvanin (approved):

carpawell:
> @carpawell 0-1 parameter methods are used without a structure, 3+ parameter methods are used with structure. Did you notice some inconsistencies?

@fyrchik, just remember that we had it grouped (even if there are no parameters at all) to make it easier to maintain client api interface changes. if some of them are not grouped then i do not see it reasonable to group them at all

PR #1131, opened 2022-02-02
Upd/Adopt SDK changes: `Address` and `ID` was moved to the separate pkgs

carpawell:
Rebase and merge only after #1101.

`object.Address` has been moved to `object/address`
`object.ID` has been moved to `object/id`


alexvanin (approved):
`oidSDK` and `addressSDK` looks better for aliases, IMO. Package name + SDK suffix.

carpawell:
> Package name + SDK suffix.

Btw, @alexvanin, @fyrchik, @cthulhu-rider, this PR also contains some alias renaming because I think that it is important to have some unspoken rules for aliasing when we have some pkgs with the same names but from different repos: I suggest using `*pkg*SDK` for `sdk` repo, `*pkg*V2` for `api` v2 repo and just `*pkg*` from internal packages.

Sometimes some aliases really confuse me when I meet them in different parts of the code _differently_. If all agree, I can rename them in the following PR.

PR #1138, opened 2022-02-07
neofs-cli: Set expiration in session token based on network info

alexvanin:
With SDK update, `neofs-cli` has to manually set expiration value of session token. Token duration should not be quite long, because session tokens are issued on each `neofs-cli` invocation. In this PR it is hardcoded to 10 epochs.

I don't think it is worth to provide interface to set manual session token duration, because session tokens are not reusable in `neofs-cli`.

### DIFF
--- a/cmd/neofs-cli/modules/object.go
+++ b/cmd/neofs-cli/modules/object.go
@@ -37,6 +37,8 @@ const (
 
 const bearerTokenFlag = "bearer"
 
+const sessionTokenLifetime = 10 // in epochs
+
 var (
 	// objectCmd represents the object command
 	objectCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


cmd/neofs-cli/modules/object.go
```diff
@@ -299,17 +301,26 @@ func prepareSessionPrmWithOwner(
...
+	cws[1] = &netInfoPrm
 
 	for i := range prms {
 		cws = append(cws, prms[i])
 	}
 
 	prepareAPIClientWithKey(cmd, key, cws...)
 
+	ni, err := internalclient.NetworkInfo(netInfoPrm)
+	exitOnErr(cmd, errf("open session: %w", err))
```
cthulhu-rider, line 320:
```suggestion
	exitOnErr(cmd, errf("read network info: %w", err))
```

cthulhu-rider (approved):

cmd/neofs-cli/modules/object.go
```diff
@@ -37,6 +37,8 @@ const (
 
 const bearerTokenFlag = "bearer"
 
+const sessionTokenDuration = 10 // in epochs
```
fyrchik, line 40: `sessionTokenLifetime`?

fyrchik (approved):

carpawell (approved):

PR #1145, opened 2022-02-07
[#1115] *: link TODOs to corresponding issues

fyrchik:
Close #1115 .

Some of them were removed, please, check if `todo` or `fixme` is still needed.

### DIFF
--- a/pkg/innerring/processors/neofs/process_assets.go
+++ b/pkg/innerring/processors/neofs/process_assets.go
@@ -86,7 +86,6 @@ func (np *Processor) processWithdraw(withdraw *neofsEvent.Withdraw) {
 	}
 
 	// create lock account
-	// fixme: check collision there, consider reversed script hash
 	lock, err := util.Uint160DecodeBytesBE(withdraw.ID()[:util.Uint160Size])
 	if err != nil {
 		np.log.Error("can't create lock account", zap.Error(err))
--- a/pkg/local_object_storage/metabase/select.go
+++ b/pkg/local_object_storage/metabase/select.go
@@ -128,7 +128,6 @@ func (db *DB) selectObjects(tx *bbolt.Tx, cid *cid.ID, fs object.SearchFilters) 
 
 		addr, err := addressFromKey([]byte(a))
 		if err != nil {
-			// TODO: storage was broken, so we need to handle it
 			return nil, err
 		}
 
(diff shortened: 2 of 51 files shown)


pkg/local_object_storage/metabase/select.go
```diff
@@ -128,7 +128,6 @@ func (db *DB) selectObjects(tx *bbolt.Tx, cid *cid.ID, fs object.SearchFilters)
 
 		addr, err := addressFromKey([]byte(a))
 		if err != nil {
-			// TODO: storage was broken, so we need to handle it
```
carpawell, line 131: think it is important

pkg/morph/event/listener.go
```diff
@@ -272,7 +272,6 @@ loop:
 				continue loop
 			}
 
-			// TODO: consider asynchronous execution
```
carpawell, line 275: think it is important

pkg/local_object_storage/metabase/select.go
fyrchik: It was fixed in #1118 on the shard level.

pkg/services/object/util/key.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 var (
-	// todo(alexvanin): should be a part of status API
+	// TODO(alexvanin): #1144 should be a part of status API
```
alexvanin, line 13: Link to the neofs-api issue. Personally, I would delete it, because issue is in-progress.

pkg/innerring/processors/neofs/process_assets.go
```diff
@@ -86,7 +86,6 @@ func (np *Processor) processWithdraw(withdraw *neofsEvent.Withdraw) {
 	}
 
 	// create lock account
-	// fixme: check collision there, consider reversed script hash
```
alexvanin, line 89: I still afraid of the collisions between lock accounts and actual user accounts, but have no good solution for that.

pkg/morph/event/listener.go
alexvanin: It can be discussed, indeed. But I propose to keep block handlers sync:
1. there is one block handler per application now
2. there is no performance issue with sync execution, we can consider that when we will need extra performance on that.

alexvanin (approved):

cmd/neofs-ir/defaults.go
```diff
@@ -23,7 +23,11 @@ func newConfig(path string) (*viper.Viper, error) {
 
 	if path != "" {
 		v.SetConfigFile(path)
-		v.SetConfigType("yml") // fixme: for now
+		if strings.HasSuffix(path, ".json") {
```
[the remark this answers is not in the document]
fyrchik: We read _default_ config there, I think one `.config/neofs.yaml` is enough.

alexvanin (approved):

PR #1170, opened 2022-02-11
Fix/One endpoint list for RPC and notifications

### LINKED ISSUE #1053: Use the same endpoint for notifications and RPC requests

carpawell:
Closes #1053.


go.mod
```diff
@@ -17,7 +17,8 @@ require (
 	github.com/nats-io/nats-server/v2 v2.7.4 // indirect
 	github.com/nats-io/nats.go v1.13.1-0.20220308171302-2f2f6968e98d
 	github.com/nspcc-dev/hrw v1.0.9
-	github.com/nspcc-dev/neo-go v0.98.0
+	github.com/nspcc-dev/neo-go v0.98.3-pre.0.20220321144433-3b639f518ebb
+	github.com/nspcc-dev/neo-go/examples/nft-nd-nns v0.0.0-20220325151650-c039133acf68
```
alexvanin, line 21: Why not neo-go v0.98.2 and why examples?

carpawell: fixed

pkg/morph/client/client.go
```diff
@@ -455,18 +455,23 @@ func toStackParameter(value interface{}) (sc.Parameter, error) {
 }
 
 // MagicNumber returns the magic number of the network
-// to which the underlying RPC node client is connected.
-//
-// Returns 0 in case of connection problems.
-func (c *Client) MagicNumber() (res uint64, err error) {
+// to which the underlying RPC node client is connected to.
```
fyrchik, line 458: Why this change? (2 `to`s?)

pkg/morph/client/notary.go
```diff
@@ -380,11 +380,13 @@ func (c *Client) NotarySignAndInvokeTX(mainTx *transaction.Transaction) error {
 		return err
 	}
 
+	magicNumber, _ := c.MagicNumber()
```
fyrchik, line 383: Could you add a comment, about why the error is ignored?

pkg/morph/client/client.go
carpawell: by mistake, fixed

pkg/morph/client/notary.go
carpawell: sure, added

pkg/morph/client/multi.go
```diff
@@ -1,102 +1,157 @@
...
+func (c *Client) notificationLoop() {
+	for {
+		select {
+		case <-c.cfg.ctx.Done():
+			_ = c.UnsubscribeAll()
+			c.close()
+
+			return
+		case <-c.closeChan:
+			_ = c.UnsubscribeAll()
```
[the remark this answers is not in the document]
carpawell: that is internal channel, not neo-go (neo-go is `case n, ok := <-c.client.Notifications:` below)

it only receives `struct{}` when `Close` has been called, so it is a normal shutdown, not because of a connection lost

pkg/morph/client/multi.go
```diff
@@ -1,102 +1,157 @@
...
+			// considered to be lost
+			if !ok {
+				c.logger.Warn("switching to the next RPC node")
+
+				if !c.switchRPC() {
+					c.logger.Error("could not establish connection to any RPC node")
+
+					// could not connect to all endpoints =>
+					// switch client to inactive mode
+					c.inactiveMode()
```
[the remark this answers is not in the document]
carpawell: no, it leads to returning `ErrConnectionLost` error on every RPC call

do not know what behavior is expected. what do you think? maybe add an option `WithErrorChannel` or `WithConnectionLostCallback`?

fyrchik (approved):

alexvanin:
Do not forget to invalidate (create new) caches for each endpoint as it is done in https://github.com/nspcc-dev/neofs-node/pull/1256

pkg/morph/client/multi.go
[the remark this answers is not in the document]
carpawell: https://github.com/nspcc-dev/neofs-node/pull/1170/commits/9458ab02a7e938bb0dc3c7a34113e32bf0854922, https://github.com/nspcc-dev/neofs-node/pull/1170/commits/8723409a5a917c711b33fe4444d8868c6495289c

carpawell:
> Do not forget to invalidate (create new) caches

@alexvanin, https://github.com/nspcc-dev/neofs-node/pull/1170/commits/36bf45143df93b7c8fea184f12a0928014b2c920

alexvanin (approved):
DNT, but we've been looking at this PR with @carpawell for quite a while.

alexvanin:
With @carpawell we did an experiment using public neo-go node of the N3 testnet side chain to see raw performance of HTTP and WS clients. In the experiment, `N` workers try to fetch 5000 blocks from the RPC using the client. There are three client configurations:
1. Websocket client
2. HTTP client with `MaxConnsPerHost = N`
3. HTTP client with `MaxConnsPerHost = 0`

In the table below you may see amount of RPS in each configuration. WS client starts to shine with 2 or more simultanious requests because it has no overhead on creating new connections to the server and it provides up to 2 times more RPS. This can be really helpful for heavily loaded dApps. 

| Workers/Client | WebSocket | HTTP(N) | HTTP(0) |
|--|--|--|--|
| 1 worker | 19 | 18 | 18
| 2 workers | 37 | 22 | 22
| 10 workers | 182±4 | 82±3 | 107±4 |
| 50 workers | 878±5 | 346±15 | 453±28 |
| 100 workers | 1520±21 | 557±60 | 711±8 |

[Code example](https://gist.github.com/carpawell/a6e89559ff014beb6623000bb73b854c)

PR #1171, opened 2022-02-11
example: Add neofs-cli config example

### LINKED ISSUE #1103: `neofs-cli control` usability
- [x] 1. One should be able to set path to wallet and its password in the Control API config file
- [x] 2. `--generate-key` flag has no sense for Control API
- [ ] 3. It would be nice to have a Control API config example with all available options and reasonable comments about them, for instance, alognside with storage node configs
- [x] 4. It might improve usability if help messages about flags will provide an infomation about flags mutual exclusion
- [x] 5. I believe it will improve usability if help message for `healthcheck` will explicitly state that by default request is formed for Storage Node. If a user needs to send healthcheck request to Inner Ring node, they need to set `--ir` flag

### LINKED ISSUE #1166: [bug in v0.27.5]: `--ir` key for control API isn't available in config
When I request healthcheck info from Control API with in Inner Ring node, CLI fails. To make it work, I need to provide `--ir` key from command line. 

## Expected Behavior

I expect that I can specify `ir: true` key in config
```
user@host ~ # cat  control.yml
endpoint: [localhost:8090](http://localhost:8090/)
wallet: /path/to/wallet.json
password: ***
ir: true

user@host ~ # /usr/local/bin/neofs-cli control healthcheck -c control.yml
Health status: READY
```

## Current Behavior
The response I get with the above config
```
user@host ~ # /usr/local/bin/neofs-cli control healthcheck -c control.yml
rpc error: rpc error: code = Unimplemented desc = unknown service control.ControlService
```
Response is ok when I use `--ir` key
```
user@host ~ # neofs-cli control healthcheck --config control.yml --ir
Health status: READY
``` 

## Your Environment
```
user@host ~ # /usr/local/bin/neofs-cli --version
Version: v0.27.5 
Build: 2022-01-31T11:48:11+00:00 
Debug: false
```

alexvanin:
Close #1103 
Close #1166

### DIFF
--- a/config/example/cli.yaml
+++ b/config/example/cli.yaml
@@ -1,0 +1,7 @@
+---
+
+wallet: wallets/wallet.json                 # path to the NEP-6 wallet file or path to private key file or WIF
+address: Nf5YDCHrrtFCbetGT8TF1kwH1SjnJjT9N1 # account address in the NEP-6 wallet, omit for default address
+password: secret                            # account password, use "" for empty password
+rpc-endpoint: s01.neofs.devenv:8080         # NeoFS API endpoint of NeoFS node
+endpoint: localhost:8090                    # Control API endpoint of NeoFS node


config/example/cli.yaml
```diff
@@ -0,0 +1,5 @@
+wallet: ./wallets/wallet.json               # path to the NEP-6 wallet file or path to private key file or WIF
+address: Nf5YDCHrrtFCbetGT8TF1kwH1SjnJjT9N1 # account address in NEP-6 wallet, omit for default address
+password: hunter2                           # account password, use "" for empty password
```
532910, line 3: `secret` or `supersecret`

config/example/cli.yaml
```diff
@@ -0,0 +1,5 @@
...
+rpc-endpoint: s01.neofs.devenv:8080         # NeoFS API endpoint of NeoFS Storage node
+endpoint: 127.0.0.1:8090                    # Control API endpoint of NeoFS Storage node
```
532910: localhost:8090

config/example/cli.yaml
```diff
@@ -0,0 +1,5 @@
+wallet: ./wallets/wallet.json               # path to the NEP-6 wallet file or path to private key file or WIF
```
532910, line 1: begin file with `---\n\n`

`./` is required for execution; files could be specified without it: `wallets/wallet.json`

config/example/cli.yaml
```diff
@@ -0,0 +1,5 @@
...
+address: Nf5YDCHrrtFCbetGT8TF1kwH1SjnJjT9N1 # account address in NEP-6 wallet, omit for default address
+password: hunter2                           # account password, use "" for empty password
+rpc-endpoint: s01.neofs.devenv:8080         # NeoFS API endpoint of NeoFS node
+endpoint: 127.0.0.1:8090                    # Control API endpoint of NeoFS node
```
532910, line 5: localhost:8090

532910 (approved):

fyrchik (approved):

PR #1195, opened 2022-02-25
Adopt changes in NeoFS SDK

cthulhu-rider:


cmd/neofs-cli/internal/client/client.go
```diff
@@ -336,28 +403,54 @@ type DeleteObjectPrm struct {
...
 //
 // Returns any error prevented the operation from completing correctly in error return.
-func DeleteObject(prm DeleteObjectPrm) (res DeleteObjectRes, err error) {
-	var delPrm client.DeleteObjectParams
+func DeleteObject(prm DeleteObjectPrm) (*DeleteObjectRes, error) {
+	var delPrm client.PrmObjectDelete
 
-	delPrm.WithAddress(prm.objAddr)
+	if id := prm.objAddr.ContainerID(); id != nil {
+		delPrm.FromContainer(*id)
```
fyrchik, line 421: As a side-note, why do we use non-pointer parameters in SDK? It seems like we mostly use pointers in other places.

cmd/neofs-cli/internal/client/client.go
```diff
@@ -384,19 +477,60 @@ func (x GetObjectRes) Header() *object.Object {
...
+
+	var hdr object.Object
+
+	if !rdr.ReadHeader(&hdr) {
+		_, err = rdr.Close()
+		return nil, fmt.Errorf("read object header: %w", err)
+	}
+
+	sz := hdr.PayloadSize()
+	if sz > 1024 {
```
fyrchik, line 520: Where does this constant comes from?

pkg/services/audit/auditor/pop.go
```diff
@@ -1,15 +1,14 @@
...
-	"github.com/nspcc-dev/neofs-sdk-go/client"
 	"github.com/nspcc-dev/neofs-sdk-go/netmap"
 	oidSDK "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"go.uber.org/zap"
 )
 
 const (
 	hashRangeNumber    = 4
-	minGamePayloadSize = hashRangeNumber * client.TZSize
+	minGamePayloadSize = hashRangeNumber * 64
```
fyrchik, line 11: Could you make a named constant or at least add a comment about `64`?

pkg/services/object/internal/client/client.go
```diff
@@ -237,97 +319,196 @@ func (x PayloadRangeRes) PayloadRange() []byte {
...
+	}
+
+	rdr, err := prm.cli.ObjectRangeInit(prm.ctx, prm.cliPrm)
+	if err != nil {
+		return nil, fmt.Errorf("init payload reading: %w", err)
+	}
+
+	data, err := io.ReadAll(rdr)
+	if err != nil {
+		return nil, fmt.Errorf("read payload: %w", err)
```
fyrchik, line 342: Maybe allocate `make([]byte, prm.Length()` like we do in other places? Less reallocations.

cmd/neofs-cli/internal/client/client.go
cthulhu-rider: Using pointers in parameter structures is inappropriate for two reasons:

1. we never modify them, so we need to mention this fact in package docs clearly;
2. pointer binding adds garbage collection overhead, which I would like to avoid if possible.

But u r right, this approach differs with other places in project. In the future, all such places will be reworked according to the principle "RO - value, RW - pointer". Also, for example, I want to get away from pointer slices.

cmd/neofs-cli/internal/client/client.go
cthulhu-rider: From the assumption that NeoFS CLI cannot afford large data buffers. I'll create a const and add docs.

pkg/services/audit/auditor/pop.go
cthulhu-rider: Used recently defined `const tz.Size`.

cmd/neofs-cli/internal/client/client.go
fyrchik: I wouldn't worry about garbage collection ovehead, as here, for example the object already exists, so it will be collected anyway. In addition the (possible) overhead is much less than network-related delays but code complexity increase is much higher IMO.

cmd/neofs-cli/internal/client/client.go
```diff
@@ -303,29 +306,93 @@ func (x *PutObjectPrm) SetPayloadReader(rdr io.Reader) {
...
+			if prm.rdr != nil {
+				prm.rdr = io.MultiReader(bytes.NewReader(data), prm.rdr)
+			} else {
+				prm.rdr = bytes.NewReader(data)
+				sz = uint64(len(data))
+			}
+		}
+
+		if prm.rdr != nil {
+			const defaultBufferSizePut = 4096 // configure?
```
fyrchik, line 353: Could you also create a task for this and link to in from here (may be referencing other places with such restrictions)?

fyrchik (approved):

cmd/neofs-cli/internal/client/client.go
cthulhu-rider: > It is also not obvious whether copy overhead is less than that of a garbage collection.

I agree, but we shouldn't try to pseudo-optimize prevent stack growth with passing pointers where it is not necessary (like in parameters setters). https://github.com/golang/go/wiki/CodeReviewComments#pass-values

cmd/neofs-cli/internal/client/client.go
```diff
@@ -336,28 +403,54 @@ type DeleteObjectPrm struct {
...
+		delPrm.WithinSession(*prm.sessionToken)
+	}
+
+	if prm.bearerToken != nil {
+		delPrm.WithBearerToken(*prm.bearerToken)
+	}
+
+	cliRes, err := prm.cli.ObjectDelete(context.Background(), delPrm)
+	if err != nil {
+		return nil, fmt.Errorf("remove object via client: %w", err)
```
carpawell, line 438: i remember that you dont like "error" words but I want to mention that "failed" or "could not" looks more clear for understanding in the logs in my opinion

cmd/neofs-cli/modules/container.go
```diff
@@ -754,8 +746,8 @@ func prettyPrintContainer(cmd *cobra.Command, cnr *container.Container, jsonEnco
 	id := container.CalculateID(cnr)
 	cmd.Println("container ID:", id)
 
-	version := cnr.Version()
-	cmd.Printf("version: %d.%d\n", version.Major(), version.Minor())
+	v := cnr.Version()
```
carpawell, line 749: what that was done for?

pkg/services/object_manager/transformer/transformer.go
```diff
@@ -148,11 +146,11 @@ func payloadHashersForObject(obj *object.RawObject) []*payloadChecksumHasher {
 		{
 			hasher: tz.New(),
 			checksumWriter: func(cs []byte) {
-				if ln := len(cs); ln != tzChecksumSize {
-					panic(fmt.Sprintf("wrong checksum length: expected %d, has %d", ln, tzChecksumSize))
+				if ln := len(cs); ln != tz.Size {
+					panic(fmt.Sprintf("wrong checksum length: expected %d, has %d", ln, tz.Size))
```
carpawell, line 150: wrong arg order, huh?

carpawell (changes requested):

cmd/neofs-cli/internal/client/client.go
cthulhu-rider: Calling context can do it, here it is a fail by design of `error` type.

cmd/neofs-cli/modules/container.go
cthulhu-rider: `version` is a package name.

carpawell (approved):

PR #1197, opened 2022-02-28
Check for object existence in parallel

### LINKED ISSUE #1146: Check for object existence in `StorageEngine.Put()` in parallel
https://github.com/nspcc-dev/neofs-node/blob/master/pkg/local_object_storage/engine/put.go#L52

fyrchik:
Close #1146 .

Benchmark results on my laptop (all shards are on a single SSD).
```
name               old time/op    new time/op    delta
Exists/2_shards-8    25.4µs ± 3%    31.7µs ±16%  +25.07%  (p=0.000 n=10+10)
Exists/4_shards-8    49.2µs ± 7%    48.1µs ±12%     ~     (p=0.579 n=10+10)
Exists/8_shards-8    93.4µs ± 1%    69.9µs ±25%  -25.22%  (p=0.000 n=10+9)

name               old alloc/op   new alloc/op   delta
Exists/2_shards-8    4.98kB ± 0%    5.22kB ± 0%   +4.83%  (p=0.000 n=10+10)
Exists/4_shards-8    9.42kB ± 1%    9.85kB ± 1%   +4.58%  (p=0.000 n=10+10)
Exists/8_shards-8    18.2kB ± 1%    19.0kB ± 0%   +4.40%  (p=0.000 n=10+9)

name               old allocs/op  new allocs/op  delta
Exists/2_shards-8       111 ± 1%       120 ± 0%   +7.37%  (p=0.000 n=10+10)
Exists/4_shards-8       203 ± 1%       215 ± 1%   +6.02%  (p=0.000 n=10+10)
Exists/8_shards-8       384 ± 1%       404 ± 0%   +5.15%  (p=0.000 n=10+9)
```

So there are no gains unless we have 8 shards. I propose to test this in a 1 shard per SSD setup and see how it goes. Anyway, we can always switch between parallel / sequential check based on the amount of shards.

### DIFF
--- a/pkg/local_object_storage/engine/engine_test.go
+++ b/pkg/local_object_storage/engine/engine_test.go
@@ -4,6 +4,7 @@ import (
 	"crypto/sha256"
 	"fmt"
 	"math/rand"
+	"os"
 	"path/filepath"
 	"sync"
 	"testing"
@@ -16,7 +17,9 @@ import (
 	"github.com/nspcc-dev/neofs-node/pkg/util/test"
 	"github.com/nspcc-dev/neofs-sdk-go/checksum"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
+	cidtest "github.com/nspcc-dev/neofs-sdk-go/container/id/test"
 	objectSDK "github.com/nspcc-dev/neofs-sdk-go/object"
+	objecttest "github.com/nspcc-dev/neofs-sdk-go/object/address/test"
 	oidSDK "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"github.com/nspcc-dev/neofs-sdk-go/owner"
 	ownertest "github.com/nspcc-dev/neofs-sdk-go/owner/test"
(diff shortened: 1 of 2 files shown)


fyrchik:
Vegman results (each metabase is on a separate SSD, each blobstor on a separate HDD). Only 4 disks are there, so I skipped test with 8 shards. `pool` version is the same as this PR but with a separate pool of goroutines (via ants).

```
name \ time/op      old          new           pool
Exists/2_shards-40  52.3µs ±23%   91.6µs ± 6%   84.2µs ± 1%
Exists/4_shards-40  72.7µs ±11%  121.8µs ± 8%  116.9µs ± 5%

name \ alloc/op     old          new           pool
Exists/2_shards-40  5.00kB ± 0%   5.03kB ± 0%   5.16kB ± 0%
Exists/4_shards-40  9.89kB ± 0%   9.93kB ± 1%  10.19kB ± 0%

name \ allocs/op    old          new           pool
Exists/2_shards-40     112 ± 1%      115 ± 1%      117 ± 1%
Exists/4_shards-40     207 ± 1%      211 ± 1%      216 ± 1%
```

Pprof CPU profiles:
[cpuold.out.gz](https://github.com/nspcc-dev/neofs-node/files/8151753/cpuold.out.gz)
[cpunew.out.gz](https://github.com/nspcc-dev/neofs-node/files/8151752/cpunew.out.gz)
[cpupool.out.gz](https://github.com/nspcc-dev/neofs-node/files/8151909/cpupool.out.gz)

It baffled at me at first, that the relative results are _worse_ in this case. I think this is because synchronization costs are comparable to the `exists` check execution time.

So I think the change in it's current form is not worth it.

alexvanin (approved):

PR #1199, opened 2022-02-28
[#1186] engine: read object directly from blobstor in case of conflicts

### LINKED ISSUE #1186: `StorageEngine` can't read object from `BlobStor` if it is missing in `Metabase`
## Context
We've encountered a problem in NeoFS mainnet when node couldn't respond with the stored object because `BlobStor` record was in the `Shard` other then one with `Metabase` record.

## Current behavior
If object is stored in `BlobStor`, but missing in `Metabase`, `Shard.Get` returns `ErrNotFound`.

## Expected behavior
`Shard` (or `StorageEngine`) tries his best and reads the object from `BlobStor` anyway.

fyrchik:
Fix #1186.

Metabase is expected to contain actual information about objects stored
in shard. If the object is present in metabase but is missing from
blobstor, peform an additional attempt to fetch it directly without
consulting metabase. Such a situation is unexpected, so error counter
is increased for the shard which has the object in the metabase. We
don't increase error counter for the shard which has the object in
blobstor, because some garbage can be expected there. In this
implementation there is no overhead for objects which are really
missing, i.e. are not present in any metabase.

### DIFF
--- a/pkg/local_object_storage/engine/error_test.go
+++ b/pkg/local_object_storage/engine/error_test.go
@@ -7,6 +7,7 @@ import (
 	"strconv"
 	"testing"
 
+	"github.com/nspcc-dev/neofs-node/pkg/core/object"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor"
 	meta "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/metabase"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/shard"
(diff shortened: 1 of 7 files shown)


pkg/local_object_storage/engine/error_test.go
```diff
@@ -123,6 +128,53 @@ func TestErrorReporting(t *testing.T) {
 	})
 }
 
+// Issue #1186.
+func TestBlobstorFailback(t *testing.T) {
+	dir, err := os.MkdirTemp("", "*")
```
alexvanin, line 133: Remove dir with `t.Cleanup`?

alexvanin (approved):

alexvanin:
For me it seems that `GetRange` and `GetRangeHash` should implement the same pair of bool flags `skipMeta` and `hasMeta` and the same traversal logic in engine components.

pkg/local_object_storage/engine/error_test.go
fyrchik: Fixed.

alexvanin (approved):

PR #1201, opened 2022-02-28
Fix/Do not reset epoch timer on `NewEpoch`

carpawell:
`NewEpoch` event is received in the next after `NewEpoch` morph call block.
It leads to the race: if that event is handled first, tick progress of the
`epochTimer` is lost and an epoch becomes one block longer.

### DIFF
--- a/pkg/innerring/processors/netmap/process_epoch.go
+++ b/pkg/innerring/processors/netmap/process_epoch.go
@@ -14,6 +14,8 @@ import (
 func (np *Processor) processNewEpoch(ev netmapEvent.NewEpoch) {
 	epoch := ev.EpochNumber()
 
+	prevEpochDuration := np.epochState.EpochDuration()
+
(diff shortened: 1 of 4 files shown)


carpawell:
That problem was decided to be solved another way.

alexvanin:
Two words about another way to solve this issue. 

The main reason of unstable epoch timer is that Inner Ring node process notification events and block events asynchronously.
Notification event tries to reset ticker state. Block events tries to increment current ticker state.

```
1. Reset()               1. Tick()
2. Tick()                2. Reset()
Internal state: 0        Internal state: 1
```

So we have inconsistency there. Possible solution is to provide `Reset()` and `Tick()` methods with current block height. Then we can synchronize behaviour.

```
1. Reset(10)               1. Tick(10)
2. Tick(10)                2. Reset(10)
Internal state: 1        Internal state: 1
```

PR #1212, opened 2022-03-03
Fix/NPE in multi client

carpawell:
Share multi cache between single clients.

### DIFF
--- a/pkg/morph/client/multi.go
+++ b/pkg/morph/client/multi.go
@@ -15,6 +15,8 @@ type multiClient struct {
 
 	sharedNotary *notary // notary config needed for single client construction
 
+	sharedCache cache
+
 	endpoints  []string
 	clientsMtx sync.Mutex
 	clients    map[string]*Client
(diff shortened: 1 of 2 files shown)


alexvanin (approved):

pkg/morph/client/multi.go
```diff
@@ -41,6 +43,7 @@ func (x *multiClient) createForAddress(addr string) (*Client, error) {
 	sCli.notary = x.sharedNotary
 
 	c := &Client{
+		cache:        x.sharedCache,
```
fyrchik, line 46: I think it is a good idea to not share this cache because this way one incorrect response can prevent other clients from working properly. I will create an issue for that.

fyrchik (approved):

pkg/morph/client/multi.go
carpawell: if you mean just creating a new cache for every single client, I have nothing against it and can add it to that PR

just need to add smth with a non-nil cache at that line

PR #1214, opened 2022-03-03
Use single `Object` type in whole project

cthulhu-rider:
Remove `Object` and `RawObject` types from `pkg/core/object` package.
Use `Object` type from NeoFS SDK Go library everywhere. Avoid using the
deprecated elements.


pkg/core/object/object.go
```diff
@@ -1,70 +1,18 @@
...
-		Object: obj,
-	}
-}
-
-// New constructs blank Object instance.
-func New() *Object {
-	return NewFromSDK(object.New())
-}
+// AddressFromObject returns address of the object.
+func AddressFromObject(obj *object.Object) *addressSDK.Address {
```
fyrchik, line 9: What about having an `Address()` in SDK?

fyrchik: Also, consider `AdressOf` as a name.

pkg/local_object_storage/writecache/writecache.go
```diff
@@ -3,8 +3,8 @@ package writecache
 import (
 	"sync"
 
-	"github.com/nspcc-dev/neofs-node/pkg/core/object"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor/fstree"
+	objectSDK "github.com/nspcc-dev/neofs-sdk-go/object"
```
fyrchik, line 7: Can we just replace the import without an alias? There is no conflict.

fyrchik (approved):

pkg/core/object/object.go
cthulhu-rider: > What about having an `Address()` in SDK?

I don't mind though it seems rather redundant due to its simplicity. In `node` it is very often used (as seen in current PR), but I have not yet seen a single use case in other projects.
More thoughts? @alexvanin @carpawell @KirillovDenis 

And I agree that `AddressOf` is more neat.

alexvanin: > I have not yet seen a single use case in other projects.

Client applications usually don't fetch address from object, indeed. Server applications (custom storage \ relay node implementations) can definitely benefit from that, but SDK is more client application oriented as for now.

pkg/local_object_storage/writecache/writecache.go
alexvanin: As far as I remember, @carpawell wanted to have the same import aliases for all SDK and V2 components? 
https://github.com/nspcc-dev/neofs-node/pull/1131#issuecomment-1031437700

carpawell: ok, if there are no conflicts, it could be done in such way

alexvanin (approved):

PR #1224, opened 2022-03-09
Fix/Handle subnet removal

### LINKED ISSUE #1162: Remove subnet from netmap candidates on subnet remove event
https://github.com/nspcc-dev/neofs-node/blob/1667ec9e6dc7f292365719f203f48282d8f9e076/pkg/innerring/subnet.go#L315

carpawell:
Closes #1162.

### DIFF
--- a/pkg/innerring/subnet.go
+++ b/pkg/innerring/subnet.go
@@ -3,16 +3,20 @@ package innerring
 import (
 	"crypto/ecdsa"
 	"crypto/elliptic"
+	"encoding/hex"
 	"errors"
 	"fmt"
 
 	"github.com/nspcc-dev/neo-go/pkg/core/mempoolevent"
 	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
+	neogoutil "github.com/nspcc-dev/neo-go/pkg/util"
 	irsubnet "github.com/nspcc-dev/neofs-node/pkg/innerring/processors/subnet"
+	netmapclient "github.com/nspcc-dev/neofs-node/pkg/morph/client/netmap"
 	morphsubnet "github.com/nspcc-dev/neofs-node/pkg/morph/client/subnet"
 	"github.com/nspcc-dev/neofs-node/pkg/morph/event"
 	subnetevents "github.com/nspcc-dev/neofs-node/pkg/morph/event/subnet"
 	"github.com/nspcc-dev/neofs-node/pkg/util"
+	"github.com/nspcc-dev/neofs-sdk-go/netmap"
 	"github.com/nspcc-dev/neofs-sdk-go/owner"
 	"github.com/nspcc-dev/neofs-sdk-go/subnet"
 	subnetid "github.com/nspcc-dev/neofs-sdk-go/subnet/id"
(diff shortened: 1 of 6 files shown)


alexvanin:
Not blocked anymore, neofs-dev-env uses neofs-contract v0.15.0 release with the fix.

pkg/innerring/subnet.go
```diff
@@ -315,5 +319,93 @@ func (s *Server) handleSubnetRemoval(e event.Event) {
...
+	if err != nil {
+		s.log.Error("getting netmap candidates",
+			zap.Error(err),
+		)
+
+		return
+	}
+
+	var removedID subnetid.ID
+	_ = removedID.Unmarshal(delEv.ID()) // parsing errors were checked above
```
cthulhu-rider, line 334: Too risky IMO. We need to get `removeID` from the place where it was checked `above`, or check for errors explicitly.

pkg/innerring/subnet.go
```diff
@@ -315,5 +319,93 @@ func (s *Server) handleSubnetRemoval(e event.Event) {
...
+		}
+
+		// not remove from netmap if there
+		// is at least one more subnet
+		removeFromNM = false
+
+		return nil
+	})
+	if err != nil {
+		s.log.Error("iterating node's subnets",
```
cthulhu-rider, line 368: `s.log` expected, maybe u wanted `log`?

pkg/innerring/subnet.go
```diff
@@ -315,5 +319,93 @@ func (s *Server) handleSubnetRemoval(e event.Event) {
...
+			zap.Error(err),
+		)
+	}
+
+	// remove from netmap if there are no
+	// more subnets for the node or if it
+	// is impossible to process its subnet
+	// information
+	if err != nil || removeFromNM {
```
cthulhu-rider, line 377: Pay attention on `IterateSubnets` docs: it returns an error if node has no or has been removed from all subnets. So we can get rid of `removeFromNM`.

Note that in this case this `if` statement and the one above can be merged.

carpawell: yes, read the doc but missed "...not included in any subnet **by the end of the loop**", thanks, fixed

pkg/innerring/subnet.go
carpawell: ok, agree

pkg/innerring/subnet.go
carpawell: oh no: i remembered why i did so: yes, i read the doc correctly but did not want to merge two cases (printing the same type of logs and so on): "could not read the subnet ID" (which seems an error) and "node does not have any more subnet " (which seems as ok situation)

on the other hand that is not that critical so maybe leave them merged, what do you think?

cthulhu-rider: `node does not have any more subnet` is expected to raise pretty rare, so I'd print it anyway.

cthulhu-rider (approved):

pkg/innerring/subnet.go
carpawell: @cthulhu-rider so maybe we need to have public `ErrNoSubnet` in SDK

PR #1231, opened 2022-03-11
Update new SDK Client interface

alexvanin:
Ready for review. I will merge after integration tests.

### DIFF
--- a/pkg/network/cache/multi.go
+++ b/pkg/network/cache/multi.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"crypto/tls"
 	"errors"
-	"io"
 	"sync"
 
 	rawclient "github.com/nspcc-dev/neofs-api-go/v2/rpc/client"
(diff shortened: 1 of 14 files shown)


pkg/network/cache/multi.go
```diff
@@ -33,21 +32,39 @@ func newMultiClient(addr network.AddressGroup, opts []client.Option) *multiClien
...
-		opts = append(opts, client.WithTLSConfig(&tls.Config{}))
+		prmDial.SetTLSConfig(&tls.Config{})
 	}
 
-	c, err := client.New(opts...)
+	if x.opts.Key != nil {
+		prmInit.SetDefaultPrivateKey(*x.opts.Key)
+	}
+
+	if x.opts.DialTimeout != 0 {
```
cthulhu-rider, line 50:
```suggestion
	if x.opts.DialTimeout > 0 {
```

cthulhu-rider (approved):

fyrchik (approved):

pkg/network/cache/multi.go
fyrchik: I think panic is ok here, node is likely misconfigured.

PR #1236, opened 2022-03-15
neofs-node: remove mutex from `lruNetCache`

fyrchik:
We already use thread-safe LRU and mutex shouldn't be taken while making
network requests.


alexvanin (approved):

carpawell:
@fyrchik, some problems with commits i guess

carpawell (approved):

Issue #1143, opened 2022-02-07
Support working in degraded mode (no metabase)


fyrchik:
https://github.com/nspcc-dev/neofs-node/blob/master/pkg/local_object_storage/blobstor/exists.go#L60
It is used in code, but unimplemented. `Blobstor.Exists()` seems to be unused, though, as `Shard.Exists` uses metabase only.
Maybe we can remove `Blobstor.Exists()` completely or use it in some way.

alexvanin:
We need exist to work in degraded mode without matabase. Let's change this issue to "support degraded mode in storage engine"

fyrchik:
Refs #1186.

PR #1237, opened 2022-03-15
shard: support working in degraded mode

fyrchik:
Refs #1143 .

Let's also discuss whether we need a support for `search`.
Also we may introduce a separate `degraded` mode to prevent slowdown when shard was moved to readonly because of maintenance.


pkg/local_object_storage/blobovnicza/get.go
```diff
@@ -1,6 +1,7 @@
 package blobovnicza
 
 import (
+	"github.com/nspcc-dev/neo-go/pkg/util/slice"
```
carpawell, line 4: is it ok to use such util from `neo-go` in the engine of the node?

maybe implement somewhere in that repo?

pkg/local_object_storage/blobstor/exists.go
```diff
@@ -29,20 +33,35 @@ func (r ExistsRes) Exists() bool {
...
-		// TODO: #1143 do smth if err != nil
+		var smallErr error
 
-		// check presence in blobovnicza
-		exists, err = b.existsSmall(prm.addr)
+		exists, smallErr = b.existsSmall(prm.addr)
+		if err != nil && (smallErr != nil || exists) {
+			b.log.Warn("error occured during object existence checking",
+				zap.Stringer("address", prm.addr),
+				zap.String("error", err.Error()))
```
carpawell, line 53: `zap.Error(err)`?

pkg/local_object_storage/blobstor/exists.go
```diff
@@ -55,8 +74,36 @@ func (b *BlobStor) existsBig(addr *addressSDK.Address) (bool, error) {
...
+		dirPath := filepath.Dir(p)
+
+		_, ok := activeCache[dirPath]
+
+		_, err := b.getObjectFromLevel(prm, p, !ok)
+		if err != nil {
+			if !errors.Is(err, object.ErrNotFound) {
+				b.log.Debug("could not get object from level",
+					zap.String("level", p),
+					zap.String("error", err.Error()))
```
carpawell, line 99: `zap.Error(err)`?

pkg/local_object_storage/blobstor/exists.go
```diff
@@ -55,8 +74,36 @@ func (b *BlobStor) existsBig(addr *addressSDK.Address) (bool, error) {
...
+	var found bool
+	err := b.iterateSortedLeaves(addr, func(p string) (bool, error) {
+		dirPath := filepath.Dir(p)
+
+		_, ok := activeCache[dirPath]
+
+		_, err := b.getObjectFromLevel(prm, p, !ok)
+		if err != nil {
+			if !errors.Is(err, object.ErrNotFound) {
+				b.log.Debug("could not get object from level",
```
carpawell, line 97: why `debug`?

pkg/local_object_storage/shard/exists.go
```diff
@@ -34,13 +36,26 @@ func (p *ExistsRes) Exists() bool {
 // Returns any error encountered that does not allow to
 // unambiguously determine the presence of an object.
 func (s *Shard) Exists(prm *ExistsPrm) (*ExistsRes, error) {
-	exists, err := s.objectExists(prm.addr)
+	exists, err := meta.Exists(s.metaBase, prm.addr)
+	if err != nil {
+		// If the shard is in degraded mode, try to consult blobstor directly.
+		// Otherwise, just return an error.
+		if s.GetMode() == ModeReadOnly {
```
carpawell, line 43: why do you try reading only in readOnly mode?

carpawell (commented):
> Also we may introduce a separate degraded mode to prevent slowdown

IMO it is important and should be implemented.

pkg/local_object_storage/blobovnicza/get.go
fyrchik: Why not? It is not that we add a new dependency like in the `api-go` case.

pkg/local_object_storage/shard/exists.go
fyrchik: That is the whole point of this PR: enable additional logic in degraded mode.

pkg/local_object_storage/blobstor/exists.go
fyrchik: We mostly use `zap.String` in engine. `zap.Error` is more verbose.

alexvanin (approved):

pkg/local_object_storage/blobstor/exists.go
fyrchik: It is logged in `Exists` if it needs to be reported.

PR #1238, opened 2022-03-15
Adopt non pointer slices from SDK

alexvanin:
This is mostly bare minimum update. There are plenty of `[]*T` in the code, especially for `[]*object.Address` types, but I didn't touch them because they are being used in storage engine method signatures.


alexvanin:
Other changes will be introduced in https://github.com/nspcc-dev/neofs-node/issues/1239

fyrchik (approved):

PR #1242, opened 2022-03-15
Fix metabase put logging

alexvanin:
Was broken in #1214.

### DIFF
--- a/pkg/local_object_storage/metabase/put.go
+++ b/pkg/local_object_storage/metabase/put.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/nspcc-dev/neo-go/pkg/io"
 	"github.com/nspcc-dev/neofs-node/pkg/core/object"
+	objectCore "github.com/nspcc-dev/neofs-node/pkg/core/object"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobovnicza"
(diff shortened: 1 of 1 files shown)


pkg/local_object_storage/metabase/put.go
```diff
@@ -462,3 +462,8 @@ func isLinkObject(obj *objectSDK.Object) bool {
 func isLastObject(obj *objectSDK.Object) bool {
 	return len(obj.Children()) == 0 && obj.Parent() != nil
 }
+
+func saddrFromObj(obj *objectSDK.Object) string {
+	return obj.ContainerID().String() + "/" + obj.ID().String()
+
```
carpawell, line 468: extra line, linter

alexvanin:
> "saddr" means string address?

Yes. Maybe there is already a func to obtain `addressSDK.Address` or `string` from object, but didn't find it.

carpawell:
> didn't find it

maybe it is a sign to add such to SDK? or address getter

fyrchik (approved):

carpawell:
> Yes. Maybe there is already a func to obtain addressSDK.Address or string from object, but didn't find it.

@alexvanin, https://github.com/nspcc-dev/neofs-node/blob/master/pkg/core/object/object.go#L9

PR #1243, opened 2022-03-15
Fix/Add immediate object notifications

### LINKED ISSUE #1228: SN does not notify about objects with 0 `__NEOFS__TICK_EPOCH`
In #1183 object notification retrieving was added but not handling object with zero `__NEOFS__TICK_EPOCH` attribute value. Such objects must produce notifications immediately after creation according to spec.

carpawell:
Closes #1228.

### DIFF
--- a/cmd/neofs-node/config.go
+++ b/cmd/neofs-node/config.go
@@ -98,6 +98,8 @@ type cfg struct {
 
 	cfgObject cfgObject
 
+	cfgNotifications cfgNotifications
+
 	metricsCollector *metrics.StorageMetrics
 
 	workers []worker
@@ -183,6 +185,12 @@ type cfgObject struct {
 	pool cfgObjectRoutines
 
 	cfgLocalStorage cfgLocalStorage
+}
+
+type cfgNotifications struct {
+	enabled      bool
+	nw           notificationWriter
+	defaultTopic string
 }
 
 type cfgLocalStorage struct {
--- a/cmd/neofs-node/object.go
+++ b/cmd/neofs-node/object.go
@@ -327,7 +327,7 @@ func initObjectService(c *cfg) {
 	)
 
 	// build service pipeline
-	// grpc | <metrics> | acl | signature | response | split
+	// grpc | <metrics> | signature | response | acl | split
 
 	splitSvc := objectService.NewTransportSplitter(
 		c.cfgGRPC.maxChunkSize,
--- a/cmd/neofs-node/object.go
+++ b/cmd/neofs-node/object.go
@@ -40,6 +40,7 @@ import (
 	apistatus "github.com/nspcc-dev/neofs-sdk-go/client/status"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 	eaclSDK "github.com/nspcc-dev/neofs-sdk-go/eacl"
+	objectSDK "github.com/nspcc-dev/neofs-sdk-go/object"
 	addressSDK "github.com/nspcc-dev/neofs-sdk-go/object/address"
 	"github.com/nspcc-dev/neofs-sdk-go/owner"
 	"github.com/nspcc-dev/neofs-sdk-go/util/signature"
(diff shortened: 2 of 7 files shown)


pkg/services/object/put/local.go
```diff
@@ -3,13 +3,20 @@ package putsvc
...
 
-	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/engine"
 	"github.com/nspcc-dev/neofs-node/pkg/services/object_manager/transformer"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
+	objectSDK "github.com/nspcc-dev/neofs-sdk-go/object"
 )
 
+// ObjectStorage is an object storage interface.
+type ObjectStorage interface {
+	// Put must save passed object to the local disk
```
cthulhu-rider, line 13: IMO `local disk` is too strong requirement since `Service` doesn't care about what type of storage is used.

cmd/neofs-node/object.go
```diff
@@ -522,3 +537,40 @@ func (c *reputationClientConstructor) Get(info coreclient.NodeInfo) (coreclient.
...
+	e  *engine.StorageEngine
+	nw notificationWriter
+	ns netmap.State
+
+	defaultTopic string
+}
+
+func (e engineWithNotifications) Put(o *objectSDK.Object) error {
+	if err := engine.Put(e.e, o); err != nil {
+		return err
```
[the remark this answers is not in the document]
alexvanin: Other notifications are triggered by iterating over objects stored in engine. It will be consistent to produce those immediate notifications on the same basis.

But it is really good question, though. /cc @realloc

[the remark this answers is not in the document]
carpawell: > is it correct?

used a rule "notify only if I managed to store the object", thought that it was expected behavior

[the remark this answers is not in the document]
carpawell: btw, i thought that the main idea about notifications is to say that some node successfully stored the object

cmd/neofs-node/object.go
```diff
@@ -327,7 +327,7 @@ func initObjectService(c *cfg) {
 	)
 
 	// build service pipeline
-	// grpc | <metrics> | acl | signature | response | split
```
fyrchik, line 330: Could you explain why this change?
It seems like the previous order was correct.

pkg/services/notificator/nats/service.go
```diff
@@ -38,12 +38,12 @@ type Option func(*opts)
 var errConnIsClosed = errors.New("connection to the server is closed")
 
 // Notify sends object address's string representation to the provided topic.
-// Uses first 4 bytes of object ID as a message ID to support 'exactly once message'
-// delivery.
+// Uses first 4 bytes of object ID as a message ID to support 'exactly-once'
```
fyrchik, line 41: It is exactly once QoS in NATS documentation.

cmd/neofs-node/object.go
```diff
@@ -522,3 +537,40 @@ func (c *reputationClientConstructor) Get(info coreclient.NodeInfo) (coreclient.
...
+	defaultTopic string
+}
+
+func (e engineWithNotifications) Put(o *objectSDK.Object) error {
+	if err := engine.Put(e.e, o); err != nil {
+		return err
+	}
+
+	ni, err := o.NotificationInfo()
+	if err == nil {
```
fyrchik, line 555: Log the error if non-nil?

fyrchik (approved):

cmd/neofs-node/object.go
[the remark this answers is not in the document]
realloc: The notification must be issued only for successfully stored the object. Otherwise, the actions following the notification will have no sense.

cmd/neofs-node/object.go
alexvanin: Debug logs, perhaps. Log records produced by individual objects tend to consume all free space, unfortunately.

cmd/neofs-node/config.go
```diff
@@ -185,6 +187,13 @@ type cfgObject struct {
 	cfgLocalStorage cfgLocalStorage
 }
 
+type cfgNotifications struct {
+	enabled bool
+	nw      notificationWriter
+
```
cthulhu-rider, line 193: Could you please sync linebreaks between fields?

cmd/neofs-node/object.go
carpawell: maybe i'm wrong, did not look deeply, just saw the order of wrapping and saw this commits:

1. that comment [appeared](https://github.com/nspcc-dev/neofs-node/commit/534ce034217d3d4bd623484309340998e1a76408);
2. then the order was [changed](https://github.com/nspcc-dev/neofs-node/commit/e74ff7c15a770570ae8318399de1b8c35fc8f384) without fixing the comment

cthulhu-rider (approved):

pkg/services/notificator/nats/service.go
carpawell: fixed

cmd/neofs-node/object.go
carpawell: it could contain two types of errors:
1. notification is not set for the object (unfortunately, `sdk` does not have dedicated err for that, only `api`; will add), we do not need to log that
2. parsing errors that are not expected since it was received from the node that approved it

cmd/neofs-node/object.go
cthulhu-rider: I don't think this comment is really useful.

cmd/neofs-node/config.go
carpawell: sure

PR #1245, opened 2022-03-16
Feat/Add object notification to CLI

### LINKED ISSUE #1229: Support object notification in CLI
Support smth like `... put --notify '15:my_topic'` that means that object should produce notification on `15` epoch to `my_topic` topic.

### LINKED ISSUE #1299: Config example misses `engine.shard.writecache.capacity` value

carpawell:
Closes #1229.

Not sure if the current description in CLI is full enough.

### DIFF
--- a/cmd/neofs-cli/modules/object.go
+++ b/cmd/neofs-cli/modules/object.go
@@ -109,6 +109,8 @@ var (
 	}
 )
 
+const notificationFlag = "notify"
+
 const (
 	hashSha256 = "sha256"
 	hashTz     = "tz"
@@ -145,6 +147,8 @@ func initObjectPutCmd() {
 	flags.Bool("disable-timestamp", false, "Do not set well-known timestamp attribute")
 	flags.Uint64VarP(&putExpiredOn, putExpiresOnFlag, "e", 0, "Last epoch in the life of the object")
 	flags.Bool(noProgressFlag, false, "Do not show progress bar")
+
+	flags.String(notificationFlag, "", "Object notification in the form of *epoch*:*topic*; '-' topic means using default")
 }
 
 func initObjectDeleteCmd() {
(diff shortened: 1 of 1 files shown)


alexvanin (approved):

cmd/neofs-cli/modules/object.go
```diff
@@ -801,6 +812,42 @@ func parseObjectAttrs(cmd *cobra.Command) ([]object.Attribute, error) {
...
+
+	raw := cmd.Flag(notificationFlag).Value.String()
+	if raw == "" {
+		return nil, nil
+	}
+
+	rawSlice := strings.Split(raw, separator)
+	if len(rawSlice) != 2 {
+		return nil, fmt.Errorf("notification must be in the form of: *epoch*%s*topic*, got %s", separator, raw)
+	}
```
fyrchik, line 828: `SplitN` would suit here more, because topics are created by user and thus can contain colons, right?

cmd/neofs-cli/modules/object.go
```diff
@@ -145,6 +147,8 @@ func initObjectPutCmd() {
 	flags.Bool("disable-timestamp", false, "Do not set well-known timestamp attribute")
 	flags.Uint64VarP(&putExpiredOn, putExpiresOnFlag, "e", 0, "Last epoch in the life of the object")
 	flags.Bool(noProgressFlag, false, "Do not show progress bar")
+
+	flags.String(notificationFlag, "", "Object notification in the form of *epoch*:*topic*; '-' topic means using default")
```
fyrchik, line 151: Is `-` reserved on the API level? Seems like empty topic would agree better on what is happening in the code.

cmd/neofs-cli/modules/object.go
carpawell: fixed

cmd/neofs-cli/modules/object.go
carpawell: > Is - reserved on the API level?

No. Just an explicit sign that we want to use the node's default topic.
/cc @alexvanin

PR #1249, opened 2022-03-17
Support status returns in storage node's server

### LINKED ISSUE #1158: Return `object.ErrAlreadyRemoved` from object service when needed
https://github.com/nspcc-dev/neofs-node/blame/master/pkg/services/object/internal/client/client.go#L138
Either do it or refactor using status codes.

### LINKED ISSUE #1247: Adopt new status codes
It seems that node does not implement new status codes such as:
- SessionTokenNotFound
- SessionTokenExpired
- ObjectAccessDenied
- ObjectNotFound
- ObjectAlreadyRemoved
- ContainerNotFound

They all described in API and available in SDK `apistatus` package.

cthulhu-rider:
* closes #1247 
* closes #1158

### DIFF
--- a/pkg/local_object_storage/blobstor/delete_big.go
+++ b/pkg/local_object_storage/blobstor/delete_big.go
@@ -21,7 +21,7 @@ type DeleteBigRes struct{}
 // Returns any error encountered that did not allow
 // to completely remove the object.
 //
-// Returns apistatus.ObjectNotFound if there is no object to delete.
+// Returns an error of type apistatus.ObjectNotFound if there is no object to delete.
 func (b *BlobStor) DeleteBig(prm *DeleteBigPrm) (*DeleteBigRes, error) {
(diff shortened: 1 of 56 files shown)


pkg/local_object_storage/blobstor/delete_big.go
```diff
@@ -21,11 +21,13 @@ type DeleteBigRes struct{}
 // Returns any error encountered that did not allow
 // to completely remove the object.
 //
-// Returns ErrNotFound if there is no object to delete.
+// Returns apistatus.ObjectNotFound if there is no object to delete.
```
fyrchik, line 24: It looks like an error here, but is really a type, I think we need to state this clearly.

fyrchik (approved):

alexvanin (approved):
Integration tests are passed.

PR #1251, opened 2022-03-18
[#1250] go.mod: Bump supported go version to 1.17

### LINKED ISSUE #1250: Update go version
go1.16 is not supported after go1.18 release.

fyrchik:
Close #1250 

### DIFF
--- a/.docker/Dockerfile.adm
+++ b/.docker/Dockerfile.adm
@@ -1,4 +1,4 @@
-FROM golang:1.16 as builder
+FROM golang:1.17 as builder
 ARG BUILD=now
 ARG VERSION=dev
 ARG REPO=repository
--- a/.docker/Dockerfile.cli
+++ b/.docker/Dockerfile.cli
@@ -1,4 +1,4 @@
-FROM golang:1.16 as builder
+FROM golang:1.17 as builder
 ARG BUILD=now
 ARG VERSION=dev
 ARG REPO=repository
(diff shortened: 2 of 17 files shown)


carpawell:
@fyrchik, could you, please, follow the [checklist](https://github.com/nspcc-dev/neofs-node/blob/master/docs/update-go-instruction.md).

alexvanin:
go1.18 is too early because neo-go v0.98.2 is not released yet?

fyrchik:
@alexvanin what is the problem? `neo-go` builds fine with any version, we can update the neo-go itself separately.

carpawell (approved):

alexvanin:
> @alexvanin what is the problem? `neo-go` builds fine with any version, we can update the neo-go itself separately.

Oups, I thought I was looking into `neofs-contract` repository for some reason. Why not go1.18 then? I expect that we update go version once a year up to the latest stable.

fyrchik:
While we can go with go1.18 (pun intended), I expect some of the distros not having 1.18 version yet (debian stable still has 1.15 for example). 1.18 can also be not that stable yet given that was released a week ago. So I think 1.17 is a safe choice here.

alexvanin:
I agree with you that go1.17 is a safe choice, but still suggest to jump up to go1.18. We can wait a couple of weeks for CI packages to catch up and merge it after that.

This autumn go1.17 will not be supported and we will face go1.18 update once again. Therefore we can do it now and forget about it until next year, I believe.

alexvanin (approved):

PR #1253, opened 2022-03-18
Support neofs-contract fixes for notary enabled env

alexvanin:
Notary disabled environment was not affected, so testnet and mainnet won't be broken. It breaks compatibility with notary enabled environment which is neofs-dev-env only right now.


alexvanin:
Should be merged after neofs-contract v0.15.0 release.

fyrchik (approved):

PR #1256, opened 2022-03-21
Improve morph client cache

### LINKED ISSUE #1213: Improve morph client cache
1. Fields should be protected by a mutex.
2. Each client in multiclient should have it's own cache. Otherwise one bad endpoint can prevent other clients from working properly.

fyrchik:
Close #1213 .

### DIFF
--- a/pkg/morph/client/multi.go
+++ b/pkg/morph/client/multi.go
@@ -15,8 +15,6 @@ type multiClient struct {
 
 	sharedNotary *notary // notary config needed for single client construction
 
-	sharedCache cache
-
 	endpoints  []string
 	clientsMtx sync.Mutex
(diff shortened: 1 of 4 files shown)


pkg/morph/client/multi.go
```diff
@@ -66,14 +72,12 @@ func (x *multiClient) iterateClients(f func(*Client) error) error {
 		}
 
 		x.clientsMtx.Lock()
-
 		c, cached := x.clients[x.endpoints[i]]
+		x.clientsMtx.Unlock()
 		if !cached {
```
carpawell, line 77: did not get: we checked if it is cached here but then check once again in the `createForAddress`

PR #1259, opened 2022-03-21
Return non-zero exit code in `acl extended create` command failures

alexvanin:


cmd/neofs-cli/modules/acl/extended/create.go
```diff
@@ -67,35 +68,35 @@ func createEACL(cmd *cobra.Command, _ []string) {
...
-		return
+		cmd.Printf("invalid container ID: %v\n", err)
+		os.Exit(1)
 	}
 
 	rulesFile, err := getRulesFromFile(fileArg)
 	if err != nil {
-		cmd.Printf("can't read rules from file : %v", err)
-		return
+		cmd.Printf("can't read rules from file: %v\n", err)
```
fyrchik, line 77: `cmd.PrintErrf` here and in other places?

alexvanin: Done.

cmd/neofs-cli/modules/acl/extended/create.go
```diff
@@ -67,51 +68,51 @@ func createEACL(cmd *cobra.Command, _ []string) {
 
 	containerID := cid.New()
 	if err := containerID.Parse(cidArg); err != nil {
-		cmd.Printf("invalid container ID: %v", err)
-		return
+		cmd.PrintErrf("invalid container ID: %v\n", err)
```
carpawell, line 71: in the old package style we would write as following:

```go
err := containerID.Parse(cidArg)
exitOnErr(cmd, errf("invalid container ID: %w", err))
```

are we doing to have such funcs in the new package style?

alexvanin: > are we doing to have such funcs in the new package style?

Sure. But such functions should be defined in separate package to avoid dependency loop. I decided not to define such package just for this fix.

PR #1262, opened 2022-03-22
Fix blobovnicza, metabase index improvements

fyrchik:
1. Functions in `Update` and `Batch` should be idempotent.
2. Reduce allocations count during index processing in `Put` / `Delete`.

### DIFF
--- a/pkg/local_object_storage/metabase/db.go
+++ b/pkg/local_object_storage/metabase/db.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"time"
 
 	v2object "github.com/nspcc-dev/neofs-api-go/v2/object"
 	"github.com/nspcc-dev/neofs-node/pkg/util/logger"
@@ -30,6 +31,9 @@ type Option func(*cfg)
 type cfg struct {
 	boltOptions *bbolt.Options // optional
 
+	boltBatchSize  int
+	boltBatchDelay time.Duration
+
 	info Info
 
 	log *logger.Logger
@@ -40,8 +44,9 @@ func defaultCfg() *cfg {
 		info: Info{
 			Permission: os.ModePerm, // 0777
 		},
-
-		log: zap.L(),
+		boltBatchDelay: bbolt.DefaultMaxBatchDelay,
+		boltBatchSize:  bbolt.DefaultMaxBatchSize,
+		log:            zap.L(),
 	}
 }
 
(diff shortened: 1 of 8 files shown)


alexvanin:
I see benchmark in this PR. Please, provide results of the bench w/ and w/o memory alloc optimization in comment below.

fyrchik:
They are included in the commit messages. Here is a comparison of the master branch with a top commit:
```
dzeta@wpc ~/r/n/p/l/metabase (fix-blobstor)> benchstat before after
name              old time/op    new time/op    delta
Put/parallel-8      1.52ms ±14%    1.47ms ±14%   -3.16%  (p=0.030 n=20+19)
Put/sequential-8    5.29ms ± 2%    5.19ms ± 4%   -2.02%  (p=0.003 n=20+20)

name              old alloc/op   new alloc/op   delta
Put/parallel-8       134kB ±12%     118kB ± 4%  -11.47%  (p=0.000 n=20+17)
Put/sequential-8     173kB ± 1%     168kB ± 1%   -2.63%  (p=0.000 n=20+20)

name              old allocs/op  new allocs/op  delta
Put/parallel-8         573 ± 5%       468 ± 1%  -18.25%  (p=0.000 n=19+20)
Put/sequential-8       820 ± 1%       787 ± 1%   -3.97%  (p=0.000 n=20+19)
```

pkg/local_object_storage/metabase/put.go
```diff
@@ -377,18 +379,58 @@ func decodeList(data []byte) (lst [][]byte, err error) {
...
+			return nil, err
+		}
+		offset += uint64(n)
+
+		next := offset + sz
+		if uint64(len(data)) < next {
+			return nil, gio.ErrUnexpectedEOF
+		}
+		lst[i] = data[offset:next]
+		offset = next
```
alexvanin, line 400: Is  optimized version of `ReadVarBytes`? Fully compatible but has less allocation because of some use-case assumptions in the code?

alexvanin (approved):

pkg/local_object_storage/metabase/put.go
```diff
@@ -377,18 +376,58 @@ func decodeList(data []byte) (lst [][]byte, err error) {
...
 	}
 	return lst, nil
 }
 
+func getVarUint(data []byte) (uint64, int, error) {
```
carpawell, line 405: how about adding tests for that?

pkg/local_object_storage/metabase/db.go
```diff
@@ -144,3 +149,21 @@ func WithPermissions(perm fs.FileMode) Option {
...
+// so it is changed right after db is opened.
+func WithBatchSize(s int) Option {
+	return func(c *cfg) {
+		c.boltBatchSize = s
+	}
+}
+
+// WithBatchDelay returns option to specify maximum time to wait before
+// the batch of concurrent transactions is processed.
+// bbolt has no such option, so it is changed right after db is opened.
```
carpawell, line 164: lets sync using new line for that disclaimer for both new options

pkg/local_object_storage/metabase/put.go
fyrchik: Yes, the on-disk format is the same, will add a test.

pkg/local_object_storage/metabase/put.go
fyrchik: Fixed.

pkg/local_object_storage/metabase/db.go
fyrchik: Fixed.

PR #1273, opened 2022-03-28
Convert some of the remaining slices to use raw struct

### LINKED ISSUE #1239: Remove remaining `[]*T` after #1238
Starting with https://github.com/nspcc-dev/neofs-api-go/issues/376 repeated fields are now non-pointer slices `[]T` instead of pointer slices `[]*T`. #1238 introduce minimal changes to support new SDK. However we can go further and adopt these changes in the node. 

- [ ] Look for `[]*T` and `...*T` pattern in code and try to replace it with `[]` and `...`. Primal example for that is `[]*object.Address` which is heavily used in storage engine. We should check if changes do not provide much impact on storage engine performance.

fyrchik:
Close #1239 .

### DIFF
--- a/cmd/neofs-cli/internal/client/client.go
+++ b/cmd/neofs-cli/internal/client/client.go
@@ -631,11 +631,11 @@ func (x *SearchObjectsPrm) SetFilters(filters object.SearchFilters) {
 
 // SearchObjectsRes groups resulting values of SearchObjects operation.
 type SearchObjectsRes struct {
-	ids []*oidSDK.ID
+	ids []oidSDK.ID
 }
 
 // IDList returns identifiers of the matched objects.
-func (x SearchObjectsRes) IDList() []*oidSDK.ID {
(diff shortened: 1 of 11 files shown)


cmd/neofs-cli/modules/container.go
```diff
@@ -290,7 +290,7 @@ var listContainerObjectsCmd = &cobra.Command{
 		objectIDs := res.IDList()
 
 		for i := range objectIDs {
-			cmd.Println(objectIDs[i])
+			cmd.Println(&objectIDs[i])
```
alexvanin, line 293: You have suggested to use `.String()` in such cases: https://github.com/nspcc-dev/neofs-node/pull/1238#discussion_r826953610

alexvanin (approved):

carpawell (approved):

cmd/neofs-cli/modules/container.go
fyrchik: Fixed.

PR #1274, opened 2022-03-29
Fix cursor output with inhumed objects

alexvanin:
Inspired by https://github.com/nspcc-dev/neofs-node/issues/1186

I decided to check if it is possible to have inhumed object in cursor output. It is possible when object and tombstone are in different shards. Considering latest fixes, I think we have to process that as well.

### DIFF
--- a/pkg/local_object_storage/engine/list.go
+++ b/pkg/local_object_storage/engine/list.go
@@ -114,7 +114,16 @@ func (e *StorageEngine) ListWithCursor(prm *ListWithCursorPrm) (*ListWithCursorR
 			continue
 		}
 
-		result = append(result, res.AddressList()...)
+		addressList := res.AddressList()
+		for j := range addressList {
+			_, err = e.exists(addressList[j])
+			if err != nil {
+				continue
+			}
+
+			result = append(result, addressList[j])
+		}
+
 		cursor.shardCursor = res.Cursor()
 		cursor.shardID = shardIDs[i]
 	}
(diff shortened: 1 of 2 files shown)


pkg/local_object_storage/engine/list.go
```diff
@@ -114,7 +114,16 @@ func (e *StorageEngine) ListWithCursor(prm *ListWithCursorPrm) (*ListWithCursorR
 			continue
 		}
 
-		result = append(result, res.AddressList()...)
+		addressList := res.AddressList()
+		for j := range addressList {
+			_, err = e.exists(addressList[j])
+			if err != nil {
+				continue
+			}
```
fyrchik, line 122: I would do it in parallel. All objects here are _likely_ to exist, so we will process all 4 shards for each object.
Alternatively, we can (1) process all 4 shards with `exists(list of objects)` and (2) even do it in parallel.
In case we are not doing it now, could you add a TODO here?

fyrchik (approved):

alexvanin:
Also gonna test in the dev-env first, because testflight in the testnet shows some performance issues. Maybe it is related to this fix, maybe not.

carpawell:
What is the status of that PR?

alexvanin:
I am not sure that overhead of many `exist` checks during object listing is worth it, but the issue still relevant.

alexvanin:
I close this PR but will keep the branch in case if this issue will be relevant in public environment.

PR #1281, opened 2022-03-30
innerring: disallow to tick timer twice on the same height

### LINKED ISSUE #1208: IR node has race condition when ticks `NewEpoch`
IR node counts blocks from the Morph and decides when it is time to start a new epoch ([NewEpoch](https://github.com/nspcc-dev/neofs-contract/blob/v0.14.2/netmap/netmap_contract.go#L324) contract call) based on that information. This process has some kind of [sync](https://github.com/nspcc-dev/neofs-node/blob/master/pkg/innerring/processors/netmap/process_epoch.go#L26) that saves from Morph delays (tx with `NewEpoch` call may be included not in the next block of the chain).

However, it leads to the race: in most cases `NewEpoch` event is handled before the `NewBlock` event and it does not affect the system. But if `NewBlock` is handled before the timers reset, IR loses its block ticking progress (it has ticked the timers and after that it has to reset them according to the `NewEpoch` event handling) and the next epoch gonna be one block longer. It does not affect the system dramatically but certainly makes some epochs longer for no logical reason unpredictably.

Some thoughts were provided in the [comment](https://github.com/nspcc-dev/neofs-node/pull/1201#issuecomment-1055392038).

fyrchik:
Close #1208.

Provide current heights as an argument to ticker.
Zero height disables any checks, thus corresponding to the old
behaviour. If non-zero height is used, ignore the tick if the height
is less than the timer tick state.

### DIFF
--- a/pkg/innerring/blocktimer.go
+++ b/pkg/innerring/blocktimer.go
@@ -66,9 +66,9 @@ func (s *Server) startBlockTimers() error {
 	return nil
 }
 
-func (s *Server) tickTimers() {
+func (s *Server) tickTimers(h uint32) {
 	for i := range s.blockTimers {
-		s.blockTimers[i].Tick()
+		s.blockTimers[i].Tick(h)
 	}
 }
 
--- a/pkg/innerring/innerring.go
+++ b/pkg/innerring/innerring.go
@@ -241,7 +241,7 @@ func (s *Server) Start(ctx context.Context, intError chan<- error) (err error) {
 				zap.Uint32("block_index", b.Index))
 		}
 
-		s.tickTimers()
+		s.tickTimers(b.Index)
 	})
 
 	if !s.withoutMainNet {
--- a/pkg/innerring/processors/netmap/process_epoch.go
+++ b/pkg/innerring/processors/netmap/process_epoch.go
@@ -23,7 +23,15 @@ func (np *Processor) processNewEpoch(ev netmapEvent.NewEpoch) {
 	}
 
 	np.epochState.SetEpochCounter(epoch)
-	if err := np.epochTimer.ResetEpochTimer(); err != nil {
+
+	h, err := np.netmapClient.Morph().TxHeight(ev.TxHash())
+	if err != nil {
+		np.log.Warn("can't get transaction height",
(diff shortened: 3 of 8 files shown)


alexvanin (commented):
While this is correct, it does not solve original issue. Consider these two tests. They reproduce possible race between `Reset()` and `Tick()` that happens in IR.

```go
func TestBlockTimer_Reset(t *testing.T) {
	var baseCounter [2]int
	blockDur := uint32(3)

	bt1 := timer.NewBlockTimer(
		func() (uint32, error) { return blockDur, nil },
		func() { baseCounter[0]++ })
	bt2 := timer.NewBlockTimer(
		func() (uint32, error) { return blockDur, nil },
		func() { baseCounter[1]++ })

	require.NoError(t, bt1.Reset())
	require.NoError(t, bt2.Reset())

	run := func(bt *timer.BlockTimer, direct bool) {
		if direct {
			bt.Reset()
			bt.Tick(1)
		} else {
			bt.Tick(1)
			bt.Reset()
		}
		bt.Tick(2)
		bt.Tick(3)
	}

	run(bt1, true)
	run(bt2, false)
	require.Equal(t, baseCounter[0], baseCounter[1])
}

func TestBlockTime_Parallel(t *testing.T) {
	var baseCounter int
	blockDur := uint32(3)
	bt := timer.NewBlockTimer(
		func() (uint32, error) { return blockDur, nil },
		func() { baseCounter++ })
	require.NoError(t, bt.Reset())

	for i := 0; i < 1_000; i++ {
		var wg sync.WaitGroup
		wg.Add(2)
		go func() {
			bt.Tick(1)
			wg.Done()
		}()
		go func() {
			bt.Reset()
			wg.Done()
		}()
		wg.Wait()
		bt.Tick(2)
		bt.Tick(3)

		require.Equal(t, baseCounter, i+1)
	}
}
```

Proposed solution to add epoch number to the `Reset()`. Reset is triggered by notification, so we can find block height using `GetApplicationLog` request, I suppose.

Maybe there are better solutions for that.

fyrchik:
@alexvanin isn't this effectively the same as `Tick(h)` + `Reset` ?

alexvanin (approved):

PR #1289, opened 2022-04-04
neofs-adm: allow to change parameters of the Policy contract

fyrchik:
Tested here (will finish testing with actual parameter update check later)  https://github.com/nspcc-dev/neofs-adm-tests/tree/set-policy):
```
$ make clone
$ NEOFS_REPO=https://github.com/fyrchik/neofs-node.git ./test.set-policy.sh neofs-adm-set-policy master
```

TODO: finish test, add to CI

### DIFF
--- a/cmd/neofs-adm/internal/modules/morph/policy.go
+++ b/cmd/neofs-adm/internal/modules/morph/policy.go
@@ -1,0 +1,63 @@
+package morph
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/nspcc-dev/neo-go/pkg/core/native/nativenames"
+	"github.com/nspcc-dev/neo-go/pkg/io"
+	"github.com/nspcc-dev/neo-go/pkg/smartcontract/callflag"
+	"github.com/nspcc-dev/neo-go/pkg/vm/emit"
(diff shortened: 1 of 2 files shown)


cmd/neofs-adm/internal/modules/morph/policy.go
```diff
@@ -0,0 +1,64 @@
...
+		default:
+			return fmt.Errorf("parameter must be one of %s, %s and %s", execFeeParam, storagePriceParam, setFeeParam)
+		}
+
+		value, err := strconv.ParseUint(kv[1], 10, 32)
+		if err != nil {
+			return fmt.Errorf("can't parse parameter value '%s': %w", args[1], err)
+		}
+
+		emit.AppCall(bw.BinWriter, policyHash, "set"+kv[0], callflag.All, big.NewInt(int64(value)))
```
carpawell, line 52: does it work ok with small `value`? as I remember small numbers has its own opcodes. or it is ok to write any number as `big.Int`?

fyrchik: `big.Int` will adjust it's behaviour appropriately. But we still can use `int64` here, `val` is 32-bit.

PR #1291, opened 2022-04-05
innerring: Don't stop the same listener twice

alexvanin:
When IR node configured without main chain, both `morphListener` and `mainnetListener` are pointing into single listener component. We should not call `Stop()` twice, because it may trigger channel closing in neo-go or other components and it can throw panic.

Related to https://github.com/nspcc-dev/neo-go/pull/2420

### DIFF
--- a/pkg/innerring/innerring.go
+++ b/pkg/innerring/innerring.go
@@ -282,7 +282,9 @@ func (s *Server) Stop() {
 	s.setHealthStatus(control.HealthStatus_SHUTTING_DOWN)
 
 	go s.morphListener.Stop()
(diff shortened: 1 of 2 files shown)


carpawell:
do we need that if it gonna be changed in the `neo-go`?

fyrchik (approved):

AnnaShaleva (approved):

AnnaShaleva:
> do we need that if it gonna be changed in the `neo-go`?

I think there's nothing wrong if you'll have this code.

carpawell (approved):

PR #1293, opened 2022-04-05
metabase: Fix out-of-range panic in `freePotentialLocks`

cthulhu-rider:


pkg/local_object_storage/metabase/lock.go
```diff
@@ -150,7 +150,7 @@ func freePotentialLocks(tx *bbolt.Tx, idCnr cid.ID, locker oid.ID) error {
 							// exclude locker
 							keyLockers = append(keyLockers[:i], keyLockers[i+1:]...)
 
-							v, err = encodeList(append(keyLockers[:i], keyLockers[i+1:]...))
+							v, err = encodeList(keyLockers)
```
carpawell, line 153: if i understand correctly, `keyLockers[i+1]` leads to panic, right?

then we have one more two lines above, no?

fyrchik (approved):

PR #1301, opened 2022-04-06
[#1213] morph/client: Return correct NNS from first attempt

alexvanin:
Fixes error:
```
neofs-node/morph.go:55  failed to set group signer scope, continue with Global  {"error": "could not check presence in NNS contract for group.neofs: returned result stack is empty"}`
```

### DIFF
--- a/pkg/morph/client/nns.go
+++ b/pkg/morph/client/nns.go
@@ -97,6 +97,7 @@ func (c *Client) NNSHash() (util.Uint160, error) {
 		c.mtx.Lock()
 		c.nnsHash = cs.Hash
 		c.mtx.Unlock()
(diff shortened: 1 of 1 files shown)


cthulhu-rider (approved):

alexvanin:
Actually it is solved by @carpawell in https://github.com/nspcc-dev/neofs-node/pull/1170/files#diff-458b0156b8a0a4ec49f3ffea2fcf929cc9ad9b1f15140632033bf60bc27d1fccR95

I think we can merge this PR instead. Thoughts @cthulhu-rider ?

alexvanin:
Fixed in https://github.com/nspcc-dev/neofs-node/pull/1170

PR #1302, opened 2022-04-07
Allow to generate chain dumps with neofs-adm

### LINKED ISSUE #722: Allow to generate chain dumps with neofs-adm
Currently, initialization stage requires running consensus nodes. This also allows us to setup environment in an already deployed network. However, for new network setup it is faster to generate a dump locally and then deploy all consensus nodes using this dump.
This can be used using `--dump` flag for both `morph init` command and `morph generate-storage-wallet`.

Proposed implementation is quite simple:
1. Send all transactions via `Client.SendRawTransaction`.
2. Allow to replace send function with writer-to-dump.

fyrchik:
Close #722.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -49,6 +49,7 @@ require (
 	github.com/google/go-querystring v1.1.0 // indirect
 	github.com/gorilla/websocket v1.4.2 // indirect
 	github.com/hashicorp/hcl v1.0.0 // indirect
+	github.com/holiman/uint256 v1.2.0 // indirect
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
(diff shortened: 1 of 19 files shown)


cthulhu-rider:
@fyrchik pls update if still actual.

go.mod
```diff
@@ -49,6 +49,7 @@ require (
 	github.com/google/go-querystring v1.1.0 // indirect
 	github.com/gorilla/websocket v1.4.2 // indirect
 	github.com/hashicorp/hcl v1.0.0 // indirect
+	github.com/holiman/uint256 v1.2.0 // indirect
```
cthulhu-rider, line 52: It it used?

cthulhu-rider (approved):

PR #1305, opened 2022-04-08
neofs-cli: use TTY to read the password

fyrchik:

### DIFF
--- a/cmd/neofs-cli/modules/root.go
+++ b/cmd/neofs-cli/modules/root.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/mitchellh/go-homedir"
 	"github.com/nspcc-dev/neo-go/cli/flags"
-	"github.com/nspcc-dev/neo-go/cli/input"
 	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
 	"github.com/nspcc-dev/neo-go/pkg/util"
 	"github.com/nspcc-dev/neo-go/pkg/wallet"
@@ -25,6 +24,7 @@ import (
 	"github.com/nspcc-dev/neofs-sdk-go/token"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	"golang.org/x/term"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


carpawell:
What is the state of that PR?

cmd/neofs-cli/modules/root.go
```diff
@@ -222,7 +222,16 @@ func getPassword() (string, error) {
...
 	}
 
-	return input.ReadPassword("Enter password > ")
+	f, err := os.OpenFile("/dev/tty", os.O_RDWR, 0)
+	if err != nil {
+		return "", err
+	}
+	defer f.Close()
+
+	_, _ = f.WriteString("Enter password > ")
```
AnnaShaleva, line 231: Check error, just in case?

AnnaShaleva:
@fyrchik, how did you test this code?

And this won't work on Windows, is neofs-cli supposed to run on Windows?

fyrchik:
@AnnaShaleva `echo .. | ./neofs-cli >file.out` should print the prompt.

Windows is not yes supported in this PR so if you figure out how to do it there, let me know!

AnnaShaleva:
> `echo .. | ./neofs-cli >file.out` should print the prompt.

BTW, echoing password and passing it to some command doesn't work. At least, in neo-go, the following attempt leads to password prompt:
```
echo q | ./bin/neo-go contract deploy -i 1-print/1-print.nef -manifest 1-print/1-print.manifest.json -r http://seed1.neo.org:10332 -w wallet.json > file.out
Enter account Ng83zo4f6MCpf9uypCR31qGott7ociXFdT password >
```

However, sshpass works fine with this:
```
sshpass -p q ./bin/neo-go contract deploy -i 1-print/1-print.nef -manifest 1-print/1-print.manifest.json -r http://seed1.neo.org:10332 -w wallet.json
Network fee: 0.00151452
System fee: 10.00104553
Total fee: 10.00256005
Relay transaction (y|N)> n
cancelled
```

carpawell:
As I understand, windows case was [solved](https://github.com/nspcc-dev/neo-go/pull/2480) in neo-go. Could that approach be applied in that PR?

fyrchik:
We should just reuse the code from neo-go. I think the PR and the task can be closed after neo-go update.

PR #1318, opened 2022-04-19
Fix/Remove object upon tombstone

### LINKED ISSUE #1226: Node can store potentially undead objects
## Context
Object removal in NeoFS is performed using special objects of type [TOMBSTONE](https://github.com/nspcc-dev/neofs-api/blob/650c367529c6fe6b215ea8d5f183b07b4ac8f9a9/object/types.proto#L25). These objects are stored physically just like all other objects. In order to solve some issues related to object removal in distributed systems, tombstones they are initially broadcast throughout the container. Nodes from a primary placement store tombstones physically, while all others only record a note that they learned about the intention to delete the object. These removal records are written in local object storage:
https://github.com/nspcc-dev/neofs-node/blob/14ea7b2c473a7be1f249225fad9f789cbd32971f/pkg/local_object_storage/metabase/inhume.go#L72

## Overview
In the current implementation, the storage node may miss the death of the object and continue to store it locally, and later distribute it throughout the container. This could be due to two reasons:

1. the node missed the appearance of the tombstone in the container (primarily due to network problems)
2. the node has marked the object as removed (`Inhume` operation mentioned above) and this entry will never expire

While `1` can't be completely solved in distributed container, `2` is caused by the loss of expiration date in local record.

## Proposed solution
We need to extend logic of local object storage and GC.  First of all we need to track expiration dates of tombstones in local object storage:
```
func (*InhumePrm) ExpiresAfter(epoch uint64)
```
After removal records became limited in lifetime, we need to be able to process expiration dates:
```
// method to be implemented
func (Grave) ExpiresAfter() uint64

// the method's usage
func (*DB) IterateOverGraveyard(h GraveHandler) error
```
These changes will allow `GC` to process grave expiration timestamps
https://github.com/nspcc-dev/neofs-node/blob/14ea7b2c473a7be1f249225fad9f789cbd32971f/pkg/local_object_storage/shard/gc.go#L185

As a consequence, an additional improvement will be the separation of expired tombstones as objects and expired grave records.

### Other objects
[LOCK](https://github.com/nspcc-dev/neofs-api/blob/650c367529c6fe6b215ea8d5f183b07b4ac8f9a9/object/types.proto#L31) objects behave similar to `TOMBSTONE` ones (except to prevent deletion), so we need to process them almost the same. We've encountered same problems in 
* https://github.com/nspcc-dev/neofs-node/pull/1181

carpawell:
Based on #1321, will rebase after its merge.

Key changes:
- `Graveyard` was separated on `Graveyard` and `Garbage` buckets since now it is possible to have an object which has both graveyard and GC records.
- When TS is received, the object covered with that TS is marked with GC immediately.
- TS info is kept in the `Graveyard` bucket once it appeared on a node.
- Every epoch all the tombstones are checked for their presence in the NeoFS network and TSs that have not been found are deleted from the `Graveyard`.
- Strange get service [initialization](https://github.com/nspcc-dev/neofs-node/pull/1318/files#diff-c9453c121fb79a56f041a2bd8b9f327c6c898ae5ce7be9db49a714446fdbdc5bR356): a get service needs a storage engine for initialization and a storage engine needs a get service.

Closes #1226.

TODO: add more tests to the engine/shard after approving the main idea of the PR.

### DIFF
--- a/pkg/local_object_storage/metabase/delete_test.go
+++ b/pkg/local_object_storage/metabase/delete_test.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/nspcc-dev/neofs-node/pkg/core/object"
 	meta "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/metabase"
+	apistatus "github.com/nspcc-dev/neofs-sdk-go/client/status"
 	cidtest "github.com/nspcc-dev/neofs-sdk-go/container/id/test"
 	objectSDK "github.com/nspcc-dev/neofs-sdk-go/object"
 	"github.com/stretchr/testify/require"
@@ -45,9 +46,6 @@ func TestDB_Delete(t *testing.T) {
 	err = meta.Inhume(db, object.AddressOf(child), object.AddressOf(ts))
 	require.NoError(t, err)
 
-	err = meta.Inhume(db, object.AddressOf(child), object.AddressOf(ts))
-	require.NoError(t, err)
-
 	// delete object
 	err = meta.Delete(db, object.AddressOf(child))
 	require.NoError(t, err)
@@ -57,13 +55,14 @@ func TestDB_Delete(t *testing.T) {
 	require.NoError(t, err)
 	require.Len(t, l, 0)
 
-	// check if they removed from graveyard
+	// check if they marked as already removed
+
 	ok, err := meta.Exists(db, object.AddressOf(child))
-	require.NoError(t, err)
+	require.Error(t, apistatus.ObjectAlreadyRemoved{})
 	require.False(t, ok)
 
 	ok, err = meta.Exists(db, object.AddressOf(parent))
-	require.NoError(t, err)
+	require.Error(t, apistatus.ObjectAlreadyRemoved{})
 	require.False(t, ok)
 }
 
--- a/pkg/local_object_storage/metabase/exists.go
+++ b/pkg/local_object_storage/metabase/exists.go
@@ -1,7 +1,6 @@
 package meta
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
 
@@ -101,24 +100,37 @@ func (db *DB) exists(tx *bbolt.Tx, addr *addressSDK.Address) (exists bool, err e
 }
 
 // inGraveyard returns:
-//  * 0 if object is not in graveyard;
-//  * 1 if object is in graveyard with GC mark;
-//  * 2 if object is in graveyard with tombstone.
+//  * 0 if object is not marked for deletion;
+//  * 1 if object with GC mark;
+//  * 2 if object is covered with tombstone.
 func inGraveyard(tx *bbolt.Tx, addr *addressSDK.Address) uint8 {
 	graveyard := tx.Bucket(graveyardBucketName)
 	if graveyard == nil {
+		// incorrect metabase state, does not make
+		// sense to check garbage bucket
 		return 0
 	}
 
 	val := graveyard.Get(addressKey(addr))
 	if val == nil {
+		garbageBCK := tx.Bucket(garbageBucketName)
+		if garbageBCK == nil {
+			// incorrect node state
+			return 0
+		}
+
+		val = garbageBCK.Get(addressKey(addr))
+		if val != nil {
+			// object has been marked with GC
+			return 1
+		}
+
+		// neither in the graveyard
+		// nor was marked with GC mark
 		return 0
 	}
 
-	if bytes.Equal(val, []byte(inhumeGCMarkValue)) {
-		return 1
-	}
-
+	// object in the graveyard
 	return 2
 }
 
--- a/pkg/local_object_storage/metabase/graveyard.go
+++ b/pkg/local_object_storage/metabase/graveyard.go
@@ -2,10 +2,10 @@ package meta
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 
 	addressSDK "github.com/nspcc-dev/neofs-sdk-go/object/address"
-	"github.com/pkg/errors"
 	"go.etcd.io/bbolt"
 )
 
--- a/pkg/local_object_storage/metabase/graveyard.go
+++ b/pkg/local_object_storage/metabase/graveyard.go
@@ -1,7 +1,6 @@
 package meta
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
 
@@ -9,53 +8,74 @@ import (
 	"go.etcd.io/bbolt"
 )
 
-// Grave represents descriptor of DB's graveyard item.
-type Grave struct {
-	gcMark bool
-
+// DeletedObject represents descriptor of the object that was
+// marked to be deleted.
+type DeletedObject struct {
 	addr *addressSDK.Address
 }
 
-// WithGCMark returns true if grave marked for GC to be removed.
-func (g *Grave) WithGCMark() bool {
-	return g.gcMark
-}
-
 // Address returns buried object address.
-func (g *Grave) Address() *addressSDK.Address {
+func (g *DeletedObject) Address() *addressSDK.Address {
 	return g.addr
 }
 
-// GraveHandler is a Grave handling function.
-type GraveHandler func(*Grave) error
+// Handler is a DeletedObject handling function.
+type Handler func(*DeletedObject) error
+
+// IterateOverGarbage iterates over all objects
+// marked with GC mark.
+//
+// If h returns ErrInterruptIterator, nil returns immediately.
+// Returns other errors of h directly.
+func (db *DB) IterateOverGarbage(h Handler) error {
+	return db.boltDB.View(func(tx *bbolt.Tx) error {
+		return db.iterateDeletedObj(tx, withGC, h)
+	})
+}
 
 // IterateOverGraveyard iterates over all graves in DB.
 //
 // If h returns ErrInterruptIterator, nil returns immediately.
 // Returns other errors of h directly.
-func (db *DB) IterateOverGraveyard(h GraveHandler) error {
+func (db *DB) IterateOverGraveyard(h Handler) error {
 	return db.boltDB.View(func(tx *bbolt.Tx) error {
-		return db.iterateOverGraveyard(tx, h)
+		return db.iterateDeletedObj(tx, grave, h)
 	})
 }
 
-func (db *DB) iterateOverGraveyard(tx *bbolt.Tx, h GraveHandler) error {
-	// get graveyard bucket
-	bktGraveyard := tx.Bucket(graveyardBucketName)
-	if bktGraveyard == nil {
+type deletedType uint8
+
+const (
+	_ deletedType = iota
+	grave
+	withGC
+)
+
+func (db *DB) iterateDeletedObj(tx *bbolt.Tx, t deletedType, h Handler) error {
+	var bkt *bbolt.Bucket
+	switch t {
+	case grave:
+		bkt = tx.Bucket(graveyardBucketName)
+	case withGC:
+		bkt = tx.Bucket(garbageBucketName)
+	default:
+		panic(fmt.Sprintf("metabase: unknown iteration object type: %d", t))
+	}
+
+	if bkt == nil {
 		return nil
 	}
 
-	// iterate over all graves
-	err := bktGraveyard.ForEach(func(k, v []byte) error {
-		// parse Grave
-		g, err := graveFromKV(k, v)
+	// iterate over all deleted objects
+	err := bkt.ForEach(func(k, v []byte) error {
+		// parse deleted object
+		delObj, err := deletedObjectFromKV(k, v)
 		if err != nil {
 			return fmt.Errorf("could not parse Grave: %w", err)
 		}
 
-		// handler Grave
-		return h(g)
+		// handler object
+		return h(delObj)
 	})
 
 	if errors.Is(err, ErrInterruptIterator) {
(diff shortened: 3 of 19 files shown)


pkg/local_object_storage/metabase/inhume.go
```diff
@@ -73,6 +73,11 @@ var errBreakBucketForEach = errors.New("bucket ForEach break")
 // if at least one object is locked.
 func (db *DB) Inhume(prm *InhumePrm) (res *InhumeRes, err error) {
 	err = db.boltDB.Update(func(tx *bbolt.Tx) error {
+		garbageBKT, err := tx.CreateBucketIfNotExists(garbageBucketName)
```
fyrchik, line 76: Do this in `Init`?

pkg/local_object_storage/metabase/graveyard.go
```diff
@@ -229,3 +229,25 @@ func graveFromKV(k, v []byte) (TombstonedObject, error) {
 		tomb: tomb,
 	}, nil
 }
+
+// DropGraves deleted Tombstoned objects from the
```
fyrchik, line 233: `s/deleted/deletes/`

pkg/services/object_manager/tombstone/getter.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// Getter is a tombstone source wrapper.
+// It checks tombstones presence via tombstone
+// source, caches it checks its expiration.
+//
+// Must be created via NewGetter function. `var` and
+// `Getter{}` declarations leads to undefined behaviour
+// and may lead to panics.
+type Getter struct {
```
fyrchik, line 33: Why is it `Getter` and not `LRUCache` or sth similar?

pkg/services/object_manager/tombstone/getter.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// GetTombstone checks the tombstone presence in the system in the
+// following order:
+//  * 1. Local LRU cache;
+//  * 2. Tombstone source.
+//
+// If a tombstone was successfully fetched (regardless of its expiration)
+// it is cached in the LRU cache.
+func (g *Getter) GetTombstone(ctx context.Context, a *addressSDK.Address, epoch uint64) bool {
```
fyrchik, line 48: Why is it called `GetTombstone`, but returns `bool`?

pkg/local_object_storage/metabase/delete_test.go
```diff
@@ -57,13 +55,13 @@ func TestDB_Delete(t *testing.T) {
 	require.NoError(t, err)
 	require.Len(t, l, 0)
 
-	// check if they removed from graveyard
+	// check if they marked as already removed
```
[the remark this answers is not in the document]
carpawell: child and parent objects (the only tested objects in that test)

originally "they" was added not by me, added a new line after that comment (usually i do that to highlight that comment is not only for the next line)

pkg/local_object_storage/metabase/graveyard_test.go
```diff
@@ -46,21 +54,46 @@ func TestDB_IterateOverGraveyard(t *testing.T) {
...
 
-	require.Equal(t, 2, counterAll)
-	require.Equal(t, []*addressSDK.Address{object.AddressOf(obj1)}, buriedTS)
-	require.Equal(t, []*addressSDK.Address{object.AddressOf(obj2)}, buriedGC)
+	require.Equal(t, 4, counterAll)
+	require.True(t, equalAddresses([]*addressSDK.Address{object.AddressOf(obj1), object.AddressOf(obj2)}, buriedTS))
+	require.True(t, equalAddresses([]*addressSDK.Address{object.AddressOf(obj3), object.AddressOf(obj4)}, buriedGC))
+}
+
+func equalAddresses(aa1 []*addressSDK.Address, aa2 []*addressSDK.Address) bool {
```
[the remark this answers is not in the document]
carpawell: wow, thx, did not know that such exists, used it

pkg/local_object_storage/metabase/util.go
```diff
@@ -21,7 +21,14 @@ bytes. Check it later.
 const invalidBase58String = "_"
 
 var (
-	graveyardBucketName       = []byte(invalidBase58String + "Graveyard")
+	// graveyardBucketName keeps rows with the objects that have been
```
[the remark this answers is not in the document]
carpawell: you mean change it to smth like "stores"?

pkg/local_object_storage/metabase/inhume.go
carpawell: if fact it is already [called](https://github.com/nspcc-dev/neofs-node/pull/1318/files#diff-1ff37983e3b5b6c80d5873c9922f36816780fb8cfbd693b52aa539543864da52R57); changed to just `Bucket`

pkg/services/object_manager/tombstone/getter.go
carpawell: i designed `Getter` as a component that wraps _something_ that could provide info about tombstones. in the first implementation, it contained `local` and `remote` TSSource interfaces: a storage engine and a wrapper over an API client but then it has changed to reusing one object service, and only after that, i added an LRU cache. so I want it to be as abstract as possible

btw, I agree that naming is awful, just could not think of something better, will be happy to hear some suggestions from you

pkg/services/object_manager/tombstone/getter.go
carpawell: one more awful naming. `TombstoneIsAvailable`, `IsAvailable`, `TobmstoneIsExpired`?

pkg/local_object_storage/metabase/util.go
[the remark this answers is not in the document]
fyrchik: yes

pkg/services/object_manager/tombstone/getter.go
fyrchik: `IsTombstoneAvailable`, `IsTombsoneAlive` look good to me.

pkg/services/object_manager/tombstone/getter.go
fyrchik: `tombstone.HealthChecker`, `tombstone.ExpirationChecker`, `tombstone.Validator`, `tombstone.ExpirationCache`?

pkg/services/object_manager/tombstone/source/source.go
```diff
@@ -0,0 +1,89 @@
...
+// Must be created via NewSource function. `var` and `Source{}`
+// declarations leads to undefined behaviour and may lead
+// to panics.
+type Source struct {
+	s *getsvc.Service
+}
+
+var (
+	errNotFound apistatus.ObjectNotFound
+	errRemoved  apistatus.ObjectNotFound
```
cthulhu-rider, line 29:
```suggestion
	errRemoved  apistatus.ObjectAlreadyRemoved
```

pkg/services/object_manager/tombstone/source/source.go
```diff
@@ -0,0 +1,89 @@
...
+	var hr headerWriter
+
+	var headPrm getsvc.HeadPrm
+	headPrm.WithAddress(a)
+	headPrm.SetHeaderWriter(&hr)
+	headPrm.SetCommonParameters(&util.CommonPrm{}) //default values are ok for that operation
+
+	err := s.s.Head(ctx, headPrm)
+	switch {
+	case errors.As(err, &errNotFound) || errors.As(err, &errRemoved):
```
cthulhu-rider, line 77: What about multi-routine access?

pkg/services/object_manager/tombstone/getter.go
carpawell: i have chosen `tombstone.ExpirationChecker`

pkg/services/object_manager/tombstone/getter.go
carpawell: i have chosen `IsTombstoneAvailable`

pkg/local_object_storage/metabase/exists.go
```diff
@@ -101,24 +100,37 @@ func (db *DB) exists(tx *bbolt.Tx, addr *addressSDK.Address) (exists bool, err e
...
-//  * 0 if object is not in graveyard;
-//  * 1 if object is in graveyard with GC mark;
-//  * 2 if object is in graveyard with tombstone.
+//  * 0 if object is not marked for deletion;
+//  * 1 if object with GC mark;
+//  * 2 if object is covered with tombstone.
 func inGraveyard(tx *bbolt.Tx, addr *addressSDK.Address) uint8 {
 	graveyard := tx.Bucket(graveyardBucketName)
 	if graveyard == nil {
+		// incorrect node state, does not make
```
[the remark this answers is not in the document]
carpawell: i meant that node that has been inited with some shard that has a metabase without some core bucket

changed to "metabase"

pkg/local_object_storage/metabase/graveyard.go
```diff
@@ -9,70 +9,245 @@ import (
...
+// GarbageObject represents descriptor of the
+// object that has been marked with GC.
+type GarbageObject struct {
 	addr *addressSDK.Address
 }
 
-// WithGCMark returns true if grave marked for GC to be removed.
-func (g *Grave) WithGCMark() bool {
-	return g.gcMark
+// Address returns buried object address.
```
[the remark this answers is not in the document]
carpawell: in fact, took that term from the tests' comments (and started to name all GCed/ TSed objects as "buried"/"deleted" and the file is still named as "graveyard" cause all those objects gonna "die" anyway). do not have my own terminology but agree that it would be better to name objects with GC as "garbage" and objects with TS as "tombstoned" everywhere in the repo

changed

pkg/local_object_storage/metabase/graveyard.go
```diff
@@ -9,70 +9,245 @@ import (
...
+	if err != nil {
+		return fmt.Errorf("could not parse grave: %w", err)
+	}
+
+	return g.h(o)
+}
+
+func (db *DB) iterateDeletedObj(tx *bbolt.Tx, h kvHandler, offset *addressSDK.Address) error {
+	var bkt *bbolt.Bucket
+	switch t := h.(type) {
```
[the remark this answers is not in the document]
carpawell: i tried that before and found that it is redundantly (IMO): 
1. handler contains info about bucket it could handle;
2. with an explicit parameter it could be possible to pass `garbage` bucket with `graveyardHandler`. that is totally incorrect but does not lead to panic or any other explicit errors => hard to debug

also, I think that sharing iteration logic is still good but I did not find any ways to do that without a type switch (btw, is a type switch that bad? maybe I do not know smth about it)

pkg/local_object_storage/metabase/graveyard.go
```diff
@@ -9,70 +9,245 @@ import (
...
+		return TombstonedObject{}, err
+	}
+
+	return TombstonedObject{
+		addr: target,
+		tomb: tomb,
+	}, nil
+}
+
+// DropGraves deletes Tombstoned objects from the
```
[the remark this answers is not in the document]
carpawell: choose the first one

pkg/local_object_storage/shard/control.go
```diff
@@ -62,8 +62,8 @@ func (s *Shard) Init() error {
 				cancelFunc: func() {},
 				handlers: []eventHandler{
 					s.collectExpiredObjects,
-					s.collectExpiredTombstones,
 					s.collectExpiredLocks,
+					s.collectExpiredTombstones,
```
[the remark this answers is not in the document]
carpawell: no, was made by a mistake, returned the order

pkg/local_object_storage/shard/gc.go
```diff
@@ -238,17 +247,54 @@ func (s *Shard) collectExpiredObjects(ctx context.Context, e Event) {
...
+		if len(tss) == tssDeleteBatch {
+			return meta.ErrInterruptIterator
+		}
+
+		return nil
 	})
-	if err != nil || len(expired) == 0 {
+
+	for {
+		log.Debug("iterating tombstones")
```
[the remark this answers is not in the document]
carpawell: it was added explicitly not for tests

think I could help sometimes: it will be printed every iteration so at least it may say about a number of tombstones

pkg/services/object_manager/tombstone/constructor.go
```diff
@@ -0,0 +1,84 @@
...
+
+	tsSource Source
+}
+
+// Option is an option of Getter's constructor.
+type Option func(*cfg)
+
+func defaultCfg() *cfg {
+	return &cfg{
+		log:       zap.L(),
```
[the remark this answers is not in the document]
carpawell: changed but note: all other components in the repo use ` zap.L()` as a default logger (that was the only reason for me to do that the same and not with `NewNop`)

pkg/services/object_manager/tombstone/getter.go
```diff
@@ -0,0 +1,94 @@
...
+	lru "github.com/hashicorp/golang-lru"
+	objectV2 "github.com/nspcc-dev/neofs-api-go/v2/object"
+	"github.com/nspcc-dev/neofs-sdk-go/object"
+	addressSDK "github.com/nspcc-dev/neofs-sdk-go/object/address"
+	"go.uber.org/zap"
+)
+
+// Source is a tombstone source interface.
+type Source interface {
+	// Tombstone must return tombstone from the source it was
```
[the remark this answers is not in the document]
carpawell: changed and added some words to the last sentence: 
```
Tombstone MUST return (nil, nil) if requested tombstone is
missing in the storage for the provided epoch.
```

pkg/services/object_manager/tombstone/source/source.go
carpawell: did not get: what problems could appear?

carpawell: got it, messed `Is` and `As`, deleted global vars, and added creating a new one on every `Tombstone` call

pkg/local_object_storage/metabase/graveyard.go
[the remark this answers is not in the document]
cthulhu-rider: No, type-switch is okay, just mentioned that it's needed only for bucket selection.

cthulhu-rider (approved):

fyrchik (approved):

PR #1321, opened 2022-04-21
Fix/GC's event handler registration

carpawell:
Morph "NewEpoch" event handling was registered in a closure over
`addNewEpochNotificationHandler` func. That may lead to the data race:
if a shard was initialized before the event registration, everything works
as planned, but if registration was made earlier, it was not able to
include GC handlers since a shard has not called `eventChanInit` yet and,
therefore, it has not registered handler yet.

Also, we may rename the option and just pass a channel instead of func. Also, we may call `eventChanInit` on the shard's creation, not on its initialization.


alexvanin (approved):

cthulhu-rider (approved):
Agree with passing the channel.

fyrchik (approved):

cthulhu-rider (approved):

Issue #1322, opened 2022-04-22
Object Tree Service


fyrchik:
# Description
Object tree service operates with arbitrary trees attached to a container. The implementation maintains a CRDT tree for each container the node is in, based on this article https://martin.klepmann.com/papers/move-op.pdf

Usecase 1: allow to alias objects by path `/path/to/file.txt` and move them without creating new objects.
Usecase 2: support versioning

# TODO
## Functional
- [x] Fully-functional implementation of non-synchronization methods (#1324)
- [x] #1326
- [x] #1328
- [x] #1329
- [ ] #1330
- [ ] #1331
- [ ] #1332
- [x] #1333
- [ ] #1334
- [x] #1401
- [x] #1442
- [ ] #1448
- [x] #1505 
- [x] #1628
- [x] #1630
- [ ] #1530
- [x] #1768

## Optimizations
- [x] #1427
- [ ] #1327
- [x] #1445
- [x] #1446
- [ ] #1447 
- [x] #1456

roman-khimov:
Tree service-related issues are to be blocked because we'll be rethinking meta.

roman-khimov:
Trees will be removed, this issue will never be done.

PR #1324, opened 2022-04-22
Initial Tree service implementation

fyrchik:
Can already be tested in devenv by providing `NEOFS_TREE_GRPC_ENDPOINT=s01.${LOCAL_DOMAIN}:8082` similar to how control service endpoint is provided.

Related #1322 .


pkg/services/tree/service.go
```diff
@@ -0,0 +1,273 @@
...
+	b := req.GetBody()
+	cid := getCID(b.GetContainerId())
+
+	nodes, err := s.forest.TreeGetByPath(cid, b.GetTreeId(), sawmill.AttributeFilename,
+		strings.Split(b.GetPath(), "/"), b.GetLatestOnly())
+	if err != nil {
+		return nil, err
+	}
+
+	info := make([]*GetNodePathResponse_Info, len(nodes))
```
alexvanin, line 174:
```suggestion
	info := make([]*GetNodePathResponse_Info, 0, len(nodes))
```

pkg/services/tree/service.go
```diff
@@ -0,0 +1,273 @@
...
+			Nodes: info,
+		},
+	}, nil
+}
+
+func (s *Service) GetSubTree(_ context.Context, req *GetSubTreeRequest) (*GetSubTreeResponse, error) {
+	b := req.GetBody()
+	cid := getCID(b.GetContainerId())
+
+	info := make([]*GetSubTreeResponse_Info, len(b.GetNodes()))
```
alexvanin, line 207:
```suggestion
	info := make([]*GetSubTreeResponse_Info, 0, len(b.GetNodes()))
```

pkg/local_object_storage/sawmill/boltdb.go
```diff
@@ -0,0 +1,521 @@
+package sawmill
+
+import (
+	"bytes"
+	"encoding/binary"
+	"math/rand"
```
alexvanin, line 6: I think it is a wrong `rand` to use.

pkg/local_object_storage/sawmill/forest.go
```diff
@@ -0,0 +1,109 @@
+package sawmill
+
+import (
+	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
+	cidSDK "github.com/nspcc-dev/neofs-sdk-go/container/id"
```
alexvanin, line 5: Two same imports there :smiley:

fyrchik: Wow, it is unexpected `Go` doesn't raise a compile error here, actually.

pkg/services/tree/service.go
fyrchik: Fixed

pkg/services/tree/service.go
fyrchik: Fixed

pkg/local_object_storage/sawmill/forest.go
fyrchik: Fixed.

pkg/local_object_storage/sawmill/boltdb.go
fyrchik: Actually, no. We don't really care about the predictability and just want to find a random spare ID.

alexvanin: Oh, I see. For me the rule of thumb is to avoid `math/rand` unless I do tests, models or local private computations. But you are right, technically it should be okay there.

cmd/neofs-node/config/tree/config.go
```diff
@@ -0,0 +1,39 @@
...
+const (
+	subsection     = "tree"
+	grpcSubsection = "grpc"
+
+	// GRPCEndpointDefault is a default endpoint of gRPC Control service.
+	GRPCEndpointDefault = ""
+)
+
+// GRPC returns structure that provides access to "grpc" subsection of
+// "control" section.
```
cthulhu-rider, line 22:
```suggestion
// "tree" section.
```

cmd/neofs-node/tree.go
```diff
@@ -0,0 +1,41 @@
...
+		tree.WithStorage(c.cfgObject.cfgLocalStorage.localStorage))
+
+	treeServer := grpc.NewServer()
+
+	c.onShutdown(func() {
+		stopGRPC("NeoFS Tree Service API", treeServer, c.log)
+		treeSvc.Shutdown()
+	})
+
+	lis, err := net.Listen("tcp", endpoint)
```
cthulhu-rider, line 32: We can also add `treeServer` and `lis` to grpc group (see `cfgGRPC` type and `serverGRPC` func).

pkg/local_object_storage/engine/tree.go
```diff
@@ -0,0 +1,78 @@
...
+	cidSDK "github.com/nspcc-dev/neofs-sdk-go/container/id"
+)
+
+var _ pilorama.Forest = (*StorageEngine)(nil)
+
+// TreeMove implements the pilorama.Forest interface.
+func (e *StorageEngine) TreeMove(cid *cidSDK.ID, treeID string, m *pilorama.Move) (*pilorama.LogMove, error) {
+	var err error
+	var lm *pilorama.LogMove
+	for _, sh := range e.sortShardsByWeight(cid) {
```
cthulhu-rider, line 14: We can also use `iterateOverSortedShards` here.

pkg/services/tree/types.proto
```diff
@@ -0,0 +1,29 @@
...
+  bytes meta = 3 [json_name = "meta"];
+  // ID of the node to move.
+  uint64 child_id = 4 [json_name = "childID"];
+}
+
+// Signature of a message.
+message Signature {
+  bytes key = 1 [json_name = "key"];
+  bytes sign = 2 [json_name = "signature"];
+}
```
cthulhu-rider, line 29: \n

pkg/local_object_storage/pilorama/interface.go
```diff
@@ -0,0 +1,35 @@
+package pilorama
+
+import cidSDK "github.com/nspcc-dev/neofs-sdk-go/container/id"
+
+// Forest represents CRDT tree.
+type Forest interface {
```
cthulhu-rider, line 6: All the methods have `Tree` name prefix, can we do without it?

cthulhu-rider (commented):
More detailed documentation in `pkg/services/tree/service.proto` would be extremely helpful.

alexvanin (approved):
LGTM for initial developement

pkg/services/tree/types.proto
fyrchik: Fixed.

pkg/local_object_storage/engine/tree.go
fyrchik: It needs an address, we only have CID, and there are no gains in terms of code size, so I decided to write a for loop

cmd/neofs-node/tree.go
fyrchik: I have copied the code for the control service. If it needs to be changed as well, let's do it in a separate task?

fyrchik:
@cthulhu-rider we have a separate task for the documentation #1334 . The reason is that the service is under active development and the interface could be changed.

pkg/local_object_storage/pilorama/interface.go
fyrchik: We can, but it would be confusing on `StorageEngine` IMO. We already have `metabase.MoveIt` and it is a completely different thing.
Another option is to have `(*StorageEngine).Pilorama()` which returns a new type implementing methods without prefix.

cmd/neofs-node/tree.go
cthulhu-rider: Agree, smth like `Unify lifetime process of gRPC servers`.

pkg/local_object_storage/engine/tree.go
cthulhu-rider: Actually we can change parameter type to `address.Address` -> `fmt.Stringer`. But I don't mind to leave for-loop.

cthulhu-rider (approved):

pkg/local_object_storage/pilorama/types.go
```diff
@@ -0,0 +1,52 @@
...
+
+const (
+	// RootID represents the ID of a root node.
+	RootID = 0
+	// TrashID is a parent for all removed nodes.
+	TrashID = math.MaxUint64
+)
+
+// ErrNotFound is returned when node id is not found or there is no nodes corresponding to the filepath.
+var ErrNotFound = errors.New("object not found")
```
KirillovDenis, line 52: It would be nice to have a way to know such error on client side without `strings.Contains`

KirillovDenis (approved):

pkg/local_object_storage/pilorama/types.go
fyrchik: It is more of a gRPC issue, the error is transmitted by network, I doubt it is possible to have something more convenient.

Also, this error is actually for missing _tree_, changed the error message.

PR #1344, opened 2022-04-29
services/tree: implement GetSubTree RPC

### LINKED ISSUE #1326: Implement `GetSubTree` for fetching serialized subtree
Implement `GetSubTree` RPC:
1. Describe serialization format. Note that gRPC server-side stream is, probably inevitable, so we can divide the tree in chunks. As an example: send a list of (child-parent) relationships. Ensure that it is easy for a client to do some processing before it receives the full tree.
2. Implement the RPC itself (input: node ID + depth, output: tree)

We probably need to support both getting by ID and by path.

fyrchik:
Close #1326 .


pkg/local_object_storage/pilorama/interface.go
```diff
@@ -20,6 +20,8 @@ type Forest interface {
 	TreeGetByPath(cid *cidSDK.ID, treeID string, attr string, path []string, latest bool) ([]Node, error)
 	// TreeGetMeta returns meta information of the node with the specified ID.
 	TreeGetMeta(cid *cidSDK.ID, treeID string, nodeID Node) (Meta, error)
+	// TreeGetChildren returns children of the node with the specified ID. The order is arbitrary.
```
cthulhu-rider, line 23: According to implementations:

- Returns `ErrTreeNotFound` if tree is missing.
- Returns both `nil` if the given node has no children.

I suggest to clarify this in interface method's docs.

pkg/services/tree/service.go
```diff
@@ -22,6 +22,9 @@ type Service struct {
 	closeCh     chan struct{}
 }
 
+// MaxGetSubTreeDepth represents maximum allowed traversal depth in GetSubTree RPC.
+const MaxGetSubTreeDepth = 10
```
cthulhu-rider, line 26: Don't forget to reflect this in protocol docs.

pkg/services/tree/service.go
```diff
@@ -266,11 +310,22 @@ func getCID(rawCID []byte) *cidSDK.ID {
 	return cidSDK.NewFromV2(&cidV2)
 }
 
-func constructMeta(arr []*KeyValue) []pilorama.KeyValue {
+func protoToMeta(arr []*KeyValue) []pilorama.KeyValue {
```
cthulhu-rider, line 313: Can `arr` contain `nil`?

pkg/local_object_storage/engine/tree.go
```diff
@@ -76,3 +76,17 @@ func (e *StorageEngine) TreeGetMeta(cid *cidSDK.ID, treeID string, nodeID pilora
...
 }
+
+// TreeGetChildren implements the pilorama.Forest interface.
+func (e *StorageEngine) TreeGetChildren(cid *cidSDK.ID, treeID string, nodeID pilorama.Node) ([]uint64, error) {
+	var err error
+	var nodes []uint64
+	for _, sh := range e.sortShardsByWeight(cid) {
+		nodes, err = sh.TreeGetChildren(cid, treeID, nodeID)
+		if err != nil {
+			continue
```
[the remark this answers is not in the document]
fyrchik: Fixed.

pkg/local_object_storage/pilorama/boltdb.go
```diff
@@ -416,6 +416,31 @@ func (t *boltForest) TreeGetMeta(cid *cidSDK.ID, treeID string, nodeID Node) (Me
...
+	var children []uint64
+
+	err := t.db.View(func(tx *bbolt.Tx) error {
+		treeRoot := tx.Bucket(bucketName(cid, treeID))
+		if treeRoot == nil {
+			return ErrTreeNotFound
+		}
+
+		b := treeRoot.Bucket(dataBucket)
+		c := b.Cursor()
```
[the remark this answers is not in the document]
fyrchik: No, root bucket is created together with it's 2 children.

pkg/local_object_storage/pilorama/interface.go
fyrchik: Fixed.

pkg/services/tree/service.go
fyrchik: `proto.Unmarshal` returns slice with empty structs, but added a check, just in case.

pkg/services/tree/service.go
fyrchik: Fixed.

cthulhu-rider (approved):

Issue #1300, opened 2022-04-06
Add an option to disable homomorhic hashing in storage nodes


alexvanin:
There is a need to toggle calculation of objects' payload homomorphic hash in NeoFS environment. Disabling can speed up object formation.

### API
- [x]  add `bool homomorphic_hashing_disabled ` to [Container](https://github.com/nspcc-dev/neofs-api/blob/650c367529c6fe6b215ea8d5f183b07b4ac8f9a9/container/types.proto#L15) message of the NeoFS API V2 protocol
- [x] create constant key to the corresponding value in network configuration stored in Morph chain (can it be defined in contract repo and imported to SDK from its packages?)
- [x] add function to modify container according to global network configuration
`func (*Container) ApplyNetworkConfig(netmap.NetworkConfig)`
- [x] add function to assert network configuration for the container
`func (Container) AssertNetworkConfig(netmap.NetworkConfig) bool`
- [x] add function to sync container with network state using API client
`func SyncContainerWithNetwork(context.Context, *container.Container, *Client) error`
P.S. to begin with, I propose leaving `Client.ContainerPut`  to accept prepared `Container` structure, and perform all preparation stages outside.

#### Container approval
- [x] make [checkPutContainer](https://github.com/nspcc-dev/neofs-node/blob/1fc9351f4feb83ba90fca247fd8ac77865caf5a2/pkg/innerring/processors/container/process_container.go#L65) method of Container processor to:
1. receive network config using Morph client
2. assert network configuration for the stored container using `AssertNetworkConfig`
3. return an error on false return

####  Data audit
- [x] make data audit [routine](https://github.com/nspcc-dev/neofs-node/blob/606dfa341437795b56a46bb6234b3a017cc5fff8/pkg/services/audit/auditor/por.go#L38) to not validate homomorphic hash of the objects if `homomorphic_hashing_disabled` flag is set in the container of the storage group under audit (maybe `Context` is the best place to store this info).

### Storage node
- [x] make [payloadSizeLimiter](https://github.com/nspcc-dev/neofs-node/blob/1fc9351f4feb83ba90fca247fd8ac77865caf5a2/pkg/services/object_manager/transformer/transformer.go#L131) to calculate homomorphic hash of the payload if `homomorphic_hashing_disabled` is set
[...]

PR #1349, opened 2022-04-29
Feat/Disable homomorhic hashing for container/IR

carpawell:
Related to #1300, #1350, #1348.

Valid only after https://github.com/nspcc-dev/neofs-node/pull/1340.


carpawell:

PR #1353, opened 2022-05-04
cli: Fix connection scheme loss during endpoint parsing

### LINKED ISSUE #1341: Need more info on `failed to receive server preface within timeout`
I try to create a container in neofs-dev-env via neofs-cli and get an error
```
rpc error: write request: rpc error: code = Unavailable desc = failed to receive server preface within timeout
```
When I read the logs I see no error messages in IR, SN and morph node. I'd like to get more information on the error cause to be able to solve it.

I'd like to see the error forwarded through all the engaged component with additional information added on every step. If it's techically possible, I'd like to get a message like "I aborted the request due to timeout, because an X component is delayed to persist some record" from CLI.

## Additional context
NeoFS: 0.28.0-rc.3
NeoGO: v0.98.2
dev-env: 19a52fd0575d1af4a4c87ceaccca2168cf71b66a

### LINKED ISSUE #1351: connection closed before server preface received
I got unexpected output when I use `neofs-cli`
```
# ./neofs-cli-amd64 accounting balance -r grpcs://st2.storage.fs.neo.org:8082 -w my-wallet-path/wallet.json
Enter password > 
rpc error: write request: rpc error: code = Unavailable desc = connection closed before server preface received
```

## Expected Behavior

Can communicate correctly with storage node
## Current Behavior

rpc error: write request: rpc error: code = Unavailable desc = connection closed before server preface received

## Your Environment
```
# ./neofs-cli-amd64 --version
Version: v0.28.0 
Build: 2022-04-29T16:17:25+00:00 
Debug: false
```

cthulhu-rider:
* closes #1351 
* relates #1341
* relates #1354

### DIFF
--- a/cmd/neofs-cli/internal/client/sdk.go
+++ b/cmd/neofs-cli/internal/client/sdk.go
@@ -2,7 +2,6 @@ package internal
 
 import (
 	"crypto/ecdsa"
-	"crypto/tls"
 	"fmt"
 
 	"github.com/nspcc-dev/neofs-node/pkg/network"
@@ -19,11 +18,7 @@ func GetSDKClient(key *ecdsa.PrivateKey, addr network.Address) (*client.Client, 
 
 	prmInit.SetDefaultPrivateKey(*key)
 	prmInit.ResolveNeoFSFailures()
-	prmDial.SetServerURI(addr.HostAddr())
-
-	if addr.TLSEnabled() {
-		prmDial.SetTLSConfig(&tls.Config{})
-	}
+	prmDial.SetServerURI(addr.URIAddr())
 
 	c.Init(prmInit)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ Changelog for NeoFS Node
 
 ## [Unreleased]
 
+### Fixed
+- Loss of the connection scheme during address parsing in NeoFS CLI (#1351)
+
 ## [0.28.0] - 2022-04-29 - Heuksando (흑산도, 黑山島)
 
 ### Added
--- a/cmd/neofs-cli/modules/control.go
+++ b/cmd/neofs-cli/modules/control.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"crypto/ecdsa"
-	"crypto/tls"
 	"fmt"
 
 	"github.com/mr-tron/base58"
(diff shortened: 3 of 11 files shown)


cmd/neofs-cli/internal/client/sdk.go
```diff
@@ -9,8 +9,31 @@ import (
...
 )
 
+// NetAddr groups information about network endpoint needed for client
+// connection.
+type NetAddr struct {
+	tlsEnabled bool
+
+	uri string
+}
+
```
fyrchik, line 19: What about having `HostAddrWithScheme` or similar on `network.Address`. Maybe even a function.

cthulhu-rider: I suggest to postpone the proposed refactor since current PR is a hotfix. See https://github.com/nspcc-dev/neofs-node/issues/1354.

alexvanin (commented):
Probably the issue is a bit deeper, because we have `SetServerURI` at https://github.com/nspcc-dev/neofs-node/blob/3a44010180d9ab4554c49b4a506b952a8d7ffd39/pkg/network/cache/multi.go#L41

I guess the bug is right there at `api-go` client: 
https://github.com/nspcc-dev/neofs-api-go/blob/870c1ffc0a9f4166ade31b293eef994a515887ab/rpc/client/options.go#L73

We should not ignore `tlsCfg` param if it was passed to the function. We should ignore `isTLS` flag in this case.

cthulhu-rider:
@alexvanin in current implementation insecure scheme overlaps TLS config even if it is explicitly provided. U mean that client should open secure connection if TLS config is set regardless of URI scheme?

alexvanin:
> in current implementation insecure scheme overlaps TLS config even if it is explicitly provided. U mean that client should open secure connection if TLS config is set regardless of URI scheme?

I am wrong about this, agree. Then https://github.com/nspcc-dev/neofs-node/pull/1353#discussion_r864794683 should be good enough. And we can drop `prmDial.SetTLSConfig` as well, because it will be set by api-go anyway.

alexvanin (approved):
Seems good, but it won't fix storage node communication as it seems to me. To fix it completely we need to return host with scheme address from `network.Address` and pass it to the client, see https://github.com/nspcc-dev/neofs-node/issues/1354#issuecomment-1117313390

cthulhu-rider:
@alexvanin 629150ea44bbb71455e2ea611fdcb7ec4e206702

cmd/neofs-cli/internal/client/sdk.go
cthulhu-rider: That was also needed to fix similar problem in Storage node app, so done in 629150ea44bbb71455e2ea611fdcb7ec4e206702.

alexvanin (approved):

fyrchik (approved):

PR #1285, opened 2022-04-01
*: Use protobuf plugin to generate stable marshalers

fyrchik:
Depends on nspcc-dev/neofs-api-go#389

Signed-off-by: Evgenii Stratonikov <evgeniy@nspcc.ru>

PR #1362, opened 2022-05-05
Implement access control in tree service

### LINKED ISSUE #1328: Implement access control in tree service
Authentication requirements:
- client API methods should be invoked by the container owner and/or with a bearer token signed by the owner. 
- future synchronization API methods should be called by storage nodes from the container (or from the previous epoch ?)

Verify and describe this in package doc-comment and implement.

fyrchik:
Close #1328 .

I used the generator from master with minor corrections. The `Makefile` will be changed after full API update in #1285.


pkg/services/tree/service.go
```diff
@@ -122,7 +123,7 @@ func (s *Service) Remove(_ context.Context, req *RemoveRequest) (*RemoveResponse
 	b := req.GetBody()
 	cid := getCID(b.GetContainerId())
 
-	err := s.verifyClient(req, cid, req.GetSignature().GetKey())
+	err := s.verifyClient(req, cid, b.GetBearerToken(), eacl.OperationPut)
```
cthulhu-rider, line 126: Are we going to define dedicated operations for Tree service?

cthulhu-rider (approved):

pkg/services/tree/service.go
fyrchik: We could, but there is no necessity currently. It is more of an auxilliary service.

pkg/services/tree/signature.go
```diff
@@ -26,15 +36,51 @@ func (s *Service) verifyClient(req interface{}, cid *cidSDK.ID, rawKey []byte) e
...
+	if bCid := bearer.EACLTable().CID(); bCid != nil && !bCid.Equal(cid) {
+		return errors.New("bearer token is created for another container")
+	}
+	if err := bearer.VerifySignature(); err != nil {
+		return fmt.Errorf("invalid bearer token signature: %w", err)
+	}
+	action := eacl.NewValidator().CalculateAction(new(eacl.ValidationUnit).
+		WithEACLTable(bearer.EACLTable()).
+		WithContainerID(cid).
+		WithRole(eacl.RoleOthers).
```
carpawell, line 70: why the role is `RoleOthers` and is fixed?

pkg/services/tree/service.go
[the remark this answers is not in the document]
fyrchik: For simplicity. `Remove` is actually just a specific `Move` from the POV of a tree, so you can just use `Move` to remove something from a tree.

pkg/services/tree/signature.go
fyrchik: The last rule in the bearer EACL should be `deny all`, otherwise `CalculateAction` returns true.
First way is to use `RoleOthers` for this purpose/
Second way is to extend `CalculateAction` with the default value or use a second boolean return signifying that `CalculateAction` returned from some rule and not by default. I think the second variant is worth implementing, thoughts? @cthulhu-rider @alexvanin

cthulhu-rider: 2nd option LGTM.

cthulhu-rider: @fyrchik lets link with https://github.com/nspcc-dev/neofs-sdk-go/issues/243?

fyrchik: Added

carpawell (approved):

PR #1365, opened 2022-05-05
Feat/optional homomorphic hashing

carpawell:
Related to #1300.

### DIFF
--- a/cmd/neofs-adm/internal/modules/config/config.go
+++ b/cmd/neofs-adm/internal/modules/config/config.go
@@ -14,17 +14,18 @@ import (
 )
 
 type configTemplate struct {
-	Endpoint          string
-	AlphabetDir       string
-	MaxObjectSize     int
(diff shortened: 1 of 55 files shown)


carpawell:
@fyrchik, please, check https://github.com/nspcc-dev/neofs-node/pull/1365/commits/d892d048cb901e63db260f6311069b3c9f3ecbc8 changes in `neofs-adm`

cthulhu-rider (approved):

carpawell:
@cthulhu-rider, @fyrchik added https://github.com/nspcc-dev/neofs-node/pull/1365/commits/543d8d983ce8edc8041bf015007fd3057b561dde that makes it possible for "new" nodes to work in "old" chains that do not have any info about the new holomorphic key in the config.

Added a default value only for `HomomorphicHashDisabled`: it seems that only that method is optional in cfg for now.

PR #1370, opened 2022-05-11
Feat/Set config via `adm`

carpawell:
Usage:

```
▶ ./bin/neofs-adm morph dump-config -r http://192.168.130.90:30333                                                                             
AuditFee:                    10000 (int)
BasicIncomeRate:             100000000 (int)
ContainerAliasFee:           500 (int)
ContainerFee:                1000 (int)
EigenTrustAlpha:             0.1 (str)
EigenTrustIterations:        4 (int)
EpochDuration:               240 (int)
HomomorphicHashingDisabled:  true (bool)
InnerRingCandidateFee:       10000000000 (int)
MaxObjectSize:               67108864 (int)
WithdrawFee:                 100000000 (int)
```

```
▶ ./bin/neofs-adm morph set-config -c config.yaml HomomorphicHashingDisabled=false AuditFee=1001 EigenTrustAlpha=0.9 test=test
Password for az wallet > 
Password for contract wallet > 
Waiting for transactions to persist...
```

```
▶ ./bin/neofs-adm morph dump-config -r http://192.168.130.90:30333                                                                                      
AuditFee:                    1001 (int)
BasicIncomeRate:             100000000 (int)
ContainerAliasFee:           500 (int)
ContainerFee:                1000 (int)
EigenTrustAlpha:             0.9 (str)
EigenTrustIterations:        4 (int)
EpochDuration:               240 (int)
HomomorphicHashingDisabled:  false (bool)
InnerRingCandidateFee:       10000000000 (int)
MaxObjectSize:               67108864 (int)
SystemDNS:                   636f6e7461696e6572 (hex)
WithdrawFee:                 100000000 (int)
test:                        74657374 (hex)
```

### DIFF
--- a/cmd/neofs-adm/internal/modules/morph/config.go
+++ b/cmd/neofs-adm/internal/modules/morph/config.go
@@ -6,6 +6,8 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"strconv"
+	"strings"
 	"text/tabwriter"
 
 	"github.com/nspcc-dev/neo-go/pkg/core/transaction"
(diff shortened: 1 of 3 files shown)


cmd/neofs-adm/internal/modules/morph/config.go
```diff
@@ -223,3 +224,84 @@ func dumpNetworkConfig(cmd *cobra.Command, _ []string) error {
...
+
+	k := viper.GetString(configKeyFlag)
+	if k == "" {
+		return errors.New("empty config key")
+	}
+
+	v := viper.GetString(configValueFlag)
+	if v == "" {
+		return fmt.Errorf("empty config value for %s key", k)
+	}
```
fyrchik, line 257: Why did you choose to use `-k` and `-v`?
We already have a similar thing in `set-policy`, where we use a list of `key=value` pairs. This allows us to change multiple parameters in a single transaction.

cmd/neofs-adm/internal/modules/morph/root.go
```diff
@@ -109,6 +109,18 @@ var (
 		RunE: forceNewEpochCmd,
 	}
 
+	setConfig = &cobra.Command{
+		Use:   "set-config",
```
fyrchik, line 113: `set-config` is rather generic (we already have configs here), what about `set-network-config`?

carpawell: > rather generic
 
agree but i did new command as an opposite command to the `dump-config`. what do you think if there are two commands that work with the same config but names as `dump-config` and `set-network-config`? should it be renamed too?

fyrchik: Yeah, I agree. Let's leave it `set-config` , given that there were no complaints for `dump-config`.

cmd/neofs-adm/internal/modules/morph/config.go
```diff
@@ -223,3 +225,85 @@ func dumpNetworkConfig(cmd *cobra.Command, _ []string) error {
...
+	// In NeoFS this is done via Notary contract. Here, however, we can form the
+	// transaction locally.
+	bw := io.NewBufBinWriter()
+	for _, arg := range args {
+		k, v, err := parseConfigValue(arg)
+		if err != nil {
+			return err
+		}
+
+		emit.AppCall(bw.BinWriter, nmHash, "setConfig", callflag.All, nil, k, v)
```
fyrchik, line 258: Could you add the comment about the `nil`? Probably the comment above can be moved here.

Issue #1329, opened 2022-04-25
Synchronize tree log storage when a node first appears in container


fyrchik:
This is an easy synchronization case: node has nothing and needs to receive the most actual tree snapshot. Note that this can be done in parallel with accepting new replicated operations.

fyrchik:
Partially done in #1373 .

All that is left is to fetch list of containers and trees and execute `Synchronize`.

PR #1373, opened 2022-05-12
Allow to synchronize tree log

fyrchik:
Related #1329.
Depends on #1362 .

### DIFF
--- a/pkg/local_object_storage/engine/tree.go
+++ b/pkg/local_object_storage/engine/tree.go
@@ -116,3 +116,22 @@ func (e *StorageEngine) TreeGetChildren(cid *cidSDK.ID, treeID string, nodeID pi
 	}
 	return nil, err
 }
+
+// TreeGetOpLog implements the pilorama.Forest interface.
+func (e *StorageEngine) TreeGetOpLog(cid *cidSDK.ID, treeID string, height uint64) (pilorama.Move, error) {
+	var err error
+	var lm pilorama.Move
+	for _, sh := range e.sortShardsByWeight(cid) {
+		lm, err = sh.TreeGetOpLog(cid, treeID, height)
+		if err != nil {
+			e.log.Debug("can't perform `GetOpLog`",
+				zap.Stringer("cid", cid),
+				zap.String("tree", treeID),
+				zap.Uint64("height", height),
+				zap.String("err", err.Error()))
+			continue
+		}
+		return lm, nil
+	}
+	return lm, err
+}
--- a/pkg/services/tree/service.go
+++ b/pkg/services/tree/service.go
@@ -319,6 +319,34 @@ func (s *Service) Apply(_ context.Context, req *ApplyRequest) (*ApplyResponse, e
 	})
 }
 
+func (s *Service) GetOpLog(req *GetOpLogRequest, srv TreeService_GetOpLogServer) error {
+	b := req.GetBody()
+	cid := getCID(b.GetContainerId())
+
+	h := b.GetHeight()
+	for {
+		lm, err := s.forest.TreeGetOpLog(cid, b.GetTreeId(), h)
+		if err != nil || lm.Time == 0 {
+			return err
+		}
+
+		err = srv.Send(&GetOpLogResponse{
+			Body: &GetOpLogResponse_Body{
+				Operation: &LogMove{
+					ParentId: lm.Parent,
+					Meta:     lm.Meta.Bytes(),
+					ChildId:  lm.Child,
+				},
+			},
+		})
+		if err != nil {
+			return err
+		}
+
+		h = lm.Time + 1
+	}
+}
+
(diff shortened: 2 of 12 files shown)


pkg/services/tree/service.go
```diff
@@ -319,6 +319,34 @@ func (s *Service) Apply(_ context.Context, req *ApplyRequest) (*ApplyResponse, e
...
 }
 
+func (s *Service) GetOpLog(req *GetOpLogRequest, srv TreeService_GetOpLogServer) error {
+	b := req.GetBody()
+	cid := getCID(b.GetContainerId())
+
+	h := b.GetHeight()
+	for {
+		lm, err := s.forest.TreeGetOpLog(cid, b.GetTreeId(), h)
+		if err != nil || lm.Time == 0 {
```
cthulhu-rider, line 329: Here I guess we expect `forest.TreeGetOpLog` to return `pilorama.Move{}, nil` if log op is not found. I only understood it from the implementation, so I suggest to clarify this return in interface method docs.

pkg/services/tree/sync.go
```diff
@@ -0,0 +1,98 @@
...
+		n.IterateAddresses(func(addr string) bool {
+			var a network.Address
+			if err := a.FromString(addr); err != nil {
+				return false
+			}
+
+			cc, err := grpc.DialContext(ctx, a.HostAddr(), grpc.WithInsecure())
+			if err != nil {
+				// Failed to connect, try the next address.
+				return false
```
cthulhu-rider, line 38: If I correctly understand we don't return an error if all addresses are "problematic", is it expected?

fyrchik: Yes, we query every node from the container, some of them can be unavailable, do our best here.

pkg/services/tree/service.go
fyrchik: Fixed.

pkg/local_object_storage/engine/tree.go
```diff
@@ -115,3 +115,21 @@ func (e *StorageEngine) TreeGetChildren(cid *cidSDK.ID, treeID string, nodeID pi
...
 }
+
+// TreeGetOpLog implements the pilorama.Forest interface.
+func (e *StorageEngine) TreeGetOpLog(cid *cidSDK.ID, treeID string, height uint64) (pilorama.Move, error) {
+	var err error
+	var lm pilorama.Move
+	for _, sh := range e.sortShardsByWeight(cid) {
+		lm, err = sh.TreeGetOpLog(cid, treeID, height)
+		if err != nil {
+			e.log.Debug("can't put node in a tree",
```
carpawell, line 126: wrong err message?

also, i think `height` may be added to the message

pkg/services/tree/sync.go
```diff
@@ -0,0 +1,98 @@
...
+	"io"
+
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/pilorama"
+	"github.com/nspcc-dev/neofs-node/pkg/network"
+	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
+	"google.golang.org/grpc"
+)
+
+// Synchronize tries to synchronize log starting from the last stored height.
+func (s *Service) Synchronize(ctx context.Context, cid *cid.ID, treeID string) error {
```
carpawell, line 16: isnt used? how was it planned to work?

pkg/services/tree/sync.go
```diff
@@ -0,0 +1,98 @@
...
+	nodes, err := s.getContainerNodes(cid)
+	if err != nil {
+		return err
```
carpawell, line 19: wrap with some additional info to help to understand that exactly went wrong?

pkg/services/tree/sync.go
carpawell: oh, found usage in another PR, ok

pkg/services/tree/sync.go
fyrchik: Fixed.

pkg/local_object_storage/engine/tree.go
fyrchik: Fixed

pkg/services/tree/sync.go
```diff
@@ -0,0 +1,98 @@
...
+				return 0, err
+			}
+			if m.Time > height {
+				height = m.Time + 1
+			} else {
+				height += 1
+			}
+		}
+		if !errors.Is(err, io.EOF) {
+			return 0, err
```
cthulhu-rider, line 94: Here and above, by returning zero, we seem to discard all potential progress. Is it possible to continue from recent `height` value in case of a break?

fyrchik: Yes, fixed.

cthulhu-rider (approved):

carpawell (approved):

PR #1375, opened 2022-05-12
neofs-adm: do not update NNS group if the key is the same

fyrchik:
If the group key is already set, do not send any transactions.


carpawell (approved):

cthulhu-rider (approved):
@fyrchik CHANGELOG, pls

PR #1392, opened 2022-05-17
Handle errors during header gathering for eACL check

### LINKED ISSUE #1386: object: Handle errors during header gathering for eACL check
After https://github.com/nspcc-dev/neofs-node/pull/1377 `headerSource` implementation of `eacl.TypedHeaderSource` ignores errors of protocol conversions (places are tagged with current issue). We need to process all these errors before `eACL` check initialization.

fyrchik:
Close #1386.

### DIFF
--- a/pkg/services/object/acl/eacl/v2/headers.go
+++ b/pkg/services/object/acl/eacl/v2/headers.go
@@ -205,14 +205,14 @@ func (h *cfg) localObjectHeaders(addr *objectSDKAddress.Address) ([]eaclSDK.Head
 func cidHeader(idCnr cid.ID) sysObjHdr {
 	return sysObjHdr{
 		k: acl.FilterObjectContainerID,
-		v: idCnr.String(),
+		v: idCnr.EncodeToString(),
 	}
 }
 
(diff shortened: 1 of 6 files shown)


pkg/services/object/acl/eacl/v2/headers.go
```diff
@@ -179,44 +186,51 @@ func (h *headerSource) objectHeaders() ([]eaclSDK.Header, bool) {
...
+	return headersFromObject(obj, addr), nil
 }
 
-func cidHeader(idCnr *cid.ID) eaclSDK.Header {
-	return &sysObjHdr{
+func cidHeader(idCnr cid.ID) sysObjHdr {
+	return sysObjHdr{
 		k: acl.FilterObjectContainerID,
-		v: cidValue(idCnr),
+		v: idCnr.String(),
```
cthulhu-rider, line 208: There was an error in `cidValue` (and other ID funcs): we need to use `EncodeToString` since `String` is unstable. Lets fix it in this PR?

fyrchik: Fixed.

cthulhu-rider (approved):

PR #1405, opened 2022-05-20
Fix/Settlement log message

### LINKED ISSUE #1374: Wrong log messages during settlement transfers
`settlementDeps` type [provides](https://github.com/nspcc-dev/neofs-node/blob/d99800ee93bbe9b0699313ba7a5ecab85699731c/pkg/innerring/settlement.go#L196) `Exchanger` for Inner Ring settlement processor.  As we can see in implementation, method always writes `for audit` transfer goal. However, this component is used for basic income processing, too. We need to write log message according to the particular transfer need.

carpawell:
Closes #1374.


pkg/innerring/settlement.go
```diff
@@ -204,6 +217,17 @@ func (s settlementDeps) ResolveKey(ni common.NodeInfo) (*owner.ID, error) {
 }
 
 func (s settlementDeps) Transfer(sender, recipient *owner.ID, amount *big.Int, details []byte) {
+	var callerContext string
+
+	switch s.settlementCtx {
```
cthulhu-rider, line 222: Oh, u've done a great job! But maybe it's easeir to have `name string` field in `settlementDeps` struct? Minus type, minus enum, minus switch.

cthulhu-rider (dismissed):

cthulhu-rider:
@carpawell  do the changes deserve to be included in the changelog, or too minor?

pkg/innerring/settlement.go
carpawell: thought about that but then did not want to have string here

but ok, agree that it is easier, fixed

carpawell:
> @carpawell do the changes deserve to be included in the changelog, or too minor?

oh, forget about that one more time. also would add changes from the #1403 PR.

cthulhu-rider (approved):

PR #1415, opened 2022-05-20
[#1335] policer: Prevent potential object loss

### LINKED ISSUE #1335: Potentially redundant deletion of a copy of an object
There is a potential situation when storage node removes the only one local copy of an object:
* `Policer` [routine](https://github.com/nspcc-dev/neofs-node/blob/b74fb2b932cf3cfc2a9b3e652ca2e032f331d635/pkg/services/policer/check.go#L16) builds placement vectors for the object and passes each vector to `processNodes`
* `processNodes` analyzes current placement vector, and if it detects redundant local copy, the corresponding callback is called

If we consider the particular placement for the object:
```
[N1, N2]
[N2]
```
then `Policer` of `N2` node will cause object removal on 1st iteration, while object is still required according to 2nd vector.

Is this placement correct? If so, then `Policer` routine must take this into account.

@fyrchik @realloc @alexvanin

## Possible soluition
Place shortage-check after loop in `checkObject`.

cthulhu-rider:
In previous implementation `Policer` considered local object copy as
redundant on processing single placement vector.

Make `Policer` to call redundant copy callback after full placement
processing. Also fix 404 error parsing.

* closes #1335

### DIFF
--- a/pkg/services/policer/check.go
+++ b/pkg/services/policer/check.go
@@ -2,12 +2,12 @@ package policer
 
 import (
 	"context"
-	"strings"
 
 	"github.com/nspcc-dev/neofs-node/pkg/core/container"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/engine"
 	headsvc "github.com/nspcc-dev/neofs-node/pkg/services/object/head"
 	"github.com/nspcc-dev/neofs-node/pkg/services/replicator"
+	"github.com/nspcc-dev/neofs-sdk-go/client"
 	"github.com/nspcc-dev/neofs-sdk-go/netmap"
 	addressSDK "github.com/nspcc-dev/neofs-sdk-go/object/address"
 	"go.uber.org/zap"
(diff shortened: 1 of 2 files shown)


pkg/services/policer/check.go
```diff
@@ -58,74 +61,81 @@ func (p *Policer) processObject(ctx context.Context, addr *addressSDK.Address) {
...
 		if p.netmapKeys.IsLocalKey(nodes[i].PublicKey()) {
-			if shortage == 0 {
-				// we can call the redundant copy callback
-				// here to slightly improve the performance
-				// instead of readability.
-				redundantLocalCopy = true
-				break
-			} else {
-				shortage--
+			if shortage > 0 {
```
fyrchik, line 97: We have this condition in the loop, how can it ever be false here?

cthulhu-rider: Forgot to remove

carpawell (approved):

carpawell (approved):

fyrchik (approved):

Issue #1398, opened 2022-05-19
Optimize node memory consumption


fyrchik:
There is a number of cases where we can optimize the memory consumption.

- [ ] After SDK move all `ID`s should be passed by value where possible
- [ ] Pass `*Prm` and `*Res` structures by value in `local_object_storage`
- [ ] #1260 
- [ ] Optimize object transformer pipelines: for a single object we consume `C * size(object)`, ideally we'd like to have `C = 1`, using streaming interfaces, each separate transformer could be profiled
- [ ] Restrict the amount of concurrent service connections, return `ErrBusy`
- [ ] All DB keys in `local_object_storage` components should be stored in some raw format, without `string` conversions. (this is a breaking change, worth trying in one place first)

Anything else?

Related #1319 .

carpawell:
> Anything else?

Optimize `GetRange` of the FSTree? Now the blobstor [reads the whole object](https://github.com/nspcc-dev/neofs-node/blob/f699e82ea7c47857fdcba8190effcc3339126a6a/pkg/local_object_storage/blobstor/get_range_big.go#L33) via `Get` method, *decompresses* it, and only after that cuts the desired part of an object. `Get` method [reads](https://github.com/nspcc-dev/neofs-node/blob/f699e82ea7c47857fdcba8190effcc3339126a6a/pkg/local_object_storage/blobstor/fstree/fstree.go#L243) all the payload in the memory even if only a small part is required.

Optimization may require moving compression on the blobovnicza and FSTree levels and [decompressing](https://github.com/nspcc-dev/neofs-node/blob/master/pkg/local_object_storage/blobstor/get_range_big.go#L44) the object only if it was really compressed.

There is already some kind of thoughts about compression on the bloboznicza side: https://github.com/nspcc-dev/neofs-node/blob/f699e82ea7c47857fdcba8190effcc3339126a6a/pkg/local_object_storage/blobstor/blobovnicza.go#L607-L610

@nspcc-dev/neofs-core

PR #1418, opened 2022-05-20
Fix/Do not use `Prm`, `Res` pointers in Storage Engine

carpawell:
Related to #1398.

### DIFF
--- a/pkg/local_object_storage/blobovnicza/delete.go
+++ b/pkg/local_object_storage/blobovnicza/delete.go
@@ -29,7 +29,7 @@ func (p *DeletePrm) SetAddress(addr oid.Address) {
 // Returns an error of type apistatus.ObjectNotFound if the object to be deleted is not in blobovnicza.
 //
 // Should not be called in read-only configuration.
-func (b *Blobovnicza) Delete(prm *DeletePrm) (*DeleteRes, error) {
+func (b *Blobovnicza) Delete(prm DeletePrm) (*DeleteRes, error) {
 	addrKey := addressKey(prm.addr)
 
 	removed := false
--- a/pkg/local_object_storage/blobstor/blobovnicza.go
+++ b/pkg/local_object_storage/blobstor/blobovnicza.go
@@ -205,8 +205,8 @@ func (b *blobovniczas) put(addr oid.Address, data []byte) (*blobovnicza.ID, erro
 //
 // If blobocvnicza ID is specified, only this blobovnicza is processed.
 // Otherwise, all blobovniczas are processed descending weight.
-func (b *blobovniczas) get(prm *GetSmallPrm) (res *GetSmallRes, err error) {
-	bPrm := new(blobovnicza.GetPrm)
+func (b *blobovniczas) get(prm GetSmallPrm) (res *GetSmallRes, err error) {
+	var bPrm blobovnicza.GetPrm
 	bPrm.SetAddress(prm.addr)
 
 	if prm.blobovniczaID != nil {
@@ -255,8 +255,8 @@ func (b *blobovniczas) get(prm *GetSmallPrm) (res *GetSmallRes, err error) {
 //
 // If blobocvnicza ID is specified, only this blobovnicza is processed.
 // Otherwise, all blobovniczas are processed descending weight.
-func (b *blobovniczas) delete(prm *DeleteSmallPrm) (res *DeleteSmallRes, err error) {
-	bPrm := new(blobovnicza.DeletePrm)
+func (b *blobovniczas) delete(prm DeleteSmallPrm) (res *DeleteSmallRes, err error) {
+	var bPrm blobovnicza.DeletePrm
 	bPrm.SetAddress(prm.addr)
 
 	if prm.blobovniczaID != nil {
@@ -310,7 +310,7 @@ func (b *blobovniczas) delete(prm *DeleteSmallPrm) (res *DeleteSmallRes, err err
 //
 // If blobocvnicza ID is specified, only this blobovnicza is processed.
 // Otherwise, all blobovniczas are processed descending weight.
-func (b *blobovniczas) getRange(prm *GetRangeSmallPrm) (res *GetRangeSmallRes, err error) {
+func (b *blobovniczas) getRange(prm GetRangeSmallPrm) (res *GetRangeSmallRes, err error) {
 	if prm.blobovniczaID != nil {
 		blz, err := b.openBlobovnicza(prm.blobovniczaID.String())
 		if err != nil {
@@ -360,7 +360,7 @@ func (b *blobovniczas) getRange(prm *GetRangeSmallPrm) (res *GetRangeSmallRes, e
 // tries to delete object from particular blobovnicza.
 //
 // returns no error if object was removed from some blobovnicza of the same level.
-func (b *blobovniczas) deleteObjectFromLevel(prm *blobovnicza.DeletePrm, blzPath string, tryActive bool, dp *DeleteSmallPrm) (*DeleteSmallRes, error) {
+func (b *blobovniczas) deleteObjectFromLevel(prm blobovnicza.DeletePrm, blzPath string, tryActive bool, dp DeleteSmallPrm) (*DeleteSmallRes, error) {
 	lvlPath := filepath.Dir(blzPath)
 
 	log := b.log.With(
@@ -423,7 +423,7 @@ func (b *blobovniczas) deleteObjectFromLevel(prm *blobovnicza.DeletePrm, blzPath
 // tries to read object from particular blobovnicza.
 //
 // returns error if object could not be read from any blobovnicza of the same level.
-func (b *blobovniczas) getObjectFromLevel(prm *blobovnicza.GetPrm, blzPath string, tryActive bool) (*GetSmallRes, error) {
+func (b *blobovniczas) getObjectFromLevel(prm blobovnicza.GetPrm, blzPath string, tryActive bool) (*GetSmallRes, error) {
 	lvlPath := filepath.Dir(blzPath)
 
 	log := b.log.With(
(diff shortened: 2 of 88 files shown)


pkg/local_object_storage/blobovnicza/delete.go
```diff
@@ -29,7 +29,7 @@ func (p *DeletePrm) SetAddress(addr *addressSDK.Address) {
 // Returns an error of type apistatus.ObjectNotFound if the object to be deleted is not in blobovnicza.
 //
 // Should not be called in read-only configuration.
-func (b *Blobovnicza) Delete(prm *DeletePrm) (*DeleteRes, error) {
+func (b *Blobovnicza) Delete(prm DeletePrm) (DeleteRes, error) {
```
cthulhu-rider, line 32: Why did you decide to return a value instead of a pointer?

carpawell: what do you mean? what that PR is doing is removing poiners in the whole SE (since it was suggested in #1398 and I really do not know why we need pointers in that cases). so you asking why does that PR exist in general?

cthulhu-rider: I understand PR existence, and accepting parameters by value is really good. But what about return values? The common approach we use is returning `(*T, error)`. What are pros to return `(T, error)`? I mean why don't we return just `T` which contains `error`? Maybe I've missed some discussion.

carpawell: > The common approach we use is returning (*T, error).

> i saw some examples in the SE where it already was as `func(prm Prm) (Res, error)` why is it common? to be able to return `nil` res?

> I mean why don't we return just T which contains error

to support common practice in `go`: start handling results only if `err == nil` (and know that operation could be finished with error and do not read docs to the `Res` struct)

> Maybe I've missed some discussion.

no, that was my decision

fyrchik: >What are pros to return `(T, error)`?

If we return a pointer, the struct will _definitely_ be allocated on heap (all these functions are big enough not to be inlined). And in case of `T` there is no allocation. It is more beneficial than passing parameters by value, actually, because parameters can still be allocated on stack as they do not outlive the called function.

>I mean why don't we return just T which contains error

This is possible, but would make the code non-idiomatic for no real reason IMO. We should still check for the error.

cthulhu-rider: Until the building itself, we do not know exactly how the data will be allocated since this is not visible from the source code. Are there any objective reasons for the change? If it's really a bottleneck, then we need to provide test proof.

We've already discussed this earlier, and previously I voted against `return (*T, error)` for the same GC optimization and suggested to accept result by pointer as a parameter. This will allow to reuse the same instance between several calls, and most definitely avoid heap allocation. But we've decided with @carpawell @alexvanin to not do it like this. At the same time I don't see returning `(T, error)` convenient.

I suggest to see what the profiler shows.

cthulhu-rider: 1. There is no "heap" in terms Go source code. In your example you make a decision according to the specific compiler. What does it mean "big enough not to be inlined"? How can I get it from the sources?

And again... I am aware and agree that returning a pointer leads rather to worse performance. Therefore, I propose a compromise: allocate a structure to record the results on the calling side and pass a pointer to it in the function.

carpawell: i reorganized that PR: it now only contains `*Prm` -> `Prm` changes

`Res` changes will be added via another PR

pkg/local_object_storage/shard/dump.go
```diff
@@ -81,7 +81,10 @@ func (s *Shard) Dump(prm *DumpPrm) (*DumpRes, error) {
 	var count int
 
 	if s.hasWriteCache() {
-		err := s.writeCache.Iterate(new(writecache.IterationPrm).WithHandler(func(data []byte) error {
+		var iterPrm writecache.IterationPrm
+
+		iterPrm.WithIgnoreErrors(prm.ignoreErrors)
+		iterPrm.WithHandler(func(data []byte) error {
```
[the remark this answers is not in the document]
carpawell: deleted

pkg/local_object_storage/metabase/delete.go
```diff
@@ -60,7 +61,7 @@ func (db *DB) Delete(prm *DeletePrm) (*DeleteRes, error) {
 				storagelog.OpField("metabase DELETE"))
 		}
 	}
-	return new(DeleteRes), err
```
fyrchik, line 63: unrelated to this commit

pkg/local_object_storage/shard/range.go
```diff
@@ -27,39 +27,34 @@ type RngRes struct {
...
+func (p *RngPrm) WithIgnoreMeta(ignore bool) {
 	p.skipMeta = ignore
-	return p
 }
 
 // Object returns the requested object part.
 //
 // Instance payload contains the requested range of the original object.
-func (r *RngRes) Object() *object.Object {
+func (r RngRes) Object() *object.Object {
```
fyrchik, line 52: Unrelated to this commit.

pkg/local_object_storage/shard/inhume.go
```diff
@@ -79,5 +75,5 @@ func (s *Shard) Inhume(prm *InhumePrm) (*InhumeRes, error) {
 		return nil, fmt.Errorf("metabase inhume: %w", err)
 	}
 
-	return new(InhumeRes), nil
+	return nil, nil
```
fyrchik, line 78: unrelated to this commit

pkg/local_object_storage/engine/container.go
```diff
@@ -70,7 +70,7 @@ func (e *StorageEngine) containerSize(prm ContainerSizePrm) (*ContainerSizeRes,
 		defer elapsed(e.metrics.AddEstimateContainerSizeDuration)()
 	}
 
-	var res ContainerSizeRes
+	res := new(ContainerSizeRes)
```
fyrchik, line 73: unrelated to this commit

pkg/local_object_storage/engine/get.go
```diff
@@ -24,16 +24,14 @@ type GetRes struct {
...
 	if p != nil {
 		p.addr = addr
 	}
-
-	return p
 }
 
 // Object returns the requested object.
-func (r *GetRes) Object() *objectSDK.Object {
+func (r GetRes) Object() *objectSDK.Object {
```
fyrchik, line 34: unrelated to this commit

pkg/local_object_storage/blobstor/blobovnicza.go
```diff
@@ -255,20 +255,21 @@ func (b *blobovniczas) get(prm *GetSmallPrm) (res *GetSmallRes, err error) {
...
-	bPrm := new(blobovnicza.DeletePrm)
+func (b *blobovniczas) delete(prm DeleteSmallPrm) (res *DeleteSmallRes, err error) {
+	var bPrm blobovnicza.DeletePrm
 	bPrm.SetAddress(prm.addr)
 
 	if prm.blobovniczaID != nil {
 		blz, err := b.openBlobovnicza(prm.blobovniczaID.String())
 		if err != nil {
-			return nil, err
+			return res, err
```
fyrchik, line 265: unrelated to this commit

pkg/local_object_storage/blobstor/blobovnicza.go
```diff
@@ -289,14 +290,14 @@ func (b *blobovniczas) delete(prm *DeleteSmallPrm) (res *DeleteSmallRes, err err
 		activeCache[dirPath] = struct{}{}
 
 		if err == nil {
-			res = new(DeleteSmallRes)
+			objectFound = true
```
fyrchik, line 293: unrelated to this commit

pkg/local_object_storage/blobstor/blobovnicza.go
```diff
@@ -343,11 +345,13 @@ func (b *blobovniczas) getRange(prm *GetRangeSmallPrm) (res *GetRangeSmallRes, e
 
 		activeCache[dirPath] = struct{}{}
 
+		objectFound = err == nil
```
fyrchik, line 348: same here

pkg/local_object_storage/engine/delete.go
```diff
@@ -47,24 +47,28 @@ func (e *StorageEngine) delete(prm *DeletePrm) (*DeleteRes, error) {
 		defer elapsed(e.metrics.AddDeleteDuration)()
 	}
 
-	shPrm := new(shard.InhumePrm)
-	existsPrm := new(shard.ExistsPrm)
+	var shPrm shard.InhumePrm
```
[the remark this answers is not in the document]
carpawell: what for?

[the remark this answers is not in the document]
fyrchik: Because after removal of return value from `With*` functions, it can be allocated on stack.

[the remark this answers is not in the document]
fyrchik: And it will stay correct if we decide to parallelize iteration over shards.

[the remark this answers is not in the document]
carpawell: > Because after removal of return value from With* functions, it can be allocated on stack.

Isn't it still true for a one var declaration before the callback call?

> And it will remain correct if we decide to parallelize iteration over shards.

agree here, moved inside the callback

carpawell:
@cthulhu-rider, @fyrchik, lets merge that, it becomes really difficult to support (rebase) it and #1460. It does not contain smth that we have not discussed yet.

cthulhu-rider (approved):
Do we want to remove nil-conditions in setters here or later?

carpawell:
> Do we want to remove nil-conditions in setters here or later?

Do you mean removing `if p != nil` in `*Prm`? Do we need it? If so, i would remove them in that PR.

PR #1422, opened 2022-05-23
Fix innerring signer scope

fyrchik:
Currently there is an error about `Array/ByteString` conversion.


carpawell:
> Also, update contract version (it is already 0.15.1 in devenv).

Did you mean updating the version in that PR?

carpawell (approved):

PR #1426, opened 2022-05-24
services/tree: do not replicate on a local node

fyrchik:

### DIFF
--- a/pkg/services/tree/replicator.go
+++ b/pkg/services/tree/replicator.go
@@ -1,11 +1,13 @@
 package tree
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/services/tree/replicator.go
```diff
@@ -59,10 +61,16 @@ func (s *Service) replicate(ctx context.Context, op movePair) error {
 		return fmt.Errorf("can't get container nodes: %w", err)
 	}
 
+	rawKey := (*keys.PublicKey)(&s.key.PublicKey).Bytes()
```
cthulhu-rider, line 64: How often `replicate` processes? If frequently, we can encode the key once.

cthulhu-rider (approved):

pkg/services/tree/replicator.go
fyrchik: Fixed.

PR #1430, opened 2022-05-25
Fix/Update `neo-go`

carpawell:
v0.98.2 => v0.98.3

That patch includes fixing panic when using WS client with a closed connection.

Did not use `0.98.5` since note releases do not contain something related to the client and usually I do not like updating just because.


fyrchik (approved):

alexvanin:
Didn't we fix it in https://github.com/nspcc-dev/neofs-node/pull/1291? Or it was not relevant after WS client update?

alexvanin:
The issue was related to storage node and #1291 fixes IR, got it.

PR #1434, opened 2022-05-25
[#1433] metabase: optimize `Select`

### LINKED ISSUE #1433: Optimize metabase filter processing
In many places (3 out of 4) we iterate over the bucket case and apply matching function: 
https://github.com/nspcc-dev/neofs-node/blob/786c920fb2c8afa2504084b5dbed39eb4c28f914/pkg/local_object_storage/metabase/select.go#L295

This can be optimized for some filters:
1. `MatchStringEqual` -- just check a single key.
2. `MatchCommonPrefix` -- prefix scan with a cursor.

fyrchik:
Close #1433.

For some filters we can scan only a subset of keys instead of checking
each key.

TODO: tests for corner cases and benchmarks


pkg/local_object_storage/metabase/db.go
```diff
@@ -80,22 +98,132 @@ func stringifyValue(key string, objVal []byte) string {
...
+		return c - 'a' + 10, true
+	case 'A' <= c && c <= 'F':
+		return c - 'A' + 10, true
+	}
+
+	return 0, false
+}
+
+// destringifyValue is the reverse operation for stringify value.
+// The last return argument returns true if the filter CAN match any value.
```
carpawell, line 116:
```suggestion
// The last return value returns true if the filter CAN match any value.
```

carpawell (approved):

carpawell (approved):

PR #1431, opened 2022-05-25
Optimize pilorama

fyrchik:
1. `Update` -> `Batch`.
2. Build index on `FileName` attribute.

PR #1439, opened 2022-05-26
services/tree: Retransmit queries to container nodes

### LINKED ISSUE #1401: Allow to redirect tree service queries to container nodes

fyrchik:
Close #1401 .

Depends on #1431 .

### DIFF
--- a/pkg/services/tree/redirect.go
+++ b/pkg/services/tree/redirect.go
@@ -1,0 +1,70 @@
+package tree
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
+	"github.com/nspcc-dev/neofs-node/pkg/network"
+	cidSDK "github.com/nspcc-dev/neofs-sdk-go/container/id"
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+)
+
+var errNoSuitableNode = errors.New("no node was found to execute the request")
+
+// forEachNode executes callback for each node in the container.
+// If the node belongs to a container, nil error is returned.
+// Otherwise, f is executed for each node, stopping if true is returned.
+func (s *Service) forEachNode(ctx context.Context, cid *cidSDK.ID, f func(c TreeServiceClient) bool) error {
+	cntNodes, err := s.getContainerNodes(cid)
+	if err != nil {
+		return fmt.Errorf("can't get container nodes for %s: %w", cid, err)
+	}
+
+	rawPub := (*keys.PublicKey)(&s.key.PublicKey).Bytes()
+	nodes := cntNodes.Flatten()
+	for _, node := range nodes {
+		if bytes.Equal(node.PublicKey(), rawPub) {
+			return nil
+		}
+	}
+
+	var called bool
+	for _, node := range nodes {
+		var stop bool
+		node.IterateAddresses(func(endpoint string) bool {
+			c, err := dialTreeService(ctx, endpoint)
+			if err != nil {
+				return false
+			}
+
+			s.log.Debug("redirecting tree service query", zap.String("endpoint", endpoint))
+			called = true
+			stop = f(c)
+			return true
+		})
+		if stop {
+			return nil
+		}
+	}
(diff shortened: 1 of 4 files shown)


pkg/services/tree/redirect.go
```diff
@@ -0,0 +1,65 @@
...
+
+var errNoSuitableNode = errors.New("no node was found to execute the request")
+
+// forEachNode executes callback for each node in the container.
+// If the node belongs to a container, nil error is returned.
+// Otherwise, f is executed for each node, stopping if true is returned.
+func (s *Service) forEachNode(ctx context.Context, cid *cidSDK.ID, f func(c TreeServiceClient) bool) error {
+	cntNodes, err := s.getContainerNodes(cid)
+	if err != nil {
+		return fmt.Errorf("can't get container nodes: %w", err)
```
carpawell, line 23: what do you think about CID in that log?

pkg/services/tree/redirect.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	rawPub := (*keys.PublicKey)(&s.key.PublicKey).Bytes()
+	for _, node := range cntNodes.Flatten() {
+		if bytes.Equal(node.PublicKey(), rawPub) {
+			return nil
+		}
+	}
+
+	nodes := cntNodes.Flatten()
```
carpawell, line 32: `cntNodes.Flatten()` is called twice in the func, move that var declaration before the first usage and reuse it here?

pkg/services/tree/redirect.go
```diff
@@ -0,0 +1,65 @@
...
+
+			s.log.Debug("redirecting tree service query", zap.String("endpoint", endpoint))
+			stop = f(c)
+			return true
+		})
+		if stop {
+			return nil
+		}
+	}
+	return errNoSuitableNode
```
carpawell, line 49: did i get it right: if i pass a callback that always returns `false` i get `errNoSuitableNode` err as a result even if there are available nodes in the container? was it expected?

pkg/services/tree/service.go
```diff
@@ -64,6 +64,18 @@ func (s *Service) Add(_ context.Context, req *AddRequest) (*AddResponse, error)
 		return nil, err
 	}
 
+	var resp *AddResponse
+	var outErr error
+	err = s.forEachNode(ctx, cid, func(c TreeServiceClient) bool {
+		resp, outErr = c.Add(ctx, req)
```
carpawell, line 70: why dont we try to check err here and continue retransmitting to another node?

pkg/services/tree/service.go
```diff
@@ -244,6 +304,24 @@ func (s *Service) GetSubTree(req *GetSubTreeRequest, srv TreeService_GetSubTreeS
...
+	var outErr error
+	err = s.forEachNode(srv.Context(), cid, func(c TreeServiceClient) bool {
+		cli, outErr = c.GetSubTree(srv.Context(), req)
+		return true
+	})
+	if err != nil {
+		return err
+	} else if outErr != nil {
+		return outErr
+	} else if cli != nil {
```
carpawell, line 317: is it possible to get `nil` cli if `outErr` == `nil`?

pkg/services/tree/redirect.go
fyrchik: Fixed

pkg/services/tree/redirect.go
fyrchik: Fixed

pkg/services/tree/redirect.go
fyrchik: Nice catch, fixed.

pkg/services/tree/service.go
fyrchik: Because the error can happen and if we are here, the connection was already established, so the error is most likely should be returned.

pkg/services/tree/service.go
fyrchik: Yes, exactly when the node belongs to a container and the operation is local. Callback is not executed at all in this case.

cthulhu-rider (approved):

Issue #1227, opened 2022-03-10
Engine drops object locking when GC physically removes lock-object


alexvanin:
## Expected Behavior
When lock-object is removed (inhumed in terms of storage engine), all corresponding regular objects should be unlocked.

## Current Behavior
1. Regular objects are unlocked only when lock-object is removed by GC.
2. GC unlocks regular objects stored in the same shard as lock-object.

## Context
In https://github.com/nspcc-dev/neofs-node/pull/1181 metabase drops object locking during `freePotentialLocks()`. This function invoked during `Inhume` of non-regular object with `GCMark`. 

To solve (1) we can invoke `freePontantialLocks()` during any `Inhume(tombstone)` invocation. However this won't affect objects which are locked in other shards. 

(2) is similar to the issue of objects with expired tombstone that we address in https://github.com/nspcc-dev/neofs-node/issues/378. When tombstone is expired, engine has to remove all corresponding objects in all shards. To solve that, shard invokes `expiredTombstonesCallback()` to go through all shards.

However expired tombstone issue can be solved differently, see proposed solution in https://github.com/nspcc-dev/neofs-node/issues/1226 We want a robust solution for locks as well, one more callback does not look appealing.

alexvanin:
Maybe engine should preemptively discover lock-object removal and initiate object unlocking in all shards? To do that we have to implement `Unlock()` method on shard \ metabase.

alexvanin:
Expected behavior according to the latest discussions:
1. Regular objects are unlocked as soon as lock objects are expired
2. Lock objects must always expire, otherwise they are not valid
3. Lock objects cannot be deleted, only expired

This makes lock object behavior consistent (no removable locks). Specification may omit such details, so it should be fixed.
Locks expiration should trigger lock index update in metabase, so regular object can be removed from now on.

As for temporary locks, we should use different scheme, e.g. alphabet-signed lock notification, that will be parsed by all storage nodes.

alexvanin:
Control service should keep an ability to set GCMark on LOCK object anyway (by @realloc )

cthulhu-rider:
Then there will be 3 types of `Inhume`:

1. with tombstone
2. regular GC mark
3. force GC mark

carpawell:
From issue body:

> When lock-object is removed (inhumed in terms of storage engine), all corresponding regular objects should be unlocked.

it was true for expired LOCKs and still true.

[...]

PR #1440, opened 2022-05-26
Fix: Unlock locked object on its lock removal

carpawell:
See https://github.com/nspcc-dev/neofs-node/issues/1227#issuecomment-1138913967.

Also, more tests in SE will be added if PR is valid after the discussion.


carpawell:
Would be added to `master` via feature branch: #1461.

PR #1454, opened 2022-05-30
Upgrade `address` package

### LINKED ISSUE #1408: Make sure that `EncodeToString` is used instead of `String` where required
In latest SDK change "stringifiable" types like IDs provide two string methods: `EncodeToString` and `String`. First one return stable protocol-level string while latter is left for human readability.

Within this issue we need to assert that `String` is used only for printing - logs and printlns.

cthulhu-rider:
* closes #1408 

Core changes:
 * avoid package-colliding variable naming
 * avoid using pointers to IDs where unnecessary
 * avoid using `idSDK` import alias pattern
 * use `EncodeToString` for protocol string calculation and `String` for
  printing

@fyrchik @carpawell all integration tests are PASS except ACL ones, but they FAIL regardless of current changes. I suggest to merge this PR and fix PR in a separate one.

### DIFF
--- a/cmd/neofs-cli/modules/util/sign_session.go
+++ b/cmd/neofs-cli/modules/util/sign_session.go
@@ -1,6 +1,8 @@
 package util
 
 import (
+	"crypto/ecdsa"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
@@ -38,8 +40,26 @@ func signSessionToken(cmd *cobra.Command, _ []string) {
 		common.ExitOnErr(cmd, "", errors.New("missing session token flag"))
 	}
 
-	var stok session.Object
-	common.ReadSessionToken(cmd, &stok, signFromFlag)
+	type iTokenSession interface {
+		json.Marshaler
+		json.Unmarshaler
+		Sign(ecdsa.PrivateKey) error
+	}
+	var errLast error
+	var stok iTokenSession
+
+	for _, el := range [...]iTokenSession{
+		new(session.Object),
+		new(session.Container),
+	} {
+		errLast = common.ReadSessionTokenErr(el, fPath)
+		if errLast == nil {
+			stok = el
+			break
+		}
+	}
+
+	common.ExitOnErr(cmd, "", errLast)
 
 	pk := key.GetOrGenerate(cmd)
 
--- a/pkg/services/container/executor.go
+++ b/pkg/services/container/executor.go
@@ -31,7 +31,12 @@ func NewExecutionService(exec ServiceExecutor) Server {
 }
 
 func (s *executorSvc) Put(ctx context.Context, req *container.PutRequest) (*container.PutResponse, error) {
-	respBody, err := s.exec.Put(ctx, req.GetMetaHeader().GetOrigin().GetSessionToken(), req.GetBody())
+	meta := req.GetMetaHeader()
+	for origin := meta.GetOrigin(); origin != nil; origin = meta.GetOrigin() {
+		meta = origin
+	}
+
+	respBody, err := s.exec.Put(ctx, meta.GetSessionToken(), req.GetBody())
 	if err != nil {
 		return nil, fmt.Errorf("could not execute Put request: %w", err)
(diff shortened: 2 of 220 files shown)


carpawell:
> avoid using idSDK import alias pattern

Why? I suggested that pattern for distinguishing between API, SDK, and internal imports.

cmd/neofs-cli/modules/util/sign_session.go
```diff
@@ -38,8 +40,25 @@ func signSessionToken(cmd *cobra.Command, _ []string) {
...
+		new(session.Object),
+		new(session.Container),
+	} {
+		errLast = common.ReadSessionTokenErr(el, fPath)
+		if errLast == nil {
+			stok = el
+		}
+	}
+
+	common.ExitOnErr(cmd, "", errLast)
```
fyrchik, line 61: This deserves a separate commit.

cmd/neofs-cli/modules/util/sign_session.go
```diff
@@ -38,8 +40,25 @@ func signSessionToken(cmd *cobra.Command, _ []string) {
...
+	var errLast error
+	var stok iTokenSession
+
+	for _, el := range [...]iTokenSession{
+		new(session.Object),
+		new(session.Container),
+	} {
+		errLast = common.ReadSessionTokenErr(el, fPath)
+		if errLast == nil {
+			stok = el
```
fyrchik, line 57: Shouldn't we `break` after?

Do we need a loop here at all? Seems like 2 simple if's will do the job.

pkg/services/container/executor.go
```diff
@@ -31,7 +31,12 @@ func NewExecutionService(exec ServiceExecutor) Server {
 }
 
 func (s *executorSvc) Put(ctx context.Context, req *container.PutRequest) (*container.PutResponse, error) {
-	respBody, err := s.exec.Put(ctx, req.GetMetaHeader().GetOrigin().GetSessionToken(), req.GetBody())
+	meta := req.GetMetaHeader()
+	for origin := meta.GetOrigin(); origin != nil; origin = meta.GetOrigin() {
+		meta = origin
+	}
+
+	respBody, err := s.exec.Put(ctx, meta.GetSessionToken(), req.GetBody())
```
fyrchik, line 39: Why is there #1285 in the commit message?

cthulhu-rider:
> I suggested that pattern for distinguishing between API, SDK, and internal imports.

We have no other `oid` or `cid` packages.

pkg/services/container/executor.go
cthulhu-rider: Typo.

cmd/neofs-cli/modules/util/sign_session.go
cthulhu-rider: Added `break`. There can be more contexts, `for` allows to easily add new ones.

carpawell:
> We have no other oid or cid packages.

That is just because we do not have that pkgs in `api-go`. I would not say it's the rule and would not say that in `V3` there won't be such pkgs. Also, IMO, it is easier to follow a common rule ("add repo suffixes") than make exceptions ("... but not for IDs"). Also, I think that it is possible (theoretically) to have an OID/CID wrapper in that repo.

What are the arguments for dropping suffixes?

carpawell:
> Am I wrong?

Theoretically, of course, no. In practice, we also say that apps should only use SDK and that is not true. Also, we may have some pkgs that only contain utils that work with `cid`, `oid` and it seems that such pkgs should be named `cid` and `oid` (the same naming is already [done](https://github.com/nspcc-dev/neofs-node/blob/cc6209e8a011974ef25a5be53ccac8c59ff0414e/pkg/services/object/metrics.go#L1) for `object` in the node repo).

cthulhu-rider:
I don't know, I can add a commit with aliasing `oid` -> `oidSDK`, `cid` -> `cidSDK` if it blocks merging.

carpawell (approved):

PR #1455, opened 2022-05-30
services/tree: Cache the list of container nodes

fyrchik:
Also fix connections to TLS addresses.

### DIFF
--- a/pkg/services/tree/container.go
+++ b/pkg/services/tree/container.go
@@ -1,0 +1,83 @@
+package tree
+
+import (
+	"bytes"
+	"fmt"
+	"sync"
+
+	"github.com/hashicorp/golang-lru/simplelru"
+	containerSDK "github.com/nspcc-dev/neofs-sdk-go/container"
+	cidSDK "github.com/nspcc-dev/neofs-sdk-go/container/id"
+	netmapSDK "github.com/nspcc-dev/neofs-sdk-go/netmap"
+)
+
+type containerCache struct {
+	sync.Mutex
+	nm  *netmapSDK.Netmap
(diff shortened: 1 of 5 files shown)


pkg/services/tree/container.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		return nil, -1, fmt.Errorf("can't get netmap: %w", err)
+	}
+
+	cnr, err := s.cnrSource.Get(cid)
+	if err != nil {
+		return nil, -1, fmt.Errorf("can't get container: %w", err)
+	}
+
+	cidStr := cid.String()
```
carpawell, line 44: think `EncodeToString` would be better to prevent magic behavior changes when we change `String` implementation

pkg/services/tree/container.go
```diff
@@ -0,0 +1,83 @@
...
+
+	s.containerCache.Lock()
+	if s.containerCache.nm != nm {
+		s.containerCache.lru.Purge()
+	} else if v, ok := s.containerCache.lru.Get(cidStr); ok {
+		item := v.(containerCacheItem)
+		if item.cnr == cnr {
```
carpawell, line 51: do you compare pointers here? could they ever be equal?

fyrchik: Yes, in the case we have received the cached value.

pkg/services/tree/container.go
fyrchik: Experimental branch currently lacks `EncodeToString`. Also I don't think `String` is a problem here, because it is a runtime cache.

carpawell (approved):

Issue #1458, opened 2022-05-31
Why gprc wants num, while shard doesn't?


532910:
https://github.com/nspcc-dev/neofs-node/blob/cc6209e8a011974ef25a5be53ccac8c59ff0414e/config/example/node.yaml#L45-L46

https://github.com/nspcc-dev/neofs-node/blob/cc6209e8a011974ef25a5be53ccac8c59ff0414e/config/example/node.yaml#L138-L139

cthulhu-rider:
`grpc` is a composite section for listening endpoints. For local object storage, `storage` section composes all the parts. Pay attention to its `shard_num` value 
https://github.com/nspcc-dev/neofs-node/blob/cc6209e8a011974ef25a5be53ccac8c59ff0414e/config/example/node.yaml#L107

It has exactly the same meaning. Do I understand correctly that you are proposing to move this value to section `shard(s)`?

PR #1474, opened 2022-06-03
Remove "num" and "shard_num" parameters from the configuration

### LINKED ISSUE #1459: How much endpoints will listen the node with the example config?
https://github.com/nspcc-dev/neofs-node/blob/cc6209e8a011974ef25a5be53ccac8c59ff0414e/config/example/node.yaml#L45-L62

fyrchik:
Close #1459, #1458.

The problem with ENV is still here, but we can circumvent it by checking if some scalar value is here, instead of checking the whole section.


cthulhu-rider (approved):
Lets fix docs of touched funcs too.

cthulhu-rider (approved):

Issue #1503, opened 2022-06-09
Provide timeout when opening bbolt databases


fyrchik:
Bolt invokes `flock` after opening a file, and can hang indefinitely if the timeout is not explicitly provided.
This happens, for example, if 2 equal paths are provided in the config.
Add some reasonable timeout and, possibly, allow to override it.

carpawell:
> possibly, allow to override it

One more config value? Isnt overhead for that purpose?

fyrchik:
Closed via #1504.

PR #1505, opened 2022-06-09
Allow to customize pilorama database

fyrchik:
Related #1503 .

`no_sync` is for performance testing and could be removed in future.
`enabled` is deliberately omitted, we will discuss the need of it before tree-service is merged into with the master branch.


carpawell:
@fyrchik, rebasing?

pkg/local_object_storage/pilorama/option.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+}
+
+func WithPerm(perm fs.FileMode) Option {
+	return func(c *cfg) {
+		c.perm = perm
+	}
+}
+
+func WithNoSync(noSync bool) Option {
```
carpawell, line 30: any comments on the options?

cthulhu-rider (approved):

PR #1510, opened 2022-06-10
placement: cache container nodes

fyrchik:
In case we have lots of objects in a single container,
`GetContainerNodes` invoked indirectly by a policer can be seen in
pprof.

### DIFF
--- a/pkg/services/object_manager/placement/netmap.go
+++ b/pkg/services/object_manager/placement/netmap.go
@@ -3,7 +3,9 @@ package placement
 import (
 	"crypto/sha256"
 	"fmt"
+	"sync"
 
+	"github.com/hashicorp/golang-lru/simplelru"
 	"github.com/nspcc-dev/neofs-node/pkg/core/netmap"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 	netmapSDK "github.com/nspcc-dev/neofs-sdk-go/netmap"
@@ -12,6 +14,12 @@ import (
 
 type netMapBuilder struct {
 	nmSrc netmap.Source
+	// mtx protects lastNm and containerCache fields.
+	mtx    sync.Mutex
+	lastNm *netmapSDK.Netmap
+	// containerCache caches container nodes by ID. It is used to skip `GetContainerNodes` invocation if
+	// neither netmap nor container has changed.
+	containerCache simplelru.LRUCache
 }
 
 type netMapSrc struct {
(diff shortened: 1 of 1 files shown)


cthulhu-rider:
Since `placement.Build` is an interface, can we implement a separate type around LRU and base `placement.Builder`?

fyrchik:
Do we have places where LRU this cache doesn't provide any benefit?
I think it would result in more code with a separate type, without providing substantial (any?) value.

pkg/services/object_manager/placement/netmap.go
```diff
@@ -45,10 +57,21 @@ func (b *netMapBuilder) BuildPlacement(cnr cid.ID, obj *oid.ID, p *netmapSDK.Pla
...
+	var cn netmapSDK.ContainerNodes
+	if nm == b.lastNm {
+		raw, ok := b.containerCache.Get(string(binCnr))
+		if ok {
+			cn = raw.(netmapSDK.ContainerNodes)
+			return BuildObjectPlacement(nm, cn, obj)
+		}
+	}
+
+	b.containerCache.Purge()
```
carpawell, line 69: is it expected to purge the `containerCache` if netmap is the last one and the cache does not contain the requested container?

pkg/services/object_manager/placement/netmap.go
```diff
@@ -45,10 +57,21 @@ func (b *netMapBuilder) BuildPlacement(cnr cid.ID, obj *oid.ID, p *netmapSDK.Pla
 	binCnr := make([]byte, sha256.Size)
 	cnr.Encode(binCnr)
 
-	cn, err := nm.GetContainerNodes(p, binCnr)
+	var cn netmapSDK.ContainerNodes
```
carpawell, line 60: why do we need that var declaration?

fyrchik: Removed

pkg/services/object_manager/placement/netmap.go
```diff
@@ -45,11 +60,28 @@ func (b *netMapBuilder) BuildPlacement(cnr cid.ID, obj *oid.ID, p *netmapSDK.Pla
...
+		b.containerCache.Purge()
+		b.mtx.Unlock()
+	}
+
 	cn, err := nm.GetContainerNodes(p, binCnr)
 	if err != nil {
 		return nil, fmt.Errorf("could not get container nodes: %w", err)
 	}
 
+	b.mtx.Lock()
```
cthulhu-rider, line 81: Can we theoretically build placement for fixed `nm` and `cnr` in case of cache miss? Seems like `GetContainerNodes` isn't synced in this case.

pkg/services/object_manager/placement/netmap.go
```diff
@@ -45,11 +60,28 @@ func (b *netMapBuilder) BuildPlacement(cnr cid.ID, obj *oid.ID, p *netmapSDK.Pla
 	binCnr := make([]byte, sha256.Size)
 	cnr.Encode(binCnr)
 
+	b.mtx.Lock()
+	if nm == b.lastNm {
```
cthulhu-rider, line 64: Theoretically we can get same network map with different pointers . Do we wan't to handle this case somehow?

fyrchik: No, this is bad, but expected. We could solve this in the cache itself, if needed.

carpawell (approved):

pkg/services/object_manager/placement/netmap.go
fyrchik: Could you elaborate a bit? If cnr is fixed, it is never evicted from the cache.

cthulhu-rider: I mean `if nm == b.lastNm && !ok` multiple routines can call `nm.GetContainerNodes`, right? If so, is it expected?

fyrchik: Yes: it is a local operation, so nothing is wrong with allowing multiple routines to use it. Also, we do not want to block for different containers.

PR #1535, opened 2022-06-21
[#1520] shard: Ignore errors on metabase refill

### LINKED ISSUE #1520: Ignore or log errors on resync metabase
It is impossible to start a node.
```
июн 15 14:31:28 yu neofs-node[2482370]: 2022/06/15 14:31:28 could not initialize shard YAwBKiTu1Q4Bvo7MFw9LUT: could not initialize func() error: fs tree iterator failure: could not decompress object data: unexpected EOF
июн 15 14:31:28 yu systemd[1]: neofs-storage.service: Main process exited, code=exited, status=1/FAILURE
```

The object in case is 6-bytes only (and 522k on another node), so the node was most likely shutdown during the replication. This is an expected scenario, we should:
1. Log more descriptive error (object address, for example)
2. Continue metabase resync.

fyrchik:
Close #1520.

### DIFF
--- a/pkg/local_object_storage/blobovnicza/blobovnicza_test.go
+++ b/pkg/local_object_storage/blobovnicza/blobovnicza_test.go
@@ -147,7 +147,7 @@ func TestIterateObjects(t *testing.T) {
 		require.NoError(t, err)
 	}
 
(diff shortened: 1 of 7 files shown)


cthulhu-rider (approved):
LGTM

PR #1539, opened 2022-06-23
neofs-adm: support storage wallet passwords from config

fyrchik:


cmd/neofs-adm/internal/modules/config/config.go
```diff
@@ -153,3 +153,11 @@ func GetPassword(v *viper.Viper, name string) (string, error) {
 	prompt := "Password for " + name + " wallet > "
 	return input.ReadPassword(prompt)
 }
+
+func GetStoragePassword(v *viper.Viper, name string) (string, error) {
+	key := "storage." + name
+	if name != "" && v.IsSet(key) {
```
cthulhu-rider, line 159: If `name == ""` we calc `key` redundantly?

cthulhu-rider (approved):

cmd/neofs-adm/internal/modules/config/config.go
fyrchik: Yes, but it is done once in a CLI and simplifies the code considerably.

PR #1544, opened 2022-06-24
Fix/Add bearer token flag to CLI

carpawell:
It led to the `flag accessed but not defined: bearer` error in the container's subcommands.

Also updated SDK with https://github.com/nspcc-dev/neofs-sdk-go/pull/277 to make the master workable.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,9 +17,9 @@ require (
 	github.com/nspcc-dev/hrw v1.0.9
 	github.com/nspcc-dev/neo-go v0.99.1-pre.0.20220609082921-2c9fb2044242
 	github.com/nspcc-dev/neo-go/pkg/interop v0.0.0-20220601120906-3bec6657f5c5 // indirect
-	github.com/nspcc-dev/neofs-api-go/v2 v2.12.3-0.20220620114558-454b5c0ed7e9
+	github.com/nspcc-dev/neofs-api-go/v2 v2.12.3-0.20220621170933-dd233c3fbc84
 	github.com/nspcc-dev/neofs-contract v0.15.1
-	github.com/nspcc-dev/neofs-sdk-go v1.0.0-rc.4.0.20220621170307-721df386c599
+	github.com/nspcc-dev/neofs-sdk-go v1.0.0-rc.4.0.20220622125136-af7e20073bc6
 	github.com/nspcc-dev/tzhash v1.5.2
 	github.com/panjf2000/ants/v2 v2.4.0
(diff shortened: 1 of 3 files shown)


cmd/neofs-cli/modules/container/root.go
```diff
@@ -41,6 +42,7 @@ func init() {
 
 	for _, containerCommand := range containerChildCommand {
 		commonflags.InitAPI(containerCommand)
+		object.InitBearer(containerCommand)
```
cthulhu-rider, line 45: Why do we need bearer token in container commands? It is applicable only for object operations

go.mod
```diff
@@ -37,6 +37,8 @@ require (
 	google.golang.org/protobuf v1.28.0
 )
 
+require gopkg.in/yaml.v3 v3.0.1
```
cthulhu-rider, line 40: Why so special?

carpawell: i guess it's because of removing `// indirect`. why? i do not know, my `go mod tidy` decided to do so

cmd/neofs-cli/modules/container/root.go
carpawell: it has `list-objects` command that calls `ObjectSearch`. dont we need bearer for that?

carpawell: mb it's ok to apply that flag not to every cmd but for `list-objects` only

just did that to continue using cli, did not think much about it

cthulhu-rider: Pretty strange to see `list-objects` in `container` command. Anyway, lets add bearer to `list-objects` only.

go.mod
cthulhu-rider: Can we manually stuff it into existing `require` block?

cmd/neofs-cli/modules/container/root.go
carpawell: > Pretty strange to see list-objects in container command.

go.mod
carpawell: oh, you meant including that to the common section. got you, moved

cmd/neofs-cli/modules/container/root.go
carpawell: > Pretty strange to see list-objects in container command.

Why? do you think that listing objects should be part of `object` cmd? well, MB

> Anyway, lets add bearer to list-objects only.

agree, moved

cthulhu-rider (approved):

PR #1552, opened 2022-06-27
shard: Turn `ModeDegraded` on metabase opening failure

### LINKED ISSUE #1549: Initialize shards properly
During node start we currently fail if the metabase cannot be initialized.
The behaviour should be the following:
1. Try opening the metabase. On failure start shard in the degraded mode.
2. Try initializing blobovniczas. On failure:
2.1. If there is only 1 shard, shutdown the node.
2.2. If there are more shards, start as usual with the shard disabled.

cthulhu-rider:
Make `Shard` to work in degraded mode if metabase is unavailable on
opening stage.

* closes #1549

About 9bf46d99818bf6f5218535fb9f555e48356d9cab: I don't think it makes sense to define `ModeDisabled` shard mode now since `StorageEngine` can't work with it. Maybe we can implement private behavior on `Shard` side when it fails all the calls if `Init` was failed, smth like `errNotInitialized`.

### DIFF
--- a/pkg/local_object_storage/shard/control.go
+++ b/pkg/local_object_storage/shard/control.go
@@ -178,11 +178,7 @@ func (s *Shard) Close() error {
 		components = append(components, s.writeCache)
 	}
 
-	components = append(components, s.pilorama, s.blobStor)
-
-	if s.GetMode() != ModeDegraded {
-		components = append(components, s.metaBase)
-	}
+	components = append(components, s.pilorama, s.blobStor, s.metaBase)
 
 	for _, component := range components {
 		if err := component.Close(); err != nil {
--- a/pkg/local_object_storage/shard/control.go
+++ b/pkg/local_object_storage/shard/control.go
@@ -9,12 +9,28 @@ import (
 	meta "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/metabase"
 	objectSDK "github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	"go.uber.org/zap"
 )
+
+func (s *Shard) handleMetabaseFailure(stage string, err error) error {
+	s.log.Error("metabase failure, switching mode",
+		zap.String("stage", stage),
+		zap.Stringer("mode", ModeDegraded),
+		zap.Error(err),
+	)
+
+	err = s.SetMode(ModeDegraded)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/local_object_storage/shard/control.go
```diff
@@ -26,7 +26,17 @@ func (s *Shard) Open() error {
 			return fmt.Errorf("could not open %T: %w", component, err)
 		}
 	}
-	return nil
+
+	return s.openMetabase()
```
fyrchik, line 30: It changes the order of opening, can we just check whether `component == s.metaBase` in cycle?
The `SetMode` could be done separately in the end.

Also, same applies to `Init`.

pkg/local_object_storage/shard/control.go
```diff
@@ -26,7 +26,17 @@ func (s *Shard) Open() error {
...
+}
+
+func (s *Shard) openMetabase() error {
+	err := s.metaBase.Open()
+	if err != nil {
+		err = s.SetMode(ModeDegraded)
```
fyrchik, line 36: I think we need to log the old error too.

cthulhu-rider:
Some tests currently panic because `ModeDegraded` mode doesn't completely prevent operations with metabase. @fyrchik is gonna fix this.

pkg/local_object_storage/shard/control.go
```diff
@@ -23,32 +39,67 @@ func (s *Shard) Open() error {
...
-		fMetabase = s.metaBase.Init
+	type initializer interface {
+		Init() error
 	}
 
-	components := []func() error{
-		s.blobStor.Init, fMetabase, s.pilorama.Init,
+	var components []initializer
+
+	if s.GetMode() == ModeDegraded {
```
fyrchik, line 72: shouldn't it be `!=`?

pkg/local_object_storage/shard/control.go
```diff
@@ -127,7 +178,11 @@ func (s *Shard) Close() error {
 		components = append(components, s.writeCache)
 	}
 
-	components = append(components, s.pilorama, s.blobStor, s.metaBase)
+	components = append(components, s.pilorama, s.blobStor)
+
+	if s.GetMode() != ModeDegraded {
+		components = append(components, s.metaBase)
```
fyrchik, line 184: We should still close it if the error was durieng `Init` stage, can we check if db is `nil` in `Close`?

fyrchik (approved):

PR #1553, opened 2022-06-27
Fix/IR chain client logs

carpawell:
It leads to the incorrect error logging in the morph client: common struct
is passed to the `main` and `morph` helper func but also is mutated b/w
calls. As a result: "mainnet chain connection has been lost" log even when
the morph client was closed (not the mainnet one).


fyrchik:
Could you explain what exactly is wrong? It seems like `p.log` pointer is copied anyway and both functions are synchronous.

carpawell:
> Could you explain what exactly is wrong?

The same struct was used for both main and side client initialization. Worked fine until we added a callback on client closing, it uses the name field from the struct i wrote above => if that struct has been mutated (it has) logs contain an incorrect message.

PR #1554, opened 2022-06-27
Fix shard mode processing

### LINKED ISSUE #1550: Change shard behaviour on reaching error threshold
1. Set positive default value if the threshold is 0 in config.
2. In degraded mode if the metabase is unavailable, writes should only be performed with the blobstor with success returned.
3. On reaching a certain threshold with read operations, move shard to the read-only mode.

fyrchik:
Close #1550 .


pkg/local_object_storage/shard/put.go
```diff
@@ -15,7 +15,10 @@ type PutPrm struct {
 }
 
 // PutRes groups the resulting values of Put operation.
-type PutRes struct{}
+type PutRes struct {
+	metaErr bool
```
carpawell, line 19: non-critical suggestion: enum?

just afraid of undefined state (e.g. both `true`)

pkg/local_object_storage/shard/get.go
```diff
@@ -77,7 +77,6 @@ func (s *Shard) Get(prm GetPrm) (*GetRes, error) {
 
 		return res.Object(), nil
 	}
-
```
carpawell, line 80: changing artifact?

pkg/local_object_storage/shard/delete.go
```diff
@@ -63,7 +64,9 @@ func (s *Shard) Delete(prm DeletePrm) (*DeleteRes, error) {
 
 	err := meta.Delete(s.metaBase, prm.addr...)
 	if err != nil {
-		return nil, err // stop on metabase error ?
+		if mode&ModeDegraded == 0 { // Skip metabase errors in degraded mode.
```
carpawell, line 67: why have we even tried to `Delete` anything in the `degraded moe?`

pkg/local_object_storage/engine/engine.go
```diff
@@ -50,7 +52,11 @@ func (e *StorageEngine) reportShardError(
 		return
 	}
 
-	err = sh.SetMode(shard.ModeDegraded)
+	if errorCount == sh.writeErrorCount {
+		err = sh.SetMode(sh.GetMode() | shard.ModeReadOnly)
```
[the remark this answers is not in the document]
fyrchik: `ReadOnly` and `Degraded` are now a flags, they are 1 and 2 respectively, so it is even backwards compatible in some way.

pkg/local_object_storage/engine/delete.go
```diff
@@ -57,7 +57,7 @@ func (e *StorageEngine) delete(prm DeletePrm) (*DeleteRes, error) {
 
 			resExists, err := sh.Exists(existsPrm)
 			if err != nil {
-				e.reportShardError(sh, "could not check object existence", err)
+				e.reportShardError(sh, sh.metaErrorCount, "could not check object existence", err)
```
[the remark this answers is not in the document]
fyrchik: No, fixed.

pkg/local_object_storage/engine/engine.go
```diff
@@ -28,18 +28,20 @@ type StorageEngine struct {
...
 
 type shardWrapper struct {
-	errorCount *atomic.Uint32
+	metaErrorCount  *atomic.Uint32
+	writeErrorCount *atomic.Uint32
 	*shard.Shard
 }
 
 // reportShardError checks that the amount of errors doesn't exceed the configured threshold.
 // If it does, shard is set to read-only mode.
```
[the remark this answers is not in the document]
fyrchik: it is an experimental branch, will fix in master

pkg/local_object_storage/shard/exists.go
```diff
@@ -14,7 +14,8 @@ type ExistsPrm struct {
 
 // ExistsRes groups the resulting values of Exists operation.
 type ExistsRes struct {
-	ex bool
+	ex   bool
+	meta bool
```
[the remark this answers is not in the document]
fyrchik: fixed

pkg/local_object_storage/shard/get.go
fyrchik: fixed

pkg/local_object_storage/shard/delete.go
fyrchik: fixed

pkg/local_object_storage/engine/engine.go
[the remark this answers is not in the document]
carpawell: just wondering what to expect on `Degraded | ReadOnly` and, more importantly, how that gonna work when we add more modes

but that is totally ok if we speaking about that only in the experimental context

pkg/local_object_storage/shard/get.go
carpawell: seems like not. not critical

PR #1557, opened 2022-06-28
neofs-cli: check if container contains locks before removal

### LINKED ISSUE #1500: neofs-cli: Check if the container has lock objects before the deletion
`neofs-cli` could fail to remove container if there are any active locks in it.
This check should be disabled with some flag like `--force` or `--yes-i-know-what-i-am-doing`.

fyrchik:
Close #1500

### DIFF
--- a/cmd/neofs-cli/modules/container/delete.go
+++ b/cmd/neofs-cli/modules/container/delete.go
@@ -1,15 +1,19 @@
 package container
 
 import (
+	"fmt"
 	"time"
 
 	internalclient "github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/client"
 	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/common"
 	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/commonflags"
 	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/key"
+	objectSDK "github.com/nspcc-dev/neofs-sdk-go/object"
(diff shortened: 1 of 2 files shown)


fyrchik:
Ready for review, but let's merge it after release.

cmd/neofs-cli/modules/container/delete.go
```diff
@@ -30,6 +35,26 @@ Only owner of the container has a permission to remove container.`,
...
+			searchPrm.SetContainerID(id)
+			searchPrm.SetFilters(fs)
+			searchPrm.SetTTL(2)
+
+			common.PrintVerbose("Searching for LOCK objects...")
+			res, err := internalclient.SearchObjects(searchPrm)
+			common.ExitOnErr(cmd, "can't search for LOCK objects: %w", err)
+
+			if len(res.IDList()) != 0 {
+				common.ExitOnErr(cmd, "", errors.New("Container wasn't removed because LOCK objects were found.\n"+
```
carpawell, line 53: why not `fmt.Errorf`?

carpawell (dismissed):
Not sure if we need to add expiration to the lock objects in that PR if the logic gonna be changed anyway: https://github.com/nspcc-dev/neofs-node/commit/9100b0ded922ec3197be90b1152a9b3da586524b

cmd/neofs-cli/modules/container/delete.go
fyrchik: Fixed

fyrchik:
@carpawell we could just disallow zero values when merging that branch.

PR #1560, opened 2022-06-29
morph/client: perform RPC switch and restore in one step

fyrchik:
Otherwise we could switch infinitely if subscription restore has failed.

### DIFF
--- a/pkg/morph/client/client.go
+++ b/pkg/morph/client/client.go
@@ -54,9 +54,11 @@ type Client struct {
 
 	cfg cfg
 
-	endpoints *endpoints
+	endpoints endpoints
 
-	// switching between rpc endpoint lock
+	// switchLock protects endpoints, inactive, and subscription-related fields.
+	// It is taken exclusively during endpoint switch and locked in shared mode
+	// on every normal call.
 	switchLock *sync.RWMutex
 
 	// channel for ws notifications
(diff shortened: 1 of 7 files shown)


pkg/morph/client/client.go
```diff
@@ -54,9 +54,9 @@ type Client struct {
 
 	cfg cfg
 
-	endpoints *endpoints
+	endpoints endpoints
 
-	// switching between rpc endpoint lock
+	// switchLock protects endpoints, inactive, and subscription-related fields.
```
carpawell, line 59: is that change necessary? my idea was to clarify that mutex locks everything (no new RPCs until the switching process is finished not only protecting the fields). could you add that idea to the comment?

fyrchik: Yes, because it protects shared data. Now it is obvious you can't use any of these fields without holding a mutex.
Added a comment about switching process.

carpawell (approved):

pkg/morph/client/fee.go
```diff
@@ -33,19 +24,17 @@ func (x customFees) feeForMethod(method string) (fixedn.Fixed8, bool) {
...
 	defaultFee fixedn.Fixed8
 
-	customFees
-}
-
-// sets default fee for all operations.
-func (x *fees) setDefault(fee fixedn.Fixed8) {
-	x.defaultFee = fee
+	// customFees represents source of customized per-operation fees.
+	customFees map[string]fixedn.Fixed8
```
cthulhu-rider, line 28: Was the purpose of deprecating `customFees` type? If so, can we remove the type?

PR #1581, opened 2022-07-05
Fix/Do not lose API version on forwarding

carpawell:
Forwarded requests contained zero version in their meta header. It did not
allow responding with API statuses (`v0.0` version considered to be older
than `v2.11`) to the forwarding node and, therefore, did not allow analyzing
responses.

### DIFF
--- a/pkg/services/object/get/v2/util.go
+++ b/pkg/services/object/get/v2/util.go
@@ -24,6 +24,7 @@ import (
 	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	versionSDK "github.com/nspcc-dev/neofs-sdk-go/version"
 	"github.com/nspcc-dev/tzhash/tz"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/services/object/get/v2/util.go
```diff
@@ -71,8 +72,14 @@ func (s *Service) toPrm(req *objectV2.GetRequest, stream objectSvc.GetObjectStre
...
 			// once compose and resign forwarding request
 			onceResign.Do(func() {
+				versionV2 := new(refs.Version)
+
+				apiVersion := versionSDK.Current()
+				apiVersion.WriteToV2(versionV2)
+
 				// compose meta header of the local server
 				metaHdr := new(session.RequestMetaHeader)
+				metaHdr.SetVersion(versionV2)
```
cthulhu-rider, line 82: Lets add a function?
```
func writeCurrentVersion(*session.RequestMetaHeader)
```

cthulhu-rider (commented):
@carpawell fix is well done, but how did you catch it? We shouldn't loose version written in the origin header, so even if we don't set version in attached meta (which is incorrect too), we should expect correct version in origin request.

fyrchik (dismissed):

pkg/services/object/get/v2/util.go
carpawell: sure, added

cthulhu-rider (approved):

PR #1570, opened 2022-06-30
Upgrade NeoFS SDK Go with changed reputation API

cthulhu-rider:

PR #1586, opened 2022-07-06
Upgrade SDK to rc.5

cthulhu-rider:
* blocked by #1570

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,8 +19,8 @@ require (
 	github.com/nspcc-dev/neo-go/pkg/interop v0.0.0-20220601120906-3bec6657f5c5 // indirect
 	github.com/nspcc-dev/neofs-api-go/v2 v2.13.0
 	github.com/nspcc-dev/neofs-contract v0.15.1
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -19,8 +19,8 @@ require (
 	github.com/nspcc-dev/neo-go/pkg/interop v0.0.0-20220601120906-3bec6657f5c5 // indirect
 	github.com/nspcc-dev/neofs-api-go/v2 v2.13.0
 	github.com/nspcc-dev/neofs-contract v0.15.1
-	github.com/nspcc-dev/neofs-sdk-go v1.0.0-rc.4.0.20220705084258-cec2373b50ae
-	github.com/nspcc-dev/tzhash v1.5.2
+	github.com/nspcc-dev/neofs-sdk-go v1.0.0-rc.5.0.20220706151041-0d862d8568a4
```
carpawell, line 22: why not pure RC5? i do not mind but commit message and PR naming confuses a little

cthulhu-rider: Because we need hotfixes after rc5.

carpawell: ok but i would highlight that in commit message

fyrchik (approved):

PR #1592, opened 2022-07-08
Do not print build time in `version`

### LINKED ISSUE #1587: Make builds reproducible
Remove build time, ensure proper Go build options are used.
Also, see https://github.com/nspcc-dev/neofs-s3-gw/issues/578

fyrchik:
Close #1587.

This makes our build more reproducible.

2 questions:
1. Do we need `Component: NeoFS CLI`  /  `Component: NeoFS Adm` etc.?
2. Do we need `GoVersion: go1.18`?

I am ok with any option.


carpawell:
> Do we need Component: NeoFS CLI / Component: NeoFS Adm etc.?

I think smth like `Aplication ...` could be useful.

> Do we need GoVersion: go1.18?

I think yes but not critical since it could be deducted from the commit.

fyrchik:
> I think yes but not critical since it could be deducted from the commit.
What do you mean by "could be deducted from the commit"?

fyrchik:
```
Version: v0.29.0-8-g527ff1b9
Component: NeoFS CLI
Debug: false
GoVersion: go1.18.3

Version: v0.29.0-8-g527ff1b9
Component: NeoFS Adm
Debug: false
GoVersion: go1.18.3

Version: v0.29.0-8-g527ff1b9
Component: NeoFS Node
Debug: false
GoVersion: go1.18.3

Version: v0.29.0-8-g527ff1b9
Component: NeoFS IR
Debug: false
GoVersion: go1.18.3

Version: v0.29.0-8-g527ff1b9
Component: NeoFS Lens
Debug: false
GoVersion: go1.18.3
```

realloc:
Suggest renaming 
Component: NeoFS Node → Component: NeoFS Storage node
Component: NeoFS IR → Component: NeoFS Inner Ring node

carpawell:
@realloc, @fyrchik, why do we call it "Component"? Is it smth common? A regular user could only use `neofs-cli`, and does not know about other binaries, why should he know that it is a "component"?

alexvanin:
Agree with @carpawell. 

By the way, can you approve, that these fields are optional except `Version` and `GoVersion`  for other applications ? Some apps doesn't have "components" or debug mode. And, in fact, only these two fields provide actual versions, what is `--version` should be all about.

fyrchik:
The point is to unify the output between different components, so let's agree on the output.
`Debug` can be appended to `Version` during the build time, so I see no problem in removing it.

I have no strong opinion on the `Component` field, as I understood it is useful to the Infra team.

carpawell:
> useful to the Infra team

Sure, I agree that it could be useful. The question was just about naming. I would call that `App`/`Application` or smth common pattern for that output (if it exists).

alexvanin:
How about that? I think Infra team is capable of parsing such output.
```
$ neofs-adm --version
NeoFS Adm
Version: v0.29.0-8-g527ff1b9 (debug)
GoVersion: go1.18.3

$ neofs-http-gw --version
NeoFS HTTP Gateway
Version: v0.21.0
GoVersion: go1.17.12
```

carpawell:
>  I would call that `App`/`Application`...

... or do not name that field at all, yes.

carpawell (dismissed):

alexvanin:
> It would be better, I think, to have `Debug` as a separate field. Adding something to version string may result in script bugs with incorrect `(debug)` part handling, especially if it will appear rarely.

If this field can be omitted in apps like gateways, I am all in.

fyrchik:
Removed debug builds.

alexvanin (approved):

PR #1594, opened 2022-07-08
neofs-adm: support better completion for `set-policy`

fyrchik:


carpawell (approved):
The issue number is missing.

carpawell (approved):

PR #1602, opened 2022-07-13
Refactor metrics and profiler sections in the config

### LINKED ISSUE #1563: replace `127.0.0.1` with `localhost` everywhere

fyrchik:
Close #1563 .

1. Use `localhost` instead of `127.0.0.1` in examples.
2. Explicitly enable both services.
3. Rename `metrics` to `prometheus` and `profiler` to `pprof`. Old sections are still used, but will be removed in future.


carpawell (dismissed):

config/example/ir.yaml
```diff
@@ -124,11 +124,11 @@ contracts:
     zhivete: f584699bc2ff457d339fb09f16217042c1a42101 # Optional: override address of zhivete contract in sidechain
 
 profiler:
-  address: 127.0.0.1:6060  # Endpoint for application pprof profiling; disabled by default
+  address: localhost:6060  # Endpoint for application pprof profiling; disabled by default
   shutdown_timeout: 30s    # Timeout for profiling HTTP server graceful shutdown
 
 metrics:
-  address: 127.0.0.1:9090  # Endpoint for application prometheus metrics; disabled by default
+  address: localhost:9090  # Endpoint for application prometheus metrics; disabled by default
```
alexvanin, line 131: What about adopting `prometheus` and `pprof` in IR, though?

fyrchik: Fixed

cthulhu-rider (approved):

config/example/node.yaml
```diff
@@ -1,12 +1,14 @@
...
-profiler:
-  address: 127.0.0.1:6060  # endpoint for Node profiling
+pprof:
+  enabled: true
+  address: localhost:6060  # endpoint for Node profiling
   shutdown_timeout: 15s  # timeout for profiling HTTP server graceful shutdown
 
-metrics:
-  address: 127.0.0.1:9090  # endpoint for Node metrics
+prometheus:
```
alexvanin, line 9: I think we started renaming because @532910 wanted to unify profiler and metrics configuration among neo-go, neofs components and service application. We decided to adopt neo-go configuration approach, correct me if I am wrong.

However there are some differences but I am not sure if they are worth discussing / synchronizing or not. 
1. neo-go uses upper case for `Pprof` and `Prometheus` and config is case sensitive.
2. neo-go does not define `shutdown_timeout` 
3. neo-go defines this on a second config level (inside `ApplicationConfiguration` section) but there is not much we can do about it, I guess.

On my opinion, the best solution is to keep it as it proposed in this PR. I just want to highlight that it will not be 100% the same in neo-go and other apps.

alexvanin (approved):

config/example/node.yaml
fyrchik: I think this was more about the big picture (`pprof` vs `profiler`). I don't like the idea to mirror `neo-go` configuration _completely_. 

BTW, viper should handle case conversions for us, so `Prometheus` should work too.

carpawell (approved):

config/example/node.yaml
fyrchik: 1. `prometheus` vs `metrics`: neo-go is more widely deployed, it is easier to change sth here. I like `prometheus` more.
2. Capital letters in neo-go is C# inheritance, I like small letters more for many reasons. I think capital letters should work now, it is just not tested.
3. To me single `address` field is more flexible. I would leave it as it is.

cc @532910 @alexvanin @realloc @roman-khimov @carpawell 
:+1: -- use single `address` (current scheme)
:100: -- use `address` + `port`

PR #1609, opened 2022-07-18
Provide priority for the endpoints

### LINKED ISSUE #1572: IR morph endpoints priority
Let's consider adding priority to the morph_rpc client pool, the same way it's done for neofs storage node endpoints in S3 and HTTP gateways.

In on-prem environment, IR node needs a way to use local Neo RPC node and switch to nearest neighbour in case of local node malfunction or overload.

fyrchik:
Close #1572.

After this, the behaviour of multiclient will be the following:
1. All nodes of equal priority are shuffled during the startup, thus iteration is done in random order (fixed at the startup time).
2. When an endpoint fails, try all other endpoints in the order of decreasing priority.

Example:
```
- address: xxx
  priority: 2
- address: yyy
  priority: 1
- address: zzz
  priority: 1
```

1. Chose xxx.
2. xxx fails -> choose yyy (let it be the random)
3. fail -> try xxx, if it is unavailable try zzz (if zzz is unavailable, finish the application).


realloc:
Can we also add a weight parameter, like we have in protocol gateways?

fyrchik:
@realloc how exactly are you going to use it here? It makes less sense for morph: we use a single endpoint and subscribe to it. Using distributing requests between multiple endpoints is error-prone, because these endpoints can have different height.

realloc:
> because these endpoints can have different height.

Agree. Have just remembered us discussing it previously.

config/example/node.env
```diff
@@ -56,11 +56,10 @@ NEOFS_CONTRACTS_PROXY=ad7c6b55b737b696e5c82c85445040964a03e97f
 # Morph chain section
 NEOFS_MORPH_DIAL_TIMEOUT=30s
 NEOFS_MORPH_DISABLE_CACHE=true
-NEOFS_MORPH_RPC_ENDPOINT="wss://rpc1.morph.fs.neo.org:40341/ws wss://rpc2.morph.fs.neo.org:40341/ws"
-
-# Main chain section (optional)
-NEOFS_MAINCHAIN_DIAL_TIMEOUT=30s
-NEOFS_MAINCHAIN_RPC_ENDPOINT="https://rpc1.n3.nspcc.ru:30341 https://rpc2.n3.nspcc.ru:30341"
+NEOFS_MORPH_RPC_ENDPOINT_0_ADDRESS="wss://rpc1.morph.fs.neo.org:40341/ws"
```
carpawell, line 59: could you, please, mention that config changes in the `CHANGELOG`?

cmd/neofs-node/config/morph/config.go
```diff
@@ -28,13 +30,27 @@ const (
...
+	for i := 0; ; i++ {
+		s := sub.Sub(strconv.FormatInt(int64(i), 10))
+		addr := config.StringSafe(s, "address")
+		if addr == "" {
+			break
+		}
+
+		es = append(es, client.Endpoint{
+			Address:  addr,
+			Priority: int(config.IntSafe(s, "priority")),
```
carpawell, line 46: just to be sure: default priority is the lowest, not the highest?

pkg/morph/client/multi.go
```diff
@@ -1,69 +1,46 @@
...
-	}
-
-	for {
-		newEndpoint, index := c.endpoints.next()
-		if index == currEndpointIndex {
-			// all the endpoint have been tried
-			// for connection unsuccessfully
-			return false
+	// Iterate endpoints in the order of decreasing priority.
+	// Skip the current endpoint.
```
carpawell, line 36: do we need to skip the current?

it could be some short network problem but we skip the current one and connect to the less prioritized RPC. also, that code wont try to reconnect at least one time if only one endpoint is presented (well, the previous code wont too, yes, but anyway)

fyrchik: This matches the old behaviour.

config/example/node.env
fyrchik: Fixed

cmd/neofs-node/config/morph/config.go
fyrchik: Fixed.

config/example/node.env
fyrchik: Fixed

carpawell (approved):

Issue #1574, opened 2022-07-01
Unify epoch expiration flag for CLI's commands


carpawell:
Some of them are named `expires-at/on`, and some of them are named `lifetime` (mb there are more options).
Some of them support relative epochs and some do not.

Examples: 
- `storagegroup put`;
- `object put`;
- `bearer create`;
- etc.

carpawell:
@fyrchik, do you mean `expires-at` is about the exact expiration epoch and `lifetime` is about the period? Or is it just about the naming convenience?

fyrchik:
I meant, that it is ok to have different flags for tokens and objects, but object-related flags should be the same.
However, I like your sugesstion, we could remove `+X` syntax in favour of the `lifetime` flag.

PR #1612, opened 2022-07-19
Unify cli flags

### LINKED ISSUE #623: cli: Do not print help on any failure
Any error returned by implemented command executor leads to additional help output. It is useful when the user incorrectly formed a command, but not more.

### Solution
Replace providing the `RunE` calls of to `cobra.Command` (errors of which are actually lead to `help`) with `Run`. With this approach command misprints are still handled, but other errors are written user friendly.
For errors `cmd.PrintErr` function group, for logs - `cmd.Print*` w/o `Err`.

fyrchik:
Related #1574.
Related #623.


cmd/neofs-cli/modules/bearer/create.go
```diff
@@ -35,14 +34,14 @@ All epoch flags can be specified relative to the current epoch with the +n synta
...
-	RunE: createToken,
+	Run: createToken,
 }
 
 func init() {
 	createCmd.Flags().StringP(eaclFlag, "e", "", "path to the extended ACL table")
 	createCmd.Flags().StringP(issuedAtFlag, "i", "", "epoch to issue token at")
 	createCmd.Flags().StringP(notValidBeforeFlag, "n", "", "not valid before epoch")
-	createCmd.Flags().StringP(expireAtFlag, "x", "", "expiration epoch")
+	createCmd.Flags().StringP(commonflags.ExpireAt, "x", "", "expiration epoch")
```
carpawell, line 44: somewhere shorthand is `x` and somewhere is `e`, lets's also sync it

cmd/neofs-cli/modules/bearer/create.go
```diff
@@ -52,33 +51,29 @@ func init() {
...
-	if err != nil {
-		return err
-	}
-	exp, expRelative, err := common.ParseEpoch(cmd, expireAtFlag)
-	if err != nil {
-		return err
-	}
+	common.ExitOnErr(cmd, "can't parse --"+issuedAtFlag+" flag: %w", err)
+
+	exp, expRelative, err := common.ParseEpoch(cmd, commonflags.ExpireAt)
```
carpawell, line 63: does it still support a relative expiration on the `expire-at` flag?

fyrchik: Yes, i think bearer command better stay the way it is, because it has 3 similar flags. Having 6 flags can be too hard for a user.

cmd/neofs-cli/modules/bearer/create.go
fyrchik: It is left so deliberatly, `e` is for `eacl` here, I think equal long names is enough.

fyrchik:
Let's just leave 623 open.

carpawell (approved):

PR #1614, opened 2022-07-19
Feat/Add health metrics

carpawell:
`neofs_node_state_health` metric name. Values have the same meaning as in [control enum](https://github.com/nspcc-dev/neofs-node/blob/198beae703289b7a62bbd38b89cb30ea7a1f87f2/pkg/services/control/types.pb.go#L83).

### DIFF
--- a/pkg/metrics/state.go
+++ b/pkg/metrics/state.go
@@ -1,0 +1,28 @@
+package metrics
+
+import "github.com/prometheus/client_golang/prometheus"
+
+const stateSubsystem = "state"
+
+type stateMetrics struct {
+	healthCheck prometheus.Gauge
+}
+
+func newStateMetrics() stateMetrics {
+	return stateMetrics{
+		healthCheck: prometheus.NewGauge(prometheus.GaugeOpts{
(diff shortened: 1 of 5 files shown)


pkg/metrics/state.go
```diff
@@ -0,0 +1,28 @@
+package metrics
+
+import "github.com/prometheus/client_golang/prometheus"
+
+const stateSubsystem = "state"
```
fyrchik, line 5: Why have you decided to introduce a new hierarchy level and not make it similar to the `epoch` metric?

fyrchik (approved):

pkg/metrics/state.go
carpawell: when i did that i thought that it could be extended with the network info (since they are both presented in `control healthcheck` command) in the future, also, i thought about moving `epoch` to that section too as it could be considered as a state. planned to see your thoughts about that

what do you think?

fyrchik: Ok, I think moving epoch from the top-level would be nice.

PR #1621, opened 2022-07-22
Add batching to the pilorama

### LINKED ISSUE #1447: Batch operations in the tree service
After #1431 we started to use bolt batches for optimizing concurrent operations execution.
However the can gain even more benefit if we also do some batching one level above.
Consider operations with the following timestamps: 4, 3, 5, 2.
Right now it will be like this:
1. do 4
2. undo 4, do 3, do 4
3. do 5
4. undo 5, undo 4, undo 3, do 2, do 3, do 4, do 5

After this PR we should behave as follows (assuming 2 values are in a single batch):
1. do 3, do 4
2. undo 4, undo 3, do 2, do 3, do 4, do 5

fyrchik:
Close #1447 .

@anatoly-bogatyrev tested -- after this change S3 speed is finally on comparable to a gRPC procotocol!

### DIFF
--- a/pkg/local_object_storage/pilorama/batch.go
+++ b/pkg/local_object_storage/pilorama/batch.go
@@ -1,0 +1,49 @@
+package pilorama
+
+import (
+	"sort"
+	"sync"
+	"time"
+
+	cidSDK "github.com/nspcc-dev/neofs-sdk-go/container/id"
+	"go.etcd.io/bbolt"
+)
+
+type batch struct {
+	forest     *boltForest
+	timer      *time.Timer
+	mtx        sync.Mutex
+	start      sync.Once
+	cid        cidSDK.ID
+	treeID     string
+	results    []chan<- error
+	operations []*Move
+}
+
+func (b *batch) trigger() {
+	b.mtx.Lock()
+	if b.timer != nil {
+		b.timer.Stop()
+		b.timer = nil
+	}
+	b.mtx.Unlock()
+	b.start.Do(b.run)
+}
+
+func (b *batch) run() {
+	sort.Slice(b.operations, func(i, j int) bool {
+		return b.operations[i].Time < b.operations[j].Time
+	})
+	err := b.forest.db.Update(func(tx *bbolt.Tx) error {
+		bLog, bTree, err := b.forest.getTreeBuckets(tx, b.cid, b.treeID)
+		if err != nil {
+			return err
+		}
+
+		var lm LogMove
+		return b.forest.applyOperation(bLog, bTree, b.operations, &lm)
+	})
+	for i := range b.operations {
+		b.results[i] <- err
+	}
+}
--- a/pkg/local_object_storage/pilorama/boltdb.go
+++ b/pkg/local_object_storage/pilorama/boltdb.go
@@ -24,6 +24,11 @@ type boltForest struct {
 
 	modeMtx sync.RWMutex
 	mode    mode.Mode
+
+	// mtx protects batches field.
+	mtx     sync.Mutex
+	batches []*batch
+
 	cfg
 }
 
@@ -318,15 +323,64 @@ func (t *boltForest) TreeApply(d CIDDescriptor, treeID string, m *Move, backgrou
 		}
 	}
 
-	return t.db.Batch(func(tx *bbolt.Tx) error {
-		bLog, bTree, err := t.getTreeBuckets(tx, d.CID, treeID)
-		if err != nil {
-			return err
+	if t.db.MaxBatchSize == 1 {
+		return t.db.Update(func(tx *bbolt.Tx) error {
+			bLog, bTree, err := t.getTreeBuckets(tx, d.CID, treeID)
+			if err != nil {
+				return err
+			}
+
+			var lm LogMove
+			return t.applyOperation(bLog, bTree, []*Move{m}, &lm)
+		})
+	}
+
+	ch := make(chan error, 1)
+	t.addBatch(d, treeID, m, ch)
+	return <-ch
+}
+
+func (t *boltForest) addBatch(d CIDDescriptor, treeID string, m *Move, ch chan error) {
+	t.mtx.Lock()
+	for i := 0; i < len(t.batches); i++ {
+		t.batches[i].mtx.Lock()
+		if t.batches[i].timer == nil {
+			t.batches[i].mtx.Unlock()
+			copy(t.batches[i:], t.batches[i+1:])
+			t.batches = t.batches[:len(t.batches)-1]
+			i--
+			continue
 		}
 
-		lm := &LogMove{Move: *m}
-		return t.applyOperation(bLog, bTree, lm)
-	})
+		found := t.batches[i].cid.Equals(d.CID) && t.batches[i].treeID == treeID
+		if found {
+			t.batches[i].results = append(t.batches[i].results, ch)
+			t.batches[i].operations = append(t.batches[i].operations, m)
+			if len(t.batches[i].operations) == t.db.MaxBatchSize {
+				t.batches[i].timer.Stop()
+				t.batches[i].timer = nil
+				t.batches[i].mtx.Unlock()
+				b := t.batches[i]
+				t.mtx.Unlock()
+				b.trigger()
+				return
+			}
+			t.batches[i].mtx.Unlock()
+			t.mtx.Unlock()
+			return
+		}
+		t.batches[i].mtx.Unlock()
+	}
+	b := &batch{
+		forest:     t,
+		cid:        d.CID,
+		treeID:     treeID,
+		results:    []chan<- error{ch},
+		operations: []*Move{m},
+	}
+	b.timer = time.AfterFunc(t.db.MaxBatchDelay, b.trigger)
+	t.batches = append(t.batches, b)
+	t.mtx.Unlock()
 }
 
 func (t *boltForest) getTreeBuckets(tx *bbolt.Tx, cid cidSDK.ID, treeID string) (*bbolt.Bucket, *bbolt.Bucket, error) {
@@ -351,15 +405,10 @@ func (t *boltForest) getTreeBuckets(tx *bbolt.Tx, cid cidSDK.ID, treeID string) 
 	return bLog, bData, nil
 }
 
-func (t *boltForest) applyOperation(logBucket, treeBucket *bbolt.Bucket, lm *LogMove) error {
+// applyOperations applies log operations. Assumes lm are sorted by timestamp.
+func (t *boltForest) applyOperation(logBucket, treeBucket *bbolt.Bucket, ms []*Move, lm *LogMove) error {
 	var tmp LogMove
 	var cKey [17]byte
-
-	var logKey [8]byte
-	binary.BigEndian.PutUint64(logKey[:], lm.Time)
-	if logBucket.Get(logKey[:]) != nil {
-		return nil
-	}
 
 	c := logBucket.Cursor()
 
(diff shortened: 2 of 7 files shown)


acid-ant (dismissed):

pkg/local_object_storage/pilorama/boltdb.go
```diff
@@ -280,15 +298,62 @@ func (t *boltForest) TreeApply(d CIDDescriptor, treeID string, m *Move, backgrou
...
+	for i := 0; i < len(t.batches); i++ {
+		t.batches[i].mtx.Lock()
+		if t.batches[i].timer == nil {
+			t.batches[i].mtx.Unlock()
+			copy(t.batches[i:], t.batches[i+1:])
+			t.batches = t.batches[:len(t.batches)-1]
+			i--
+			continue
+		}
+		valid := t.batches[i].cid.Equals(d.CID) && t.batches[i].treeID == treeID
```
carpawell, line 329: does `valid` mean `found` in that context?

pkg/local_object_storage/pilorama/boltdb.go
```diff
@@ -280,15 +298,62 @@ func (t *boltForest) TreeApply(d CIDDescriptor, treeID string, m *Move, backgrou
...
+		t.batches[i].mtx.Unlock()
+	}
+	b := &batch{
+		db:         t,
+		cid:        d.CID,
+		treeID:     treeID,
+		results:    []chan<- error{ch},
+		operations: []*Move{m},
+	}
+	b.timer = time.AfterFunc(time.Millisecond*20, b.trigger)
```
carpawell, line 354: does 20ms _really_ help? have you tried other values?

carpawell (approved):

pkg/local_object_storage/pilorama/boltdb.go
```diff
@@ -331,45 +391,43 @@ func (t *boltForest) applyOperation(logBucket, treeBucket *bbolt.Bucket, lm *Log
...
+				return err
+			}
+			if err := t.do(logBucket, treeBucket, cKey[:], &tmp); err != nil {
+				return err
+			}
+			key, value = c.Next()
 		}
-		key, value = c.Next()
 	}
-
```
roman-khimov, line 372: Meh.

pkg/local_object_storage/pilorama/boltdb.go
```diff
@@ -280,15 +298,62 @@ func (t *boltForest) TreeApply(d CIDDescriptor, treeID string, m *Move, backgrou
...
+			t.batches = t.batches[:len(t.batches)-1]
+			i--
+			continue
+		}
+		valid := t.batches[i].cid.Equals(d.CID) && t.batches[i].treeID == treeID
+		if valid {
+			t.batches[i].results = append(t.batches[i].results, ch)
+			t.batches[i].operations = append(t.batches[i].operations, m)
+			if len(t.batches[i].operations) == t.db.MaxBatchSize {
+				t.batches[i].timer = nil
```
roman-khimov, line 334: With the timer still ticking in the background?

pkg/services/tree/service.go
```diff
@@ -483,13 +485,16 @@ func (s *Service) Apply(_ context.Context, req *ApplyRequest) (*ApplyResponse, e
...
+	s.replicateLocalCh <- applyOp{
+		treeID:        req.GetBody().GetTreeId(),
+		CIDDescriptor: pilorama.CIDDescriptor{CID: cid, Position: pos, Size: size},
+		Move: pilorama.Move{
+			Parent: op.GetParentId(),
+			Child:  op.GetChildId(),
+			Meta:   meta,
+		},
+	}
+	return &ApplyResponse{Body: &ApplyResponse_Body{}, Signature: &Signature{}}, nil
```
roman-khimov, line 497: So the caller won't know if `TreeApply` is to fail?

pkg/services/tree/replicator.go
```diff
@@ -36,6 +42,12 @@ func (s *Service) replicationWorker() {
 		select {
 		case <-s.closeCh:
 			return
+		case op := <-s.replicateLocalCh:
+			err := s.forest.TreeApply(op.CIDDescriptor, op.treeID, &op.Move, false)
+			if err != nil {
+				s.log.Error("failed to apply replicated operation",
+					zap.String("err", err.Error()))
+			}
 		case task := <-s.replicationTasks:
```
roman-khimov, line 51: So we have some network (taking loooong time) operations here...

pkg/services/tree/replicator.go
```diff
@@ -36,6 +42,12 @@ func (s *Service) replicationWorker() {
 		select {
 		case <-s.closeCh:
 			return
+		case op := <-s.replicateLocalCh:
```
roman-khimov, line 45: ... and local ones (supposed to be fast) here? What's gonna happen if we're waiting for the network operation to finish (5s * number of endpoints)?

roman-khimov (commented):
What happens if we lose a batch of operations, how will this affect state consistency wrt objects available/their versions?

pkg/local_object_storage/pilorama/boltdb.go
roman-khimov: What's the expected number of operations per second and `MaxBatchSize`?

fyrchik: What do you mean by `20ms really help`?
The purpose of this PR is to batch operations _over a single container_, this is what helps.
Actually, I don't remember where 20ms came from (10ms is default in bbolt). Anyway, we should have `db.MaxBatchDelay` here.

Fixed

pkg/local_object_storage/pilorama/boltdb.go
```diff
@@ -23,6 +23,13 @@ type boltForest struct {
 
 	modeMtx sync.Mutex
 	mode    mode.Mode
+
+	// mtx protects batches field.
+	mtx     sync.Mutex
+	batches []*batch
```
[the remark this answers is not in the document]
fyrchik: Do you mean why we use custom batches instead of `Batch`?

Inserting operations in a log is costly: we need to rollback and reapply some operations. If we insert 2 operations, with this PR we need to rollback+reapply only once, because of the custom logic.

pkg/services/tree/service.go
fyrchik: Yes, it is not important:
1. The caller has no real options to handle this error.
2. Background synchronization will do it's job and sync any missing operation.

pkg/services/tree/replicator.go
fyrchik: Yeah, actually it is even worse: we could deadlock here under a very high load. Fixed.

pkg/local_object_storage/pilorama/boltdb.go
fyrchik: Fixed

pkg/local_object_storage/pilorama/boltdb.go
```diff
@@ -110,9 +117,16 @@ func (t *boltForest) Open(readOnly bool) error {
...
+	if t.closeCh != nil {
+		return nil
+	}
+
+	t.closeCh = make(chan struct{})
+
 	if t.db.IsReadOnly() {
 		return nil
 	}
+
```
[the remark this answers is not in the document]
fyrchik: Fixed

pkg/local_object_storage/pilorama/boltdb.go
fyrchik: Fixed

pkg/local_object_storage/pilorama/boltdb.go
fyrchik: Fixed

pkg/local_object_storage/pilorama/boltdb.go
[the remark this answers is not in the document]
roman-khimov: I'm rather talking about `DB.Begin()` and managing DB transactions manually. It'd be much simpler, but of course the result depends on how effective your time sorting in `(*batch).run` is. My expectation would be that it's not critical (so that most of the time operations are already well-ordered), but this entirely depends on the operations flow.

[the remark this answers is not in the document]
fyrchik: The operation flow is comprised from 2 components:
1. Background sync.
2. Replication of user operations.

While operations from (2) should be near the end of the log, under high load the order of the last e.g. 100 operations is completely random.

Could you elaborate about `DB.Begin()`? How does it solve a problem?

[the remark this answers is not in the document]
roman-khimov: Whether it solves the problem or not depends on the core issue. If it's DB sync (which would be my primary suspect), then `DB.Begin()` can help a lot, because you'll open up a transaction, push a set of changes into it and then commit (with the same logic, either `MaxBatchSize` or `MaxBatchDelay`). If it's rollbacks, then sure, transactions may be of little help.

pkg/local_object_storage/pilorama/boltdb.go
carpawell: > What do you mean by 20ms really help?

i meant that i can say nothing about that duration (is it short or long) so i am interested if you tried any other values and if you measured how different values affect performance, if yes, i would add a constant with a comment or make that configurable (`db.MaxBatchDelay` is also ok for me). in other words, "What's the expected number of operations per second?" as @roman-khimov asked.

pkg/local_object_storage/pilorama/boltdb.go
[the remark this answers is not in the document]
fyrchik: The issue is that inserting N operation in random order has O(N^2) complexity.
This commit ammortizes the cost, so that if N operations are grouped in a single batch we have O(N) complexity.
For `append` I think native `bbolt.Batch` already does the job.

Basically, we block for too much time (because of multiple ineficcient undo+do in a row).
Sync can also be an issue, but it is not obvious how to approach it (we already use native `bbolt.Batch` everywhere).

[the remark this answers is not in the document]
roman-khimov: > For append I think native bbolt.Batch already does the job.

I wouldn't be so sure.

> Concurrent Batch calls are opportunistically combined into larger
transactions. Batch is only useful when there are multiple goroutines
calling it.

In fact, if you're doing all changes in a single thread it slows you down.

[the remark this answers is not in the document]
fyrchik: We do it in different goroutines (each tree service RPC is executed in parallel)

pkg/local_object_storage/pilorama/batch.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	b.mtx.Unlock()
+	b.start.Do(b.run)
+}
+
+func (b *batch) run() {
+	sort.Slice(b.operations, func(i, j int) bool {
+		return b.operations[i].Time < b.operations[j].Time
+	})
+	err := b.db.db.Update(func(tx *bbolt.Tx) error {
```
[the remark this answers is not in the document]
fyrchik: Fixed

pkg/local_object_storage/pilorama/boltdb.go
fyrchik: We have no expected number of operations, because this should work in any circumstances. I think bolt default is fine here. I am also thinking about making this setting different from the "normal batch size/delay", but we need tests for this, will do in another PR if it helps.

carpawell (approved):
No changelog.

acid-ant (dismissed):

pkg/local_object_storage/pilorama/boltdb.go
roman-khimov: You have to have some expectations, this service has some performance limits whether you like it or not. BoltDB itself is not infinite wrt to how many reads/writes per second it can deliver. If you know your limits and expectations, you can tune parameters like this better (or evaluate other approaches).

PR #1627, opened 2022-07-26
Refactor/Add tree service to the config framework

carpawell:


config/example/node.yaml
```diff
@@ -60,10 +60,10 @@ grpc:
...
 
 tree:
-  enabled: false
-  cache_size: 10
-  replication_worker_count: 64
-  replication_channel_capacity: 64
+  enabled: true
+  cache_size: 15
+  replication_worker_count: 32
+  replication_channel_capacity: 32
```
fyrchik, line 66: The support is experimental, I don't think we should `enabled: true` in the example.

cmd/neofs-node/config/tree/config.go
```diff
@@ -0,0 +1,77 @@
...
+	// DefaultCacheSize is a default cache size of the tree service.
+	DefaultCacheSize = 10
+
+	// DefaultReplicationChannelCapacity is a default tree service
+	// replication channel capacity.
+	DefaultReplicationChannelCapacity = 64
+
+	// DefaultReplicationWorkerCount is a default tree service
+	// replication worker count.
+	DefaultReplicationWorkerCount = 64
```
fyrchik, line 17: Let's not introduce defaults also here? We already have some in the service itself.

carpawell: ok, made them return 0

config/example/node.yaml
carpawell: yes, the only reason why i did so is unit tests that distinguish default and non-default values

is it ok to keep the default values in examples?

carpawell: i meant `enabled: false`

we can not test that if our example files contain default values

fyrchik: Ok, I think we can make it `true`, we have `testnet`/`mainnet` examples anyway.

fyrchik (approved):

PR #1631, opened 2022-07-27
Fix neofs-adm group processing during an update

fyrchik:
1. Assign the group to all contracts.
2. Assign group to the alphabet ones too.


cmd/neofs-adm/internal/modules/morph/initialize_deploy.go
```diff
@@ -188,7 +189,15 @@ func (c *initializeContext) updateContracts() error {
 
 		params := c.getAlphabetDeployItems(i, len(c.Wallets))
 		emit.Array(w.BinWriter, params...)
-		emit.Opcodes(w.BinWriter, opcode.LDSFLD1, opcode.LDSFLD0)
+
+		alphaCs.Manifest.Groups = baseGroups
```
carpawell, line 193: could you, please, clarify a little why we first do `baseGroups := alphaCs.Manifest.Groups`, do nothing with this vars and then do `alphaCs.Manifest.Groups = baseGroups`

carpawell (approved):

PR #1633, opened 2022-07-27
Allow to deploy arbitrary contracts with neofs-adm

### LINKED ISSUE #1629: Make neofs-adm deploy additional smart contracts
We need to have `neofs-adm morph deploy` command to deploy a set of custom smart contracts, the same way as we do for `neofs-contracts` basic pack. This feature is required for on-prem installations having extended functionality deployed with Alphabet multisig.

fyrchik:
Close #1629 .
Tests are here https://github.com/nspcc-dev/neofs-adm-tests/blob/main/test.deploy-single.sh

### DIFF
--- a/cmd/neofs-adm/internal/modules/morph/deploy.go
+++ b/cmd/neofs-adm/internal/modules/morph/deploy.go
@@ -21,6 +21,7 @@ import (
 const (
 	contractPathFlag = "contract"
 	updateFlag       = "update"
+	customZoneFlag   = "domain"
 )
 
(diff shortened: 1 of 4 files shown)


realloc:
Can we also add a custom domain option for custom contracts? That's not a requirement, just a minor UX improvement.

fyrchik:
Do you mean name or both name and TLD?

realloc:
Yes, allow setting custom ЕДЫ instead of default `.neofs`

carpawell (approved):

PR #1634, opened 2022-07-27
Fix/Do not return expired objects

### LINKED ISSUE #1617: Check expiration on `Get`/`GetRange`/etc
Now there is a gap b/w `NewEpoch` event and an object expiration (marking with GCMark). It leads to confusing behavior: the object is expired but still available until GC comes and marks it as garbage.

We could check the expiration on every "get" request.

carpawell:
Closes #1617.

### DIFF
--- a/pkg/local_object_storage/metabase/db.go
+++ b/pkg/local_object_storage/metabase/db.go
@@ -24,6 +24,13 @@ type matcher struct {
 	matchBucket func(*bbolt.Bucket, string, string, func([]byte, []byte) error) error
 }
 
+// EpochState is an interface that provides access to the
+// current epoch number.
+type EpochState interface {
+	// CurrentEpoch must return current epoch height.
+	CurrentEpoch() uint64
+}
+
 // DB represents local metabase of storage node.
 type DB struct {
 	*cfg
@@ -50,6 +57,8 @@ type cfg struct {
 	info Info
 
 	log *logger.Logger
+
+	epochState EpochState
 }
 
 func defaultCfg() *cfg {
@@ -69,6 +78,10 @@ func New(opts ...Option) *DB {
 
 	for i := range opts {
 		opts[i](c)
+	}
+
+	if c.epochState == nil {
+		panic("metabase: epoch state is not specified")
 	}
 
 	return &DB{
@@ -309,5 +322,12 @@ func WithMaxBatchDelay(d time.Duration) Option {
 		if d != 0 {
 			c.boltBatchDelay = d
 		}
+	}
+}
+
+// WithEpochState return option to specify a source of current epoch height.
+func WithEpochState(s EpochState) Option {
+	return func(c *cfg) {
+		c.epochState = s
 	}
 }
--- a/pkg/local_object_storage/metabase/delete.go
+++ b/pkg/local_object_storage/metabase/delete.go
@@ -58,9 +58,10 @@ func (db *DB) Delete(prm DeletePrm) (DeleteRes, error) {
 
 func (db *DB) deleteGroup(tx *bbolt.Tx, addrs []oid.Address) error {
 	refCounter := make(referenceCounter, len(addrs))
+	currEpoch := db.epochState.CurrentEpoch()
 
 	for i := range addrs {
-		err := db.delete(tx, addrs[i], refCounter)
+		err := db.delete(tx, addrs[i], refCounter, currEpoch)
 		if err != nil {
 			return err // maybe log and continue?
 		}
@@ -78,7 +79,7 @@ func (db *DB) deleteGroup(tx *bbolt.Tx, addrs []oid.Address) error {
 	return nil
 }
 
-func (db *DB) delete(tx *bbolt.Tx, addr oid.Address, refCounter referenceCounter) error {
+func (db *DB) delete(tx *bbolt.Tx, addr oid.Address, refCounter referenceCounter, currEpoch uint64) error {
 	// remove record from the garbage bucket
 	garbageBKT := tx.Bucket(garbageBucketName)
 	if garbageBKT != nil {
@@ -89,7 +90,7 @@ func (db *DB) delete(tx *bbolt.Tx, addr oid.Address, refCounter referenceCounter
 	}
 
 	// unmarshal object, work only with physically stored (raw == true) objects
-	obj, err := db.get(tx, addr, false, true)
+	obj, err := db.get(tx, addr, false, true, currEpoch)
 	if err != nil {
 		if errors.As(err, new(apistatus.ObjectNotFound)) {
 			return nil
--- a/pkg/local_object_storage/metabase/exists.go
+++ b/pkg/local_object_storage/metabase/exists.go
@@ -3,7 +3,10 @@ package meta
 import (
 	"errors"
 	"fmt"
+	"strconv"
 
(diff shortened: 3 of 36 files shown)


pkg/local_object_storage/metabase/exists.go
```diff
@@ -52,7 +57,7 @@ func (db *DB) Exists(prm ExistsPrm) (res ExistsRes, err error) {
 
 func (db *DB) exists(tx *bbolt.Tx, addr oid.Address) (exists bool, err error) {
 	// check graveyard first
-	switch inGraveyard(tx, addr) {
+	switch objectStatus(tx, addr, db.epochState.CurrentEpoch()) {
```
fyrchik, line 60: Calling this inside the transaction can introduce unexpected logical errors: `CurrentEpoch` can return different values inside a single transaction. Can we provide this as a parameter instead?

pkg/local_object_storage/metabase/exists.go
```diff
@@ -86,14 +93,50 @@ func (db *DB) exists(tx *bbolt.Tx, addr oid.Address) (exists bool, err error) {
...
+
+	// bucket with objects that have expiration attr
+	expirationBucket := tx.Bucket(attributeBucketName(addr.Container(), objectV2.SysAttributeExpEpoch))
+	if expirationBucket != nil {
+		// bucket that contains objects that expire in the current epoch
+		currEpochBkt := expirationBucket.Bucket([]byte(strconv.FormatUint(currEpoch, 10)))
+		if currEpochBkt != nil {
+			// we can do nothing with errors from db
+			// => consider object as not expired
+			_ = currEpochBkt.ForEach(func(k, _ []byte) error {
```
fyrchik, line 120: Why simple `Get` doesn't work?

pkg/local_object_storage/metabase/exists.go
```diff
@@ -86,14 +93,50 @@ func (db *DB) exists(tx *bbolt.Tx, addr oid.Address) (exists bool, err error) {
...
+
+	if expired {
+		return 3
+	}
+
 	graveyardBkt := tx.Bucket(graveyardBucketName)
 	garbageBkt := tx.Bucket(garbageBucketName)
+
 	addrKey := addressKey(addr)
+
```
fyrchik, line 139: unrelated to the commit

pkg/local_object_storage/metabase/exists.go
fyrchik: Also, storage errors _should_ be returned to the caller.

pkg/local_object_storage/metabase/exists.go
```diff
@@ -86,14 +93,50 @@ func (db *DB) exists(tx *bbolt.Tx, addr oid.Address) (exists bool, err error) {
...
+	// expired previously for less than the one epoch duration
+
+	rawOID := []byte(addr.Object().EncodeToString())
+	var expired bool
+
+	// bucket with objects that have expiration attr
+	expirationBucket := tx.Bucket(attributeBucketName(addr.Container(), objectV2.SysAttributeExpEpoch))
+	if expirationBucket != nil {
+		// bucket that contains objects that expire in the current epoch
+		currEpochBkt := expirationBucket.Bucket([]byte(strconv.FormatUint(currEpoch, 10)))
```
fyrchik, line 116: What about the objects that expired in the previous epoch?

pkg/local_object_storage/writecache/init.go
```diff
@@ -63,7 +67,15 @@ func (c *cache) isFlushed(addr oid.Address) bool {
 	existsPrm.SetAddress(addr)
 
 	mRes, err := c.metabase.Exists(existsPrm)
-	if err != nil || !mRes.Exists() {
+	if err != nil {
+		if errors.Is(err, object.ErrObjectIsExpired) || errors.As(err, new(apistatus.ObjectAlreadyRemoved)) {
+			return true
+		}
+
+		return false
```
fyrchik, line 75: `return errors.Is ...` ?

pkg/local_object_storage/metabase/db.go
```diff
@@ -322,3 +322,10 @@ func WithMaxBatchDelay(d time.Duration) Option {
 		}
 	}
 }
+
+// WithEpochState return option to specify a source of current epoch height.
```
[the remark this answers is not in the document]
carpawell: moved that change to the previous commit

pkg/local_object_storage/metabase/exists.go
carpawell: sure, fixed (although it required many changes in private methods)

pkg/local_object_storage/metabase/exists.go
carpawell: > Why simple Get doesn't work?

no idea, fixed

> Also, storage errors should be returned to the caller.

yes, but that func returns just `uint8`. fixed to `Get` call anyway

pkg/local_object_storage/metabase/exists.go
carpawell: ok, removed that new lines

pkg/local_object_storage/metabase/exists.go
carpawell: i've added comment to that func:

```
// we check only if the object is expired in the current
// epoch since it is considered the only corner case: the
// GC is expected to collect all the objects that have
// expired previously for less than the one epoch duration
```

i expect that GC is able to collect all the expired objects in one epoch (one hour currently), so the only problem objects are the object that expire in the current epoch but have not been collected by the GC yet. theoretically, we could check more than one epoch (we do not have index that could say what objects are expired, only the objects that has *exact* expiration epoch), so we should know how deep to check OR we should add new expiration index

i thought about force new epoch event that could lead to some problems. we may add some threshold for that cases (e.g. check if an object was not expired in the last `5` epoch)

what do you think?

pkg/local_object_storage/metabase/delete.go
```diff
@@ -89,7 +90,7 @@ func (db *DB) delete(tx *bbolt.Tx, addr oid.Address, refCounter referenceCounter
 	}
 
 	// unmarshal object, work only with physically stored (raw == true) objects
-	obj, err := db.get(tx, addr, false, true)
+	obj, err := db.get(tx, addr, false, true, currEpoch)
```
fyrchik, line 93: How do we ever delete expired object from the metabase?

carpawell: third arg (`false` here) controls checking object status. if an object is presented, it will be deleted, despite its status

KirillovDenis:
Resolve conflicts please

pkg/local_object_storage/metabase/db.go
```diff
@@ -311,3 +322,10 @@ func WithMaxBatchDelay(d time.Duration) Option {
...
+func WithEpochState(s epochState) Option {
```
cthulhu-rider, line 327: `epochState` is needed to be exported.

cthulhu-rider (dismissed):

pkg/local_object_storage/metabase/db.go
carpawell: fixed

cthulhu-rider (approved):

PR #1638, opened 2022-07-29
shard: fix `GetRange` for objects stored in the write-cache

### LINKED ISSUE #1635: Sometimes "object hash --range" and "object range" return response for entire object
1. Create container without specifying ACL
2. Put object to the container
3. Put same object to the container with headers
4. Call `object hash --range 0:10` on object from p.3. Sometimes it returns hash of entire object rather than hash of specified range.
5. Call `object range --range 0:10` on object from p.3. Sometimes it returns content of entire object rather than hash of specified range.

Please, find attached [calls of CLI commands](https://github.com/nspcc-dev/neofs-node/files/9209076/cli-log.txt)

## Your Environment
The issue was not observed on devenv, but occurs in cloud deployment of the system

* Version used:
v0.30.0

fyrchik:
Close #1635 

### DIFF
--- a/pkg/local_object_storage/blobstor/blobovnicza.go
+++ b/pkg/local_object_storage/blobstor/blobovnicza.go
@@ -628,7 +628,7 @@ func (b *blobovniczas) getObjectRange(blz *blobovnicza.Blobovnicza, prm GetRange
 	to := from + prm.rng.GetLength()
 	payload := obj.Payload()
 
-	if uint64(len(payload)) < to {
+	if pLen := uint64(len(payload)); to < from || pLen < from || pLen < to {
 		var errOutOfRange apistatus.ObjectOutOfRange
 
 		return GetRangeSmallRes{}, errOutOfRange
--- a/pkg/local_object_storage/shard/range.go
+++ b/pkg/local_object_storage/shard/range.go
@@ -3,6 +3,8 @@ package shard
 import (
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobovnicza"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor"
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/writecache"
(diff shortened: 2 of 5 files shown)


pkg/local_object_storage/shard/range.go
```diff
@@ -102,8 +104,26 @@ func (s *Shard) GetRange(prm RngPrm) (RngRes, error) {
...
+	wc := func(c writecache.Cache) (*object.Object, error) {
+		res, err := c.Get(prm.addr)
+		if err != nil {
+			return nil, err
+		}
+
+		payload := res.Payload()
+		from := rng.GetOffset()
+		to := from + rng.GetLength()
+		if uint64(len(payload)) < to {
```
carpawell, line 116:
```suggestion
		if uint64(len(payload)) < from || uint64(len(payload)) < to {
```
?

fyrchik: Fixed

pkg/local_object_storage/blobstor/blobovnicza.go
```diff
@@ -628,7 +628,7 @@ func (b *blobovniczas) getObjectRange(blz *blobovnicza.Blobovnicza, prm GetRange
 	to := from + prm.rng.GetLength()
 	payload := obj.Payload()
 
-	if uint64(len(payload)) < to {
+	if pLen := uint64(len(payload)); to < from || pLen < from || pLen < to {
```
carpawell, line 631: `to` equals `from` (uint) + `len` (uint) so i doubt it could be less than `from`

carpawell (approved):

pkg/local_object_storage/blobstor/blobovnicza.go
fyrchik: Try `uint64(10) + math.MaxUint64`

carpawell: oh, overflow...

PR #1643, opened 2022-07-30
innerring: exit if we cannot bind to the control endpoint

### LINKED ISSUE #1641: if control port is busy, node silently hangs
# neofs-ir --version                                 
Version: v0.28.1

fyrchik:
Close #1641 .

Return listen errors in a synchronous fashion.
Another solution would be to use buffered channel, but this is not
scalable: for each new similar runner we would need to extend the
buffer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Changelog for NeoFS Node
 ### Fixed
 
 - Losing request context in eACL response checks (#1595)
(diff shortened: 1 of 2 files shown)


carpawell:
Does that bug also mean that if we _somehow_ got some errors almost simultaneously, we won't stop the app due to the unbuffered err channel?

KirillovDenis (approved):

realloc (approved):

carpawell (approved):

carpawell (approved):

Issue #1615, opened 2022-07-21
Consider automatic swtich to the morph RPC with the highest priority


fyrchik:
After our main endpoint has gone down, we would like to eventually switch back to it even if there are no errors with the other endpoint.

carpawell:
What should be the criteria for the switch? Just time duration? Should it be configured?

If the client has a few endpoints with the same (highest) priority, should it try to connect to the first one after internal sorting? Or all the endpoints with the highest priority should be considered equal?

PR #1648, opened 2022-08-02
Fix/Change morph priority order

carpawell:
To mirror the behaviour of the gateways.


KirillovDenis (dismissed):

carpawell:
Changed: `0`/default -> `1` priority.

cmd/neofs-node/config/morph/config.go
```diff
@@ -41,9 +44,14 @@ func RPCEndpoint(c *config.Config) []client.Endpoint {
 			break
 		}
 
+		priority := int(config.IntSafe(s, "priority"))
+		if priority == 0 {
```
KirillovDenis, line 48: Shouldn't we use `if priority <= 0`?

config/example/node.env
```diff
@@ -63,9 +63,8 @@ NEOFS_CONTRACTS_PROXY=ad7c6b55b737b696e5c82c85445040964a03e97f
 NEOFS_MORPH_DIAL_TIMEOUT=30s
 NEOFS_MORPH_DISABLE_CACHE=true
 NEOFS_MORPH_RPC_ENDPOINT_0_ADDRESS="wss://rpc1.morph.fs.neo.org:40341/ws"
-NEOFS_MORPH_RPC_ENDPOINT_0_PRIORITY=2
 NEOFS_MORPH_RPC_ENDPOINT_1_ADDRESS="wss://rpc2.morph.fs.neo.org:40341/ws"
-NEOFS_MORPH_RPC_ENDPOINT_1_PRIORITY=1
+NEOFS_MORPH_RPC_ENDPOINT_1_PRIORITY=2
```
KirillovDenis, line 67: Let's add priority to the `ENDPOINT_0` explicitly too

docs/storage-node-configuration.md
```diff
@@ -146,10 +146,10 @@ morph:
...
-| `priority` | `int`    | `0`           | Priority of an endpoint. Endpoint with a higher priority has more chance of being used. Endpoints with equal priority are iterated over randomly. |
+| Parameter  | Type     | Default value | Description                                                                                                                                                                   |
+|------------|----------|---------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `address`  | `string` |               | _WebSocket_ N3 endpoint.                                                                                                                                                      |
+| `priority` | `int`    | `1`           | Priority of an endpoint. Endpoint with a higher priority (lower configuration value) has more chance of being used. Endpoints with equal priority are iterated over randomly. |
```
KirillovDenis, line 152: As I understand priority - "Endpoint with a higher priority `must be / will be` used until it's healthy" rather than `has more chance`

pkg/innerring/innerring.go
```diff
@@ -959,16 +959,24 @@ func createClient(ctx context.Context, p *chainParams, errChan chan<- error) (*c
...
+
 	section := p.name + ".endpoint.client"
 	for i := 0; ; i++ {
 		addr := p.cfg.GetString(fmt.Sprintf("%s.%d.%s", section, i, "address"))
 		if addr == "" {
 			break
 		}
 
+		priority := p.cfg.GetInt(section + ".priority")
+		if priority == 0 {
```
KirillovDenis, line 973:
```suggestion
		if priority <= 0 {
```

docs/storage-node-configuration.md
carpawell: that sentence means that it is possible to have more than one endpoint with the same priority; they are shuffled so it is not guaranteed that the first gonna be used

moreover, we are [planning](https://github.com/nspcc-dev/neofs-node/issues/1615) to have some mechanism that would try to return to the most prioritized endpoint but it is not implemented yet. so it is all about the probability

config/example/node.env
carpawell: added zero priority to the `0` endpoint

docs/storage-node-configuration.md
realloc: Weighted probability

KirillovDenis (approved):

PR #1654, opened 2022-08-03
Fix/Do not print false warning about deprecated config

carpawell:


cmd/neofs-ir/main.go
```diff
@@ -141,10 +142,18 @@ func initHTTPServers(cfg *viper.Viper, log *logger.Logger) []*httputil.Server {
...
 		if timeout == 0 {
 			timeout = cfg.GetDuration(item.oldPrefix + ".shutdown_timeout")
-			printDeprecatedMsg = true
+			if timeout != 0 {
+				printDeprecatedMsg = true
+			}
+		}
+
+		if timeout == 0 {
+			timeout = time.Second * defaultTimeout
```
fyrchik, line 156: Why not make `defaultTimeout * time.Second` the constant itself?

cmd/neofs-ir/defaults.go
```diff
@@ -38,10 +38,8 @@ func defaultConfiguration(cfg *viper.Viper) {
 	cfg.SetDefault("logger.level", "info")
 
 	cfg.SetDefault("profiler.address", "localhost:6060")
-	cfg.SetDefault("profiler.shutdown_timeout", "30s")
 
 	cfg.SetDefault("metrics.address", "localhost:9090")
```
fyrchik, line 42: This used the old values because of the need to support both sections.
I think we can remove already deprecated sections in the next release and simplify the code considerably.

carpawell:
@fyrchik, changed the idea of the PR: just dropped the deprecated naming.

KirillovDenis (approved):

realloc (approved):

PR #1672, opened 2022-08-09
Fix/Panic on write-cache `Delete` operation

### LINKED ISSUE #1664: Storage node panics after deleting object via s3 gateway
We run test to copy object within single s3 gateway. The test fails with error described at https://github.com/nspcc-dev/neofs-s3-gw/issues/649
Based on the logs s3 gateway routes request to buky storage node and it fails with panic error.

Interestingly, in several runs the error occurs only with boto3, but not with aws cli.

1. Create a new s3 bucket
2. Put 2 objects into the bucket
3. Copy first object within the bucket
4. Delete the copied object from the bucket.

## Expected Behavior
The object is deleted.

## Current Behavior
The request to s3 gateway fails with error and buky storage node fails with panic error.
Please, find attached [log](https://github.com/nspcc-dev/neofs-node/files/9270850/buky_log-storage_1610.txt) of buky storage server. The panic can be observed at 16:13:20, 16:14:00, 17:40:39

Timestamps 16:13:20, 16:14:00 are related to the first run of the test.
Timestamp 17:40:39 is related to the second run of the test.

## Your Environment
4 cloud VMs with 4 storage nodes and 4 s3 gateways.
On each VM s3 gateway is configured to use storage node on the same VM with priority=1 and fallback to other storage nodes with priority=2.

* Version used:
v0.30.2

carpawell:
If an object is found in the Write-cache and is placed at the end of
the in-memory cache, the memory counter update operation tries to dereference
the index that is out of the sliced array. Moreover, even if panic does not appear,
the counter is updated with the wrong value.

Closes #1664. The wrong place for that line has been chosen in #568.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Changelog for NeoFS Node
 ### Changed
 
 ### Fixed
+- Panic on write-cache's `Delete` operation (#1664)
 
 ### Removed
 
--- a/pkg/local_object_storage/writecache/delete.go
+++ b/pkg/local_object_storage/writecache/delete.go
@@ -26,9 +26,9 @@ func (c *cache) Delete(addr oid.Address) error {
 	c.mtx.Lock()
 	for i := range c.mem {
 		if saddr == c.mem[i].addr {
+			c.curMemSize -= uint64(len(c.mem[i].data))
 			copy(c.mem[i:], c.mem[i+1:])
 			c.mem = c.mem[:len(c.mem)-1]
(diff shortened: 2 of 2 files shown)


anatoly-bogatyrev (dismissed):

cthulhu-rider:
@carpawell pls don't forget to add the fix into support releases.

cthulhu-rider (approved):

carpawell:
> Btw UPD changelog?

Updated

anatoly-bogatyrev (approved):

PR #1694, opened 2022-08-17
Update logo for dark theme

mike-petrov:
Signed-off-by: Mikhail Petrov [mike@nspcc.ru](mailto:mike@nspcc.ru)


mike-petrov:
I added the attributes of dark github theme `[data-color-mode=dark][data-dark-theme*=dark]`, but in the README.md we can not use svg (only img, for security reasons) so the attributes will not work.

Issue #1652, opened 2022-08-03
`neofs-cli container set-eacl` hangs without error report


aprasolova:
Sometimes when I run `neofs-cli container set-eacl --await ...` it hangs for 30 seconds, then fails without any error message neither in console nor in SN log. CLI exit code is 0.

```
$ neofs-cli --rpc-endpoint s01.neofs.devenv:8080 --wallet wallet.json container set-eacl \
    --cid TitkoZ6znKvAEpxSMevWeayo8YJVGzeFKXYMdV1ux4T \
    --table eacl_table.json --config empty_passwd.yml --await 	
awaiting...
```
The command was run at 18:51:50.459. These log messages related to eACL setting I have found in the `s01.neofs.devenv` log
```
2022-08-03T15:51:50.896Z	debug	client/notary.go:522	notary request invoked	{"method": "setEACL", "valid_until_block": 800, "fallback_valid_for": 40, "tx_hash": "a08d4fc58f4b8af831ea02fb324a72d560b9c1b3f711d840e0dce1fb2d272f3e"}
2022-08-03T15:51:51.522Z	debug	event/listener.go:319	event parser not set	{"script hash LE": "4da19a534afb5578d7739f688733ab8289215507", "event type": "SetEACLSuccess"}
2022-08-03T15:52:20.916Z	debug	client/notary.go:522	notary request invoked	{"method": "setEACL", "valid_until_block": 800, "fallback_valid_for": 40, "tx_hash": "48584209947be355a4ca961b3271c306817e4544ea91042f3fabde09254aa5bc"}
2022-08-03T15:52:21.579Z	debug	event/listener.go:319	event parser not set	{"script hash LE": "4da19a534afb5578d7739f688733ab8289215507", "event type": "SetEACLSuccess"}
```
When I requested two transactions from log from `morph-chain.neofs.devenv`, got no response
```
$ curl -XPOST -d@data http://morph-chain.neofs.devenv:30333
{"id":1,"jsonrpc":"2.0","error":{"code":-100,"message":"Unknown transaction"}}
$ cat data
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "getrawtransaction",
  "params": ["0x48584209947be355a4ca961b3271c306817e4544ea91042f3fabde09254aa5bc"]
}
```

## Your Environment
- neofs-dev-env `374724f04f4de8bdd494ac9d6a9319071b20406a`
- neofs-cli `v0.30.2`

[...]

PR #1696, opened 2022-08-17
cli/container: Pre-check ACL extensibility in `set-eacl` cmd

cthulhu-rider:
* related #1652

This doesn't improve UX of the async command `set-eacl` completely, but covers predictable cases.


cmd/neofs-cli/modules/container/set_eacl.go
```diff
@@ -36,6 +41,19 @@ Container ID in EACL table will be substituted with ID from the CLI.`,
 		pk := key.GetOrGenerate(cmd)
 		cli := internalclient.GetSDKClientByFlag(cmd, pk, commonflags.RPC)
 
+		if flagVarsSetEACL.preCheck {
```
fyrchik, line 44: It is hard to know about the problem, even having the description of this flag.
What about replacing it with `--no-precheck`? This way user will be satisfied by default and infra team can provide the flag to speed up things.
What do you think? @aprasolova @carpawell @realloc @532910

fyrchik (commented):
Other than the comment, LGTM

cmd/neofs-cli/modules/container/set_eacl.go
realloc: Good idea. Checking by default and having a flag to change this behaviour looks logical.

carpawell: agree

cthulhu-rider: Inverted the flag.

PR #1712, opened 2022-08-19
Feat/Object count metrics

carpawell:
```
# HELP neofs_node_object_counter Objects counters per shards
# TYPE neofs_node_object_counter gauge
neofs_node_object_counter{shard="A1AJ1cgEdGGYC2kGTGQxQJ"} 5
neofs_node_object_counter{shard="S9b69efduvCprPU4FWKirH"} 3
```

Does not contain the node's total object count since on the shard level it is quite easy to control objects but on the SE level it is not always clear whether a shard contains an object or not:
1. GC works on the shard level
2. SE's `Delete` always calls shard's `Inhume` not `Delete`

It is possible to solve (e.g. some SE'e atomic counter that could be passed to every shard) but IMO it is easier to calculate on a metric collector side.

### DIFF
--- a/pkg/local_object_storage/metabase/VERSION.md
+++ b/pkg/local_object_storage/metabase/VERSION.md
@@ -31,7 +31,7 @@ This file describes changes between the metabase versions.
   - Keys and values
     - `id` -> shard id as bytes
     - `version` -> metabase version as little-endian uint64
-    - `counter` -> shard's object counter as little-endian uint64
+    - `phy_counter` -> shard's physical object counter as little-endian uint64
 
 ### Unique index buckets
 - Buckets containing objects of REGULAR type
@@ -83,10 +83,6 @@ This file describes changes between the metabase versions.
   - Key: split ID
   - Value: list of object IDs
 
-
-## Version 2
-
-- Added shard's object counter to the info bucket 
 
 ## Version 1
 
--- a/pkg/local_object_storage/metabase/VERSION.md
+++ b/pkg/local_object_storage/metabase/VERSION.md
@@ -31,6 +31,7 @@ This file describes changes between the metabase versions.
   - Keys and values
     - `id` -> shard id as bytes
     - `version` -> metabase version as little-endian uint64
+    - `counter` -> shard's object counter as little-endian uint64
 
 ### Unique index buckets
 - Buckets containing objects of REGULAR type
@@ -83,7 +84,9 @@ This file describes changes between the metabase versions.
   - Value: list of object IDs
 
 
-# History
+## Version 2
+
+- Added shard's object counter to the info bucket 
 
 ## Version 1
 
--- a/pkg/local_object_storage/metabase/counter.go
+++ b/pkg/local_object_storage/metabase/counter.go
@@ -1,0 +1,59 @@
+package meta
+
+import (
+	"encoding/binary"
+
+	"go.etcd.io/bbolt"
+)
+
+var shardCounterKey = []byte("counter")
+
+// ObjectCounter returns object count that metabase has
+// tracked since it was opened and initialized.
+//
+// Returns only the errors that do not allow reading counter
+// in Bolt database.
+func (db *DB) ObjectCounter() (counter uint64, err error) {
+	err = db.boltDB.View(func(tx *bbolt.Tx) error {
+		b := tx.Bucket(shardInfoBucket)
+		if b != nil {
+			data := b.Get(shardCounterKey)
+			if len(data) == 8 {
+				counter = binary.LittleEndian.Uint64(data)
+			}
+		}
+
+		return nil
+	})
+
+	return
+}
+
+// updateCounter updates the object counter. Tx MUST be writable.
+// If inc == `true`, increases the counter, decreases otherwise.
+func (db *DB) updateCounter(tx *bbolt.Tx, delta uint64, inc bool) error {
+	b := tx.Bucket(shardInfoBucket)
+	if b == nil {
+		return nil
+	}
+
+	var counter uint64
+
+	data := b.Get(shardCounterKey)
+	if len(data) == 8 {
+		counter = binary.LittleEndian.Uint64(data)
+	}
+
+	if inc {
+		counter += delta
+	} else if counter <= delta {
+		counter = 0
+	} else {
+		counter -= delta
+	}
+
+	newCounter := make([]byte, 8)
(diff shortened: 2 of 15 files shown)


realloc:
Agree, the total node's object count can be handled externally.

Is it possible to add a separate per shard counter for logically available objects? For example, some objects may be inhumed, but not physically removed from the graveyard yet, so we could see the difference in the monitoring system.

realloc:
Yes, there should be one counter with all types of objects available and one for all types of objects physically stored.

pkg/local_object_storage/metabase/VERSION.md
```diff
@@ -85,3 +82,14 @@ This file describes changes between the metabase versions.
   - Name: container ID + `_splitid`
   - Key: split ID
   - Value: list of object IDs
+
+
+## Version 1
+
+- Added shard's object counter to the info bucket
```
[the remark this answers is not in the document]
carpawell: i thought that meta versioning is about saying whether a user needs to resync the metabase or not

if meta from that PR is inited with the old db file, its object count will be incorrect (0). or do you mean that we may iterate all the objects and update the counter?

pkg/local_object_storage/metabase/delete.go
```diff
@@ -53,50 +64,61 @@ func (db *DB) Delete(prm DeletePrm) (DeleteRes, error) {
...
+		}
+
+		if removed {
+			rawDeleted++
 		}
 	}
 
+	err := db.updateCounter(tx, rawDeleted, false)
+	if err != nil {
+		return 0, fmt.Errorf("could not decrease object counter: %w", err)
```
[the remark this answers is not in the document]
carpawell: agree, removed `return`, added error log

pkg/local_object_storage/metabase/counter.go
```diff
@@ -0,0 +1,65 @@
...
+			counter = binary.LittleEndian.Uint64(data)
+		case 0:
+		default:
+			return nil
+		}
+
+		if inc {
+			counter += delta
+		} else {
+			if counter <= delta {
```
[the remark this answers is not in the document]
carpawell: merged but usually, i prefer not to use `else if`

pkg/local_object_storage/metabase/counter.go
```diff
@@ -0,0 +1,65 @@
...
+		var counter uint64
+		newCounter := make([]byte, 8)
+
+		data := b.Get(shardCounterKey)
+		switch len(data) {
+		case 8:
+			counter = binary.LittleEndian.Uint64(data)
+		case 0:
+		default:
+			return nil
```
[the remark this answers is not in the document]
carpawell: well... ok, added panic

@fyrchik, ping

pkg/local_object_storage/metabase/counter.go
```diff
@@ -0,0 +1,66 @@
...
+	data := b.Get(shardCounterKey)
+	switch len(data) {
+	case 8:
+		counter = binary.LittleEndian.Uint64(data)
+	case 0:
+	default:
+		panic(fmt.Errorf(
+			"unexpected len of object counter value: %d", len(data)),
+		)
+	}
```
fyrchik, line 52: I think just checking `len(data) == 8` would be enough.

pkg/local_object_storage/metabase/delete.go
```diff
@@ -60,11 +61,22 @@ func (db *DB) deleteGroup(tx *bbolt.Tx, addrs []oid.Address) error {
...
+
+		if removed {
+			rawDeleted++
+		}
+	}
+
+	err := db.updateCounter(tx, rawDeleted, false)
+	if err != nil {
+		db.log.Error("could not decrease object counter",
+			zap.Error(err))
```
fyrchik, line 90: No, let's not log anything inside of a transaction. It is potentially a blocking call.

pkg/local_object_storage/metabase/put.go
```diff
@@ -142,6 +143,14 @@ func (db *DB) put(
 		}
 	}
 
+	if !isParent {
+		err = db.updateCounter(tx, 1, true)
+		if err != nil {
+			db.log.Error("could not increase object counter: %w",
+				zap.Error(err))
```
fyrchik, line 150: Here too. I think we can (should) return the error to the caller, `updateCounter` has no logical errors.

pkg/metrics/object.go
```diff
@@ -254,3 +270,15 @@ func (m objectServiceMetrics) AddPutPayload(ln int) {
 func (m objectServiceMetrics) AddGetPayload(ln int) {
 	m.getPayload.Add(float64(ln))
 }
+
+func (m objectServiceMetrics) ChangeBy(shardID string, delta int) {
```
fyrchik, line 274: `Change` what? Could you add `ObjectCounter` to the name?

pkg/metrics/object.go
```diff
@@ -254,3 +270,15 @@ func (m objectServiceMetrics) AddPutPayload(ln int) {
...
+	m.shardMetrics.With(prometheus.Labels{shardIDLabelKey: shardID}).Add(float64(delta))
+}
+
+func (m objectServiceMetrics) IncObjectCounter(shardID string) {
+	m.shardMetrics.With(prometheus.Labels{shardIDLabelKey: shardID}).Inc()
+}
+
+func (m objectServiceMetrics) DecObjectCounter(shardID string) {
+	m.shardMetrics.With(prometheus.Labels{shardIDLabelKey: shardID}).Dec()
+}
```
fyrchik, line 288: Why did you decide to stay on 3 methods instead of a single `AddToObjectCounter` which covers both `inc` and `dec`?

pkg/metrics/object.go
carpawell: removed, only `AddToObjectCounter(shardID string, delta int)` and `SetObjectCounter(shardID string, v uint64)` are kept on the metrics side; wrappers still have `Inc*`, `Dec*` methods

pkg/metrics/object.go
carpawell: kept only two: `Set*` and `Add*`

pkg/local_object_storage/metabase/delete.go
carpawell: missed that it is inside a bolt TX

return back the returned code: no logging, direct err to a caller only

fyrchik (approved):

PR #1713, opened 2022-08-19
Feat/Return to the most prioritized morph endpoint

### LINKED ISSUE #1615: Consider automatic swtich to the morph RPC with the highest priority

carpawell:
Closes #1615.

The switching period could be configurable OR could be kinda flexible (try to wait more time if a node refuses connection more than one time) but it complicates the code and I do not think that it is worth it.
Also, we could not just try to connect to the most prioritized RPC but simply increase priority if the most prioritized RPC does not allow a connection. Not sure if we need it.

Checked using https://github.com/carpawell/neofs-dev-env/tree/local/add-rpc-morph-node.


carpawell:
Too dangerous PR, waits `v0.32.0` release.

pkg/morph/client/multi.go
```diff
@@ -120,6 +132,79 @@ func (c *Client) notificationLoop() {
 	}
 }
 
+func (c *Client) switchToMostPrioritized() {
+	const period = 30 * time.Second
```
fyrchik, line 136: I think 30 second is a really small time for this kind of thing, something like 10 minutes is more suitable IMO. cc @realloc

pkg/morph/client/multi.go
```diff
@@ -120,6 +132,79 @@ func (c *Client) notificationLoop() {
...
+			for i, e := range c.endpoints.list {
+				if i == c.endpoints.curr {
+					continue
+				}
+
+				if e.Priority < currPriority {
+					tryE = e.Address
+					break
+				}
+			}
```
fyrchik, line 159: Endpoints are already sorted by priority, so we can just try each node in list until the current one.
This also helps in a situation when we failed to connect to the most prioritized node: in this case the next should be tried in the same loop.

pkg/morph/client/multi.go
```diff
@@ -63,6 +61,16 @@ func (c *Client) switchRPC() bool {
...
 		}
 
+		c.client = cli
+		c.rpcActor = act
+		c.gasToken = gas
+
+		if !c.switchIsActive.Load() &&
+			c.endpoints.list[c.endpoints.curr].Priority != c.endpoints.list[0].Priority {
+			c.switchIsActive.Store(true)
+			go c.switchToMostPrioritized()
```
fyrchik, line 71: Do we create a goroutine each time `switchRPC` is called? I think this should be a background worker.

pkg/morph/client/multi.go
realloc: I'd suggest moving all similar constants to a config file anyway.

fyrchik: Agree, but the default still should make sense.

realloc: From the test cases, we often see 2 minutes as a requirement.

pkg/morph/client/multi.go
carpawell: > Do we create a goroutine each time switchRPC is called?

yes if it has not been created yet: my idea was that a node with a non-prioritized connection is smth really rare, so i do not want to have a goroutine if i know that i do not need it

sounds strange but that was a 1-month-younger me. what do you think?

pkg/morph/client/multi.go
carpawell: changed to 2 min

> moving all similar constants to a config file anyway

should it be done in that PR? not sure if it is useful. mb add when it has use cases?

pkg/morph/client/multi.go
carpawell: > Endpoints are already sorted by priority, so we can just try each node in list until the current one.

If we have `[a, b, c, d]` nodes with `[1, 2, 2, 2]` priorities and are connected to `d`, it is useless for us to reconnect to any node that is not `a`. At least this is how i understand the priorities

> we failed to connect to the most prioritized node: in this case the next should be tried in the same loop

well, that is one of the questions i was not sure about. ok, added that

pkg/morph/client/multi.go
fyrchik: It is definitely useful, because we must be able to control such things.

carpawell: added

pkg/morph/client/multi.go
```diff
@@ -120,6 +132,88 @@ func (c *Client) notificationLoop() {
...
+			copy(endpointsCopy, c.endpoints.list)
+
+			currEndpoint := c.endpoints.curr
+			currPriority := c.endpoints.list[c.endpoints.curr].Priority
+			highestPriority := c.endpoints.list[0].Priority
+			c.switchLock.RUnlock()
+
+			if currPriority == highestPriority {
+				// already connected to
+				// the most prioritized
```
fyrchik, line 156: What about doing it in `defer`?

acid-ant (approved):

docs/storage-node-configuration.md
```diff
@@ -132,18 +132,20 @@ contracts:
...
+|-------------------|-----------------------------------------------------------|------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `dial_timeout`    | `duration`                                                | `5s`             | Timeout for dialing connections to N3 RPCs.                                                                                                                         |
+| `cache_ttl`       | `duration`                                                | Morph block time | Sidechain cache TTL value (min interval between similar calls).<br/>Negative value disables caching.<br/>Cached entities: containers, container lists, eACL tables. |
+| `switch_interval` | `duration`                                                | `2m`             | Time interval between the attempts to connect to the highest priority RPC node if the connection is not established yet.                                            |
```
fyrchik, line 147: `switch` is alphabetially greater than `rpc_endpoint`, make it the last?

pkg/morph/client/constructor.go
```diff
@@ -60,6 +63,7 @@ func defaultConfig() *cfg {
 		signer: &transaction.Signer{
 			Scopes: transaction.Global,
 		},
+		switchInterval: defaultSwitchInterval,
```
fyrchik, line 66: Can we make it `0` (disabled) in `morph/client` and ensure to provide interval in both IR and neofs-node?

pkg/morph/client/multi.go
```diff
@@ -120,6 +132,88 @@ func (c *Client) notificationLoop() {
...
+				c.switchIsActive.Store(false)
+				return
+			}
+
+			for i, e := range endpointsCopy {
+				if i == currEndpoint || currPriority == e.Priority {
```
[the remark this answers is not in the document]
carpawell: part of the old code, dropped

pkg/morph/client/multi.go
```diff
@@ -120,6 +132,88 @@ func (c *Client) notificationLoop() {
...
+
+			currEndpoint := c.endpoints.curr
+			currPriority := c.endpoints.list[c.endpoints.curr].Priority
+			highestPriority := c.endpoints.list[0].Priority
+			c.switchLock.RUnlock()
+
+			if currPriority == highestPriority {
+				// already connected to
+				// the most prioritized
+				c.switchIsActive.Store(false)
```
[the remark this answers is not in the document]
carpawell: good, added

PR #1714, opened 2022-08-19
Feat/Add meta to `neofs-lens`

carpawell:
Adds `meta` subcommand. Also, includes refactoring that separates `inspect` and `list` commands on `write-cache` and `blobovnicza` subcommands.


fyrchik (approved):

cmd/neofs-lens/internal/meta/list-graveyard.go
```diff
@@ -0,0 +1,49 @@
...
+	db := meta.New(
+		meta.WithPath(vPath),
+		meta.WithBoltDBOptions(&bbolt.Options{
+			ReadOnly: true,
+			Timeout:  100 * time.Millisecond,
+		}),
+		meta.WithEpochState(epochState{}),
+	)
+
+	common.ExitOnErr(cmd, common.Errf("could not open metabase: %w", db.Open(true)))
```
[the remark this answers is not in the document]
carpawell: sure, added such func for all the components

cmd/neofs-lens/internal/meta/root.go
```diff
@@ -0,0 +1,32 @@
...
+
+var (
+	vAddress string
+	vPath    string
+)
+
+type epochState struct{}
+
+func (s epochState) CurrentEpoch() uint64 {
+	return math.MaxUint64
```
cthulhu-rider, line 21: Looks like a fake, what is this workaround for?

cthulhu-rider (commented):
For future pls don't blend feature code with refactor

carpawell:
> For future pls don't blend feature code with refactor

Usually, I do not do PRs that way. That happened because of the order of my changes: first, i added the meta but then realized that it was incredibly hard to use it without a refactor (and separate the refactor from the feature). In fact, commits are quite independent now. 

Rebased CHANGELOG conflicts.

cmd/neofs-lens/internal/meta/root.go
carpawell: after #1634 the meta panics if no epoch state is presented. so that is just for returning all the objects in it

we can add a flag that specifies the current epoch but not sure if it is useful

cthulhu-rider (approved):

Issue #1707, opened 2022-08-18
Blobstor may "lose" objects on config changes


carpawell:
## Context

`blobovnicza` stores objects in its buckets according to object sizes: it chooses the minimum number (the power of 2) that satisfies the [object limit](https://github.com/nspcc-dev/neofs-node/blob/4e361e2ab41602e369edd4feb1dafad99269664c/config/example/node.yaml#L133). It creates buckets for objects starting from `32<<10` = 32kb, e.g. if the size limit is `100kb`, tree buckets gonna be created: [0:32kb), [32kb:64kb), [64kb:128kb). That buckets will be used for storing and searching objects. Searching will try to iterate buckets according to the [*current* configuration](https://github.com/nspcc-dev/neofs-node/blob/4e361e2ab41602e369edd4feb1dafad99269664c/pkg/local_object_storage/blobovnicza/iterate.go#L31) of the blobovnicza.

## Problem

Consider situation:
1. A Node was configured to store objects with `size<=N`, `N%2==0`
2. It gets objects with `N/2 < size <= N` and stores them in the last of the supported buckets
3. The node goes offline
4. Its config changes and now it gonna store "small" objects only if their `size<=N/2`
5. It gets `Get` with the object from `2.` and could not return it since it does not know about the last bucket it has

Problems:
1. Current implementation could not return object
2. Object is stored but does not satisfy object placement (seems like such objects should be migrated to the FSTree with corresponding metabase changes)

## Ideas
1. Could be solved by just iterating all the buckets it has (but somehow logic with the defined bucket names were implemented, mb it has some nuances, /cc @cthulhu-rider)
2. We could migrate or not migrate objects. If some objects are expected to be stored in a bolt base it could be unexpected to find them in FSTree after the config change (/cc @realloc)

cthulhu-rider:
* `2` will be handled within https://github.com/nspcc-dev/neofs-node/issues/1780

PR #1718, opened 2022-08-22
Implement generic tests for storage components

fyrchik:
DRAFT, TODO:
- [x] Check `IgnoreErrors` flag for iteration over blobstor components.
- [ ] Fix `Raw` flag to `Get` to return _only_ raw data.
- [ ] Add a test for testing put/get      with and without compression.
- [x] Control tests for blobstor components.
- [x] Check that `Put` fails for blobstor components in readonly, but Get succeeds.
- [ ] Test concurrent Put/Get/Delete for testing with `-race` flag.
- [x] Remove all already existing testing code, that is covered by generic tests.

@carpawell @cthulhu-rider @realloc I feel like something is missing for `Open`/`SetMode`/`Close` tests, may be you can suggest some other things that could possibly happen?


fyrchik:
Ready for review. Decided to postpone other points -- some of them require refactoring, others can affect stability, and the remaining need to be first discussed. I will create tasks.

pkg/local_object_storage/blobstor/internal/blobstortest/delete.go
```diff
@@ -0,0 +1,82 @@
...
+
+func TestDelete(t *testing.T, cons Constructor, min, max uint64) {
+	s := cons(t)
+	require.NoError(t, s.Open(false))
+	require.NoError(t, s.Init())
+	t.Cleanup(func() { require.NoError(t, s.Close()) })
+
+	objects := prepare(t, 4, s, min, max)
+
+	t.Run("delete non-existent", func(t *testing.T) {
```
carpawell, line 20: can we run all the "not found" tests before the `prepare` call? just to be sure that object is missing

carpawell (approved):
if it is ready, change commit messages then?

carpawell:
> maybe you can suggest some other things that could possibly happen?

@fyrchik, [changing](https://github.com/nspcc-dev/neofs-node/issues/1707) the configuration b/w opens?

fyrchik:
The interface doesn't allow this, and, as you noticed, the proper support has not yet been implemented.

pkg/local_object_storage/blobstor/internal/blobstortest/delete.go
fyrchik: The purpose it to run this on a dirty storage so that we don't just delete a random object.

pkg/local_object_storage/shard/control.go
```diff
@@ -50,9 +50,17 @@ func (s *Shard) Open() error {
 		components = append(components, s.pilorama)
 	}
 
-	for _, component := range components {
+	for i, component := range components {
 		if err := component.Open(false); err != nil {
 			if component == s.metaBase {
+				// We must first open all other components to avoid
```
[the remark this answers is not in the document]
fyrchik: We could, but first need to think whether it affects anything. Personally, I think the order doesn't matter in terms of readability, as in future we would like to handle e.g. write cache failures.

carpawell (approved):

Issue #1711, opened 2022-08-19
Simplify neofs-adm code after neo-go v0.99.2 update


fyrchik:
`actor` package can be useful.
See https://github.com/nspcc-dev/neofs-node/pull/1705#discussion_r949013572

fyrchik:
Morph code can use it too.

carpawell:
Do not forget https://github.com/nspcc-dev/neofs-node/pull/1736#discussion_r957750999.
Use notary actors after https://github.com/nspcc-dev/neo-go/pull/2665.

roman-khimov:
I think we can just create another issue to update again with 0.99.3 release, while here concentrate on 0.99.2 features (that probably solve 80% of problems).

PR #1736, opened 2022-08-29
Upd/Adopt neo-go 0.99.2

carpawell:
Related to #1711.

Notary actors will be added after the https://github.com/nspcc-dev/neo-go/pull/2665.

### DIFF
--- a/cmd/neofs-adm/internal/modules/storagecfg/root.go
+++ b/cmd/neofs-adm/internal/modules/storagecfg/root.go
@@ -6,7 +6,6 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"math/big"
 	"math/rand"
 	"net"
 	"net/url"
@@ -25,10 +24,10 @@ import (
 	"github.com/nspcc-dev/neo-go/pkg/encoding/address"
 	"github.com/nspcc-dev/neo-go/pkg/encoding/fixedn"
 	"github.com/nspcc-dev/neo-go/pkg/rpcclient"
-	"github.com/nspcc-dev/neo-go/pkg/smartcontract"
+	"github.com/nspcc-dev/neo-go/pkg/rpcclient/actor"
+	"github.com/nspcc-dev/neo-go/pkg/rpcclient/nep17"
 	"github.com/nspcc-dev/neo-go/pkg/smartcontract/trigger"
 	"github.com/nspcc-dev/neo-go/pkg/util"
-	"github.com/nspcc-dev/neo-go/pkg/vm/vmstate"
 	"github.com/nspcc-dev/neo-go/pkg/wallet"
 	netutil "github.com/nspcc-dev/neofs-node/pkg/network"
 
@@ -322,23 +321,17 @@ func depositGas(cmd *cobra.Command, acc *wallet.Account, network string) {
 	sideClient := initClient(n3config[network].MorphRPC)
 	balanceHash, _ := util.Uint160DecodeStringLE(n3config[network].BalanceContract)
 
-	res, err := sideClient.InvokeFunction(balanceHash, "balanceOf", []smartcontract.Parameter{{
-		Type:  smartcontract.Hash160Type,
-		Value: acc.Contract.ScriptHash(),
-	}}, nil)
-	fatalOnErr(err)
-
-	if res.State != vmstate.Halt.String() {
-		fatalOnErr(fmt.Errorf("invalid response from balance contract: %s", res.FaultException))
+	sideActor, err := actor.NewSimple(sideClient, acc)
+	if err != nil {
+		fatalOnErr(fmt.Errorf("creating actor over side chain client: %w", err))
 	}
 
-	var balance *big.Int
-	if len(res.Stack) != 0 {
-		balance, _ = res.Stack[0].TryInteger()
-	}
+	sideGas := nep17.NewReader(sideActor, balanceHash)
+	accSH := acc.Contract.ScriptHash()
 
-	if balance == nil {
-		fatalOnErr(errors.New("invalid response from balance contract"))
+	balance, err := sideGas.BalanceOf(accSH)
+	if err != nil {
+		fatalOnErr(fmt.Errorf("side chain balance: %w", err))
 	}
 
 	ok := getConfirmation(false, fmt.Sprintf("Current NeoFS balance is %s, make a deposit? y/[n]: ",
@@ -359,11 +352,17 @@ func depositGas(cmd *cobra.Command, acc *wallet.Account, network string) {
 	gasHash, err := mainClient.GetNativeContractHash(nativenames.Gas)
 	fatalOnErr(err)
 
-	tx, err := mainClient.CreateNEP17TransferTx(acc, neofsHash, gasHash, amount.Int64(), 0, nil, nil)
-	fatalOnErr(err)
+	mainActor, err := actor.NewSimple(mainClient, acc)
+	if err != nil {
+		fatalOnErr(fmt.Errorf("creating actor over main chain client: %w", err))
+	}
 
-	txHash, err := mainClient.SignAndPushTx(tx, acc, nil)
-	fatalOnErr(err)
+	mainGas := nep17.New(mainActor, gasHash)
+
+	txHash, _, err := mainGas.Transfer(accSH, neofsHash, amount, nil)
+	if err != nil {
+		fatalOnErr(fmt.Errorf("sending TX to the NeoFS contract: %w", err))
+	}
 
 	cmd.Print("Waiting for transactions to persist.")
 	tick := time.NewTicker(time.Second / 2)
--- a/cmd/neofs-adm/internal/modules/storagecfg/root.go
+++ b/cmd/neofs-adm/internal/modules/storagecfg/root.go
@@ -350,7 +350,9 @@ func depositGas(cmd *cobra.Command, acc *wallet.Account, network string) {
 	neofsHash, _ := util.Uint160DecodeStringLE(n3config[network].NeoFSContract)
 
 	gasHash, err := mainClient.GetNativeContractHash(nativenames.Gas)
-	fatalOnErr(err)
+	if err != nil {
+		fatalOnErr(fmt.Errorf("gas contract hash: %w", err))
+	}
 
 	mainActor, err := actor.NewSimple(mainClient, acc)
 	if err != nil {
--- a/pkg/morph/client/client.go
+++ b/pkg/morph/client/client.go
@@ -15,6 +15,8 @@ import (
 	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
 	"github.com/nspcc-dev/neo-go/pkg/encoding/fixedn"
 	"github.com/nspcc-dev/neo-go/pkg/rpcclient"
+	"github.com/nspcc-dev/neo-go/pkg/rpcclient/actor"
+	"github.com/nspcc-dev/neo-go/pkg/rpcclient/nep17"
 	sc "github.com/nspcc-dev/neo-go/pkg/smartcontract"
 	"github.com/nspcc-dev/neo-go/pkg/smartcontract/trigger"
 	"github.com/nspcc-dev/neo-go/pkg/util"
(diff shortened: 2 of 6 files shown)


pkg/morph/client/util.go
```diff
@@ -92,3 +94,21 @@ func StringFromStackItem(param stackitem.Item) (string, error) {
...
+	return func(r *result.Invoke, t *transaction.Transaction) error {
+		if r.State != HaltState {
+			return wrapNeoFSError(&notHaltStateError{state: r.State, exception: r.FaultException})
+		}
+
+		if len(r.Script) == 0 {
+			return wrapNeoFSError(errEmptyInvocationScript)
+		}
+
+		if add != 0 {
```
roman-khimov, line 108: If it's zero, adding won't hurt anyway.

pkg/morph/client/notary.go
```diff
@@ -167,22 +169,20 @@ func (c *Client) DepositNotary(amount fixedn.Fixed8, delta uint32) (res util.Uin
 		return util.Uint256{}, fmt.Errorf("can't get GAS script hash: %w", err)
 	}
 
-	txHash, err := c.client.TransferNEP17(
-		c.acc,
+	gasToken := nep17.New(c.rpcActor, gas)
```
roman-khimov, line 172: This can be simplified post-0.99.2 with `gas.New()`, but works fine for now.

pkg/morph/client/constructor.go
```diff
@@ -82,6 +84,12 @@ func New(key *keys.PrivateKey, opts ...Option) (*Client, error) {
 		panic("empty private key")
 	}
 
+	acc := wallet.NewAccountFromPrivateKey(key)
+	accAddr, err := address.StringToUint160(acc.Address)
```
roman-khimov, line 88: `key.GetScriptHash()` to avoid error check, but I know this can be improved in the `wallet.Account` itself.

pkg/morph/client/client.go
```diff
@@ -356,7 +330,13 @@ func (c *Client) GasBalance() (res int64, err error) {
 		return 0, err
 	}
 
-	return c.client.NEP17BalanceOf(gas, c.acc.PrivateKey().GetScriptHash())
+	gasToken := nep17.New(c.rpcActor, gas)
```
roman-khimov, line 315: It seems to me that you're creating this wrapper again and again in the same `*Client`, most likely it can be done once during init and then reused for the whole `Client` lifetime.

cmd/neofs-adm/internal/modules/morph/initialize.go
```diff
@@ -363,57 +363,45 @@ loop:
...
+		if err != nil {
+			return fmt.Errorf("could not get %s account for %d wallet: %w",
+				committeeAccountName, i, err)
+		}
 
-	// Calculate network fee again, because tx size has changed.
-	_, accounts, err := getSigners(c.CommitteeAcc, cosigners)
-	if err != nil {
-		panic(fmt.Errorf("BUG: can't calculate network fee: %w", err))
+		signers = append(signers, actor.SignerAccount{
```
fyrchik, line 386: Judging by the neo-go code this wouldn't create a valid transaction, have you tested it?
https://github.com/nspcc-dev/neo-go/blob/master/pkg/rpcclient/actor/actor.go#L138

Is it the right way to use multisignature? @roman-khimov

fyrchik: Also, `localClient` needs some love, `initialize_test.go` can help here.

cmd/neofs-adm/internal/modules/morph/initialize_nns.go
```diff
@@ -88,7 +88,7 @@ func (c *initializeContext) updateNNSGroup(nnsHash util.Uint160, pub *keys.Publi
 	if err != nil || sysFee == 0 {
 		return err
 	}
-	return c.sendCommitteeTx(bw.Bytes(), sysFee, true)
+	return c.sendCommitteeTx(bw.Bytes(), true)
```
fyrchik, line 91: I wouldn't work this way, the system fee is too big.

cmd/neofs-adm/internal/modules/morph/local_client.go
```diff
@@ -179,6 +179,21 @@ func (l *localClient) SignAndPushInvocationTx(_ []byte, _ *wallet.Account, _ int
 	panic("unexpected call")
 }
 
+func (l *localClient) GetVersion() (*result.Version, error) {
+	// not used by `morph init` command
+	panic("unexpected call")
```
fyrchik, line 184: It is expected now, `initialize_test.go` fails with panic here.

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -357,13 +350,21 @@ func depositGas(cmd *cobra.Command, acc *wallet.Account, network string) {
...
+	mainActor, err := actor.NewSimple(mainClient, acc)
+	if err != nil {
+		fatalOnErr(fmt.Errorf("creating actor over main chain client: %w", err))
+	}
 
-	txHash, err := mainClient.SignAndPushTx(tx, acc, nil)
-	fatalOnErr(err)
+	mainGas := nep17.New(mainActor, gasHash)
+
+	txHash, _, err := mainGas.Transfer(accSH, neofsHash, amount, nil)
```
AnnaShaleva, line 364: BTW, you can use the second returned value (`ValidUntilBlock`) for precise transaction acceptance tracking (instead of 20-seconds timer). If `VUB`-th block is accepted but transaction's application log is not yet available, then the transaction won't be accepted at all.

cmd/neofs-adm/internal/modules/morph/n3client.go
```diff
@@ -30,12 +31,16 @@ type Client interface {
...
 	NEP17BalanceOf(util.Uint160, util.Uint160) (int64, error)
+	InvokeContractVerify(contract util.Uint160, params []smartcontract.Parameter, signers []transaction.Signer, witnesses ...transaction.Witness) (*result.Invoke, error)
+	InvokeFunction(contract util.Uint160, operation string, params []smartcontract.Parameter, signers []transaction.Signer) (*result.Invoke, error)
 	InvokeScript([]byte, []transaction.Signer) (*result.Invoke, error)
 	SendRawTransaction(*transaction.Transaction) (util.Uint256, error)
 	GetCommittee() (keys.PublicKeys, error)
 	CalculateNotaryFee(uint8) (int64, error)
+	CalculateNetworkFee(tx *transaction.Transaction) (int64, error)
 	AddNetworkFee(*transaction.Transaction, int64, ...*wallet.Account) error
 	SignAndPushInvocationTx([]byte, *wallet.Account, int64, fixedn.Fixed8, []rpcclient.SignerAccount) (util.Uint256, error)
```
fyrchik, line 45: Do we still need all of this? Smaller interface here will help us to support `localClient` more easily.

cmd/neofs-adm/internal/modules/morph/initialize.go
carpawell: > have you tested it?

yes but only with 1/1 multisig. not sure it gonna work correctly with more then one signature

cmd/neofs-adm/internal/modules/morph/n3client.go
roman-khimov: Yeah, `localClient` can benefit a lot from a slimmer interface. That was one my thoughts as well.

cmd/neofs-adm/internal/modules/storagecfg/root.go
roman-khimov: BTW, isn't it 30 seconds? Maybe it's related to nspcc-dev/neo-go#2663?

roman-khimov: Yeah, it's very risky, any minimal delay in block acceptance can cause a timeout this way.

fyrchik: https://github.com/nspcc-dev/neofs-node/issues/1741

roman-khimov:
BTW, I don't see the removal of `toStackParameter` anywhere in this PR, while this should be possible to do.

pkg/morph/client/util.go
carpawell: dropped

pkg/morph/client/constructor.go
carpawell: used `key.GetScriptHash()`

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -357,13 +350,21 @@ func depositGas(cmd *cobra.Command, acc *wallet.Account, network string) {
 	neofsHash, _ := util.Uint160DecodeStringLE(n3config[network].NeoFSContract)
 
 	gasHash, err := mainClient.GetNativeContractHash(nativenames.Gas)
-	fatalOnErr(err)
+	if err != nil {
+		fatalOnErr(fmt.Errorf("gas contract hash: %w", err))
+	}
```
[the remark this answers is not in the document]
carpawell: moved to a separate commit

pkg/morph/client/util.go
```diff
@@ -92,3 +94,21 @@ func StringFromStackItem(param stackitem.Item) (string, error) {
...
+
+func addFeeCheckerModifier(add int64) func(r *result.Invoke, t *transaction.Transaction) error {
+	return func(r *result.Invoke, t *transaction.Transaction) error {
+		if r.State != HaltState {
+			return wrapNeoFSError(&notHaltStateError{state: r.State, exception: r.FaultException})
+		}
+
+		if len(r.Script) == 0 {
+			return wrapNeoFSError(errEmptyInvocationScript)
+		}
```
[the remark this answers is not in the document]
carpawell: missed that, dropped that check

cmd/neofs-adm/internal/modules/morph/initialize.go
carpawell: returned signing as it was done before

cmd/neofs-adm/internal/modules/morph/initialize_nns.go
carpawell: will continue that work in a separate commit

cmd/neofs-adm/internal/modules/morph/local_client.go
carpawell: will continue that work in a separate commit

cmd/neofs-adm/internal/modules/morph/n3client.go
carpawell: will continue that work in a separate PR

carpawell:
1. Will continue work related to an `adm` enhancement in a separate PR
2. Kept only morph changes and `adm` changes that relate to a GAS balance operations
3. Added `nolint` to `AddNetworkFee` since it relates
4. Checked both notary and non-notary envs

> I don't see the removal of toStackParameter anywhere in this PR

@roman-khimov, notary calls [require](https://github.com/nspcc-dev/neofs-node/blob/df88afeef3af1dd63ce1f05f55b69ec5591035eb/pkg/morph/client/notary.go#L445) stack prms. Could be solved by creating actors but that should be done in every call since it is not clear what signers are in that TX.

roman-khimov:
>  it is not clear what signers are in the current TX.

From a quick glance, you have only two possibilities there -> can create two actors and use them as needed. Reusing the same set of signers is exactly what is expected of a backend code and what invokers/actors make easy to do.

fyrchik (approved):

AnnaShaleva (approved):

carpawell:
> From a quick glance, you have only two possibilities there

@roman-khimov, in practice -- yes, but the Alphabet is not a constant list in theory. So I suggest to think about it one more time when notary calls are refactored.

roman-khimov:
It's not updated often and you probably know well when it happens, so some `UpdateAlphabetAccount` or whatever else can solve the problem. Actors work best when they're long-lived, if you have to create one for every call then it's not that big of an improvement (although it would still be for notary calls because of attribute logic).

roman-khimov (approved):

fyrchik (approved):

Issue #1733, opened 2022-08-29
File descriptor overflow (1700000+ per node) with blobovnizas' depth greater than 3


anatoly-bogatyrev:
- File descriptor overflow (1700000+ per node) with blobovnizas' depth greater than 3.

anatoly-bogatyrev:
Issue can be closed, when https://github.com/nspcc-dev/neofs-node/pull/1737 will be merged. PR has been tested.

PR #1737, opened 2022-08-30
Fix active blobovnicza caching

### LINKED ISSUE #1691: Shard is going to degraded mode under load
Steps:
* Run four storage node
* create preset for 24kb objects 100 containers x 1000 objects in each
* run load with params 100 threads / r/w = 70/30 / objects size = 24kb, sleep for write 0.25s, sleep for read 0.1s from three servers in parallel.

Expected behaviour: load work properly.

Current behaviour: after 10 seconds we see that errors on shards is getting more that threshold and shard is going to degraded. After some time, the node restarts.

root@az:~# neofs-cli --endpoint localhost:8091 -w /etc/neofs/storage/wallet.json control shards list | grep Mode 
Enter password > 
Mode: degraded-read-only
Mode: degraded-read-only
Mode: read-write
Mode: degraded-read-only

Logs and node config and attachments.

[logs + config.zip](https://github.com/nspcc-dev/neofs-node/files/9351137/logs.%2B.config.zip)

root@az:~# neofs-node --version
NeoFS Storage node
Version: v0.31.0
GoVersion: go1.18.4

root@az:~# neo-go --version
NeoGo
Version: 0.99.1
GoVersion: go1.18.4

fyrchik:
Close #1691 .
Maybe related to #1733 .

### DIFF
--- a/pkg/local_object_storage/blobstor/blobovniczatree/blobovnicza.go
+++ b/pkg/local_object_storage/blobstor/blobovniczatree/blobovnicza.go
@@ -179,19 +179,26 @@ func (b *Blobovniczas) updateAndGet(p string, old *uint64) (blobovniczaWithIndex
 	defer b.activeMtx.Unlock()
 
 	// check 2nd time to find out if it blobovnicza was activated while thread was locked
-	if tryActive, ok := b.active[p]; ok && tryActive.blz == active.blz {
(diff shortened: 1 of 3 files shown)


carpawell (approved):
"the logic becomes to complicate because" msg in the second commit

carpawell (approved):

PR #1743, opened 2022-08-31
Fix/Non-notary calls

carpawell:
Some methods add "IR" suffix to their names in notary enabled envs
because of contract logic. It was broken due to incorrect notary state
reading (tryNotary != notary is enabled).

Was broken by a mistake in #1224. Could appear only in notary disabled envs in `v0.28.1+` versions, that is why we have not noticed that.

### DIFF
--- a/pkg/morph/client/static.go
+++ b/pkg/morph/client/static.go
@@ -34,7 +34,7 @@ type staticOpts struct {
 //
 // See also TryNotary.
 func (s *StaticClient) WithNotary() bool {
-	return s.tryNotary
+	return s.client.IsNotaryEnabled()
 }
 
 // IsAlpha returns Alphabet status of the client.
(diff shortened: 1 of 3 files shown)


alexvanin (approved):

fyrchik (approved):

pkg/morph/client/static.go
```diff
@@ -34,7 +34,7 @@ type staticOpts struct {
 //
 // See also TryNotary.
 func (s *StaticClient) WithNotary() bool {
-	return s.tryNotary
+	return s.client.IsNotaryEnabled()
```
cthulhu-rider, line 37: What about `TryNotary` option?

carpawell: about trying notary invokes (not all the calls should be notary even in notary-enabled mode)

`notaryEnabled && tryNotary` => invoke notary call
`notaryDisabled && !tryNotary` => regular call

PR #1745, opened 2022-09-01
Writecache fixes

fyrchik:
There are 2 problems with the current flush logic:
1. Objects are evicted from the cache even if PUT to the main storage failed.
2. There is a race condition, so that where an object could be silently evicted from LRU without removal from the storage.

In this PR we change this logic in 2 aspects:
1. Removing evicted items is done in the eviction callback.
2. The removal itself is done in batches to avoid blocking client threads.


cmd/neofs-node/config/engine/shard/writecache/config.go
```diff
@@ -55,22 +55,6 @@ func (x *Config) Path() string {
...
-	s := config.SizeInBytesSafe(
-		(*config.Config)(x),
-		"memcache_capacity",
-	)
-
-	if s > 0 {
-		return s
-	}
-
-	return MemSizeDefault
```
carpawell, line 71: remove that const too?

pkg/local_object_storage/writecache/writecache.go
```diff
@@ -99,6 +97,12 @@ func New(opts ...Option) Cache {
 		opts[i](&c.options)
 	}
 
+	// Make the LRU cache contain which take approximately 3/4 of the maximum space.
+	// Assume small and big objects are stored in 50-50 proportion.
+	c.maxFlushedMarksCount = int(c.maxCacheSize/c.maxObjectSize+c.maxCacheSize/c.smallObjectSize) * 3 / 8
```
carpawell, line 102: can we write it as `int(...+...) / 2 * 3 / 4` to make it easier for a reader?

carpawell (dismissed):

cmd/neofs-node/config/engine/shard/writecache/config.go
fyrchik: Fixed

pkg/local_object_storage/writecache/writecache.go
fyrchik: Fixed

carpawell (approved):

PR #1753, opened 2022-09-05
services/tree: do not restrict depth in `GetSubTree`

fyrchik:
Previously, the depth was restricted because with BFS the amount of
nodes we have in memory blows up exponentially. With DFS is is linear,
so we can process trees of arbitrary depth.

Refs nspcc-dev/neofs-s3-gw#624


KirillovDenis (approved):

KirillovDenis:
Shouldn't we get rid of [this](https://github.com/fyrchik/neofs-node/blob/66d3f09ec7c07c86a5a3d8b1cbba58ba6056eff5/pkg/services/tree/service.go#L29)?

cthulhu-rider (dismissed):

alexvanin (approved):

cthulhu-rider (approved):

Issue #1530, opened 2022-06-16
Allow tree structure modification during network upgrade


alexvanin:
Tree service client applications, like S3 Gateway, might develop over time and produce some incompatible changes:
- rename key in the node meta (see https://github.com/nspcc-dev/neofs-s3-gw/issues/533),
- update value format in the node meta,
- split one node into several nodes,
- merge several nodes into one,
- many more.

Client applications don't want to break compatibility during update, so they want to make changes in tree structure. 

This is applicable for private installations with manual software update. Client side may prepare a transition app / script that will update tree structure. To do that, tree service should allow tree service operations for network admin, like `Add`, `GetSubTree`, `Remove`, and, probably, extra method to list all trees in the storage node (maybe it is available in control service for tree sync?). 

I see two options there:
1. Tree service pass auth check (bearer token) if request is signed by storage node wallet
2. Tree service pass auth check (bearer token) if request is signed by multi-signature wallet of Alphabet nodes

Maybe there are better options for that.

fyrchik:
So we need to do two things:
1. Extend access control checks with local node key
2. Allow to list/drop arbitrary trees via the control service in addition to sync.

RIght?

alexvanin:
In this proposal yes and yes.

Maybe you find out better solution on how to manage trees for system admin.

roman-khimov:
Trees will be removed, this issue will never be done.

PR #1761, opened 2022-09-07
services/tree: allow to drop trees from the database

### LINKED ISSUE #1630: Support droping a tree from the database
Remove tree in a single transaction.
Use-cases:
1. S3 buckets removal
2. Tree removal after nodes has exited from a container

cc @alexvanin @realloc

fyrchik:
Close #1630.

I think removing based on morph event covers our main use-case.
Do we need an RPC to remove trees without removing the container? @alexvanin @KirillovDenis

### DIFF
--- a/pkg/local_object_storage/pilorama/interface.go
+++ b/pkg/local_object_storage/pilorama/interface.go
@@ -34,6 +34,9 @@ type Forest interface {
 	// TreeGetOpLog returns first log operation stored at or above the height.
 	// In case no such operation is found, empty Move and nil error should be returned.
 	TreeGetOpLog(cid cidSDK.ID, treeID string, height uint64) (Move, error)
+	// TreeDrop drops a tree from the database.
+	// If the tree is not found, ErrTreeNotFound should be returned.
+	TreeDrop(cid cidSDK.ID, treeID string) error
 }
 
 type ForestStorage interface {
--- a/cmd/neofs-node/tree.go
+++ b/cmd/neofs-node/tree.go
@@ -2,9 +2,14 @@ package main
 
 import (
 	"context"
+	"errors"
 
 	treeconfig "github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/tree"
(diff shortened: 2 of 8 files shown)


alexvanin:
Not sure about RPC, but it definitely will be useful in the context of tree service compatibility updates for https://github.com/nspcc-dev/neofs-node/issues/1530

Can / should we do cleanup with new tree removal RPC at [bucket removal](https://github.com/nspcc-dev/neofs-s3-gw/blob/6108f8d92952e0a8c8ee52daa3dfa86acdf9fe0f/api/handler/delete.go#L263)? @KirillovDenis

KirillovDenis:
> Can / should we do cleanup with new tree removal RPC at [bucket removal](https://github.com/nspcc-dev/neofs-s3-gw/blob/6108f8d92952e0a8c8ee52daa3dfa86acdf9fe0f/api/handler/delete.go#L263)? @KirillovDenis

As I understand container removal leads to tree removal on storage-node side without any additional requests. So we shouldn't use new RPC.

fyrchik:
Ok, so let's have it without an RPC, and extend control service in #1530 when it will be needed.

pkg/local_object_storage/pilorama/interface.go
```diff
@@ -34,6 +34,8 @@ type Forest interface {
 	// TreeGetOpLog returns first log operation stored at or above the height.
 	// In case no such operation is found, empty Move and nil error should be returned.
 	TreeGetOpLog(cid cidSDK.ID, treeID string, height uint64) (Move, error)
+	// TreeDrop drops a tree from the database.
```
cthulhu-rider, line 37: I think we should doc `ErrTreeNotFound` return case.

Also IMO `database` does not fit well under the description of the interface, because rather tends to a specific implementation. But it's not critical.

fyrchik: It is tricky, because storage engine can't always return this error (write errors are more important). But, still documenting errors is a good thing, fixed.

cthulhu-rider (approved):

Issue #1721, opened 2022-08-22
The network map is updated on one of the four nodes only


anatoly-bogatyrev:
An unspecified problem with updating the network map on different nodes.
The network map is processed by only one node out of four. 
In this case, after restarting the storage nodes, all nodes receive the current network map.

Step 1:
All nodes have been starter.
Netmap in all nodes:
```
root@az:/etc/neofs/storage# neofs-cli control netmap-snapshot --endpoint localhost:8091 --wallet wallet.json
Enter password > 
Epoch: 926
Node 1: 2A38rHJ982uLUnaj1BHUGHRzAGB7FNnE9DZpXPrHjJUKd ONLINE [/dns4/node4.neofs/tcp/8080]
	Continent: Europe
	Country: Russia
	CountryCode: RU
	Deployed:  NSPCC
	Location: Sheremetyevo Apt/Moscow
	Price: 10
	UN-LOCODE: RU SVO
```

Step 2:
Tick epoch:
```
$ neofs-adm morph force-new-epoch -c configuration/config.yaml
Current epoch: 926, increase to 927.
Waiting for transactions to persist...
```

Step 3:
Nodes 1-3 - Netmap has not been updated.
```
root@az:/etc/neofs/storage# neofs-cli control netmap-snapshot --endpoint localhost:8091 --wallet wallet.json
Enter password > 
Epoch: 926
Node 1: 2A38rHJ982uLUnaj1BHUGHRzAGB7FNnE9DZpXPrHjJUKd ONLINE [/dns4/node4.neofs/tcp/8080]
	Continent: Europe
	Country: Russia
	CountryCode: RU
	Deployed:  SBER
	Location: Sheremetyevo Apt/Moscow
	Price: 10
	UN-LOCODE: RU SVO
```
Node 4 - Netmap has been updated as expected:
```
root@glagoli:/etc/neofs/storage# neofs-cli control netmap-snapshot --endpoint localhost:8091 --wallet wallet.json
Enter password > 
Epoch: 927
Node 1: wfGE7cCjrD7rPRLsoRum6uYDvT8Eohr1zKCHz6rkRbBJ ONLINE [/dns4/node3.neofs/tcp/8080]
	Continent: Europe
	Country: Russia
	CountryCode: RU
	Deployed:  NSPCC
	Location: Sheremetyevo Apt/Moscow
	Price: 10
	UN-LOCODE: RU SVO
[...]

PR #1763, opened 2022-09-07
node/netmap: Write log messages about `NewEpoch` events

cthulhu-rider:
* inspired by explorations performed within https://github.com/nspcc-dev/neofs-node/issues/1721

Within the issue I made a hypothesis that node had missed some ticks of the epoch timer. However it was difficult to explicitly prove or disprove this assumption.

These changes make storage nodes to log new epochs. From now it'll be easy to track the node timeline.

### DIFF
--- a/cmd/neofs-node/morph.go
+++ b/cmd/neofs-node/morph.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 
 	"github.com/nspcc-dev/neo-go/pkg/core/block"
+	"github.com/nspcc-dev/neo-go/pkg/core/state"
 	"github.com/nspcc-dev/neo-go/pkg/util"
 	morphconfig "github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/morph"
 	"github.com/nspcc-dev/neofs-node/pkg/core/netmap"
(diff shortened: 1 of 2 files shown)


fyrchik (approved):

pkg/morph/subscriber/subscriber.go
```diff
@@ -133,6 +133,10 @@ func (s *subscriber) routeNotifications(ctx context.Context) {
 					continue
 				}
 
+				s.log.Debug("new notification event from sidechain",
```
carpawell, line 136: why only the regular notification?

cmd/neofs-node/morph.go
```diff
@@ -211,7 +212,16 @@ func listenMorphNotifications(c *cfg) {
 		})
 	}))
 
-	setNetmapNotificationParser(c, newEpochNotification, netmapEvent.ParseNewEpoch)
+	setNetmapNotificationParser(c, newEpochNotification, func(src *state.ContainedNotificationEvent) (event.Event, error) {
+		res, err := netmapEvent.ParseNewEpoch(src)
+		if err == nil {
+			c.log.Info("new epoch event from sidechain",
```
carpawell, line 218: do we really need that to be "Info"?

cthulhu-rider: I think yes, epoch ticks is a main heartbeat of the node, and node admin should always have the ability to track the timeline. Don't you think so?

pkg/morph/subscriber/subscriber.go
cthulhu-rider: What else do we need? New blocks are logged. Other things don't touch epoch ticks.

cmd/neofs-node/morph.go
fyrchik: I think epoch events are crucial for NeoFS network to function, so it makes sense to log this with `Info`.

carpawell: hm, IMO, that is more like debug. in normal situations i do not expect an admin to be informed about epoch via the logs (via the metrics or the monitoring service -- yes), only if he tries to debug smth. but ok

pkg/morph/subscriber/subscriber.go
carpawell: i just was thinking about that switch context

> New blocks are logged.

yes, but not here, not in that switch, so that is why i asked what is the difference b/w that cases. IMO, we should either log notifications on the receiver side or log all the notifications in that switch. but not critical

carpawell (approved):

cmd/neofs-node/morph.go
carpawell: on the other hand, we have the epoch metric and do not have a block metric. that could be assumed as a criterion for being smth "info"

cthulhu-rider: We still don't have strict requirement for log levels. Epoch ticks is the main thing that showa node's heartbits in logs.

PR #1766, opened 2022-09-07
blobovnicza/get: Iterate over all buckets regardless of config

### LINKED ISSUE #1707: Blobstor may "lose" objects on config changes

cthulhu-rider:
Re-configuration of Blobovnicza's object size limit must not affect already stored objects. In previous implementation `Blobovnicza` didn't see objects stored in buckets which became too big after size limit re-configuration. For example, lets consider 1st configuration with 64KB size limit for stored objects. Lets assume that we stored object of 64KB size, and re-configured `Blobovnicza` with 32KB limit. After reboot object should be still available, but actually it isn't. This is caused by `Get` operation algorithm which iterates over configured size ranges only, and doesn't process any other existing size bucket. By the way, increasing of the object size limit didn't lead to the problem even in previous implementation.

Make `Blobovnicza.Get` method to iterate over all size buckets regardless of current configuration. This covers the described scenario.

* related #1707

### DIFF
--- a/pkg/local_object_storage/blobovnicza/get_test.go
+++ b/pkg/local_object_storage/blobovnicza/get_test.go
@@ -1,0 +1,75 @@
+package blobovnicza
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	oidtest "github.com/nspcc-dev/neofs-sdk-go/object/id/test"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBlobovnicza_Get(t *testing.T) {
+	t.Run("re-configure object size limit", func(t *testing.T) {
+		filename := filepath.Join(t.TempDir(), "blob")
+
+		var blz *Blobovnicza
+
+		t.Cleanup(func() {
+			blz.Close()
+			os.RemoveAll(filename)
+		})
+
+		fnInit := func(szLimit uint64) {
+			if blz != nil {
+				require.NoError(t, blz.Close())
+			}
+
+			blz = New(
+				WithPath(filename),
+				WithObjectSizeLimit(szLimit),
+			)
+
+			require.NoError(t, blz.Open())
+			require.NoError(t, blz.Init())
+		}
+
+		// initial distribution: [0:32K] (32K:64K]
+		fnInit(2 * firstBucketBound)
+
+		addr := oidtest.Address()
+		obj := make([]byte, firstBucketBound+1)
+
+		var prmPut PutPrm
+		prmPut.SetAddress(addr)
+		prmPut.SetMarshaledObject(obj)
+
+		// place object to [32K:64K] bucket
+		_, err := blz.Put(prmPut)
+		require.NoError(t, err)
+
+		var prmGet GetPrm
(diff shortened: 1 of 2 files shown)


pkg/local_object_storage/blobovnicza/get_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+
+	oidtest "github.com/nspcc-dev/neofs-sdk-go/object/id/test"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBlobovnicza_Get(t *testing.T) {
+	t.Run("re-configure object size limit", func(t *testing.T) {
```
fyrchik, line 13: You have a single `t.Run` in this test, why not just change the name of the test?

fyrchik (approved):

pkg/local_object_storage/blobovnicza/get_test.go
cthulhu-rider: I stick to the approach of creating a test function for an operation, not for a scenario. Do u think it's worth to have a separate function?

fyrchik: I mean we don't yet have other tests, so this looks a bit strange.

carpawell (approved):
I'm okay with that solution except that we do not migrate the objects. I'm not sure if we need it since I have not thought about it deeply.

@realloc, @fyrchik

PR #1772, opened 2022-09-08
Fix/Acl in tree service

### LINKED ISSUE #1628: Improve tree service access control
#1328 implemented tree service access control. However there is an issue with such approach: tree service allows access only if request is made by container owner or it contains bearer token. This limits S3 Gateway use cases.

S3 Gateway operates with bearer tokens of the client's buckets/containers. If client tries to access someone else's bucket/container, then it produces request without bearer token. If the object is public, then extended ACL table allows to return object to S3 Gateway without bearer token, which is expected behaviour. 

Tree service doesn't work this way and won't return response to S3 Gateway without bearer token.

We should consider the same access control rules for both object service and tree service.
Or provide additional rules in extended ACL specifically for tree service.

It is blocking https://github.com/nspcc-dev/neofs-s3-gw/issues/505 and https://github.com/nspcc-dev/neofs-s3-gw/issues/504

carpawell:
Close #1628.


pkg/services/tree/signature.go
```diff
@@ -23,9 +26,20 @@ type message interface {
 	SetSignature(*Signature)
 }
 
-// verifyClient verifies that the request for a client operation was either signed by owner
-// or contains a valid bearer token.
-func (s *Service) verifyClient(req message, cid cidSDK.ID, rawBearer []byte, op eacl.Operation) error {
+func aclErr(op acl.Op) error {
```
cthulhu-rider, line 29: `basicACLErr` would look clearer, wouldn't it?

pkg/services/tree/signature.go
```diff
@@ -23,9 +26,20 @@ type message interface {
...
+	return fmt.Errorf("access to operation %s is denied by basic ACL check", op)
+}
+
+func eaclErr(op eacl.Operation, err error) error {
+	return fmt.Errorf("access to operation %s is denied by extended ACL check: %w", op, err)
+}
+
+// verifyClient verifies that the request for a client operation was either signed by an
```
cthulhu-rider, line 37: Not sure this sentence is grammatically correct :eyeglasses:

pkg/services/tree/signature.go
```diff
@@ -132,3 +159,50 @@ func signMessage(m message, key *ecdsa.PrivateKey) error {
...
+	switch op {
+	case acl.OpObjectGet:
+		eaclOp = eacl.OperationGet
+	case acl.OpObjectPut:
+		eaclOp = eacl.OperationPut
+	default:
+		panic(fmt.Sprintf("unexpected tree service ACL operation: %s", op))
+	}
+
+	return
```
cthulhu-rider, line 194: I'd move it into `switch`, same for `eACLRole`.

pkg/services/tree/signature.go
```diff
@@ -132,3 +159,50 @@ func signMessage(m message, key *ecdsa.PrivateKey) error {
 
 	return nil
 }
+
+func roleFromReq(cnr *core.Container, req message) (acl.Role, error) {
+	role := acl.RoleOthers
+	owner := cnr.Value.Owner()
+
+	// No error is expected because `VerifyDataWithSource` checks the signature.
```
cthulhu-rider, line 167: `VerifyDataWithSource` seems out-of-context here. Why do you mention it? Is it expected to have already been called?

pkg/services/tree/signature.go
carpawell: incorrectly changed that comment while was changing the code. fixed, could you check it one more time?

pkg/services/tree/signature.go
carpawell: in fact, it was just [copied](https://github.com/nspcc-dev/neofs-node/blob/8e8265d5ea4e7c878815b832c523d8d9bc6ecc8c/pkg/services/tree/signature.go#L42-L43), not my code. dropped

carpawell:
Added a commit that changes eACL checks logic: do not check filters because if their meaningless in that context. Kept the previous behavior: a request could pass the eACL checks only if a table contains an explicit rule that allows that signer to make that (mapped to an object) request.

pkg/services/tree/signature.go
```diff
@@ -36,53 +50,66 @@ func (s *Service) verifyClient(req message, cid cidSDK.ID, rawBearer []byte, op
...
-	if !bt.AssertContainer(cid) {
-		return errors.New("bearer token is created for another container")
-	}
-	if !bt.VerifySignature() {
-		return errors.New("invalid bearer token signature")
-	}
+	eaclOp := eACLOp(op)
+
+	var tb eacl.Table
+	if len(rawBearer) != 0 && basicACL.AllowedBearerRules(op) {
```
fyrchik, line 72: I think it might be useful to log with DEBUG level if `len(rawBearer) != 0` and `basicACL.AllowedBearerRules(op)` is false. What do you think?

pkg/services/tree/signature.go
```diff
@@ -200,3 +188,57 @@ func eACLRole(role acl.Role) eacl.Role {
...
+		// check target
+		if !targetMatches(record, role, signer) {
+			continue
+		}
+
+		switch a := record.Action(); a {
+		case eacl.ActionAllow:
+			return nil
+		case eacl.ActionDeny:
+			return eACLErr(op, errors.New("DENY eACL rule"))
```
fyrchik, line 215: Can we use private var for this `DENY eACL` error, to avoid allocations?

pkg/services/tree/signature.go
carpawell: agree, added debug log for that case

pkg/services/tree/signature.go
carpawell: sure, added vars for all errors in that file

carpawell:
Also added some comments about ACL checks to the `.proto` file.

KirillovDenis (approved):
LGTM, checked for https://github.com/nspcc-dev/neofs-s3-gw/issues/505

fyrchik (approved):

masterSplinter01 (approved):

PR #1775, opened 2022-09-12
go.mod: Update neofs-sdk-go

fyrchik:


carpawell (commented):
A few comments to the commit why do we need that?

carpawell (approved):

Issue #1731, opened 2022-08-24
Allow to move data from one shard to others


fyrchik:
During disk replacement it is needed to move all objects somewhere. `control dump-shard` is nice, but it would be better to move everything locally.
After this is implemented we could try replicating objects to other container nodes in case there are no suitable shards.

fyrchik:
Pilorama should be synchronized each epoch after #1929 .

PR #1777, opened 2022-09-12
Allow to move data away from shard

fyrchik:
Refs #1731


cmd/neofs-cli/modules/control/evacuate_shard.go
```diff
@@ -0,0 +1,56 @@
...
+
+	cli := getClient(cmd, pk)
+
+	var resp *control.EvacuateShardResponse
+	var err error
+	err = cli.ExecRaw(func(client *client.Client) error {
+		resp, err = control.EvacuateShard(client, req)
+		return err
+	})
+	if resp != nil {
```
cthulhu-rider, line 36: Do we need this `if`? I guess exactly one non-nil error is returned, so we can just check `common.ExitOnErr` before.

pkg/local_object_storage/engine/evacuate.go
```diff
@@ -0,0 +1,146 @@
...
+const defaultEvacuateBatchSize = 100
+
+type pooledShard struct {
+	hashedShard
+	pool util.WorkerPool
+}
+
+var errMustHaveTwoShards = errors.New("amount of shards must be > 2")
+
+func (e *StorageEngine) Evacuate(prm EvacuateShardPrm) (EvacuateShardRes, error) {
```
cthulhu-rider, line 49: I'd expect some docs with error cases, but not critical.

cthulhu-rider (approved):

pkg/local_object_storage/engine/evacuate.go
```diff
@@ -0,0 +1,144 @@
...
+		})
+	}
+	e.mtx.RUnlock()
+
+	weights := make([]float64, 0, len(shards))
+	for i := range shards {
+		weights = append(weights, e.shardWeight(shards[i].Shard))
+	}
+
+	if !sh.GetMode().ReadOnly() {
```
carpawell, line 83: can we move that check higher?

cmd/neofs-cli/modules/control/evacuate_shard.go
fyrchik: Fixed

pkg/local_object_storage/engine/evacuate.go
fyrchik: Fixed

pkg/local_object_storage/engine/evacuate.go
```diff
@@ -0,0 +1,146 @@
...
+				}
+				return res, err
+			}
+
+			hrw.SortSliceByWeightValue(shards, weights, hrw.Hash([]byte(lst[i].EncodeToString())))
+			for j := range shards {
+				if shards[j].ID().String() == sid {
+					continue
+				}
+				ok := e.putToShard(shards[j].Shard, i, shards[j].pool, lst[i], getRes.Object())
```
carpawell: is the `j` shard with the `i` index a trick here to prevent `ToMoveIt` operation? if yes, does it really have to be `i` (obj loop index, not the shard)?

pkg/local_object_storage/engine/evacuate.go
```diff
@@ -0,0 +1,146 @@
...
+	var c *meta.Cursor
+	var res EvacuateShardRes
+	for {
+		listPrm.WithCursor(c)
+
+		// TODO (@fyrchik): #1731 this approach doesn't work in degraded modes
+		//  because ListWithCursor works only with the metabase.
+		listRes, err := sh.Shard.ListWithCursor(listPrm)
+		if err != nil {
+			return res, nil
```
carpawell, line 99: is that (listing error) the only success return from the infinite loop?

pkg/local_object_storage/engine/evacuate.go
fyrchik: Done

pkg/local_object_storage/engine/evacuate.go
fyrchik: Fixed

pkg/local_object_storage/engine/evacuate.go
fyrchik: Yes, fixed error check

carpawell (approved):

Issue #1740, opened 2022-08-26
Sometimes contracts are not deployed with timeout


Alexpast1423:
Sometimes i have a timeout when deploy contracts on create cluster
i'm use 
* neofs-contract-v0.15.4
* /usr/bin/neo-go --version
NeoGo
Version: 0.99.2-2-g536303ef
GoVersion: go1.18.4

On deploy i have error: 
```
Stage 1: transfer GAS to alphabet nodes.
Waiting for transactions to persist...
Error: timeout while waiting for transaction to persist
Waiting for transactions to persist...
```

And on node: 
```
root@az:~# service neogo-morph-cn status
● neogo-morph-cn.service - NeoGo N3-morph-cn node
     Loaded: loaded (/etc/systemd/system/neogo-morph-cn.service; enabled; vendor preset: enabled)
     Active: active (running) since Fri 2022-08-26 01:00:45 UTC; 5h 48min ago
   Main PID: 4985 (neo-go)
      Tasks: 11 (limit: 9510)
     Memory: 63.0M
        CPU: 25.169s
     CGroup: /system.slice/neogo-morph-cn.service
             └─4985 /usr/bin/neo-go node --mainnet --config-path=/etc/neofs/morph-cn

авг 26 06:49:29 az neo-go[4985]: 2022-08-26T06:49:29.103Z        INFO        received PrepareResponse        {"validator": 3}
авг 26 06:49:29 az neo-go[4985]: 2022-08-26T06:49:29.103Z        INFO        received Commit        {"validator": 3}
авг 26 06:49:29 az neo-go[4985]: 2022-08-26T06:49:29.103Z        INFO        received Commit        {"validator": 1}
авг 26 06:49:29 az neo-go[4985]: 2022-08-26T06:49:29.103Z        INFO        approving block        {"height": 2494, "hash": "7a8c9b5c0d4ba72ac1d0716c57a5e9b23ce641fbde08530673db702b1f618687", "tx_count": 0, "merkle": "000000000000000000>
авг 26 06:49:29 az neo-go[4985]: 2022-08-26T06:49:29.104Z        INFO        initializing dbft        {"height": 2495, "view": 0, "index": 2, "role": "Backup"}
авг 26 06:49:29 az neo-go[4985]: 2022-08-26T06:49:29.330Z        INFO        Error encountered with rpc request        {"code": -100, "method": "getcontractstate", "params": "[1]"}
[...]

PR #1785, opened 2022-09-14
neofs-adm: use `ValidUntilBlock` to wait for tx persist

### LINKED ISSUE #1741: Set neofs-adm timeout values based on protocol settings
Currently we hardcode timeout in `awaitTx` https://github.com/nspcc-dev/neofs-node/blob/180e5e938f55a54393c84404f6ea2a9ce1dd270c/cmd/neofs-adm/internal/modules/morph/n3client.go#L80

It would be nice to set it depending on the protocol settings.

fyrchik:
Close #1741.
Related #1740 .

### DIFF
--- a/cmd/neofs-adm/internal/modules/morph/initialize.go
+++ b/cmd/neofs-adm/internal/modules/morph/initialize.go
@@ -315,7 +315,7 @@ func (c *initializeContext) getSigner(tryGroup bool) transaction.Signer {
 }
 
 func (c *clientContext) awaitTx(cmd *cobra.Command) error {
-	if len(c.Hashes) == 0 {
+	if len(c.SentTxs) == 0 {
 		return nil
 	}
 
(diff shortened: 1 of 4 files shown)


cmd/neofs-adm/internal/modules/morph/initialize.go
```diff
@@ -325,53 +325,58 @@ func (c *clientContext) awaitTx(cmd *cobra.Command) error {
...
+		for range tick.C {
+			// We must fetch current height before application log, to avoid race condition.
+			currBlock, err = c.GetBlockCount()
+			if err != nil {
+				return fmt.Errorf("can't fetch current block height: %w", err)
+			}
+			res, err := c.GetApplicationLog(txs[i].hash, &at)
+			if err == nil {
+				if retErr == nil && len(res.Executions) > 0 && res.Executions[0].VMState != vmstate.Halt {
+					retErr = fmt.Errorf("tx %d persisted in %s state: %s",
```
carpawell, line 373: not about that PR: why do we stop if a TX is expired and won't be added to the chain but don't stop if a TX has persisted but with `HALT` state?
and, moreover, we return only one `retErr` and do not log (at least) others

cmd/neofs-adm/internal/modules/morph/initialize.go
```diff
@@ -325,53 +325,58 @@ func (c *clientContext) awaitTx(cmd *cobra.Command) error {
...
-func awaitTx(cmd *cobra.Command, c Client, hashes []util.Uint256) error {
+func awaitTx(cmd *cobra.Command, c Client, txs []hashVUBPair) error {
 	cmd.Println("Waiting for transactions to persist...")
 
-	// improve TX awaiting process:
-	// https://github.com/nspcc-dev/neofs-node/issues/1741
 	const pollInterval = time.Second
-	const waitDuration = 30 * time.Second
 
 	tick := time.NewTicker(pollInterval)
```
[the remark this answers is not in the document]
fyrchik: Well, it is tricky because even with 5 sec block/timer, we would wait uselessly wait half of the interval. I think 1 second is ok.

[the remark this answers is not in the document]
carpawell: well, i thought about `interval/2` but 1 sec is also not smth bad

cmd/neofs-adm/internal/modules/morph/initialize.go
fyrchik: I don't remember the reason, could you create an issue for discussion.

carpawell: #1788

carpawell (approved):

Issue #1128, opened 2022-02-01
Remove WIF support in `--wallet` argument


532910:
I believe there should be no way to pass a sensitive data as a command line argument.

alexvanin:
`--wif` is removed in v0.27.5. You still able to provide WIF with `--wallet` but it is less encouraged this way. Indeed, it is not secure, but it is useful for testing (e.g. neofs-testcases uses it a lot). 

Config is tightly coupled with arguments, it will be hard to keep WIF in config but remove it from arguments. 

I guess we will drop WIF support completely one day.

carpawell:
We are ready.

fyrchik:
I think NEP2 support can also be dropped, I haven't never seen it being actually used anywhere.

PR #1789, opened 2022-09-16
[nspcc-dev#1128] cli: Remove WIF and NEP2 support in --wallet argument

acid-ant:
Closes: #1128


cmd/neofs-cli/internal/key/raw.go
```diff
@@ -16,13 +16,8 @@ import (
 
 var errCantGenerateKey = errors.New("can't generate new private key")
 
-// Get returns private key from the following sources:
-// 1. WIF
-// 2. Raw binary key
-// 3. Wallet file
-// 4. NEP-2 encrypted WIF.
+// Get returns private key from wallet file only.
```
carpawell, line 19: raw binary keys are still supported, no?

cmd/neofs-cli/internal/key/wallet.go
```diff
@@ -15,7 +15,7 @@ import (
 
 // Key-related errors.
 var (
-	ErrInvalidKey      = errors.New("provided key is incorrect")
+	ErrInvalidKey      = errors.New("provided key is incorrect, WIF and NEP2 are unsupported")
```
carpawell, line 18: do we need to mention that in err message?

carpawell (commented):
1. Could you, please, add a commit sign-off? We have a CI step for that and it has failed in that PR.
2. We also should fix flag description. E.g. `neofs-cli container create -h`: `-w, --wallet string         WIF (NEP-2) string or path to the wallet or binary key`

cmd/neofs-cli/internal/key/wallet.go
realloc, line 18: It may be better to mention the expected format first.

cmd/neofs-cli/internal/key/raw.go
```diff
@@ -16,13 +16,8 @@ import (
 
 var errCantGenerateKey = errors.New("can't generate new private key")
 
-// Get returns private key from the following sources:
-// 1. WIF
-// 2. Raw binary key
```
realloc, line 21: Raw binary still may be usefull. We want to prevent key leak via WIF copy-paste in shell and similar insecure scenarios first.

cmd/neofs-cli/internal/key/raw.go
```diff
@@ -32,23 +27,16 @@ func Get(cmd *cobra.Command) *ecdsa.PrivateKey {
...
-	p, err := getKeyFromFile(keyDesc)
-	if err == nil {
-		return p, nil
-	}
-
-	w, err := wallet.NewWalletFromFile(keyDesc)
-	if err == nil {
-		return FromWallet(w, viper.GetString(commonflags.Account))
-	}
+	if _, err := os.Stat(keyDesc); err == nil {
```
realloc, line 30: Won't we lose FS errors here and always return `invalid key`?

realloc:
Please avoid merge commits at rebase at all cost.

cmd/neofs-cli/internal/key/raw.go
fyrchik: If we have only wallet/raw key left, both a file, and read-error should be the same as stat one.

cmd/neofs-cli/internal/key/raw.go
acid-ant: yes, still supported, updated.

cmd/neofs-cli/internal/key/raw.go
acid-ant: updated.

cmd/neofs-cli/internal/key/wallet.go
acid-ant: thanks, done.

acid-ant: done

cmd/neofs-cli/internal/key/raw.go
acid-ant: Thanks, done.

fyrchik (dismissed):

carpawell (dismissed):

carpawell:
@acid-ant, for future: we also try to keep [CHANGELOG.md](https://github.com/nspcc-dev/neofs-node/blob/3df98ce7ba67da83631e36c2e52c7f5f85b850f7/CHANGELOG.md) in an actual state in every PR.

realloc:
Is there a related issue? If not, could you please take care of labels and projects?

acid-ant:
> Is there a related issue? If not, could you please take care of labels and projects?

This is a related issue https://github.com/nspcc-dev/neofs-node/issues/1128

acid-ant:
> @acid-ant, for future: we also try to keep [CHANGELOG.md](https://github.com/nspcc-dev/neofs-node/blob/3df98ce7ba67da83631e36c2e52c7f5f85b850f7/CHANGELOG.md) in an actual state after every PR.

Updated. Please review.

fyrchik (approved):

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ Changelog for NeoFS Node
 ### Fixed
 
 ### Removed
-
+- Remove WIF and NEP2 support in --wallet argument
```
carpawell, line 13: the last edit:

```suggestion
- WIF and NEP2 support in `neofs-cli`'s --wallet flag (#1128)
```

- we specify an issue number if exists or a current PR number if we work without an issue
- that repo contains 5 (and could more in the future) executables so it could be helpful to mention what exactly binary has changed its behavior

carpawell (changes requested):

CHANGELOG.md
acid-ant: Thanks, addressed.

carpawell (approved):

Issue #1420, opened 2022-05-23
How to check the relationship between requested object with session token object?


cthulhu-rider:
In current implementation ACL service uses object ID from session token context for access control. As I remember, this was done in order to grant access to parties which assemble the object (request generation). However, storage node does not check the relationship between the object from the token and the explicitly requested object. This gives rise to potential access control violations in the system when a token issued to another object is used to obtain an object.

I propose to discuss possible approaches to fix this problem. Here are some solutions from my mind:

0. Let it be as it is
1. Explicit checking of object association (this generally requires network requests, which can lead to new access questions)
2. Gather split chains and write them into session token (removes the problem completely, at the same time complicates the formation of a token)

cthulhu-rider:
Discussed with @realloc @fyrchik @carpawell 

2nd looks attractive. We need to make object list in object session context `repeated` (backward compatible change). We also need to provide iteration over object split-chain similar to the assembly of an object.

With this approach we will tighten authorization and access control of generated requests.

alexvanin:
From the service side, the cost of 2nd option is to have one/two extra requests. Is it possible to take advantage of https://github.com/nspcc-dev/neofs-api/issues/220 in trusted environment and avoid setting object.IDs?

To issue session token we need to find all children, if they exist. Probably the fastest way is to find linking object and get info from there. It will be nice to have search filter to that.

Another option is to get `SplitInfoError`. In this case, if I remember correctly, app should send `object.Head` request to the container node. Don't think that it is a good idea to work with placements and network map manually.

roman-khimov:
Related to #2667.

roman-khimov:
I think this is solved now. @cthulhu-rider?

PR #1791, opened 2022-09-16
object/acl: Fix correlation of object session to request

cthulhu-rider:
In previous implementation of `neofs-node` app object session was not checked for substitution of the object related to it. Also, for access checks, the session object was substituted instead of the one from the request. This, on the one hand, made it possible to inherit the session from the parent object for authorization for certain actions. On the other hand, it covered the mentioned object substitution, which is a critical vulnerability.

Next changes are applied to processing of all Object service requests:
 - check if object session relates to the requested object
 - use requested object in access checks.

Disclosed problem of object context inheritance will be solved within #1420.

* relates #1420

### DIFF
--- a/pkg/services/object/acl/v2/util.go
+++ b/pkg/services/object/acl/v2/util.go
@@ -93,24 +93,12 @@ func originalSessionToken(header *sessionV2.RequestMetaHeader) (*sessionSDK.Obje
 	return &tok, nil
 }
 
-func getObjectIDFromRequestBody(body interface{}) (*oid.ID, error) {
-	var idV2 *refsV2.ObjectID
-
-	switch v := body.(type) {
-	default:
-		return nil, nil
-	case interface {
-		GetObjectID() *refsV2.ObjectID
-	}:
-		idV2 = v.GetObjectID()
-	case interface {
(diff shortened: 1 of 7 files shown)


pkg/services/object/acl/v2/util.go
```diff
@@ -93,64 +93,33 @@ func originalSessionToken(header *sessionV2.RequestMetaHeader) (*sessionSDK.Obje
...
-		GetAddress() *refsV2.Address
-	}:
-		idV2 = v.GetAddress().GetObjectID()
-	}
+// decodeRequestedObject decodes oid.ID from refsV2.ObjectID message if latter
+// is presented, otherwise returns an error iff message is required.
+func decodeRequestedObject(m *refsV2.ObjectID, required bool) (*oid.ID, error) {
+	if m == nil {
+		if required {
+			return nil, errors.New("missing object ID")
```
fyrchik, line 107: It seems like we use this func in only 2 places, may be it is better to have `nil` check on the call-site and remove the second parameter?

fyrchik (approved):

fyrchik (approved):

PR #1105, opened 2022-01-24
Add control API for shard evacuation

fyrchik:
Depends on #1089 .
Close #1086.

PR #1800, opened 2022-09-19
Allow to replicate object to other nodes in `control.EvacuateShard`

fyrchik:
Related #1105

### DIFF
--- a/pkg/local_object_storage/engine/evacuate.go
+++ b/pkg/local_object_storage/engine/evacuate.go
@@ -8,12 +8,15 @@ import (
 	meta "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/metabase"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/shard"
 	"github.com/nspcc-dev/neofs-node/pkg/util"
+	objectSDK "github.com/nspcc-dev/neofs-sdk-go/object"
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"go.uber.org/zap"
 )
 
 // EvacuateShardPrm represents parameters for the EvacuateShard operation.
 type EvacuateShardPrm struct {
 	shardID      *shard.ID
+	handler      func(oid.Address, *objectSDK.Object) error
 	ignoreErrors bool
 }
 
@@ -32,7 +35,13 @@ func (p *EvacuateShardPrm) WithIgnoreErrors(ignore bool) {
 	p.ignoreErrors = ignore
 }
 
+// WithFaultHandler sets handler to call for objects which cannot be saved on other shards.
+func (p *EvacuateShardPrm) WithFaultHandler(f func(oid.Address, *objectSDK.Object) error) {
+	p.handler = f
+}
+
 // Count returns amount of evacuated objects.
+// Objects for which handler returned no error are also assumed evacuated.
 func (p EvacuateShardRes) Count() int {
 	return p.count
 }
@@ -58,7 +67,7 @@ func (e *StorageEngine) Evacuate(prm EvacuateShardPrm) (EvacuateShardRes, error)
 		return EvacuateShardRes{}, errShardNotFound
 	}
 
-	if len(e.shards) < 2 {
+	if len(e.shards) < 2 && prm.handler == nil {
 		e.mtx.RUnlock()
 		return EvacuateShardRes{}, errMustHaveTwoShards
 	}
@@ -138,13 +147,17 @@ func (e *StorageEngine) Evacuate(prm EvacuateShardPrm) (EvacuateShardRes, error)
 				}
 			}
 
-			// TODO (@fyrchik): #1731 try replicating to another node.
-			//  The suggestion is to have prm.handler which is called
-			//  if a Put has failed.
+			if prm.handler == nil {
+				// Do not check ignoreErrors flag here because
+				// ignoring errors on put make this command kinda useless.
+				return res, fmt.Errorf("%w: %s", errPutShard, lst[i])
+			}
 
-			// Do not check ignoreErrors flag here because
-			// ignoring errors on put make this command kinda useless.
-			return res, fmt.Errorf("%w: %s", errPutShard, lst[i])
+			err = prm.handler(lst[i], getRes.Object())
+			if err != nil {
+				return res, err
+			}
+			res.count++
 		}
 
 		c = listRes.Cursor()
--- a/pkg/services/control/server/evacuate.go
+++ b/pkg/services/control/server/evacuate.go
@@ -86,11 +86,11 @@ func (s *Server) replicate(addr oid.Address, obj *objectSDK.Object) error {
 	}
 
 	var res replicatorResult
-	task := new(replicator.Task).
-		WithObject(obj).
-		WithObjectAddress(addr).
-		WithCopiesNumber(1).
-		WithNodes(nodes)
+	var task replicator.Task
+	task.SetObject(obj)
+	task.SetObjectAddress(addr)
+	task.SetCopiesNumber(1)
+	task.SetNodes(nodes)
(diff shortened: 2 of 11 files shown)


cthulhu-rider (dismissed):

pkg/local_object_storage/engine/evacuate.go
```diff
@@ -32,6 +35,11 @@ func (p *EvacuateShardPrm) WithIgnoreErrors(ignore bool) {
 	p.ignoreErrors = ignore
 }
 
+// WithHandler sets handler to call for objects which cannot be saved on other shards.
+func (p *EvacuateShardPrm) WithHandler(f func(oid.Address, *objectSDK.Object) error) {
```
carpawell, line 39: maybe expand naming with "Error" somehow? not clear what exactly that handler should handle

pkg/local_object_storage/engine/evacuate.go
```diff
@@ -138,13 +146,17 @@ func (e *StorageEngine) Evacuate(prm EvacuateShardPrm) (EvacuateShardRes, error)
...
+			}
 
-			// Do not check ignoreErrors flag here because
-			// ignoring errors on put make this command kinda useless.
-			return res, fmt.Errorf("%w: %s", errPutShard, lst[i])
+			err = prm.handler(lst[i], getRes.Object())
+			if err != nil {
+				return res, err
+			}
+			res.count++
```
carpawell, line 159: so technically that that counter is not about the object being evacuated successfully now OR the handler MUST evacuate the object somewhere?

pkg/services/control/server/evacuate.go
```diff
@@ -39,3 +49,62 @@ func (s *Server) EvacuateShard(_ context.Context, req *control.EvacuateShardRequ
...
+func (s *Server) replicate(addr oid.Address, obj *objectSDK.Object) error {
+	nm, err := s.netMapSrc.GetNetMap(0)
+	if err != nil {
+		return err
+	}
+
+	cid, ok := obj.ContainerID()
+	if !ok {
+		// Container ID is expected to be here. If for some reason it is not,
+		// we log error and return nil to prevent situations
```
carpawell, line 62: log or not log?

also, could it be the first check before any network interaction?

pkg/services/control/server/evacuate.go
```diff
@@ -39,3 +49,62 @@ func (s *Server) EvacuateShard(_ context.Context, req *control.EvacuateShardRequ
...
+	if err != nil {
+		return fmt.Errorf("can't build a list of container nodes")
+	}
+
+	nodes := placement.FlattenNodes(ns)
+	bs := (*keys.PublicKey)(&s.key.PublicKey).Bytes()
+	for i := 0; i < len(nodes); i++ {
+		if bytes.Equal(nodes[i].PublicKey(), bs) {
+			copy(nodes[i:], nodes[i+1:])
+			nodes = nodes[:len(nodes)-1]
```
carpawell, line 85: break here?

pkg/services/control/server/evacuate.go
```diff
@@ -39,3 +49,62 @@ func (s *Server) EvacuateShard(_ context.Context, req *control.EvacuateShardRequ
...
+		}
+	}
+
+	var res replicatorResult
```
carpawell, line 89: what if `nodes` is empty after the current node exclusion?

pkg/services/control/server/evacuate.go
```diff
@@ -39,3 +49,62 @@ func (s *Server) EvacuateShard(_ context.Context, req *control.EvacuateShardRequ
...
+	task := new(replicator.Task).
+		WithObject(obj).
+		WithObjectAddress(addr).
+		WithCopiesNumber(1).
+		WithNodes(nodes)
+	s.replicator.HandleTask(context.TODO(), task, &res)
```
carpawell, line 95: what if other nodes already have that object? i think that is quite a common case, seems like it is a useless operation then

pkg/local_object_storage/engine/evacuate.go
fyrchik: It doesn't handle errors, it is called for objects which can't be put to other shards, which is what comment says.

pkg/local_object_storage/engine/evacuate.go
fyrchik: Fixed a comment for `Count`

pkg/services/control/server/evacuate.go
fyrchik: I don't think we need to complicate this, such errors should be caught in other places.
Moved network map fetching.

pkg/services/control/server/evacuate.go
fyrchik: We could have the same node in different replicas. User must ensure it is not like this, and user can't be trusted.

pkg/services/control/server/evacuate.go
fyrchik: The `res.count == 0` after `HandleTask` and we will return an error.

pkg/services/control/server/evacuate.go
fyrchik: No, we try the fullfill the policy -- we store an object, but will drop it. So we replicate manually here.

pkg/local_object_storage/engine/evacuate.go
carpawell: > It doesn't handle errors

yeah, i got it, did not think much about naming just wanted to highlight that `Handler` could be misunderstood since it is not a general handler: it is not called on every object, only called on "error" object. i would name that "FaultHandler" or smth like that

let it be up to you

pkg/services/control/server/evacuate.go
carpawell: what about "log"? mb i get it wrong, what exactly logging that comment is telling about? general? not logging that err here?

pkg/services/control/server/evacuate.go
carpawell: oh, thought that `FlattenNodes` does not contain repeats, my bad

pkg/services/control/server/evacuate.go
carpawell: `HandleTask` has some debug logging even with an empty task so i meant it could be cached earlier, but ok

pkg/services/control/server/evacuate.go
carpawell: > we try the fullfill the policy

are you sure? in the replicator we [check](https://github.com/nspcc-dev/neofs-node/blob/c7f85994e550d4c990c08d1ea6529ec99c954770/pkg/services/policer/check.go#L127) if a node contain the object and [select](https://github.com/nspcc-dev/neofs-node/blob/c7f85994e550d4c990c08d1ea6529ec99c954770/pkg/services/policer/check.go#L147) only the nodes that do not have an object before the `HandleTask` call

seems that you take *all* the nodes from the placement, exclude the current one, and try to put the object to the first one *one* time, that is useless if `REP > 1`

i may be wrong

pkg/local_object_storage/engine/evacuate.go
fyrchik: Agree, fixed.

pkg/services/control/server/evacuate.go
fyrchik: Removed "log", the purpose of this comment is to justify returning `nil` even though missing CID is an error.

pkg/services/control/server/evacuate.go
fyrchik: This is tricky, in this situation node is expected to be in a maintenance mode, so most likely policer will do the job itself.
Here we only try to _ensure_ that an object is _not lost_. Maintaining a proper amount of copies is a policer task.

carpawell (approved):

cthulhu-rider (approved):

PR #1809, opened 2022-09-21
Fix/Do not boot up if metabase is outdated

carpawell:


acid-ant (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ Changelog for NeoFS Node
 
 ### Fixed
 
+- Outdated metabase start up (#1809)
```
fyrchik, line 12: "fail startup if metabase has old version"?

pkg/local_object_storage/metabase/version.go
```diff
@@ -12,14 +12,25 @@ const version = 2
...
 
+// ErrOutdatedVersion is returned on initializing
+// an existing metabase that is not compatible with
+// the current code version.
+type ErrOutdatedVersion struct {
+	stored uint64
+}
+
+func (e ErrOutdatedVersion) Error() string {
+	return fmt.Sprintf("invalid version: expected=%d, stored=%d", version, e.stored)
```
fyrchik, line 23: I don't like complex error types, what about `var ErrOutdatedVersion = errors.New("invalid version")` + `return fmt.Errorf("%w: expected...", ErrOutdatedVersion, ...)`?

cthulhu-rider: Here I agree with @fyrchik at least since we don't decode the error structure. We could use `errors.Is`.

fyrchik (approved):

PR #1817, opened 2022-09-27
Use `ExternalAddr` attribute for inter-node communication

fyrchik:


acid-ant (approved):

pkg/core/netmap/nodes.go
```diff
@@ -17,12 +17,15 @@ func (x Node) PublicKey() []byte {
 // IterateAddresses iterates over all announced network addresses
 // and passes them into f. Handler MUST NOT be nil.
 func (x Node) IterateAddresses(f func(string) bool) {
+	for _, addr := range (netmap.NodeInfo)(x).ExternalAddresses() {
```
cthulhu-rider, line 20: So firstly we are going to process external addresses everywhere, is it expected for intra-container communication? I thought "old" addresses are preferred for this purpose.

cthulhu-rider:
@fyrchik are you sure you've changed all the needed places? I see a few more places of `NumberOfNetworkEndpoints` and `IterateNetworkEndpoints`.

`pkg/core/netmap.Node` type was borned to become a complete type mediator for SDK `netmap.NodeInfo`, but we haven't finished the job yet. Lets do it within a separate issue?

realloc:
IIRC, the ExternalAddr attribute should be used for client (sdk?) connections to nodes. For inter-node communication we have to use addressed from netmap and use ExternalAddr as a fallback only, if there is an option allowing such behaviour.

fyrchik:
There are 3 different approaches:
1. Don't touch the address structure, and decide if we want to use external addressses on the callsite (client cache).
2. Modify network.Address to have `external` field and take it into account during sorting.
3. Filter external addresses during `placement.Node` construction.

I decided to go with the first, because this way there are fewer places with filtering logic, so less bugs. Also, it will be easier to change `allow_external` setting on SIGHUP.

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ Changelog for NeoFS Node
 ### Added
 - Serving `NetmapService.NetmapSnapshot` RPC (#1793)
 - `netmap snapshot` command of NeoFS CLI (#1793)
-
+- `apiclient.allow_external` config flag to fallback to node external addresses (#1817)
```
carpawell, line 9: do we need to add that to the `Update from...` section too?

pkg/core/client/client.go
```diff
@@ -52,6 +54,16 @@ func (x NodeInfo) AddressGroup() network.AddressGroup {
 	return x.addrGroup
 }
 
+// SetExternalAddressGroup sets a group of network addresses.
```
carpawell, line 57: should it have _the same_ comment as `SetAddressGroup`?

pkg/core/netmap/nodes.go
```diff
@@ -17,6 +17,9 @@ func (x Node) PublicKey() []byte {
...
+		f(addr)
```
carpawell, line 21: no break if `f(addr) == true`?

CHANGELOG.md
fyrchik: I think no, it is a completely new feature, old nodes are expected to continue working.

pkg/core/client/client.go
fyrchik: Fixed

pkg/core/netmap/nodes.go
fyrchik: Fixed

CHANGELOG.md
carpawell: yes, but how will an admin know about that? i try to write any config changes to that section. but ok

carpawell (dismissed):

CHANGELOG.md
realloc: I suggest adding this to Update section. With new behaviour node admins can explicitly declare external addresses for their node, not only listen corresponding addresses.

fyrchik: Yes, but do we need `Added` section in this case? Every added feature can be used in some way, I've though `Updating from` was for things requiring special care.
Fixed anyway.

realloc: If we need Ops attention, we add/duplicate something in `Updating from`. For this particular case we need to highlight it.

carpawell (approved):

PR #1820, opened 2022-09-28
 neofs-adm: Add wallet-address flag in refill command

alexvanin:


cmd/neofs-adm/internal/modules/morph/generate.go
```diff
@@ -139,38 +141,55 @@ func refillGas(cmd *cobra.Command, gasFlag string, createWallet bool) error {
 	if err != nil {
 		return err
 	}
-	if storageWalletPath == "" {
-		return fmt.Errorf("missing wallet path (use '--%s <out.json>')", storageWalletFlag)
+	// wallet address is not part of the config
+	walletAddress, err := cmd.Flags().GetString(walletAddressFlag)
```
fyrchik, line 145: error could be ignored here, it is non-nil only if the flag has not been defined

cmd/neofs-adm/internal/modules/morph/generate.go
```diff
@@ -139,38 +141,55 @@ func refillGas(cmd *cobra.Command, gasFlag string, createWallet bool) error {
...
-	var w *wallet.Wallet
+	var gasReceiver util.Uint160
 
-	if createWallet {
-		w, err = wallet.NewWallet(storageWalletPath)
+	if len(walletAddress) != 0 {
+		gasReceiver, err = address.StringToUint160(walletAddress)
+		if err != nil {
+			return fmt.Errorf("invalid wallet address %s: %w", walletAddress, err)
+		}
```
fyrchik, line 156: In this case we should mark wallet path and address flags as mutually exclusive.

alexvanin: It's more like overrides `storage-wallet` flag. Updated flag description in `root.go`.

fyrchik: I mean if it is used `storage-wallet` is ignored, and they are never used together, why not make them mutually exclusive?

alexvanin: Added mutual exclusive check. I don't want to fix the code below, because `refillGas` is being used in credential generation routine where `wallet-address` does not exist. So I tried to avoid any major changes here.

acid-ant (approved):

fyrchik (approved):

PR #1821, opened 2022-09-28
Fix description of nodeinfo command

vdomnich-yadro:
Description updated because the command returns information about the remote node that is listening to the `--rpc-endpoint`


fyrchik (dismissed):

fyrchik:
Could you also prefix the commit with `[#...] neofs-cli: Fix description ...`?

fyrchik (approved):

fyrchik (approved):

PR #1823, opened 2022-09-28
[#1697] services/object: Return proper error if session token is missing

### LINKED ISSUE #1697: `Object not found` while deleting object with Session Token
Deletion of an object with a session token via a node which IS in the container (where the object is put) and NOT granted a session token fails with an unexpected error. While other operations using a session token via other nodes run as expected.

## Expected Behavior
Deletion fails with `rpc error: client failure: status: code = 4096 message = session token not found`.

## Current Behavior
Deletion fails with `rpc error: remove object via client: status: code = 2049 message = object not found`.

1. Create a wallet and make a deposit;
2. Created a private container (with a placement policy allowing one rep and one loc) and put an object thereinto;
3. Issue a session token with the wallet generated above to trust a certain node;
4. Try to delete the object via a node which IS IN the container and NOT granted a session token.

## Context
While writing a test for session token via `neofs-cli`.
https://github.com/EliChin/neofs-testcases/blob/feature/pytest_session/pytest_tests/testsuites/session_token/test_object_st.py

## Your Environment
neofs-node: 0.31.0
neo-go: 0.99.2
neofs-node:Ubuntu 0.22.4

fyrchik:
Close #1697 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Changelog for NeoFS Node
 
 ### Fixed
 - Description of command `netmap nodeinfo` (#1821)
+- Proper status for object.Delete if session token is missing (#1697)
 
 ### Removed
 - Remove WIF and NEP2 support in `neofs-cli`'s --wallet flag (#1128)
--- a/pkg/services/object/delete/delete.go
+++ b/pkg/services/object/delete/delete.go
@@ -3,11 +3,24 @@ package deletesvc
 import (
 	"context"
 
+	"github.com/nspcc-dev/neofs-node/pkg/services/object/util"
 	"go.uber.org/zap"
 )
 
 // Delete serves requests to remote the objects.
 func (s *Service) Delete(ctx context.Context, prm Prm) error {
+	// If session token is not found we will fail during tombstone PUT.
+	// Here we fail immediately to ensure no unnecessary network communication is done.
+	if tok := prm.common.SessionToken(); tok != nil {
+		_, err := s.keyStorage.GetKey(&util.SessionInfo{
+			ID:    tok.ID(),
+			Owner: tok.Issuer(),
+		})
+		if err != nil {
+			return err
+		}
+	}
+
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ Changelog for NeoFS Node
 
 ### Fixed
 - Description of command `netmap nodeinfo` (#1821)
+- Proper status for object.Delete is session token is missing (#1697)
```
carpawell, line 21:
```suggestion
- Proper status for object.Delete if session token is missing (#1697)
```

cthulhu-rider (dismissed):

pkg/services/object/delete/delete.go
```diff
@@ -3,11 +3,24 @@ package deletesvc
...
 
+	"github.com/nspcc-dev/neofs-node/pkg/services/object/util"
 	"go.uber.org/zap"
 )
 
 // Delete serves requests to remote the objects.
 func (s *Service) Delete(ctx context.Context, prm Prm) error {
+	// If session token is not found we will fail during tombstone PUT.
+	// Here we fail immediately to ensure no unnecessary network communication is done.
+	if tok := prm.common.SessionToken(); tok != nil {
```
carpawell, line 14: A few thoughts about it:

1. does that make object removing operation require a session token?
2. if `1.` is "yes", we need either remove such [branches](https://github.com/nspcc-dev/neofs-node/blob/5284ac53f906a0c1abd8f487245e077fa688c07a/pkg/services/object/delete/exec.go#L251-L255). if "no" (a node could remove public RW objects), then that change is incorrect

CHANGELOG.md
fyrchik: Fixed

pkg/services/object/delete/delete.go
carpawell: incorrectly understood that change.

PR #1825, opened 2022-09-30
writecache: flush cache when moving to the DEGRADED mode

fyrchik:
Degraded mode allows us to operate without an SSD, thus writecache should be unavailable in this mode.


acid-ant (approved):

pkg/local_object_storage/writecache/flush.go
```diff
@@ -59,7 +59,7 @@ func (c *cache) runFlushLoop() {
 	}()
 }
 
-func (c *cache) flush() {
+func (c *cache) flushDB() {
```
carpawell, line 62: why _that_ renaming?

carpawell (approved):

pkg/local_object_storage/writecache/flush.go
fyrchik: Because we already have `OpenDB`

carpawell (approved):

PR #1830, opened 2022-10-03
Refactor/Config in SN

carpawell:
Blocked by:
- [x] https://github.com/nspcc-dev/neofs-node/pull/1815

Related to #1770.

PR #1815, opened 2022-09-26
Feat/Reload config on SIGHUP

carpawell:
Related to #1770.

1. Storage interface could be unified: `Open` and `Reload` could have the same args list but that will require additional refactoring, not sure if we need that.
2. It now seems like we have duplication of configuration structs: one placed in the `main` pkg and another in the `cmd/neofs-node/config` dir. They could be merged somehow if we need them to be together.

PR #1831, opened 2022-10-03
Feat/Reread log level

carpawell:
Blocked by:
- [x] https://github.com/nspcc-dev/neofs-node/pull/1815
- [x] https://github.com/nspcc-dev/neofs-node/pull/1830

Consider `logger` refactor in a separate PR.

### DIFF
--- a/cmd/neofs-node/config.go
+++ b/cmd/neofs-node/config.go
@@ -492,7 +492,7 @@ func initCfg(appCfg *config.Config) *cfg {
 	)
 	fatalOnErr(err)
 
-	log, err := logger.NewLogger(logPrm)
+	log, err := logger.NewLogger(&logPrm)
 	fatalOnErr(err)
 
 	var netAddr network.AddressGroup
--- a/cmd/neofs-node/config.go
+++ b/cmd/neofs-node/config.go
@@ -289,7 +289,7 @@ type internals struct {
 
 	appCfg *config.Config
 
-	log *zap.Logger
+	log *logger.Logger
 
 	wg      *sync.WaitGroup
 	workers []worker
--- a/cmd/neofs-node/config.go
+++ b/cmd/neofs-node/config.go
@@ -87,6 +87,10 @@ type applicationConfiguration struct {
 	// has already been read
 	_read bool
 
+	LoggerCfg struct {
+		level string
+	}
+
 	EngineCfg struct {
 		errorThreshold uint32
 		shardPoolSize  uint32
@@ -181,6 +185,12 @@ func (a *applicationConfiguration) readConfig(c *config.Config) error {
 	}
 
 	a._read = true
+
+	// Logger
+
+	a.LoggerCfg.level = loggerconfig.Level(c)
+
+	// Storage Engine
 
 	a.EngineCfg.errorThreshold = engineconfig.ShardErrorThreshold(c)
 	a.EngineCfg.shardPoolSize = engineconfig.ShardPoolSize(c)
@@ -346,10 +356,17 @@ type shared struct {
 	metricsCollector *metrics.NodeMetrics
 }
 
+// dynamicConfiguration stores parameters of the
+// components that supports runtime reconfigurations
+type dynamicConfiguration struct {
+	logger *logger.Prm
+}
+
 type cfg struct {
 	applicationConfiguration
 	internals
 	shared
+	dynamicConfiguration
 
 	// configuration of the internal
 	// services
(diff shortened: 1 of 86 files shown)


acid-ant (dismissed):

pkg/util/logger/logger.go
```diff
@@ -9,10 +9,10 @@ import (
 
 // Logger represents a component
 // for writing messages to log.
-//
-// It is a type alias of
-// go.uber.org/zap.Logger.
-type Logger = zap.Logger
+type Logger struct {
+	*zap.Logger
```
cthulhu-rider, line 13: IMO it doesn't make any sense to do such wrapper. I think we should either use `zap.Logger` or completely autonomous `Logger` instance. The latter should be a pure mediator encapsulating `zap.Logger.`

carpawell: > IMO it doesn't make any sense

yes, agree. but i do not want to bring refactoring as long as it is possible, see comment below

carpawell:
We talked with @cthulhu-rider about our double productive work here and in #1832 and i merged our thoughts here.
I _do not_ like the idea of refactoring the logger in PR that adds SIGHUP support (that PR is already bigger than i expected it to be). I _do not_ like the idea of having two constructors for both cases (dynamic and static config, suggest using one and just drop the `Prm` if you do not want to reconfigure the `Logger`). I _do_ like the idea that a component should not provide `Reload` public method to any of its callers (e.g. logger is passed to almost each of our components and any of them is possible to call `Reload` then).

So my suggestion: merge that as it is (with the approach review of course) with its ugly `logger.Logger` refactor and consider it as inevitable preparation work for the future refactor. Paraphrase #1832 to sound like "Refactor of the Logger component", discuss its moving to SDK, and so on. @cthulhu-rider has done hard work and i have nothing against it. Also, every future dynamic configuration should look as (much as possible) it is done for the Logger in that PR.

/cc @fyrchik @acid-ant

pkg/util/logger/logger.go
```diff
@@ -15,44 +15,95 @@ type Logger struct {
...
+// ErrLoggerNotConnected is returned when a Prm has not been connected to any
+// Logger but Prm.Reload had been called.
+var ErrLoggerNotConnected = errors.New("parameters are not connected to any Logger")
+
+// Reload reloads configuration of a connected instance of the Logger.
+// Returns ErrLoggerNotConnected if no connection has been performed.
+// Returns any reconfiguration errors from the Logger directly.
+func (p Prm) Reload() error {
+	if p._log == nil {
+		return ErrLoggerNotConnected
```
fyrchik, line 60: When this can happen? Shouldn't we create a new logger in this case?

cmd/neofs-node/validate.go
```diff
@@ -7,13 +7,26 @@ import (
...
 
 // validateConfig validates storage node configuration.
 func validateConfig(c *config.Config) error {
+	// logger configuration validation
+
+	var loggerPrm logger.Prm
+
+	err := loggerPrm.SetLevelString(loggerconfig.Level(c))
+	if err != nil {
+		return fmt.Errorf("logger level: %w", err)
```
fyrchik, line 25: invalid logger level?

cmd/neofs-node/config.go
```diff
@@ -346,10 +356,17 @@ type shared struct {
 	metricsCollector *metrics.NodeMetrics
 }
 
+// dynamicConfiguration stores parameters of the
+// components that supports runtime reconfigurations
+type dynamicConfiguration struct {
```
fyrchik, line 361: Do we expect to store shard parameters and morph here in future?

carpawell: yes, i expect to store them (and any component that supports cfg reload) here and it would be perfect if all of them will have the same interface `Reload() err`

pkg/util/logger/logger.go
carpawell: > When this can happen?

```go
var Prm logger.Prm
prm.Reload() // here
```

we can panic since it is totally incorrect usage or return the static err

> Shouldn't we create a new logger in this case?

do you mean creating a new logger inside the `Reload` method and placing a new instance using `p._log` OR it is the general question of what a caller should do?

fyrchik: I mean do we have such code anywhere? If this error is not reachable we can panic here, may be even implicitly.

carpawell: > do we have such code anywhere

not yet (there are talks about moving the logger to SDK or a separate repo and sharing it b/w our repos)

> we can panic here

added a panic, but explicit, with the message

cthulhu-rider (approved):

fyrchik (approved):

PR #1843, opened 2022-10-06
Feat/Sync tree on startup

carpawell:
Relates #1329.

### DIFF
--- a/cmd/neofs-node/config.go
+++ b/cmd/neofs-node/config.go
@@ -42,6 +42,7 @@ import (
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/writecache"
 	"github.com/nspcc-dev/neofs-node/pkg/metrics"
 	"github.com/nspcc-dev/neofs-node/pkg/morph/client"
+	containerClient "github.com/nspcc-dev/neofs-node/pkg/morph/client/container"
 	nmClient "github.com/nspcc-dev/neofs-node/pkg/morph/client/netmap"
 	"github.com/nspcc-dev/neofs-node/pkg/morph/event"
 	netmap2 "github.com/nspcc-dev/neofs-node/pkg/morph/event/netmap"
@@ -305,6 +306,10 @@ type internals struct {
 	workers []worker
 	closers []func()
 
+	// onlineStateHandlers are executed in a separate
+	// goroutine on every !ONLINE -> ONLINE state transition
+	onlineStateHandlers []func(context.Context)
+
 	apiVersion   version.Version
 	healthStatus *atomic.Int32
 	// is node under maintenance
@@ -346,6 +351,8 @@ type shared struct {
 	// current network map
 	netMap       atomicstd.Value // type netmap.NetMap
 	netMapSource netmapCore.Source
+
+	cnrClient *containerClient.Client
 
 	respSvc *response.Service
 
@@ -831,6 +838,13 @@ func (c *cfg) LocalNodeInfo() (*netmapV2.NodeInfo, error) {
 // Called with nil when storage node is outside the NeoFS network map
 // (before entering the network and after leaving it).
 func (c *cfg) handleLocalNodeInfo(ni *netmap.NodeInfo) {
+	if c.cfgNetmap.state.controlNetmapStatus() != control.NetmapStatus_ONLINE &&
+		ni != nil && ni.IsOnline() {
+		for _, h := range c.onlineStateHandlers {
+			go h(c.ctx)
+		}
+	}
+
 	c.cfgNetmap.state.setNodeInfo(ni)
 }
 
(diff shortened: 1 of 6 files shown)


pkg/services/tree/replicator.go
```diff
@@ -130,6 +130,71 @@ func (s *Service) pushToQueue(cid cidSDK.ID, treeID string, op *pilorama.LogMove
 	}
 }
 
+func (s *Service) synchronize(ctx context.Context) {
```
fyrchik, line 133: Let's put it in `sync.go`? Also it would be nice to parametrize it by `CID`. When a node first appears in container without shutdown is also a valid use-case.

pkg/services/tree/replicator.go
```diff
@@ -130,6 +130,71 @@ func (s *Service) pushToQueue(cid cidSDK.ID, treeID string, op *pilorama.LogMove
...
+	if !s.nodeState.IsOnline() {
+		s.log.Info("tree service awaits for node inclusion to netmap...")
+
+		const waitInterval = 30 * time.Second
+		t := time.NewTimer(waitInterval)
```
fyrchik, line 138: Is it a `NewTimer`, not `NewTIcker`?

pkg/services/tree/replicator.go
```diff
@@ -130,6 +130,71 @@ func (s *Service) pushToQueue(cid cidSDK.ID, treeID string, op *pilorama.LogMove
 	}
 }
 
+func (s *Service) synchronize(ctx context.Context) {
+	if !s.nodeState.IsOnline() {
```
fyrchik, line 134: May be it is better to use `tree.SynchronizeAll` in some callback executed when an epoch has ticked?

carpawell:
@fyrchik, moved the code out of the `tree` pkg.

carpawell:
Could be now rechecked with #1874.

cmd/neofs-node/tree.go
```diff
@@ -53,3 +65,32 @@ func initTreeService(c *cfg) {
...
+func syncTrees(ctx context.Context, treeSvc *tree.Service, cnrCli *containerClient.Client, log *zap.Logger) {
+	log.Info("synchronizing trees...")
+
+	ids, err := cnrCli.List(nil)
+	if err != nil {
+		log.Error("trees are not synchronized", zap.Error(err))
+		return
+	}
+
+	// TODO: fetch all the trees via a new tree RPC
```
fyrchik, line 78: Issue?

cmd/neofs-node/tree.go
```diff
@@ -38,6 +40,16 @@ func initTreeService(c *cfg) {
...
 	}))
 
+	syncTreeFunc := func(ctx context.Context) {
+		syncTrees(ctx, c.treeService, c.shared.cnrClient, c.log)
+	}
+
+	if c.cfgNetmap.state.controlNetmapStatus() == control.NetmapStatus_ONLINE {
+		c.workers = append(c.workers, newWorkerFromFunc(syncTreeFunc))
+	} else {
+		c.addOnlineStateHandler(syncTreeFunc)
```
fyrchik, line 50: What happens if we start ONLINE, go OFFLINE and then ONLINE again?

cmd/neofs-node/config.go
```diff
@@ -803,6 +810,13 @@ func (c *cfg) LocalNodeInfo() (*netmapV2.NodeInfo, error) {
 // Called with nil when storage node is outside the NeoFS network map
 // (before entering the network and after leaving it).
 func (c *cfg) handleLocalNodeInfo(ni *netmap.NodeInfo) {
+	if c.cfgNetmap.state.controlNetmapStatus() != control.NetmapStatus_ONLINE &&
+		ni != nil && ni.IsOnline() {
+		for _, h := range c.onlineStateHandlers {
```
fyrchik, line 815: Right now we have only 1 use case, may be just call `syncTree` here?

cmd/neofs-node/netmap.go
```diff
@@ -81,8 +81,13 @@ func (s *networkState) setNodeInfo(ni *netmapSDK.NodeInfo) {
 	s.controlNetStatus.Store(ctrlNetSt)
 }
 
-func (s *networkState) controlNetmapStatus() control.NetmapStatus {
-	return s.controlNetStatus.Load().(control.NetmapStatus)
+func (s *networkState) controlNetmapStatus() (res control.NetmapStatus) {
+	v := s.controlNetStatus.Load()
+	if v != nil {
```
fyrchik, line 86: Can we instead store `Undefined` or `OFFLINE` during initialization?

fyrchik:
#1329 also involves some logic during new epoch processing -- node can appear in some container because of netmap changes.

acid-ant (approved):

cmd/neofs-node/netmap.go
carpawell: yes, that looks better, changed

cmd/neofs-node/tree.go
carpawell: good question

moved out `c.addOnlineStateHandler(syncTreeFunc)` from the `else` case: online handler will be refistered anyway

cmd/neofs-node/tree.go
carpawell: added #1902

fyrchik (approved):

PR #1862, opened 2022-10-10
Add healthcheck method to the Tree service

### LINKED ISSUE #1768: Tree service discovery
Tree service is an optional service. Not sure it will stay optional, but now it is. Clients want to discover if tree service is enabled on remote node for early failures (see https://github.com/nspcc-dev/neofs-s3-gw/issues/699). Clients can't rely on `Dial` invocations, because tree service lives on the same socket as NeoFS API. 

I see such options how tree service can be discovered
1) Provide dummy RPC to check if server returns error
2) Use existing RPC as dummy to check if server returns error
3) Enable [server reflection](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md).

I think (3) is an overhead and more suitable for debug purposes rather than using it in production.
(1) looks better than (2) for me, but requires protocol update.

acid-ant:
Closes #1768 

### DIFF
--- a/pkg/services/tree/service.proto
+++ b/pkg/services/tree/service.proto
@@ -47,6 +47,8 @@ service TreeService {
   rpc Apply (ApplyRequest) returns (ApplyResponse);
   // GetOpLog returns a stream of logged operations starting from some height.
   rpc GetOpLog(GetOpLogRequest) returns (stream GetOpLogResponse);
+  // Healthcheck is a dummy rpc to check service availability
+  rpc Healthcheck(HealthcheckRequest) returns (HealthcheckResponse);
 }
 
 message AddRequest {
@@ -328,3 +330,23 @@ message GetOpLogResponse {
   // Response signature.
   Signature signature = 2;
 };
+
+message HealthcheckResponse {
+  message Body {
+  }
+
+  // Response body.
+  Body body = 1;
+  // Response signature.
+  Signature signature = 2;
+};
+
+message HealthcheckRequest {
+  message Body {
+  }
+
+  // Request body.
+  Body body = 1;
+  // Request signature.
+  Signature signature = 2;
+}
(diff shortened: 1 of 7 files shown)


pkg/services/tree/service.proto
```diff
@@ -47,6 +48,8 @@ service TreeService {
   rpc Apply (ApplyRequest) returns (ApplyResponse);
   // GetOpLog returns a stream of logged operations starting from some height.
   rpc GetOpLog(GetOpLogRequest) returns (stream GetOpLogResponse);
+  // Healthcheck is a dummy rpc to check is service available
+  rpc Healthcheck(google.protobuf.Empty) returns (google.protobuf.Empty);
```
carpawell, line 52: we usually [add](https://github.com/nspcc-dev/neofs-node/blob/2d43892fc9bfc0067689b5ef1370873c9bb5a7b1/pkg/services/tree/service.proto#L140) empty request structs explicitly for clear extensibility

/cc @fyrchik, @cthulhu-rider

pkg/services/tree/service.proto
```diff
@@ -47,6 +48,8 @@ service TreeService {
   rpc Apply (ApplyRequest) returns (ApplyResponse);
   // GetOpLog returns a stream of logged operations starting from some height.
   rpc GetOpLog(GetOpLogRequest) returns (stream GetOpLogResponse);
+  // Healthcheck is a dummy rpc to check is service available
```
carpawell, line 51: please, double check that the comment is correct, seems like it should be `check if ... is ...` construct

acid-ant: Thanks, rephrased.

pkg/services/tree/service.proto
acid-ant: Added custom response/request.

cthulhu-rider (dismissed):
Pls pay attention to missing linebreak in `.proto` file.

carpawell (approved):
Except for the missing empty line.

carpawell (approved):

fyrchik (approved):

alexvanin (approved):

PR #1872, opened 2022-10-10
Allow to use multiple shard IDs in control service

### LINKED ISSUE #1867: Allow to specify a list of shard IDs in a control service
`evacuate` can benefit from this because we won't do unnecessary work (first migrate from 1 to 2, then from 2 to network)
Because shards can share an SSD, `set-mode` can be easier with multiple ID.

We can also have a denoted value "all" to mean "list of all shards".

fyrchik:
Close #1867 .


cmd/neofs-cli/modules/control/shards_set_mode.go
```diff
@@ -36,7 +37,8 @@ func initControlSetShardModeCmd() {
 	flags := setShardModeCmd.Flags()
 
 	flags.String(controlRPC, controlRPCDefault, controlRPCUsage)
-	flags.String(shardIDFlag, "", "ID of the shard in base58 encoding")
+	flags.StringSlice(shardIDFlag, nil, "List of shard IDs in base58 encoding")
```
carpawell, line 40: is that a compatible change? (i think yes, just to be sure)

cmd/neofs-cli/modules/control/shards_set_mode.go
```diff
@@ -99,3 +103,25 @@ func getShardID(cmd *cobra.Command) []byte {
...
+	if len(sidList) == 0 {
+		common.ExitOnErr(cmd, "", fmt.Errorf("either --%s or --%s flag must be provided", shardIDFlag, shardAllFlag))
+	}
+
+	res := make([][]byte, 0, len(sidList))
+	for i := range sidList {
+		raw, err := base58.Decode(sidList[i])
+		common.ExitOnErr(cmd, "incorrect shard ID encoding: %w", err)
+
+		res = append(res, raw)
```
carpawell, line 123: can we check their uniqueness here? and mb in the control/shard services

pkg/services/control/server/helpers.go
```diff
@@ -0,0 +1,20 @@
+package control
+
+import "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/shard"
+
+func (s *Server) getShardIDList(raw [][]byte) []*shard.ID {
```
carpawell, line 5: if there is some bug in the `neofs-cli` or the req has been sent by another tool and we get an empty list of the shards to evacuate, do we... evacuate all of them?

cmd/neofs-cli/modules/control/shards_set_mode.go
fyrchik: Yes, old commands will continue to work

pkg/services/control/server/helpers.go
fyrchik: Yes, but it should be simple -- we check the length

pkg/local_object_storage/engine/evacuate.go
```diff
@@ -94,72 +99,90 @@ func (e *StorageEngine) Evacuate(prm EvacuateShardPrm) (EvacuateShardRes, error)
...
-							zap.String("from", sid),
-							zap.Stringer("to", shards[j].ID()),
-							zap.Stringer("addr", lst[i]))
-
-						res.count++
+
+				hrw.SortSliceByWeightValue(shards, weights, hrw.Hash([]byte(lst[i].EncodeToString())))
+				for j := range shards {
+					if _, ok := shardMap[shards[j].ID().String()]; ok {
+						fmt.Println("skip", shards[j].ID().String())
```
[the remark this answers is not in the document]
fyrchik: fixed

cmd/neofs-cli/modules/control/shards_set_mode.go
fyrchik: fixed

pkg/services/control/server/helpers.go
carpawell: it scares me, i dunno why

but ok

carpawell (dismissed):

carpawell (approved):

carpawell (approved):

Issue #1332, opened 2022-04-25
Support interaction with tree service from `neofs-cli`


fyrchik:
- [x] `Add`
- [x] `AddByPath`
- [ ] `Remove`
- [ ] `Move`
- [x] `GetNodeByPath`
- [ ] `GetSubTree`
- [ ] `Healthcheck`
- [ ] `GetOpLog`
- [x] `ListTrees`

carpawell:
All the Tree Service RPC should have corresponding CLI commands. `Apply` is an exception.

acid-ant:
How about common command for the whole list of services on node? Like this:
```
@:~/workspace/neofs-node$ neofs-cli --rpc-endpoint 192.168.130.71:8080  node services
tree service status: N/A
@:~/workspace/neofs-node$ neofs-cli --rpc-endpoint 192.168.130.71:8080  node services
tree service status: available
@:~/workspace/neofs-node$

```

vkarak1:
There are several suggestions / comments which could help to make life better to work with tree service:

1. It would be great to see details in neofs-storage service log regarding tree syncing, for now I was able to see only the following types of messages:
```
debug        tree/sync.go:280        syncing container trees...        {"cid": "5S1mE1yrhm7Ww3JEiWSazTwJbDCvDwyqYB6bSpmnmy9N
debug        tree/sync.go:291        container trees have been synced        {"cid": "5S1mE1yrhm7Ww3JEiWSazTwJbDCvDwyqYB6bSpmnmy9N"}
```
Adding details for example which tree / path and etc  has been synchronized will be great for testing purposes.
 2. I has been working on testing tree service where I turned on node which contained S3 container and I expected to see something useful in neofs-node logs (something like 'container_id / tree name / is updating / from which node' and etc), but there is no any mention about synchonization  of tree service of this S3 container.
 3. Since tree service is available only for S3, it would be great to allow `neofs-cli tree ` working with bucket name. It would be great to specify --cid OR --bm (bucket name) - where one of them is required.

roman-khimov:
Trees will be removed, this issue will never be done.

PR #1874, opened 2022-10-10
Feat/Add tree cmds to CLI

carpawell:
Related to #1332.

### DIFF
--- a/cmd/neofs-cli/modules/tree/add.go
+++ b/cmd/neofs-cli/modules/tree/add.go
@@ -1,0 +1,95 @@
+package tree
+
+import (
+	"crypto/sha256"
+	"fmt"
+	"strings"
+
+	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/common"
+	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/commonflags"
+	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/key"
+	"github.com/nspcc-dev/neofs-node/pkg/services/tree"
+	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
+	"github.com/spf13/cobra"
+)
+
+var addCmd = &cobra.Command{
+	Use:   "add",
+	Short: "Add a node to the tree service",
+	Run:   add,
+	PersistentPreRun: func(cmd *cobra.Command, _ []string) {
+		commonflags.Bind(cmd)
+	},
+}
+
+func initAddCmd() {
+	commonflags.Init(addCmd)
+	initCTID(addCmd)
+
+	ff := addCmd.Flags()
+	ff.StringSlice(metaFlagKey, nil, "Meta pairs in the form of Key1=[0x]Value1,Key2=[0x]Value2")
+	ff.Uint64(parentIDFlagKey, 0, "Parent node ID")
+
+	_ = cobra.MarkFlagRequired(ff, commonflags.RPC)
+}
+
+func add(cmd *cobra.Command, _ []string) {
+	pk := key.GetOrGenerate(cmd)
+
+	var cnr cid.ID
+	err := cnr.DecodeString(cmd.Flag(containerIDFlagKey).Value.String())
+	common.ExitOnErr(cmd, "decode container ID string: %w", err)
+
+	tid, _ := cmd.Flags().GetString(treeIDFlagKey)
+	pid, _ := cmd.Flags().GetUint64(parentIDFlagKey)
+
+	meta, err := parseMeta(cmd)
+	common.ExitOnErr(cmd, "meta data parsing: %w", err)
+
+	ctx := cmd.Context()
+
+	cli, err := _client(ctx)
+	common.ExitOnErr(cmd, "client: %w", err)
+
+	rawCID := make([]byte, sha256.Size)
+	cnr.Encode(rawCID)
+
+	req := new(tree.AddRequest)
+	req.Body = &tree.AddRequest_Body{
+		ContainerId: rawCID,
+		TreeId:      tid,
+		ParentId:    pid,
+		Meta:        meta,
+		BearerToken: nil, // TODO: #1891 add token handling
+	}
+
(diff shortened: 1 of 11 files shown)


pkg/services/tree/signature.go
```diff
@@ -143,7 +143,7 @@ func verifyMessage(m message) error {
 	return nil
 }
 
-func signMessage(m message, key *ecdsa.PrivateKey) error {
+func SignMessage(m message, key *ecdsa.PrivateKey) error {
```
fyrchik, line 146: It needs a comment now.

cmd/neofs-cli/modules/tree/add.go
```diff
@@ -0,0 +1,119 @@
...
+	pk := key.GetOrGenerate(cmd)
+
+	var cnr cid.ID
+	err := cnr.DecodeString(cmd.Flag(cidFlagKey).Value.String())
+	common.ExitOnErr(cmd, "decode container ID string: %w", err)
+
+	tid := cmd.Flag(tidFlagKey).Value.String()
+	if tid == "" {
+		common.ExitOnErr(cmd, "", errors.New("empty tree ID"))
+	}
```
fyrchik, line 48: Can we just make the flag required? I believe tree service supports empty names now.

cmd/neofs-cli/modules/tree/add.go
```diff
@@ -0,0 +1,119 @@
...
+	rawCID := make([]byte, sha256.Size)
+	cnr.Encode(rawCID)
+
+	req := new(tree.AddRequest)
+	req.Body = &tree.AddRequest_Body{
+		ContainerId: rawCID,
+		TreeId:      tid,
+		ParentId:    pid,
+		Meta:        meta,
+		BearerToken: nil, // TODO: add token handling
```
fyrchik, line 69: Issue?

cmd/neofs-cli/modules/tree/add.go
```diff
@@ -0,0 +1,119 @@
...
+
+		pairs = append(pairs, &pair)
+	}
+
+	return pairs, nil
+}
+
+func bytesFromString(s string) ([]byte, error) {
+	switch {
+	case strings.HasPrefix(s, "0x"):
```
fyrchik, line 110: Do we need this in CLI? Especially when we don't have similar logic in get operations now.

cmd/neofs-cli/modules/tree/root.go
```diff
@@ -21,6 +23,11 @@ const (
 	pidFlagKey = "pid"
 
 	metaFlagKey = "meta"
+
+	pathFlagKey          = "path"
+	pathAttributeFlagKey = "pattr"
+
+	latestOnlyFlagKey = "latest-only"
```
fyrchik, line 32: To me `--latest` looks more convenient.

cmd/neofs-cli/modules/tree/get_by_path.go
```diff
@@ -0,0 +1,102 @@
...
+
+	common.ExitOnErr(cmd, "message signing: %w", tree.SignMessage(req, pk))
+
+	resp, err := cli.GetNodeByPath(ctx, req)
+	common.ExitOnErr(cmd, "rpc call: %w", err)
+
+	nn := resp.GetBody().GetNodes()
+	if len(nn) == 0 {
+		common.ExitOnErr(cmd, "", errors.New("node not found"))
+	}
```
fyrchik, line 89: I don't think it is an error, it is a valid response.

carpawell: do you mean exit with `0` not with the `1`?

cmd/neofs-cli/modules/tree/add.go
carpawell: > Do we need this in CLI

i do not have info about how values in the Tree's KV gonna (planned to) be used, so could not answer that question at all

> Especially when we don't have similar logic in get operations now.

i am ready to add that but dont know how that is expected to work (how to distinguish string bytes from the raw bytes)

in a word -- dropped

cmd/neofs-cli/modules/tree/add.go
carpawell: added #1891

cmd/neofs-cli/modules/tree/add.go
carpawell: > I believe tree service supports empty names now

ok, but what for? dropped that limitation

pkg/services/tree/signature.go
carpawell: thought about that but racked my brains what words should be here

added, check, please

cmd/neofs-cli/modules/tree/get_by_path.go
fyrchik: Yes, an we can print nothing in this case. Maybe print `node not found` in verbose mode?

cmd/neofs-cli/modules/tree/root.go
```diff
@@ -0,0 +1,43 @@
...
+
+	initAddCmd()
+	initGetByPathCmd()
+	initAddByPathCmd()
+}
+
+const (
+	cidFlagKey = "cid"
+	tidFlagKey = "tid"
+	pidFlagKey = "pid"
```
fyrchik, line 25: Not immediately obvious, maybe `parentIDFlagKey`?

fyrchik (approved):

cmd/neofs-cli/modules/tree/add_by_path.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/key"
+	"github.com/nspcc-dev/neofs-node/pkg/services/tree"
+	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
+	"github.com/nspcc-dev/neofs-sdk-go/object"
+	"github.com/spf13/cobra"
+)
+
+var addByPathCmd = &cobra.Command{
+	Use:   "add-by-path",
+	Short: "Get a node by its path",
```
acid-ant, line 18: Get --> Add

acid-ant (approved):

PR #1880, opened 2022-10-11
Use hex format to print storage node ID

### LINKED ISSUE #1765: Synchronize printing string format of storage node ID
Currently we print storage node ID (which is actually its public key) in different formats. For example:
https://github.com/nspcc-dev/neofs-node/blob/cd71de69a08d87a63fedd7bc4ead54c4ddac8791/pkg/innerring/processors/netmap/process_peers.go#L60-L66
https://github.com/nspcc-dev/neofs-node/blob/cd71de69a08d87a63fedd7bc4ead54c4ddac8791/cmd/neofs-cli/modules/control/snapshot.go#L64-L65

While both formats can be casted to each other, it's not really convenient to compare the values visually. I propose to work out for this case and others like it a string format at the convention level. Why conventions? For example, for object/container IDs, a string format is declared at the protocol level.

To conveniently follow the convention, it is proposed to create a special function in the SDK, which will be used whenever it is required to comply with the generally accepted format.

Btw which format do you prefer? For me its hex :sheep:. @fyrchik @carpawell

acid-ant:
Closes #1765

### DIFF
--- a/cmd/neofs-node/notificator.go
+++ b/cmd/neofs-node/notificator.go
@@ -1,9 +1,9 @@
 package main
 
 import (
+	"encoding/hex"
 	"fmt"
 
-	"github.com/mr-tron/base58"
 	nodeconfig "github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/node"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/engine"
 	"github.com/nspcc-dev/neofs-node/pkg/morph/event"
@@ -110,7 +110,7 @@ func (n notificationWriter) Notify(topic string, address oid.Address) {
 func initNotifications(c *cfg) {
 	if nodeconfig.Notification(c.appCfg).Enabled() {
 		topic := nodeconfig.Notification(c.appCfg).DefaultTopic()
-		pubKey := base58.Encode(c.cfgNodeInfo.localInfo.PublicKey())
+		pubKey := hex.EncodeToString(c.cfgNodeInfo.localInfo.PublicKey())
 
 		if topic == "" {
 			topic = pubKey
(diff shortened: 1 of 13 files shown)


cmd/neofs-node/notificator.go
```diff
@@ -109,7 +109,7 @@ func (n notificationWriter) Notify(topic string, address oid.Address) {
 func initNotifications(c *cfg) {
 	if nodeconfig.Notification(c.appCfg).Enabled() {
 		topic := nodeconfig.Notification(c.appCfg).DefaultTopic()
-		pubKey := base58.Encode(c.cfgNodeInfo.localInfo.PublicKey())
+		pubKey := hex.EncodeToString(c.cfgNodeInfo.localInfo.PublicKey())
```
cthulhu-rider, line 112: Lets maybe unify the format by creating some function like
```
func StringifyNodeKey(netmap.NodeInfo) string {}
```
We can either place it to:
1. https://github.com/nspcc-dev/neofs-node/tree/master/pkg/core/netmap
2. https://github.com/nspcc-dev/neofs-sdk-go/tree/master/netmap

I also suggest to create key encoding function
```
func StringifyKeyBinary([]byte) string
```
and use it everywhere. It will allow to easily change string format if needed.

carpawell (approved):

cmd/neofs-node/notificator.go
acid-ant: Added usage of netmap.StringifyPublicKey().

cthulhu-rider: May be smth went wrong, I don't see change here

cthulhu-rider (approved):

PR #1879, opened 2022-10-11
services/object: Fix child check in GET

cthulhu-rider:
* closes #1878

PR #1884, opened 2022-10-12
services/object: fallback to GET in GET_RANGE

fyrchik:
Current spec allows denying GET_RANGE requests from other storage nodes. However, GET should always be allowed and it is enough to perform GET_RANGE locally


fyrchik:
Depends on #1879 .

pkg/services/object/get/util.go
```diff
@@ -143,12 +145,35 @@ func (c *clientWrapper) getObject(exec *execCtx, info coreclient.NodeInfo) (*obj
...
+				obj, err := c.get(exec, key)
+				if err != nil {
+					return nil, err
+				}
+
+				payload := obj.Payload()
+				from := rng.GetOffset()
+				to := from + rng.GetLength()
+
+				if pLen := uint64(len(payload)); to < from || pLen < from || pLen < to {
```
cthulhu-rider, line 161: Do we need to check `pLen < from`?

cthulhu-rider (approved):

PR #1897, opened 2022-10-14
Improve shard modes' docs

cthulhu-rider:


cmd/neofs-cli/modules/control/shards_set_mode.go
```diff
@@ -39,12 +44,14 @@ func initControlSetShardModeCmd() {
...
+	modes := make([]string, 0, len(mShardModes))
+	for strMode := range mShardModes {
+		modes = append(modes, "'"+strMode+"'")
+	}
+
 	flags.String(shardModeFlag, "",
-		fmt.Sprintf("New shard mode keyword ('%s', '%s', '%s')",
-			shardModeReadWrite,
-			shardModeReadOnly,
-			shardModeDegraded,
```
fyrchik, line 46: BTW what about removing `degraded` from help? It is not really intended to be used (`degraded-read-only` should) as it has it's has it's own problems. So we still support it but not suggest it to a user.

cthulhu-rider: Done

cmd/neofs-cli/modules/control/shards_set_mode.go
```diff
@@ -36,12 +83,15 @@ func initControlSetShardModeCmd() {
 	flags := setShardModeCmd.Flags()
 	flags.StringSlice(shardIDFlag, nil, "List of shard IDs in base58 encoding")
 	flags.Bool(shardAllFlag, false, "Process all shards")
+
+	modes := make([]string, 0)
+
+	iterateSafeShardModes(func(strMode string) {
+		modes = append(modes, "'"+strMode+"'")
```
carpawell, line 90: use `fmt` pkg here?

carpawell (approved):

cmd/neofs-cli/modules/control/shards_set_mode.go
fyrchik: It is shorter like this

carpawell: yes but looks not that obvious (more "nonhuman" symbols like `"`, `+` IMO)

cthulhu-rider: Here we just wanna wrap string into apostrophies, I don't think formatting is worth it.

docs/storage-node-configuration.md
```diff
@@ -167,13 +167,14 @@ Contains configuration for each shard. Keys must be consecutive numbers starting
...
-| `writecache`      | [Writecache config](#writecache-subsection) |               | Write-cache configuration.               |
-| `metabase`        | [Metabase config](#metabase-subsection)     |               | Metabase configuration.                  |
-| `blobstor`        | [Blobstor config](#blobstor-subsection)     |               | Blobstor configuration.                  |
-| `gc`              | [GC config](#gc-subsection)                 |               | GC configuration.                        |
+| Parameter         | Type                                        | Default value | Description                                                                                               |
+|-------------------|---------------------------------------------|---------------|-----------------------------------------------------------------------------------------------------------|
+| `mode`            | `string enum`                               | `read-write`  | Shard Mode.<br/>Possible values:  `read-write`, `read-only`, `degraded`, `degraded-read-only`, `disabled` |
```
fyrchik, line 172: I think `string` is enough, "possible values" already tells us that this is an enum.

cmd/neofs-cli/modules/control/shards_set_mode.go
```diff
@@ -36,12 +83,15 @@ func initControlSetShardModeCmd() {
...
+	})
+
 	flags.String(shardModeFlag, "",
-		fmt.Sprintf("New shard mode keyword ('%s', '%s', '%s')",
-			shardModeReadWrite,
-			shardModeReadOnly,
-			shardModeDegraded,
-		),
+		fmt.Sprintf("New shard mode keyword (%s)", strings.Join(modes, ", ")),
```
fyrchik, line 94: While we are here, what about removing `keyword` from the message?

cthulhu-rider: Done.

docs/storage-node-configuration.md
cthulhu-rider: Fixed.

fyrchik (approved):

PR #1910, opened 2022-10-17
Add some linters

fyrchik:
Take almost all of them from neo-go and make the config similar.
There is #1911 left, it will be done in a separate task.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -53,6 +53,6 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.48.0
+          version: v1.50.0
           args: --timeout=5m
           only-new-issues: true
--- a/cmd/neofs-cli/internal/commonflags/flags.go
+++ b/cmd/neofs-cli/internal/commonflags/flags.go
@@ -44,11 +44,11 @@ const (
 )
 
 // Init adds common flags to the command:
-// - GenerateKey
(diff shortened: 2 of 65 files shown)


cthulhu-rider (approved):

.github/workflows/go.yml
```diff
@@ -53,6 +53,6 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.48.0
+          version: latest
```
carpawell, line 56: are you sure about that? it could lead to the "make PR, go sleep, wake up, PR is outdated because of linter release despite the same code"

cmd/neofs-cli/internal/commonflags/flags.go
```diff
@@ -49,6 +49,8 @@ const (
 // - Account
 // - RPC
 // - Timeout
+//
+//nolint:godot
```
carpawell, line 53: i think we can add `;` and `.` to that list. or the linter requires dots on every line? same for other cases

.github/workflows/go.yml
fyrchik: Fixed

cmd/neofs-cli/internal/commonflags/flags.go
fyrchik: No, it requires dots in the end only.
Fixed it here.

carpawell (approved):

Issue #1877, opened 2022-10-11
We need to make the change API process more clear for changing tests.


anikeev-yadro:
After API was changed we had output json changed too, as result our tests which use parsing output json had broken.
For example:
the old output was:
```
COMMAND: 'neofs-cli --config /tests/neofs_cli_configs/empty_passwd.yml object head --rpc-endpoint 'obj-nn-test-node1.spb.yadro.com:8081' --wallet '/tests/TemporaryDir/9e36c117-375a-4bb9-98d4-4069ed75bd85.json' --cid '5bx2Sf6QGH1cwPYygKXbnci54c2r4B7U57nyKqJKSn3F' --oid 'DBLVipfwZ2RidS8VHtEHuytTWPxWm7RCQCrmwfGoaNxt' --json'
OUTPUT:
...
"object":{
		"verb":"DELETE", 
		"address":{
			"containerID":{
					"value":"RGK2R6ck2MYpZwA0fDZjzkEhLNh+jQPy1ppvinOvM34="
					}, 
			"objectID":{"value":"L0Npxk176tmA5R8TG1KUgrKg3PDmyeH4CJ/wwp3oTfA="}
			}}}, 
...
```
the new output is:
```
COMMAND: 'neofs-cli --config /tests/wallet_config.yml object head --rpc-endpoint 'obj-nn-test-node2.spb.yadro.com:8081' --wallet '/tests/TemporaryDir/05e48067-b192-429f-973d-d336a449a5d6.json' --cid '43qM2vL1PEUgJ8JDaRqD9YjZW5D8q6QWxUgWbwy6hjib' --oid '3Gyij4fKgWMM5ut75FjP6bpoUWQmYQpVgE1UYhRxKXqU' --json'
OUTPUT:
...
		"object":{
			"verb":"DELETE", 
			"target":{"container":{"value":"LUzi/6K59Svl6oq02HEtf5o1pzVNHtn/4MJe9/rCQWQ="}, 
			"objects":[{"value":"q7OWeKaYaCr+q11OkmA0JiFRaSHCAGOF9K2OsehOeR0="}]}
		}}, 
...
```

cthulhu-rider:
From one side, we do not guarantee backward compatibility in JSON format. From the other side, I think we need to improve JSON UX in tools like NeoFS CLI.

Possible solutions:
0. do nothing and leave it to the user's attention to protocol changes
1. provide intermediate NeoFS CLI JSON format for some NeoFS entities which
[...]

PR #1923, opened 2022-10-18
cli/control: Pre-check maintenance mode in `set-status`

### LINKED ISSUE #1916: if MM is off globally, neofs-cli should not acept requst to change node's state to MM
## Expected Behavior
"neofs-cli set-status --status maintenance" returns: MM is off globally, cannot set status to MM

## Current Behavior
1) neofs-cli control set-status --status maintenance:
    Network status update request successfully sent
2) neofs-cli object put
    **rpc error:** client failure: rpc error: code = Unknown desc = could not create Put object streamer: could not create Put object 
    streamer: status: code = 1027 message = node is under maintenance

1. neofs-adm morph set-config MaintenanceModeAllowed=false -c config.yaml
3. neofs-adm  morph force-new-epoch -c config.yaml
4. `neofs-cli --endpoint localhost:8091 -w /etc/neofs/storage/wallet.json control set-status --status maintenance
Enter password >
Network status update request successfully sent.`
5. `neofs-cli -r node1:8080 -w /etc/neofs/storage/wallet.json object put --file /home/service/object.sample --no-progress --cid 4SVKDe8ckt7qznoU1bRdNJzUpmrpF9kJRbSDtcAZ7Jvs
Enter password >
rpc error: client failure: rpc error: code = Unknown desc = could not create Put object streamer: could not create Put object streamer: status: code = 1027 message = node is under maintenance`

## Your Environment
root@az:/home/service# neofs-node -version
NeoFS Storage node
Version: v0.32.0-125-gbcf3df35
GoVersion: go1.18.4

cthulhu-rider:
* closes #1916

and also
1. Maybe it's worth to provide `--force` flag to skip the check.
2. I would think about serving NeoFS API on Control endpoints too: in this PR I couldn't use a single endpoint to do all the ops.


fyrchik:
I think it is definitely worth to add `force`, because we may want to go into maintenance even if the network disallows us.

cmd/neofs-cli/modules/control/set_netmap_status.go
```diff
@@ -54,6 +57,32 @@ func setNetmapStatus(cmd *cobra.Command, _ []string) {
 		status = control.NetmapStatus_OFFLINE
 	case netmapStatusMaintenance:
 		status = control.NetmapStatus_MAINTENANCE
+
+		common.PrintVerbose("Reading network settings to check allowance of \"%s\" mode...", st)
```
fyrchik, line 61: Why did you decide to do it from the CLI and not from the storage node?
It is a bit strange to provide 2 endpoints to this command.

fyrchik (approved):

cmd/neofs-cli/modules/control/set_netmap_status.go
cthulhu-rider: Because it's good to firstly implement anything on client side, and later consider server-side support. Doing validation on the node side sounds quite good. I guess we don't need `force` flag in this case right?

fyrchik: I agree, but in this case we complicate user experience by requiring to provide 2 endpoints.
At the same time `force` flag in control service can allow to move node in "local" maintenance mode even when it is disabled globally with a proper message, and do nothing in other case.

pkg/services/control/service.proto
```diff
@@ -75,7 +75,16 @@ message SetNetmapStatusRequest {
...
+        // If status is MAINTENANCE, the node checks whether maintenance is
+        // allowed in the network settings. In case of prohibition, the request
+        // is denied. Otherwise, node switches to local maintenance state. To
+        // force local maintenance, use `force_maintenance` flag.
         NetmapStatus status = 1;
+
+        // MAINTENANCE status validation skip flag. If set, node starts local
+        // maintenance regardless of network settings. The flag MUST NOT be
+        // set for any other status.
+        bool force_maintenance = 2;
```
fyrchik, line 87: What is the reason for restricting it like this? IMO `force` is nice by itself and can easily be extended in future.

cmd/neofs-node/netmap.go
```diff
@@ -361,6 +360,33 @@ func (c *cfg) SetNetmapStatus(st control.NetmapStatus) error {
 	return c.updateNetMapState(func(*nmClient.UpdatePeerPrm) {})
 }
 
+func (c *cfg) ForceMaintenance() error {
+	return c.setMaintenanceStatus(true)
+}
+
+func (c *cfg) setMaintenanceStatus(force bool) error {
+	netSettings, err := c.cfgNetmap.wrapper.ReadNetworkConfiguration()
```
fyrchik, line 368: Why have you decided to check network settings here and not in the control service?

cthulhu-rider: It is a Control service, I just distinguish handling of the request payload from message processing (signatures and so on).

pkg/services/control/service.proto
fyrchik: As we already noticed, sometimes it is tricky https://github.com/nspcc-dev/neofs-node/issues/1877 . Changing description is easier, but I don't insist.

cthulhu-rider: > Changing description is easier

I agree, but if we don't restrict using force flag with other statuses, some client can set it just because he can. If we later specify the forcing behavior of this status in docs, it can break the client behavior.

cmd/neofs-node/netmap.go
carpawell: > It is a Control service

what do you mean?

cthulhu-rider: The method implements service needs.

fyrchik (approved):

PR #1924, opened 2022-10-18
Add support to store/restore/delete binary objects

### LINKED ISSUE #1338: Support storing binary objects
Currently there is no way to store the binary object using NeoFS CLI. This feature can be useful for users which want to store object in container again or restore it after accidental loss.

acid-ant:
Closes #1338

With flag `--binary` user can point to the file with serialized object structure - id + signature + header + payload
Added for development purposes.

### DIFF
--- a/cmd/neofs-cli/modules/object/put.go
+++ b/cmd/neofs-cli/modules/object/put.go
@@ -1,7 +1,9 @@
 package object
 
 import (
+	"bytes"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -40,12 +42,11 @@ func initObjectPutCmd() {
 
 	flags := objectPutCmd.Flags()
 
-	flags.String("file", "", "File with object payload")
-	_ = objectPutCmd.MarkFlagFilename("file")
-	_ = objectPutCmd.MarkFlagRequired("file")
+	flags.String(fileFlag, "", "File with object payload")
+	_ = objectPutCmd.MarkFlagFilename(fileFlag)
+	_ = objectPutCmd.MarkFlagRequired(fileFlag)
 
-	flags.String("cid", "", "Container ID")
-	_ = objectPutCmd.MarkFlagRequired("cid")
+	flags.String(commonflags.CIDFlag, "", commonflags.CIDFlagUsage)
 
 	flags.String("attributes", "", "User attributes in form of Key1=Value1,Key2=Value2")
 	flags.Bool("disable-filename", false, "Do not set well-known filename attribute")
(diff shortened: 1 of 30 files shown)


cthulhu-rider:
Do we declare merging behavior of the command? I mean if we provide filepath to serialized object, and specifiy, for example, attributes, they are set in the resulting object.

If so, IMO this complicates the command. I would prefer to have indiviual commands for object formation and storing.

@fyrchik

cmd/neofs-cli/modules/object/put.go
```diff
@@ -117,19 +138,26 @@ func putObject(cmd *cobra.Command, _ []string) {
...
+		prm.SetPayloadReader(payloadReader)
 	} else {
-		fi, err := f.Stat()
-		if err != nil {
-			cmd.PrintErrf("Failed to get file size, progress bar is disabled: %v\n", err)
-			prm.SetPayloadReader(f)
-		} else {
-			p = pb.New64(fi.Size())
+		if binary {
+			p = pb.New64(int64(len(obj.Payload())))
```
carpawell, line 144:
```suggestion
			p = pb.New(len(obj.Payload()))
```

cmd/neofs-cli/modules/object/put.go
```diff
@@ -55,22 +56,43 @@ func initObjectPutCmd() {
...
 	pk := key.GetOrGenerate(cmd)
 
 	var ownerID user.ID
-	user.IDFromKey(&ownerID, pk.PublicKey)
-
 	var cnr cid.ID
-	readCID(cmd, &cnr)
 
-	filename := cmd.Flag("file").Value.String()
+	var payloadReader io.Reader
```
carpawell, line 74: can we unite the declaration with the definition of the `payloadReader` var?

cmd/neofs-cli/modules/object/util.go
```diff
@@ -23,6 +25,14 @@ const (
 
 	rawFlag     = "raw"
 	rawFlagDesc = "Set raw request option"
+	fileFlag    = "file"
+	binaryFlag  = "binary"
+
+	cidFlag      = "cid"
```
carpawell, line 31: maybe share them with `common` pkg? (or via another PR)

cmd/neofs-cli/modules/object/put.go
acid-ant: Refactored

cmd/neofs-cli/modules/object/util.go
acid-ant: Moved into commonflags.

carpawell (approved):

fyrchik (approved):

fyrchik:
Could you add `neofs-cli: ` prefix to a commit message?

PR #1928, opened 2022-10-18
Feat/Add `TreeService.GetTrees` RPC

### LINKED ISSUE #1902: TreeService.ListTrees method
Could (and should) be used in tree service synchronizations: all the methods require `treeID` but no treeID-getters are implemented in the tree API.

carpawell:
Closes #1902. Relates to #1332.


pkg/services/tree/service.go
```diff
@@ -522,6 +522,51 @@ func (s *Service) GetOpLog(req *GetOpLogRequest, srv TreeService_GetOpLogServer)
...
+		var resp *GetTreesResponse
+		var outErr error
+		err = s.forEachNode(ctx, ns, func(c TreeServiceClient) bool {
+			resp, outErr = c.GetTrees(ctx, req)
+			return outErr == nil
+		})
+		if err != nil {
+			return nil, err
+		}
+		return resp, outErr
```
cthulhu-rider, line 555: Does `forEachNode` guarantee that if `err == nil` then `resp != nil`?

cmd/neofs-cli/modules/tree/get_trees.go
```diff
@@ -0,0 +1,64 @@
...
+		},
+	}
+
+	common.ExitOnErr(cmd, "message signing: %w", tree.SignMessage(req, pk))
+
+	resp, err := cli.GetTrees(ctx, req)
+	common.ExitOnErr(cmd, "rpc call: %w", err)
+
+	for _, treeID := range resp.GetBody().GetIds() {
+		fmt.Println(treeID)
```
cthulhu-rider, line 62: Maybe print on `cmd`?

pkg/services/tree/sync.go
```diff
@@ -19,7 +19,7 @@ import (
 var ErrNotInContainer = errors.New("node is not in container")
 
 // Synchronize tries to synchronize log starting from the last stored height.
-func (s *Service) Synchronize(ctx context.Context, cid cid.ID, treeID string) error {
+func (s *Service) Synchronize(ctx context.Context, cid cid.ID, treeID *string) error {
```
cthulhu-rider, line 22: Upd docs?

pkg/services/tree/sync.go
```diff
@@ -29,43 +29,89 @@ func (s *Service) Synchronize(ctx context.Context, cid cid.ID, treeID string) er
...
+			outErr = err
+		}
+
+		if outErr != nil {
+			return fmt.Errorf("could not fetch tree ID list: %w", outErr)
+		}
+	}
+
+	for _, id := range treesToSync {
+		var d pilorama.CIDDescriptor
```
cthulhu-rider, line 76: Init ouside loop?

pkg/local_object_storage/pilorama/boltdb.go
```diff
@@ -557,6 +557,31 @@ func (t *boltForest) TreeGetChildren(cid cidSDK.ID, treeID string, nodeID Node)
 	return children, err
 }
 
+// TreeGetTrees implements the Forest interface.
+func (t *boltForest) TreeGetTrees(cid cidSDK.ID) ([]string, error) {
```
fyrchik, line 561: `TreeListAll`?

pkg/local_object_storage/pilorama/interface.go
```diff
@@ -37,6 +37,9 @@ type Forest interface {
 	// TreeDrop drops a tree from the database.
 	// If the tree is not found, ErrTreeNotFound should be returned.
 	TreeDrop(cid cidSDK.ID, treeID string) error
+	// TreeGetTrees returns all the tree IDs that have been added to the
+	// passed container ID. Nil slice should be returned if no tree found.
+	TreeGetTrees(cid cidSDK.ID) ([]string, error)
```
fyrchik, line 42: Can we build `neofs-node` on this commit?

pkg/services/tree/service.go
```diff
@@ -522,6 +522,51 @@ func (s *Service) GetOpLog(req *GetOpLogRequest, srv TreeService_GetOpLogServer)
 	}
 }
 
+func (s *Service) GetTrees(ctx context.Context, req *GetTreesRequest) (*GetTreesResponse, error) {
```
fyrchik, line 525: I don't like `Get` here, we have `GetOpLog` (which _is_ a "gettree" operation) and `GetAddByPath`. What about `List`?

cmd/neofs-node/tree.go
```diff
@@ -81,7 +81,7 @@ func syncTrees(ctx context.Context, treeSvc *tree.Service, cnrCli *containerClie
 
 	for _, id := range ids {
 		for _, tID := range wellKnownTrees {
-			err = treeSvc.Synchronize(ctx, id, tID)
+			err = treeSvc.Synchronize(ctx, id, &tID)
```
fyrchik: Taking a pointer to a loop variable can bite us unexpectedly if we ever go async inside of a `Synchronize`. What about `&wellKnownTrees[i]`?

pkg/services/tree/sync.go
fyrchik, line 22: Let's not complicate this function further. It has a clear purpose -- synchronize a single tree, the caller can list the trees and loop, it is not hard. If there are more callers, I would prefer a separate method/function.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ Changelog for NeoFS Node
 
 ### Added
 - `--timeout` flag in `neofs-cli control` commands (#1917)
+- `TreeService.GetTrees` RPC; all trees synchronization on bootstrap (#1902)
```
fyrchik, line 8: Why not 2 different lines? It is ok to refer to the same issue.

pkg/services/tree/service.go
carpawell: no, why do you ask? `forEachNode` knows nothing about `resp`

cmd/neofs-cli/modules/tree/get_trees.go
carpawell: fixed

pkg/services/tree/sync.go
carpawell: was changed according to another comment

pkg/services/tree/sync.go
carpawell: fixed

pkg/local_object_storage/pilorama/boltdb.go
carpawell: renamed: `ListTrees` in RPC and `TreeList` in the `Forest` intreface

pkg/local_object_storage/pilorama/interface.go
carpawell: i was surprised how "useful" `Forest` intreface is (defined in tree storage pkg, implemented there too, used for tests and is forced to be implemented by the `engine` (sub)pkgs). so the answer is "no" cause i did not want to make all changes in a one huge commit and im always think that small atomic commits > being possible to perform correct building on every commit. but ok, ive added dummy implementation in the engine here, and now every commit could be build

pkg/services/tree/service.go
carpawell: renamed: `ListTrees` in RPC and `TreeList` in the `Forest` intreface

cmd/neofs-node/tree.go
carpawell: good point. changed (it was a temporary change though)

pkg/services/tree/sync.go
carpawell: fixed. but it has become less readable IMO

CHANGELOG.md
carpawell: ok, separated them

pkg/services/tree/service.go
cthulhu-rider: I mean this exact `forEachNode`.

carpawell: Do we need that guarantee? That is how tree requests are implemented, i just copied the code. The behavior is the same as in the generated code.

cthulhu-rider: If we return (nil, nil), panic occurs

carpawell: > If we return (nil, nil)

then generated code has returned `(nil, nil)` to us

pkg/local_object_storage/pilorama/interface.go
fyrchik: One could argue that if we cannot build after a commit, is it really _atomic_? Building on any commit is a must-have, for example, because `git bisect` can accidentally checkout on it..

pkg/services/tree/sync.go
```diff
@@ -34,6 +39,85 @@ func (s *Service) Synchronize(ctx context.Context, cid cid.ID, treeID string) er
...
+
+	nodes = append(nodes[:pos], nodes[pos+1:]...) // exclude that node
+	if len(nodes) == 0 {
+		return errNoOtherNodes
+	}
+
+	return s.synchronizeTree(ctx, cid, treeID, nodes, d)
+}
+
+func (s *Service) synchronizeTree(ctx context.Context, cid cid.ID,
```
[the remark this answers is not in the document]
carpawell: cant argue with that

PR #1929, opened 2022-10-18
Feat/Sync trees when a node first time appears in a container

### LINKED ISSUE #1329: Synchronize tree log storage when a node first appears in container

carpawell:
Closes #1329. Waits for #1928.

### DIFF
--- a/pkg/services/tree/service.go
+++ b/pkg/services/tree/service.go
@@ -24,6 +24,9 @@ type Service struct {
 	replicationTasks chan replicationTask
 	closeCh          chan struct{}
 	containerCache   containerCache
+
+	syncChan chan struct{}
+	cnrMap   map[cidSDK.ID]struct{}
 }
 
 var _ TreeServiceServer = (*Service)(nil)
@@ -48,6 +51,8 @@ func New(opts ...Option) *Service {
 	s.replicateCh = make(chan movePair, s.replicatorChannelCapacity)
 	s.replicationTasks = make(chan replicationTask, s.replicatorWorkerCount)
 	s.containerCache.init(s.containerCacheSize)
+	s.cnrMap = make(map[cidSDK.ID]struct{})
+	s.syncChan = make(chan struct{})
 
 	return &s
 }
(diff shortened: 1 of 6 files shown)


pkg/services/tree/sync.go
```diff
@@ -200,3 +200,94 @@ func (s *Service) synchronizeSingle(ctx context.Context, d pilorama.CIDDescripto
 		height = newHeight
 	}
 }
+
+// SynchronizeAll forces tree service to synchronize all the trees according to
+// netmap information. Must not be called before Service.Start.
+// Note: blocking operation if the service is already synchronizing.
+func (s *Service) SynchronizeAll() {
+	s.syncChan <- struct{}{}
```
fyrchik, line 208: What about returning error if this cannot be done immediately? In case the synchronization takes too long we will only make things worse like this.

pkg/services/tree/service.go
```diff
@@ -48,18 +51,25 @@ func New(opts ...Option) *Service {
...
+	go s.syncLoop(ctx)
+
+	// initial sync
+	s.syncChan <- struct{}{}
 }
 
 // Shutdown shutdowns the service.
 func (s *Service) Shutdown() {
 	close(s.closeCh)
+	close(s.syncChan)
```
fyrchik, line 72: Can we (1) drain it here first and then check `<-s.closeCh` in syncAll before blocking. This way we avoid possible panic.

pkg/services/tree/service.go
```diff
@@ -24,6 +24,9 @@ type Service struct {
 	replicationTasks chan replicationTask
 	closeCh          chan struct{}
 	containerCache   containerCache
+
+	syncChan chan struct{}
+	cnrMap   map[cidSDK.ID]struct{}
```
fyrchik, line 29: What exactly does it contain? Do we need it as a map and not as a slice?

pkg/services/tree/sync.go
carpawell: good, added

pkg/services/tree/service.go
carpawell: > What exactly does it contain?

containers that the service is tracking, all the containers that are stored are considered to be in sync

all the containers that do not include the node would be dropped, all the containers that include the node but are not found in that `map` will be synced, and all the containers that could not be synced successfully will be tried to be synced in the next `Sync...All()`

> Do we need it as a map and not as a slice?

yes, we use a rand access when deciding what exact container should be synced on the `NewEpoch` event

fyrchik (approved):

PR #1952, opened 2022-10-21
[#1936] network/cache: Handle dial errors

fyrchik:
Close #1936.

After an SDK update `Dial` can return an error.

Signed-off-by: Evgenii Stratonikov <evgeniy@morphbits.ru>

PR #1957, opened 2022-10-24
Move shard to read-only if cannot move to degraded

fyrchik:
Depends on #1952 .

### DIFF
--- a/pkg/local_object_storage/engine/engine.go
+++ b/pkg/local_object_storage/engine/engine.go
@@ -40,9 +40,10 @@ func (e *StorageEngine) reportShardError(
 	msg string,
 	err error,
 	fields ...zap.Field) {
+	sid := sh.ID()
 	errCount := sh.errorCount.Inc()
 	e.log.Warn(msg, append([]zap.Field{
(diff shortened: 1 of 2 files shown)


pkg/local_object_storage/engine/engine.go
```diff
@@ -53,12 +54,25 @@ func (e *StorageEngine) reportShardError(
 
 	err = sh.SetMode(mode.DegradedReadOnly)
 	if err != nil {
-		e.log.Error("failed to move shard in degraded mode",
+		e.log.Error("failed to move shard in degraded-read-only mode, moving to read-only",
+			zap.Stringer("shard_id", sid),
 			zap.Uint32("error count", errCount),
 			zap.Error(err))
+
+		err = sh.SetMode(mode.ReadOnly)
```
cthulhu-rider, line 62: Can you please explain why setting `ReadOnly` can succeed after the `DegradedReadOnly` failed (not obvious at first glance)?

Also maybe it's worth to mention the fallback in `reportShardError` docs.

cthulhu-rider (approved):

carpawell (approved):

PR #1965, opened 2022-10-25
Fix removal of the missing object in RO writecache

### LINKED ISSUE #1905: 'flush cache' command reported OK, but actually cache was not flushed because it was in READ-ONLY mode.
Now we have I hidden write-cache mode. From the system administrator side it is not correclty. Because the admin cannot see the overall status of the system. And he have not obviously errors in log

## Expected Behavior
If write-cache is in READ-ONLY mode we should report somethink like this "Operation not possible in write-cache READ-ONLY mode".

## Current Behavior
We returned ```Write-cache has been flushed```, but actually write-cache was in READ-ONLY mode and we got a lot of errors in log.

1. Run the command to flush cache while it is in READ-ONLY mode:
```
root@az:/srv/neofs# sudo neofs-cli --endpoint 127.0.0.1:8091 -w /etc/neofs/storage/wallet.json control shards flush-cache --id MnaJVa2bAqANyXabcajLWQ,KNqV3sT3nSFQKMan7w54QH
Enter password >
Write-cache has been flushed.
```
2. See the rrros in log:
```
Oct 17 08:49:42 az sudo[1559]:     root : TTY=pts/0 ; PWD=/srv/neofs ; USER=root ; COMMAND=/usr/bin/neofs-cli --endpoint 127.0.0.1:8091 -w /etc/neofs/storage/wallet.json control shards flush-cache --id MnaJVa2bAqANyXabcajLWQ,KNqV3sT3nSFQKMan7w54QH
Oct 17 08:49:45 az neofs-node[644]: 2022-10-17T08:49:45.312Z        error        shard/delete.go:43        can't delete object from write cache        {"shard_id": "MnaJVa2bAqANyXabcajLWQ", "error": "write-cache is in read-only mode"}
Oct 17 08:49:45 az neofs-node[644]: 2022-10-17T08:49:45.313Z        error        shard/delete.go:43        can't delete object from write cache        {"shard_id": "MnaJVa2bAqANyXabcajLWQ", "error": "write-cache is in read-only mode"}
Oct 17 08:49:45 az neofs-node[644]: 2022-10-17T08:49:45.313Z        error        shard/delete.go:43        can't delete object from write cache        {"shard_id": "MnaJVa2bAqANyXabcajLWQ", "error": "write-cache is in read-only mode"}
Oct 17 08:49:45 az neofs-node[644]: 2022-10-17T08:49:45.313Z        error        shard/delete.go:43        can't delete object from write cache        {"shard_id": "MnaJVa2bAqANyXabcajLWQ", "error": "write-cache is in read-only mode"}
Oct 17 08:49:45 az neofs-node[644]: 2022-10-17T08:49:45.313Z        error        shard/delete.go:43        can't delete object from write cache        {"shard_id": "MnaJVa2bAqANyXabcajLWQ", "error": "write-cache is in read-only mode"}
...
```

**Versions:**
```
NeoFS Storage node
Version: v0.32.0-121-gade19077
GoVersion: go1.18.4
```
## Your Environment
Server setup and configuration:
cloud, 4 VMs, 4 SN, 4 http qw, 4 s3 gw

Operating System and version (uname -a):
linux vedi 5.10.0-16-amd64 https://github.com/nspcc-dev/neofs-node/issues/1 SMP Debian 5.10.127-1 (2022-06-30) x86_64 GNU/Linux

cthulhu-rider:
* closes #1905

### DIFF
--- a/pkg/local_object_storage/shard/delete.go
+++ b/pkg/local_object_storage/shard/delete.go
@@ -1,8 +1,11 @@
 package shard
 
 import (
+	"errors"
+
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor/common"
 	meta "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/metabase"
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/writecache"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"go.uber.org/zap"
 )
@@ -39,7 +42,7 @@ func (s *Shard) Delete(prm DeletePrm) (DeleteRes, error) {
 	for i := range prm.addr {
 		if s.hasWriteCache() {
 			err := s.writeCache.Delete(prm.addr[i])
-			if err != nil && !IsErrNotFound(err) {
+			if err != nil && !IsErrNotFound(err) && !errors.Is(err, writecache.ErrReadOnly) {
 				s.log.Warn("can't delete object from write cache", zap.String("error", err.Error()))
 			}
 		}
--- a/pkg/local_object_storage/shard/delete.go
+++ b/pkg/local_object_storage/shard/delete.go
@@ -40,7 +40,7 @@ func (s *Shard) Delete(prm DeletePrm) (DeleteRes, error) {
 		if s.hasWriteCache() {
 			err := s.writeCache.Delete(prm.addr[i])
 			if err != nil && !IsErrNotFound(err) {
-				s.log.Error("can't delete object from write cache", zap.String("error", err.Error()))
+				s.log.Warn("can't delete object from write cache", zap.String("error", err.Error()))
 			}
 		}
 
--- a/pkg/local_object_storage/writecache/writecache.go
+++ b/pkg/local_object_storage/writecache/writecache.go
@@ -23,6 +23,12 @@ type Info struct {
 type Cache interface {
 	Get(address oid.Address) (*object.Object, error)
 	Head(oid.Address) (*object.Object, error)
+	// Delete removes object referenced by the given oid.Address from the
+	// Cache. Returns any error encountered that prevented the object to be
+	// removed.
+	//
+	// Returns apistatus.ObjectNotFound if object is missing in the Cache.
+	// Returns ErrReadOnly if the Cache is currently in the read-only mode.
 	Delete(oid.Address) error
 	Iterate(IterationPrm) error
 	Put(common.PutPrm) (common.PutRes, error)


pkg/local_object_storage/shard/delete.go
```diff
@@ -40,7 +40,7 @@ func (s *Shard) Delete(prm DeletePrm) (DeleteRes, error) {
 		if s.hasWriteCache() {
 			err := s.writeCache.Delete(prm.addr[i])
 			if err != nil && !IsErrNotFound(err) {
-				s.log.Error("can't delete object from write cache", zap.String("error", err.Error()))
+				s.log.Info("can't delete object from write cache", zap.String("error", err.Error()))
```
carpawell, line 43: isn't it `Error`? `RO` mode is ok to be in `Info` but what about others?

cthulhu-rider: I don't think this should be an `error` level since garbage removal problems are not critical and don't require immediate attention. So, IMO we should leave no higher than `warn` level here. Thoughts @fyrchik?

fyrchik: Both `Error` and `Warn` look ok to me.

pkg/local_object_storage/writecache/delete.go
```diff
@@ -11,41 +14,71 @@ import (
...
-	c.modeMtx.RLock()
-	defer c.modeMtx.RUnlock()
-	if c.readOnly() {
-		return ErrReadOnly
-	}
-
+	// Pre-check object location. If object is missing, we return immediately
+	// regardless of current cache mode. Otherwise, we need to check writing
+	// mode for removal.
+	var exists, inDB bool
```
fyrchik, line 21: Can we just check for `ErrReadOnly` on the caller side?

cthulhu-rider: So we fail with read-only error even if object is missing?

carpawell:
@cthulhu-rider, conflicts

pkg/local_object_storage/writecache/delete.go
fyrchik: Yes, I think it is ok. Well, we can also perform `Exists` on the caller side or filter `ErrReadOnly`.

cthulhu-rider: So do you suggest to:
1. exclude `ErrReadOnly` from shard logs
2. pre-check object existence in shard layer as you said in the issue
> I suggest lowering the error level here and also check if an object exists first.

pkg/local_object_storage/writecache/delete.go
```diff
@@ -1,51 +1,85 @@
...
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"go.etcd.io/bbolt"
 )
 
 // Delete removes object from write-cache.
 //
 // Returns an error of type apistatus.ObjectNotFound if object is missing in write-cache.
+// Returns ErrReadOnly if the object can't be removed due to read-only mode.
 func (c *cache) Delete(addr oid.Address) error {
-	c.modeMtx.RLock()
```
fyrchik, line 14: `modeMtx` should be taken for the whole duration -- it also serves as a synchronization point for `Close`, `Flush` etc.

pkg/local_object_storage/writecache/delete.go
fyrchik: Yes, `ErrReadOnly` can be excluded.
Also IMO it can be returned befor existence check.

pkg/local_object_storage/writecache/delete.go
cthulhu-rider: Not actualy anymore.

fyrchik (approved):

carpawell:
should the changelog be updated?

Issue #1969, opened 2022-10-25
Investigate and document all possible disk errors that shoud count in shard error counter


fyrchik:
As an example we have recently encountered `too many open files` error because of wrong system configuration.
Should we count them in the shard error counter?

Currently we count all errors from `Open`/`Write` as "physical" errors. Maybe we should behave differently in some situations.

fyrchik:
As a first step towards this goal I propose to define custom error type with `IsLogical` method. The engine can then simply call `IsLogical` and increase error counter if not.

PR #1982, opened 2022-10-26
local_object_storage: Add a type for logical errors

### LINKED ISSUE #1588: Distinguish between logical and storage errors in the metabase
Currently all errors are returned via the `error` return value. Consider `Exists`, it can return `apistatus.ErrAlreadyRemoved` and some others. There are multiple problems with this approach:
1. Error is complicate and error-prone.
2. It doesn't reflect the reality: if the object was removed it is an expected situation, there is nothing wrong with it.

The proposal is to define local status enum in the metabase and use it for all operations. It should contain each possible logical condition and could be accessed via `res.Reason()` method. Some examples:
1. checking for existence of the removed object
2. deleting locked object
3. deleting special object

On the shard level we can have a simple routine which translates this enum to a proper status code if needed.

fyrchik:
Close #1588.

All logic errors are wrapped in `logicerr.Logical` type and do not affect shard error counter.
This should (1) fix all bugs about shard being unexpectedly (AND wrongly) in degraded and (2) make it easier to make some errors "logical". For example, for #1969 we could just filter `write` syscall result and wrap some of them.


acid-ant (approved):

pkg/core/object/errors.go
```diff
@@ -1,8 +1,12 @@
...
+	"errors"
+
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/util/logicerr"
+)
 
 // ErrObjectIsExpired is returned when the requested object's
 // epoch is less than the current one. Such objects are considered
 // as removed and should not be returned from the Storage Engine.
-var ErrObjectIsExpired = errors.New("object is expired")
+var ErrObjectIsExpired = logicerr.Wrap(errors.New("object is expired"))
```
cthulhu-rider, line 12: Lets replace this eror to `local_object_storage` space (here or later)? Seems pretty strange to see it here.

pkg/local_object_storage/blobstor/common/errors.go
```diff
@@ -1,10 +1,14 @@
...
+import (
+	"errors"
+
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/util/logicerr"
+)
 
 // ErrReadOnly MUST be returned for modifying operations when the storage was opened
 // in readonly mode.
-var ErrReadOnly = errors.New("opened as read-only")
+var ErrReadOnly = logicerr.Wrap(errors.New("opened as read-only"))
```
cthulhu-rider, line 11: Maybe provide similar constructor `logicerr.New(string) error`? Otherwise, I would expect wrapping on the `return` side instead of the definition place.

cthulhu-rider (approved):

pkg/local_object_storage/blobstor/common/errors.go
fyrchik: The intention was to keep the package as small as possible. But I guess `New` is not big and would save us some code.

fyrchik: Also, it is intentionally used in the definition -- this way it is harder to misuse or forget about something.
This wrapper is used only for error reporting and here the _essence_ of the error matters, not where it ocurred.

fyrchik: Fixde

pkg/core/object/errors.go
fyrchik: Fixed

cthulhu-rider (approved):

Issue #1812, opened 2022-09-23
System behaviour when evacuating REP 1 objects


fyrchik:
We want to behave correctly when evacuating objects storing only on 1 node.
Here are 2 situations:
1. Move node to maintenance -> evacuate objects
2. Evacuate objects -> move node to maintenance

In (1) the objects could be unavailable between the first and the second step.
In (2) the objects could become unavailable if policer on another node removes them right after the evacuation.

So we need to take special care here. The proposal is following:
1. Allow node to answer some requests in maintenance mode if possible.
2. Change its state in the netmap.
3. Make sure policer takes netmap into account even if a node answers requests: if an object is stored on a node that is under maintenance in a single replica, make sure it is replicated.

fyrchik:
Actually, this may be an SDK issue -- we could extend replica selection algorithm to take state into account during HRW sorting.

fyrchik:
Let's make it a bug -- we do not want policer to remove a single available copy.
However, policer still needs to avoid _creating_ additional copies.

carpawell:
@fyrchik, was it closed via the #1987?

PR #1987, opened 2022-10-26
[#1812] policer: Do not remove copies if there are maintenance nodes

fyrchik:


pkg/services/policer/check.go
```diff
@@ -144,12 +144,18 @@ func (p *Policer) processNodes(ctx *processPlacementContext, addr oid.Address,
 	nodes []netmap.NodeInfo, shortage uint32, checkedNodes *nodeCache) {
 	prm := new(headsvc.RemoteHeadPrm).WithObjectAddress(addr)
 
+	// Number of copies that are stored on non-maintenance nodes.
```
cthulhu-rider, line 147: maybe `on maintenance`?

cthulhu-rider (approved):

pkg/services/policer/check.go
fyrchik: fixed

PR #1996, opened 2022-10-28
engine: always select proper shard for a tree

fyrchik:
Currently there is a possibility for modifying operations to fail because of I/O errors and a new tree to be created on another shard. This commit adds existence check for modifying operations. Read operations remain as they are, not to slow things. `TreeDrop` is an exception, because this is a tree removal and trying multiple shards is not an unwanted behaviour.

### DIFF
--- a/pkg/local_object_storage/pilorama/forest.go
+++ b/pkg/local_object_storage/pilorama/forest.go
@@ -209,3 +209,10 @@ func (f *memoryForest) TreeList(cid cidSDK.ID) ([]string, error) {
 
 	return res, nil
 }
+
+// TreeExists implements the pilorama.Forest interface.
+func (f *memoryForest) TreeExists(cid cidSDK.ID, treeID string) (bool, error) {
+	fullID := cid.EncodeToString() + "/" + treeID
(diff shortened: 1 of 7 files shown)


pkg/local_object_storage/pilorama/forest.go
```diff
@@ -209,3 +209,10 @@ func (f *memoryForest) TreeList(cid cidSDK.ID) ([]string, error) {
 
 	return res, nil
 }
+
+// TreeExists implements the pilorama.Forest interface.
+func (f *memoryForest) TreeExists(cid cidSDK.ID, treeID string) (bool, error) {
+	fullID := cid.String() + "/" + treeID
```
carpawell, line 215:
```suggestion
	fullID := cid.EncodeToString() + "/" + treeID
```

carpawell (approved):
Changelog conficts

pkg/local_object_storage/pilorama/forest.go
fyrchik: Fixed

Issue #1971, opened 2022-10-25
object-lock command improvement suggestions


vkarak1:
## Is your feature request related to a problem? Please describe.
Could you please consider two suggestions to the object-lock command:

1. Object lock can not be run while shard is in read-only state and neofs-cli returns unfriendly message, is that possible to improve the message in described case? Now it displays:
   `Store lock object in NeoFS: client failure: status: code = 1024 message = lock operation failed`
2. It would be great to specify object_id and container_id in object-lock command using usual "--cid" and "--oid" prefixes. Now the command looks the following:
`  neofs-cli object lock -r node1:8080 -w /etc/neofs/storage/wallet.json --expire-at 10 893dGYPLgRNSiqVBW61J7aP1xLvsYzGxQViNZ6smmUx5 8sRdhwH6ERxjC4cpXEiNcVYaeYudGynpzxsD6wfCJSGk`

NeoFS Storage node
Version: v0.33.0-31-ge1be0180
GoVersion: go1.18.4

carpawell:
Closed via #1998.

carpawell:
I misunderstand the `1.` point, it is about an error description, not about an error chain.

The problem is in [that](https://github.com/nspcc-dev/neofs-node/blob/b122312e4b0f03c443b12378027fb7f1f27fd69b/pkg/local_object_storage/engine/lock.go#L50) line: we just do not have enough information about the errors we get. I guess it was done on purpose: we iterate over _some shard set_ and we do not have any error that could wrap some other errors, so it is just not clear what error we can return here: the first, the last of try to add some API error that could wrap all errors from all shards and return that "encapsulated error set".

The case that led to that error creation was RO mode for every node's shard. So what message could be printed here? RO shard? That would mean that the last shard failed because of RO but no info about the previous ones. What if the first one (the desired one) has RO mode? print nothing cause that is not an error and we _did_ lock an object?

/cc @fyrchik, @vkarak1, @acid-ant, @cthulhu-rider

PR #1998, opened 2022-10-28
Fix/Unify CID and OID flags provision

carpawell:
Related to #1971.

### DIFF
--- a/cmd/neofs-cli/modules/object/lock.go
+++ b/cmd/neofs-cli/modules/object/lock.go
@@ -21,22 +21,22 @@ import (
 
 // object lock command.
 var objectLockCmd = &cobra.Command{
-	Use:   "lock CONTAINER OBJECT...",
+	Use:   "lock",
 	Short: "Lock object in container",
 	Long:  "Lock object in container",
-	Args:  cobra.MinimumNArgs(2),
-	Run: func(cmd *cobra.Command, args []string) {
-		var cnr cid.ID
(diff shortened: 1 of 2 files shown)


cmd/neofs-cli/modules/object/lock.go
```diff
@@ -21,22 +21,22 @@ import (
...
+	Run: func(cmd *cobra.Command, _ []string) {
+		cidRaw, _ := cmd.Flags().GetString("cid")
 
-		err := cnr.DecodeString(args[0])
+		var cnr cid.ID
+		err := cnr.DecodeString(cidRaw)
 		common.ExitOnErr(cmd, "Incorrect container arg: %v", err)
 
-		argsList := args[1:]
+		oidsRaw, _ := cmd.Flags().GetStringSlice("oid")
```
cthulhu-rider, line 34: We need a named const for cid and oid, they are used 3 times.

acid-ant (approved):

cmd/neofs-cli/modules/object/lock.go
carpawell: https://github.com/nspcc-dev/neofs-node/pull/1924#discussion_r1001618644

All of our pkgs are facing that problem, suggest fixing that in a separate PR, will rebase that PR if the PR from the link is merges before

fyrchik (approved):

PR #2014, opened 2022-11-02
Add pre-commit

realloc:
Adding [pre-commit](https://pre-commit.com/) framework support.

It covers our basic needs, but there is a room for improvement:
- Add GitHub issue existence check
- Add version match check for tag/VERSION/deb


acid-ant (approved):

alexvanin:
Is it possible to test pre-commit.ci in this PR?

realloc:
> Is it possible to test pre-commit.ci in this PR?

Hope so, working on this.

realloc:
> Is it possible to test pre-commit.ci in this PR?

pre-commit.ci doesn't add nice sections in Checks tab. Trying separate action.

If both ways fail, we can add check status with external Jenkins, as we did some time ago.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,16 @@
...
+    rev: v4.3.0
+    hooks:
+    - id: check-added-large-files
+    - id: check-case-conflict
+    - id: check-executables-have-shebangs
+    - id: check-shebang-scripts-are-executable
+    - id: check-merge-conflict
+    - id: check-json
+    - id: check-xml
+    - id: check-yaml
```
fyrchik, line 12: If any of this fails, can we see all errors or only the first?

Makefile
```diff
@@ -132,6 +132,14 @@ test:
 lint:
 	@golangci-lint --timeout=5m run
 
+# Activate pre-commit hooks
+pre-commit:
+	pre-commit install -t pre-commit -t commit-msg
```
fyrchik, line 137: If I change a commit message during rebase, will all hooks be executed on only message check?

.gitlint
```diff
@@ -0,0 +1,10 @@
+[general]
+fail-without-commits=true
+contrib=CC1
+
+[title-match-regex]
+regex=^\[\#[0-9]*\]\s
```
fyrchik, line 6: `^\[\#[0-9]+\]\s.+$`?

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,17 @@
+name: pre-commit
+
+on:
+  pull_request:
+  push:
+    branches: [main]
```
fyrchik, line 6: `master`?
Can we use regex for support?

fyrchik:
Also, can we globally exclude autogenerated files from most of the checks? I can imagine a situation where a linter will complain.

.pre-commit-config.yaml
realloc: Should see all errors. Will check again.

Makefile
realloc: No, it should not.

realloc:
> Also, can we globally exclude autogenerated files from most of the checks? I can imagine a situation where a linter will complain.

We can. For the `neofs-node`, what do you propose to exclude?

realloc:
@alexvanin Let's keep precommit.ci for now. It looks like a normal GH Actions check. We can do all cosmetically fancy stuff later with the dedicated Jenkins.

fyrchik:
@realloc at least all `**/*.pb.go` files.

fyrchik (approved):

PR #2019, opened 2022-11-07
Fix/Big object removal

### LINKED ISSUE #1699: GC could not delete object that is split
Our mainnet node `asuka` reported such logs:

```
asuka neofs-node[71361]: 2022-08-17T16:10:04.625Z        warn        shard/gc.go:216        could not delete the objects        {"error": "object not found, split info has been provided"}
```

It seems like GC must not get such errors while removing objects.

1. Add object address to the error;
2. Do not [stop removing objects](https://github.com/nspcc-dev/neofs-node/blob/d99800ee93bbe9b0699313ba7a5ecab85699731c/pkg/local_object_storage/metabase/delete.go#L72) after fall (one of the shard kept many objects with GC and could do nothing with that);
3. Recheck that removing split objects works correctly.

NeoFS version: `v0.28.1`

carpawell:
It is not an error: removing virtual object is expected and should be just skipped. Getting a virtual object with `raw` flag is considered as an impossible action, all the virtual objects removals will be handled via their children's removals implicitly.

Closes #1699.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Changelog for NeoFS Node
 - `neofs-cli container nodes`'s output (#1991)
 - Do not panic with bad inputs for `GET_RANGE` (#2007)
 - Correctly select the shard for applying tree service operations (#1996)
+- Physical child object removal by GC (#1699)
 
 ### Removed
 ### Updated
--- a/pkg/local_object_storage/metabase/delete_test.go
+++ b/pkg/local_object_storage/metabase/delete_test.go
@@ -39,9 +39,9 @@ func TestDB_Delete(t *testing.T) {
 	require.NoError(t, err)
 	require.Len(t, l, 1)
 
-	// try to remove parent unsuccessfully
+	// try to remove parent, should be no-op, error-free
 	err = metaDelete(db, object.AddressOf(parent))
-	require.Error(t, err)
+	require.NoError(t, err)
 
 	// inhume parent and child so they will be on graveyard
 	ts := generateObjectWithCID(t, cnr)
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Changelog for NeoFS Node
 - Open FSTree in sync mode by default (#1992)
 - `neofs-cli container nodes`'s output (#1991)
 - Do not panic with bad inputs for `GET_RANGE` (#2007)
+- Big objects removal (#1699)
```
fyrchik, line 14: Could you be a bit more specific here? We have many similar bugs regarding big object removal.

pkg/local_object_storage/metabase/delete_test.go
```diff
@@ -39,9 +39,9 @@ func TestDB_Delete(t *testing.T) {
 	require.NoError(t, err)
 	require.Len(t, l, 1)
 
-	// try to remove parent unsuccessfully
+	// try to remove parent, should be no-op, error-free
```
fyrchik, line 42: Have you checked that `control drop-object` is not broken?

fyrchik (approved):

acid-ant (dismissed):

CHANGELOG.md
carpawell: sure, expanded

pkg/local_object_storage/metabase/delete_test.go
carpawell: depends on parent object removal expectations

fyrchik (approved):

pkg/local_object_storage/metabase/delete_test.go
carpawell: checked, removes all the child objects on parent object removal

PR #2044, opened 2022-11-11
[SUPPORT] Fix/Panic caused by flag redefinition

carpawell:
Oups. Missclick while copy-pasting in #2039 (or just wrong autocomplete by IDE, do not remember).


carpawell:
BTW, some build CI step would prevent it.

fyrchik (approved):

PR #2050, opened 2022-11-12
[SUPPORT] Fix/Lock object resync in a metabase

### LINKED ISSUE #1502: Lock objects do not fill metabase on `resync_metabase`
Seems like it is true. SE [does not](https://github.com/nspcc-dev/neofs-node/blob/68a01121936170abae893d8772bd2629e6322c5d/pkg/local_object_storage/shard/control.go#L77) call [`Lock`](https://github.com/nspcc-dev/neofs-node/blob/68a01121936170abae893d8772bd2629e6322c5d/pkg/local_object_storage/metabase/lock.go#L31).

Needed to be verified. If true, we are losing every lock relationship on every metabase dropping.

carpawell:
Closes #1502.

### DIFF
--- a/pkg/local_object_storage/engine/inhume.go
+++ b/pkg/local_object_storage/engine/inhume.go
@@ -9,6 +9,7 @@ import (
 	apistatus "github.com/nspcc-dev/neofs-sdk-go/client/status"
 	objectSDK "github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	"go.uber.org/zap"
 )
 
 // InhumePrm encapsulates parameters for inhume operation.
@@ -79,6 +80,18 @@ func (e *StorageEngine) inhume(prm InhumePrm) (InhumeRes, error) {
 	}
 
 	for i := range prm.addrs {
+		if !prm.forceRemoval {
+			locked, err := e.isLocked(prm.addrs[i])
+			if err != nil {
+				e.log.Warn("removing an object without full locking check",
+					zap.Error(err),
+					zap.Stringer("addr", prm.addrs[i]))
+			} else if locked {
+				var lockedErr apistatus.ObjectLocked
+				return InhumeRes{}, lockedErr
+			}
+		}
+
 		if prm.tombstone != nil {
 			shPrm.SetTarget(*prm.tombstone, prm.addrs[i])
 		} else {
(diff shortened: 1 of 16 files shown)


pkg/services/policer/check.go
```diff
@@ -162,6 +166,12 @@ func (p *Policer) processNodes(ctx *processPlacementContext, addr oid.Address,
 		)
 	}
 
+	if typ == object.TypeLock {
+		// all nodes of a container must store the `LOCK` objects
+		// for correct object removal protection
+		shortage = uint32(len(nodes))
```
fyrchik, line 172: Could you explain it a bit here?
1. We broadcast them on PUT
2. They can't be deleted anyway.

pkg/local_object_storage/engine/inhume.go
```diff
@@ -79,6 +80,18 @@ func (e *StorageEngine) inhume(prm InhumePrm) (InhumeRes, error) {
 	}
 
 	for i := range prm.addrs {
+		if !prm.forceRemoval {
+			locked, err := e.isLocked(prm.addrs[i])
+			if err != nil {
+				return InhumeRes{}, fmt.Errorf("could not check locking status: %w", err)
```
fyrchik, line 86: While I agree this could be a correct behaviour, I think it is a rather big change to me merged right before the release.
Can we check what whatever is possible here and skip missing shards?
1. We do similar things for `Exists`/Tombstones in degraded mode.
2. We still check many shards, so it would be better than our current behaviour.

carpawell: so you mean just checking all the shards we can and removing the objects if all the _available_ objects allow doing that?

pkg/services/policer/check.go
carpawell: > We broadcast them on PUT

yes, but we do not check `LOCK`'s placement after creation (e.g. a node could be turned off for a while when a `LOCK` was put)

> They can't be deleted anyway.

but they would be tried to be deleted on every policer's iteration

the interesting thing about it is that the answers to both of your questions are related: we either try to keep `LOCK`s on every container node or try to remove them unsuccessfully by GC (well, ok, the 3rd option is to change the code significantly)

fyrchik: I mean could you add this info to a comment?

pkg/local_object_storage/engine/inhume.go
fyrchik: Yes

pkg/services/policer/check.go
carpawell: :hushed:

carpawell: added

pkg/local_object_storage/engine/inhume.go
carpawell: ok, added logging and shard error counter increment, check, please

fyrchik (approved):

acid-ant (approved):

PR #2061, opened 2022-11-15
Feat/Add parallel bench for `meta.Get`

carpawell:


acid-ant (approved):

fyrchik (commented):
While we are here, can we make these benchmarks provide a bit more useful numbers?
These benchmarks do different things: the parallel one tries to get the same object repeatedly, the second one returns  the time for multiple objects.
I propose to perform `Get(addrs[i%len(addrs))` in the first one and similarly use atomic counter (or thread-local counter, `addrs` is read-only) in the parallel one.

Also I suggest providing batch size in both of the benchmarks -- parallel one can use `runtime.NumCPU`, while the `serial` one needs 1.

carpawell:
@fyrchik, i did not plan them to be the same but ok, unified.

sami-nspcc:

pkg/local_object_storage/metabase/get_test.go
```diff
@@ -181,48 +187,59 @@ func BenchmarkGet(b *testing.B) {
...
-	})
-
-	var err error
-	var getPrm meta.GetPrm
-	getPrm.SetAddress(addrs[len(addrs)/2])
+	db, addrs := prepareDb(runtime.NumCPU())
 
 	b.Run("parallel", func(b *testing.B) {
 		b.ReportAllocs()
+		b.SetParallelism(1)
```
fyrchik, line 212: It is 1 by default, why this line?

fyrchik (approved):

sami-nspcc:
I have recieved a notification of a new pull request. I am starting the build of images and binaries for further testing. Build number is 3747

sami-nspcc:
Test run is finished. Please download the tarball from <a href="https://http.t5.fs.neo.org/get/FszwwAx3VznDysQGxBUMRj1zj3zp8VcohorVDnunrviV/F5YVaTpb8RSC8v3D9Viuii185fPd33SKsTGXRC76F23Q">link</a>. Untar and use `allure open` to watch the report

PR #2113, opened 2022-11-29
Fix/Container size estimations

carpawell:
The lead node should take into account container replication factor, the size of the container vector and the vectors number: the average estimation report by the container nodes does not describe container size, it only could say how much space an average container node uses and such an estimation works correctly for the `REP 1 CBF 1` containers only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ Changelog for NeoFS Node
 - Write-cache flush duplication (#2074)
 - Ignore error if a transaction already exists in a morph client (#2075)
 - ObjectID signature output in the CLI (#2104)
+- Container size estimation value (#2113)
 
 ### Removed
 ### Updated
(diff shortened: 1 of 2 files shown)


sami-nspcc:
I have recieved a notification of a new pull request. I am starting the build of images and binaries for further testing. Build number is 3729

sami-nspcc:
I have recieved a notification of a new pull request. I am starting the build of images and binaries for further testing. Build number is 3730

sami-nspcc:

fyrchik:
What problem are you trying to solve here?

alexvanin:
> it only could say how much space an average container node uses 

Well that's sort of expected, as far as I remember. Contract value represents average load of container data per node. Then it is used to calculate basic income value for all nodes in the container.

If you change it, check settlement code.

https://github.com/nspcc-dev/neofs-node/blob/79130f781e276bcf04e21ff1db4bba2c5cefbcaa/pkg/innerring/processors/settlement/basic/collect.go#L51-L67

fyrchik:
Works as intended, let's discuss desirable formula for income calculation in a separate issue.

PR #2135, opened 2022-12-06
cli: Use `--timeout` flag as a command exec timeout

### LINKED ISSUE #2124: Check `--await` and `--timeout` interaction
Check how they interact and describe the behavior.
Maybe add `--await-timeout`

cthulhu-rider:
* closes #2124

### DIFF
--- a/cmd/neofs-cli/internal/client/sdk.go
+++ b/cmd/neofs-cli/internal/client/sdk.go
@@ -21,22 +21,22 @@ var errInvalidEndpoint = errors.New("provided RPC endpoint is incorrect")
 
 // GetSDKClientByFlag returns default neofs-sdk-go client using the specified flag for the address.
 // On error, outputs to stderr of cmd and exits with non-zero code.
-func GetSDKClientByFlag(cmd *cobra.Command, key *ecdsa.PrivateKey, endpointFlag string) *client.Client {
-	cli, err := getSDKClientByFlag(cmd, key, endpointFlag)
+func GetSDKClientByFlag(ctx context.Context, cmd *cobra.Command, key *ecdsa.PrivateKey, endpointFlag string) *client.Client {
+	cli, err := getSDKClientByFlag(ctx, cmd, key, endpointFlag)
 	if err != nil {
 		common.ExitOnErr(cmd, "can't create API client: %w", err)
 	}
 	return cli
 }
 
-func getSDKClientByFlag(cmd *cobra.Command, key *ecdsa.PrivateKey, endpointFlag string) (*client.Client, error) {
+func getSDKClientByFlag(ctx context.Context, cmd *cobra.Command, key *ecdsa.PrivateKey, endpointFlag string) (*client.Client, error) {
 	var addr network.Address
 
 	err := addr.FromString(viper.GetString(endpointFlag))
 	if err != nil {
 		return nil, fmt.Errorf("%v: %w", errInvalidEndpoint, err)
 	}
-	return GetSDKClient(context.TODO(), cmd, key, addr)
+	return GetSDKClient(ctx, cmd, key, addr)
 }
 
 // GetSDKClient returns default neofs-sdk-go client.
--- a/cmd/neofs-cli/internal/client/sdk.go
+++ b/cmd/neofs-cli/internal/client/sdk.go
@@ -7,9 +7,9 @@ import (
 	"crypto/rand"
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/common"
-	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/commonflags"
 	"github.com/nspcc-dev/neofs-node/pkg/network"
 	"github.com/nspcc-dev/neofs-sdk-go/client"
 	neofsecdsa "github.com/nspcc-dev/neofs-sdk-go/crypto/ecdsa"
@@ -51,14 +51,16 @@ func GetSDKClient(ctx context.Context, cmd *cobra.Command, key *ecdsa.PrivateKey
 	prmInit.ResolveNeoFSFailures()
 	prmDial.SetServerURI(addr.URIAddr())
 	prmDial.SetContext(ctx)
-	if timeout := viper.GetDuration(commonflags.Timeout); timeout > 0 {
-		// In CLI we can only set a timeout for the whole operation.
-		// By also setting stream timeout we ensure that no operation hands
-		// for too long.
-		prmDial.SetTimeout(timeout)
-		prmDial.SetStreamTimeout(timeout)
 
-		common.PrintVerbose(cmd, "Set request timeout to %s.", timeout)
+	deadline, ok := ctx.Deadline()
+	if ok {
+		if timeout := time.Until(deadline); timeout > 0 {
+			// In CLI we can only set a timeout for the whole operation.
+			// By also setting stream timeout we ensure that no operation hands
+			// for too long.
+			prmDial.SetTimeout(timeout)
+			prmDial.SetStreamTimeout(timeout)
+		}
 	}
 
 	c.Init(prmInit)
--- a/cmd/neofs-cli/internal/commonflags/flags.go
+++ b/cmd/neofs-cli/internal/commonflags/flags.go
@@ -84,6 +84,19 @@ func Bind(cmd *cobra.Command) {
 	_ = viper.BindPFlag(Timeout, ff.Lookup(Timeout))
 }
 
+// GetCommandContextWithAwait works like GetCommandContext but uses specified
+// await timeout if 'timeout' flag is omitted and given boolean await flag is
+// set.
+func GetCommandContextWithAwait(cmd *cobra.Command, awaitFlag string, awaitTimeout time.Duration) (context.Context, context.CancelFunc) {
+	if !viper.IsSet(Timeout) {
+		if await, _ := cmd.Flags().GetBool(awaitFlag); await {
+			return getCommandContext(cmd, awaitTimeout)
+		}
+	}
+
+	return GetCommandContext(cmd)
+}
+
 // GetCommandContext returns cmd context with timeout specified in 'timeout' flag
 // if the flag is set.
 func GetCommandContext(cmd *cobra.Command) (context.Context, context.CancelFunc) {
--- a/cmd/neofs-cli/internal/commonflags/flags.go
+++ b/cmd/neofs-cli/internal/commonflags/flags.go
@@ -1,6 +1,7 @@
 package commonflags
 
 import (
+	"context"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -81,4 +82,23 @@ func Bind(cmd *cobra.Command) {
 	_ = viper.BindPFlag(Account, ff.Lookup(Account))
 	_ = viper.BindPFlag(RPC, ff.Lookup(RPC))
 	_ = viper.BindPFlag(Timeout, ff.Lookup(Timeout))
+}
+
+// GetCommandContext returns cmd context with timeout specified in 'timeout' flag
+// if the flag is set.
+func GetCommandContext(cmd *cobra.Command) (context.Context, context.CancelFunc) {
+	return getCommandContext(cmd, viper.GetDuration(Timeout))
+}
+
+func getCommandContext(cmd *cobra.Command, timeout time.Duration) (context.Context, context.CancelFunc) {
+	parentCtx := cmd.Context()
+	if parentCtx == nil {
+		parentCtx = context.Background()
+	}
+
+	if timeout <= 0 {
+		return parentCtx, func() {}
+	}
+
+	return context.WithTimeout(parentCtx, timeout)
 }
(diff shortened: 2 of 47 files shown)


sami-nspcc:
I have recieved a notification of a new pull request. I am starting the build of images and binaries for further testing. Build number is 3776

sami-nspcc:

carpawell:
How do you solve problems from https://github.com/nspcc-dev/neofs-node/issues/2124#issuecomment-1337511336? The most important, i think, is setting timeout lower that a block duration. It is really implicit and a general user may not know about such a limitation so he would create a container and would not understand that he had succeeded (and probably would create one more container).

cthulhu-rider:
User understands timeout of the whole operation only, it shouldn't care about what's going on inside.

cmd/neofs-cli/modules/container/util.go
```diff
@@ -12,8 +12,6 @@ import (
 
 const (
 	attributeDelimiter = "="
-
-	awaitTimeout = 120 // in seconds
```
roman-khimov, line 16: This is still relevant as a concept. Commands should have longer default timeouts if invoked with `--await`.

cmd/neofs-cli/internal/client/sdk.go
```diff
@@ -29,33 +28,37 @@ func GetSDKClientByFlag(cmd *cobra.Command, key *ecdsa.PrivateKey, endpointFlag
...
+	cli, err := GetSDKClient(cmd, key, addr)
 	if err != nil {
 		common.ExitOnErr(cmd, "can't create API client: %w", err)
 	}
 	return cli
 }
 
 // GetSDKClient returns default neofs-sdk-go client.
-func GetSDKClient(ctx context.Context, cmd *cobra.Command, key *ecdsa.PrivateKey, addr network.Address) (*client.Client, error) {
+func GetSDKClient(cmd *cobra.Command, key *ecdsa.PrivateKey, addr network.Address) (*client.Client, error) {
```
roman-khimov, line 39: But why removing `ctx` from it? We can have it as a parameter, especially given that it's already there.

cmd/neofs-cli/internal/commonflags/flags.go
```diff
@@ -82,3 +83,26 @@ func Bind(cmd *cobra.Command) {
 	_ = viper.BindPFlag(RPC, ff.Lookup(RPC))
 	_ = viper.BindPFlag(Timeout, ff.Lookup(Timeout))
 }
+
+// AcceptTimeout makes given cobra.Command to accept 'timeout' flag as global
+// execution timeout. If cobra.Command.Run is specified, AcceptTimeout precedes
+// it with time limited context setting.
+func AcceptTimeout(cmd *cobra.Command) {
```
roman-khimov, line 90: Why not `PreRun` for it?

cmd/neofs-cli/modules/container/delete.go
```diff
@@ -104,17 +106,21 @@ Only owner of the container has a permission to remove container.`,
...
 			getPrm.SetContainer(id)
 
-			for i := 0; i < awaitTimeout; i++ {
+			for {
 				time.Sleep(1 * time.Second)
 
-				_, err := internalclient.GetContainer(cmd.Context(), getPrm)
+				select {
+				case <-ctx.Done():
+					common.ExitOnErr(cmd, "", errDeleteTimeout)
```
roman-khimov, line 114: Also, these "waited for too long" cases should have a special exit code

roman-khimov: Consider rephrasing `errDeleteTimeout` (and alike) as well to say that "the operation was requested, but timeout has happened while waiting for its result".

cmd/neofs-cli/internal/commonflags/flags.go
cthulhu-rider: cuz then context cancellation must be done in `PostRun`, and afaik cobra doesn't provide mechanism to sync stuff b/w pre-run/post-run funcs. Well, we can use global variables, and i guess they'll fit, but i tried to avoid them

cmd/neofs-cli/modules/container/util.go
cthulhu-rider: i don't see what's the relevance. Timeout controls time period of command execution regardless of the set of internal actions. As a user, I'd expect unconditional default timeout since it's easier to understand and I can't provide await timeout explicitly

cmd/neofs-cli/internal/client/sdk.go
cthulhu-rider: this corrects the code that passed `context.TODO()` and simplifies usage cuz all callers pass `cmd.Context()` anyway

cmd/neofs-cli/modules/container/delete.go
cthulhu-rider: > Also, these "waited for too long" cases should have a special exit code

overall i agree, but think it's better to do within separate issue since previously there were no dedicated codes for these cases

cthulhu-rider: > Consider rephrasing

seems good, done

cmd/neofs-cli/modules/container/util.go
roman-khimov: Timeouts can be different for different actions and in this case they should be. 15s is just not enough in many `--await` cases (consider mainnet). This doesn't introduce any new timeout, it's the same old one, just a different default value (XXX without `--await` and YYY with `--await`, pretty easy). The effect of `--await` is felt by users immediately, so you won't need to explain anything, commands need a bit more time in this case.

cmd/neofs-cli/internal/client/sdk.go
roman-khimov: https://pkg.go.dev/context
> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:

I know we violate this right and left, but at least we can try not making the situation worse (these is `ctx` already).

cmd/neofs-cli/modules/container/util.go
cthulhu-rider: i got it. Suggest to leave `15s` for "simple" ops and `1m` for async-await, thoughts?

cmd/neofs-cli/internal/client/sdk.go
cthulhu-rider: i'd say cobra violates `Do not store Contexts inside a struct type` and we just follow this, but ok, i'll leave context as prm

cmd/neofs-cli/modules/container/util.go
roman-khimov: OK for me.

cthulhu-rider: done

cmd/neofs-cli/internal/client/sdk.go
cthulhu-rider: done

cmd/neofs-cli/modules/tree/get_by_path.go
```diff
@@ -31,7 +31,7 @@ func initGetByPathCmd() {
 	// tree service does not allow any attribute except
 	// the 'FileName' but that's a limitation of the
 	// current implementation, not the rule
-	//ff.String(pathAttributeFlagKey, "", "Path attribute")
+	// ff.String(pathAttributeFlagKey, "", "Path attribute")
```
roman-khimov, line 34: Not required.

cmd/neofs-cli/modules/tree/get_by_path.go
```diff
@@ -59,14 +61,14 @@ func getByPath(cmd *cobra.Command, _ []string) {
 
 	latestOnly, _ := cmd.Flags().GetBool(latestOnlyFlagKey)
 	path, _ := cmd.Flags().GetString(pathFlagKey)
-	//pAttr, _ := cmd.Flags().GetString(pathAttributeFlagKey)
+	// pAttr, _ := cmd.Flags().GetString(pathAttributeFlagKey)
```
roman-khimov, line 64: And here.

cmd/neofs-cli/modules/tree/get_by_path.go
```diff
@@ -59,14 +61,14 @@ func getByPath(cmd *cobra.Command, _ []string) {
...
 
 	req := new(tree.GetNodeByPathRequest)
 	req.Body = &tree.GetNodeByPathRequest_Body{
 		ContainerId:   rawCID,
 		TreeId:        tid,
 		PathAttribute: object.AttributeFileName,
-		//PathAttribute: pAttr,
+		// PathAttribute: pAttr,
```
roman-khimov, line 71: And here.

cmd/neofs-cli/modules/container/delete.go
roman-khimov: #2380.

cmd/neofs-cli/modules/object/put.go
```diff
@@ -83,7 +86,7 @@ func putObject(cmd *cobra.Command, _ []string) {
 		buf, err := os.ReadFile(filename)
 		common.ExitOnErr(cmd, "unable to read given file: %w", err)
 		objTemp := object.New()
-		//TODO(@acid-ant): #1932 Use streams to marshal/unmarshal payload
+		// TODO(@acid-ant): #1932 Use streams to marshal/unmarshal payload
```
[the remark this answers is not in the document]
cthulhu-rider: fixed all

cmd/neofs-cli/modules/container/util.go
```diff
@@ -55,3 +55,7 @@ func getSession(cmd *cobra.Command) *session.Container {
 
 	return &res
 }
+
+func getAwaitContext(cmd *cobra.Command) (context.Context, context.CancelFunc) {
+	return commonflags.GetCommandContextWithAwait(cmd, "await", time.Minute)
```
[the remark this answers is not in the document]
roman-khimov: Better now, but I think unless we have it somewhere on
```
$ ./bin/neofs-cli container create --help
Create new container and register it in the NeoFS. 
It will be stored in sidechain when inner ring will accepts it.

Usage:
  neofs-cli container create [flags]

Flags:
      --address string        Address of wallet account
  -a, --attributes strings    Comma separated pairs of container attributes in form of Key1=Value1,Key2=Value2
      --await                 Block execution until container is persisted
      --basic-acl string      HEX encoded basic ACL value or keywords like 'public-read-write', 'private', 'eacl-public-read' (default "private")
      --disable-timestamp     Disable timestamp container attribute
  -f, --force                 Skip placement validity check
  -h, --help                  help for create
      --name string           Container name attribute
  -p, --policy string         QL-encoded or JSON-encoded placement policy or path to file with it
  -r, --rpc-endpoint string   Remote node address (as 'multiaddr' or '<host>:<port>')
      --session string        Filepath to a JSON- or binary-encoded token of the container PUT session
      --subnet string         String representation of container subnetwork
  -t, --timeout duration      Timeout for the operation (default 15s)
      --ttl uint32            TTL value in request meta header (default 2)
  -w, --wallet string         Path to the wallet
  -x, --xhdr strings          Request X-Headers in form of Key=Value

Global Flags:
  -c, --config string   Config file (default is $HOME/.config/neofs-cli/config.yaml)
  -v, --verbose         Verbose output
```
it'd still be very unobvious to users.

roman-khimov (approved):

cmd/neofs-cli/modules/container/create.go
```diff
@@ -134,17 +134,21 @@ It will be stored in sidechain when inner ring will accepts it.`,
 			getPrm.SetClient(cli)
 			getPrm.SetContainer(id)
 
-			for i := 0; i < awaitTimeout; i++ {
+			for {
 				time.Sleep(1 * time.Second)
```
notimetoname, line 138: change it to a channel now (here and in other places)? as i understand it will block sigint for 1 sec

notimetoname (commented):
Can we at least _try_ to warn a user in cases when he tries to create a container with a 5s timeout in 15s block duration env?

notimetoname:
> it shouldn't care about what's going on inside

So we make him care with that PR. If he specifies a too short duration he is not able to understand undefined results.

roman-khimov:
> If he specifies a too short duration he is not able to understand undefined results.

I think he would get some "container creation was requested, but timeout has happened while waiting for the outcome", is it that hard to decipher what's the problem?

notimetoname:
> is it that hard to decipher what's the problem?

Missed error text changes. Well, I like it more now but still sometimes it can be successful and sometimes not. No one knows will the next call finishes or not. If we are talking about containers, a user won't know its ID (or it will be too hard than just waiting for a command to finish).

So my point is if it is possible to warn a user that some calls are useless by design, I would do it. About the rest, I am ok.

roman-khimov:
> If we are talking about containers, a user won't know its ID

We can print it, that'd be useful, true.

> if it is possible to warn a user that some calls are useless by design, I would do it

We can compare timeouts against block time and print a warning, yes.

cthulhu-rider:
> a user won't know its ID

actually he does https://github.com/nspcc-dev/neofs-node/blob/04ce9e3e8a46c6cf425e9cbb5ec193b1bcb6d679/cmd/neofs-cli/modules/container/create.go#L125

in `delete` and `set-eacl` ID is an argument, so user definitely knows it.

i noticed that `set-eacl` doesn't print message about accepted request at all. Will add

notimetoname:
> actually he does

My bad.

notimetoname (approved):

PR #2136, opened 2022-12-06
cli: Autogenerate random key even without `-g` flag where possible 

### LINKED ISSUE #2123: Generate new private key silently if it is necessary in `neofs-cli`
Avoid user from providing option `-g`, leave only `-w`.

cthulhu-rider:
* closes #2123

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Changelog for NeoFS Node
 - TLS RPC support for IR consensus mode (#2322)
 
 ### Changed
+- NeoFS CLI generates random private key automatically if wallet is omitted (#2123)
 
 ### Fixed
 - Inability to restore RPC connection after the second disconnect (#2325)
--- a/cmd/neofs-cli/internal/key/raw.go
+++ b/cmd/neofs-cli/internal/key/raw.go
@@ -25,8 +25,17 @@ func Get(cmd *cobra.Command) *ecdsa.PrivateKey {
 	return pk
 }
 
+var errMissingFlag = errors.New("missing key flag")
+
+// get decodes ecdsa.PrivateKey from the wallet file located in the path
+// stored by commonflags.WalletPath viper key. Returns errMissingFlag
+// if the viper flag is not set.
 func get(cmd *cobra.Command) (*ecdsa.PrivateKey, error) {
 	keyDesc := viper.GetString(commonflags.WalletPath)
+	if keyDesc == "" {
+		return nil, errMissingFlag
+	}
+
 	data, err := os.ReadFile(keyDesc)
 	if err != nil {
 		return nil, fmt.Errorf("%w: %v", ErrFs, err)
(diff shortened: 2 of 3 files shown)


sami-nspcc:
I have recieved a notification of a new pull request. I am starting the build of images and binaries for further testing. Build number is 3777

sami-nspcc:

carpawell (commented):
@fyrchik, @realloc, have we decided to depricate `-g` flag?

sami-nspcc:
Test run is finished. Please download the tarball from <a href="https://http.t5.fs.neo.org/get/FszwwAx3VznDysQGxBUMRj1zj3zp8VcohorVDnunrviV/Cc5vS76fjQcYnzQnVSJgePaH93YFuCPYXfrSEyqNWo6g">link</a>. Untar and use `allure open` to watch the report

realloc:
I strongly oppose the `-g` deprecation. In cases where we have a config file for `nepfs-cli` there will be no way to use generated key. This not only may break test scenarios, but also lead to ambiguous situations.

realloc (changes requested):
Instead of deprecation, maybe we should explicitly print out what key is used in verbose mode and still allow to use `-g` for overriding existing wallet settings from configuration file or ENV variables.

cthulhu-rider:
@realloc I misinterpreted the task, so I'll remove deprecation mark.

cmd/neofs-cli/internal/key/raw.go
```diff
@@ -51,12 +60,23 @@ func GetOrGenerate(cmd *cobra.Command) *ecdsa.PrivateKey {
...
+
+	common.PrintVerbose("Wallet is not specified, generating random private key for command processing...")
+
+	priv, err := keys.NewPrivateKey()
+	if err != nil {
+		return nil, fmt.Errorf("%w: %v", errCantGenerateKey, err)
 	}
-	return get()
+
+	common.PrintVerbose("Private key generated successfully.")
```
carpawell, line 79: can we know its public part?

carpawell (commented):
Does that PR mean sending requests with a random key every time `-w` is not specified (even by a mistake)?

carpawell:
Conflicts.

cmd/neofs-cli/internal/key/raw.go
```diff
@@ -51,12 +60,23 @@ func GetOrGenerate(cmd *cobra.Command) *ecdsa.PrivateKey {
 }
 
 func getOrGenerate(cmd *cobra.Command) (*ecdsa.PrivateKey, error) {
-	if viper.GetBool(commonflags.GenerateKey) {
```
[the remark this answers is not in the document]
roman-khimov:
```
Global Flags:
  -c, --config string   Config file (default is $HOME/.config/neofs-cli/config.yaml)
```
Maybe it's unwrapped into the path/pass already, I have not checked.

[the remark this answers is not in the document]
roman-khimov: User may have some config file, but still want to run a specific command in "anonymous" mode. From what I see now he'll just receive the key from `get()` without any possibility to override. That's the reason to keep `commonflags.GenerateKey` at the same time, it doesn't make much sense otherwise.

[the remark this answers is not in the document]
cthulhu-rider: done

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ Changelog for NeoFS Node
 - TLS RPC support for IR consensus mode (#2322)
 
 ### Changed
+- NeoFS CLI generates random private key automatically if wallet is omitted (#2123)
```
notimetoname, line 11: for all the commands? even `object put`?

cthulhu-rider: for those that use `key.GetOrGenerate`. There are several command using `key.Get` which doesn't fallback to random key (e.g. `container put` can't exec w/o wallet obviously)

notimetoname: oh, i see. containers are safe but `object put/delete`, `container set-ecl` are not. well do not totally agree but not that critical

notimetoname (approved):

CHANGELOG.md
cthulhu-rider: a little race on the answer with you

container cannot be created w/o wallet. Object can: it will be owned by anonymous and imo it's pretty expected, if u wanna own the object - provide wallet

PR #2138, opened 2022-12-07
[#2130] adm: Add dump-nns and restore-nns commands

### LINKED ISSUE #2130: neofs-adm: add export/import NNS domains
Add new commands:
* `dump-domains` (maybe `dump-nns`)
* `restore-domains` (maybe `restore-nns`)

KirillovDenis:
close #2130

### DIFF
--- a/cmd/neofs-adm/internal/modules/morph/nns.go
+++ b/cmd/neofs-adm/internal/modules/morph/nns.go
@@ -1,0 +1,446 @@
+package morph
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/nspcc-dev/neo-go/pkg/io"
+	"github.com/nspcc-dev/neo-go/pkg/neorpc/result"
+	"github.com/nspcc-dev/neo-go/pkg/rpcclient/invoker"
+	"github.com/nspcc-dev/neo-go/pkg/rpcclient/unwrap"
+	"github.com/nspcc-dev/neo-go/pkg/smartcontract/callflag"
+	"github.com/nspcc-dev/neo-go/pkg/util"
+	"github.com/nspcc-dev/neo-go/pkg/vm/emit"
+	"github.com/nspcc-dev/neo-go/pkg/vm/stackitem"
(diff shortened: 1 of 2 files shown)


KirillovDenis:
Currently, this PR sub-optimal in terms of tx sending (we await txs for every domains, this should be fixed a little bit later).

Also there are the following problems/questions:
* NNS contract doesn't have methods to get owner and admin for specific domain. Should we try to get this info using raw contract storage (it seems that some owners should be changed after redeployment (e.g. for `container` domain))?
* If domain already has some records should we 1) add missed records, 2) overwrite all records 3) skip restoring this domain ?

/cc @alexvanin @fyrchik

carpawell:
@roman-khimov, @cthulhu-rider, I have checked this PR, TBH, I would redo some things and also it is from an archive repo now. Do not understand the 1st question, we have (and seems like always had `OwnerOf` method). The second one is a regular question, we just need a consensus here (I think "add missed records" is the way).

Suggest closing this PR, or merging it and continuing immediately with an extensions via another PRs.

cmd/neofs-adm/internal/modules/morph/nns.go
```diff
@@ -0,0 +1,446 @@
...
+		if err != nil {
+			return fmt.Errorf("can't check if domain is already restored '%s': %w", domain.Name, err)
+		}
+
+		bw.Reset()
+
+		if isAvailable {
+			soa := domain.SOA
+			emit.AppCall(bw.BinWriter, ch, "register", callflag.All,
+				soa.Name, wCtx.ConsensusAcc.Contract.ScriptHash(), soa.Email, soa.Refresh, soa.Retry, soa.Expire, soa.TTL)
```
roman-khimov, line 293: I think it's broken here. Domains have owners and we're not keeping them.

roman-khimov (commented):
1. This won't merge without update
2. Some things need to be fixed, but they will take some time
3. I don't believe in dump/restore concept, it's very suspicious to me and I'm not sure we can easily survive this.
4. I've never seen this being done in practice and never needed it.
5. We'll have https://github.com/neo-project/neo/issues/3463 eventually, so chain migration doesn't seem to be valuable at all.
6. If needed, the code won't go away and we can return to it.

PR #2139, opened 2022-12-07
object/put: use `sync.Pool` for temporary payloads

fyrchik:


acid-ant (approved):

carpawell (commented):
Could you explain the idea a little?

carpawell (approved):
My fault: i did not notice `getPayload` call and thought that `Pool` was used write-only.

PR #2142, opened 2022-12-07
Fix/Shard mode switch

### LINKED ISSUE #1956: Get rid of race-conditions with shard mode checking
Example: https://github.com/nspcc-dev/neofs-node/blob/master/pkg/local_object_storage/shard/gc.go#L181
Shard mode mutex should be taken on read for the whole operation.

carpawell:
Closes #1956.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@ Changelog for NeoFS Node
 - Synchronizing a tree now longer reports an error for a single-node container (#2154)
 - Prevent leaking goroutines in the tree service (#2162)
 - Do not search for LOCK objects when delete container when session provided (#2152)
+- Race conditions on shard's mode switch (#1956)
 
 ### Removed
 - `-g` option from `neofs-cli control ...` and `neofs-cli container create` commands (#2089)
--- a/pkg/local_object_storage/shard/gc.go
+++ b/pkg/local_object_storage/shard/gc.go
@@ -187,7 +187,10 @@ func (gc *gc) stop() {
 // with GC-marked graves.
 // Does nothing if shard is in "read-only" mode.
 func (s *Shard) removeGarbage() {
-	if s.GetMode() != mode.ReadWrite {
+	s.m.RLock()
+	defer s.m.RUnlock()
+
+	if s.info.Mode != mode.ReadWrite {
 		return
 	}
 
@@ -221,7 +224,7 @@ func (s *Shard) removeGarbage() {
 	deletePrm.SetAddresses(buf...)
 
 	// delete accumulated objects
-	_, err = s.Delete(deletePrm)
+	_, err = s.delete(deletePrm)
 	if err != nil {
 		s.log.Warn("could not delete the objects",
 			zap.String("error", err.Error()),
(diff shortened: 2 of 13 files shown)


pkg/local_object_storage/shard/gc.go
```diff
@@ -214,9 +216,12 @@ func (s *Shard) removeGarbage() {
```
acid-ant, line 216: Should we unlock here too?

acid-ant (approved):

pkg/local_object_storage/shard/gc.go
carpawell: for sure. rechecked all the places

CHANGELOG.md
```diff
@@ -58,6 +58,7 @@ Changelog for NeoFS Node
 - Spawning useless `GETRANGE` with zero length for a big object (#2101)
 - Incomplete object put errors do contain the deepest error's message (#2092)
 - Prioritize internal addresses for clients (#2156)
+- Races on shard's mode switch (#1956)
```
fyrchik, line 61: It's race conditions, not races.

pkg/local_object_storage/shard/gc.go
fyrchik: Can we move all code to a separate function and use `defer` instead? It is ok to have 1-2 different `Unlock`, we have 4.

pkg/local_object_storage/shard/gc.go
```diff
@@ -284,17 +298,25 @@ func (s *Shard) collectExpiredTombstones(ctx context.Context, e Event) {
 	for {
 		log.Debug("iterating tombstones")
 
-		if s.GetMode().NoMetabase() {
+		s.m.RLock()
```
fyrchik, line 301: Same here.

CHANGELOG.md
carpawell: fixed but what's the difference?

pkg/local_object_storage/shard/tree.go
```diff
@@ -13,32 +13,44 @@ var ErrPiloramaDisabled = logicerr.New("pilorama is disabled")
...
 	}
 	return s.pilorama.TreeAddByPath(d, treeID, attr, path, meta)
 }
 
 // TreeApply implements the pilorama.Forest interface.
 func (s *Shard) TreeApply(d pilorama.CIDDescriptor, treeID string, m *pilorama.Move) error {
+	s.m.RLock()
+	defer s.m.RUnlock()
+
 	if s.pilorama == nil {
```
fyrchik, line 49: We can do this before locking a mutex.

pkg/local_object_storage/shard/gc.go
carpawell: `expiredTombstonesCallback` could call the same shard and deadlock once again

pkg/local_object_storage/shard/tree.go
carpawell: sure, just want to have the same logic as _the same as possible_ (lock in the same place (the first lines of the methods))

fixed

acid-ant (approved):

pkg/local_object_storage/shard/gc.go
fyrchik: You can split it into `delete` (no mutex) and `Delete` (take mutex + do `delete`).

carpawell: @fyrchik, agree, fixed. unfortunately, i do not know how to solve that easily for our callbacks that are `engine` level config in fact. do you have any ideas?

fyrchik (approved):

PR #2148, opened 2022-12-09
Add metric `readonly` to get shards mode

### LINKED ISSUE #2022: Metrics: Need to add metric to get shards mode
Metrics: Need to add metric to get shards state.

The goal is to be able to get information about the state of shards on storage nodes and send an alert when the state changes to degrading ones.

acid-ant:
How it looks in prometheus:
```
# HELP neofs_node_object_readonly Shards state
# TYPE neofs_node_object_readonly gauge
neofs_node_object_readonly{shard="4SVWUS1TajZ9ZwPRbUaNMY"} 1
neofs_node_object_readonly{shard="NoxKcaHtMd4rQp8DjqJiP5"} 0
```
Close #2022

### DIFF
--- a/pkg/local_object_storage/engine/shards.go
+++ b/pkg/local_object_storage/engine/shards.go
@@ -45,6 +45,10 @@ func (m *metricsWithID) DecObjectCounter(objectType string) {
 	m.mw.AddToObjectCounter(m.id, objectType, -1)
 }
 
+func (m *metricsWithID) SetReadonly(readonly bool) {
+	m.mw.SetReadonly(m.id, readonly)
+}
+
 // AddShard adds a new shard to the storage engine.
 //
 // Returns any error encountered that did not allow adding a shard.
@@ -58,6 +62,10 @@ func (e *StorageEngine) AddShard(opts ...shard.Option) (*shard.ID, error) {
 	err = e.addShard(sh)
 	if err != nil {
 		return nil, fmt.Errorf("could not add %s shard: %w", sh.ID().String(), err)
+	}
+
+	if e.cfg.metrics != nil {
+		e.cfg.metrics.SetReadonly(sh.ID().String(), sh.GetMode() != mode.ReadWrite)
 	}
 
 	return sh.ID(), nil
--- a/pkg/local_object_storage/shard/metrics_test.go
+++ b/pkg/local_object_storage/shard/metrics_test.go
@@ -10,6 +10,7 @@ import (
 	meta "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/metabase"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/pilorama"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/shard"
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/shard/mode"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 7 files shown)


pkg/local_object_storage/shard/mode.go
```diff
@@ -68,3 +69,14 @@ func (s *Shard) GetMode() mode.Mode {
 
 	return s.info.Mode
 }
+
+// UpdateReadonlyMetric updates "readonly" metric in prometheus.
+func (s *Shard) UpdateReadonlyMetric() {
```
fyrchik, line 74: Why can't we do it in `SetMode`? It is the only place where we change shard mode and we don't need any additional logic on the caller.

acid-ant: The idea is to have metric for shard mode right after the startup, not only when shard mode changed somehow.

carpawell: could it be an unexported method that is called on startup and on mode changes?

acid-ant: I've removed UpdateReadonlyMetric(), added IsReadOnly() and refactored a bit SetMode and AddShard. How about this?

pkg/local_object_storage/shard/mode.go
```diff
@@ -68,3 +71,8 @@ func (s *Shard) GetMode() mode.Mode {
 
 	return s.info.Mode
 }
+
+// IsReadOnly return true if shard not in mode ReadWrite.
+func (s *Shard) IsReadOnly() bool {
```
fyrchik, line 76: First, we should take a mutex, second we already have `GetMode`, I think it solves the job.

pkg/metrics/object.go
```diff
@@ -158,6 +159,15 @@ func newObjectServiceMetrics() objectServiceMetrics {
 		},
 			[]string{shardIDLabelKey, counterTypeLabelKey},
 		)
+
+		shardsReadonly = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Namespace: namespace,
+			Subsystem: objectSubsystem,
+			Name:      "readonly",
+			Help:      "Shards state",
```
fyrchik, line 167: `Shard state`?

pkg/local_object_storage/shard/metrics_test.go
```diff
@@ -50,6 +50,10 @@ func (m metricsStore) DecObjectCounter(objectType string) {
 	m.AddToObjectCounter(objectType, -1)
 }
 
+func (m metricsStore) SetShardReadonly(readonly bool) {
+	m.SetShardReadonly(readonly)
```
fyrchik, line 54: Infinite recursion here? Also let's make it `ReadOnly` -- the same way it is done in `mode` package.

pkg/metrics/object.go
acid-ant: Fixed.

pkg/local_object_storage/shard/mode.go
acid-ant: Refactored.

pkg/local_object_storage/shard/metrics_test.go
acid-ant: Thanks, fixed and test updated.

pkg/local_object_storage/engine/shards.go
```diff
@@ -60,6 +64,10 @@ func (e *StorageEngine) AddShard(opts ...shard.Option) (*shard.ID, error) {
 		return nil, fmt.Errorf("could not add %s shard: %w", sh.ID().String(), err)
 	}
 
+	if e.cfg.metrics != nil {
+		e.cfg.metrics.SetReadonly(sh.ID().String(), sh.GetMode() != mode.ReadWrite)
```
fyrchik, line 68: Let's use `GetMode().ReadOnly()` -- we don't use `mode.Degraded` at all, but if we want, we shouldn't report `readonly`, another metric can be done. But again, `Degraded` is not used.

fyrchik: Ok, we use this for alerting, let's postpone it.

PR #2158, opened 2022-12-14
policer: Reduce default cache size

fyrchik:
We use cache to avoid policing the same object multiple times in a short time span (< 30 seconds). If we have 200_000 objects in a blobstor, it is a bit useless -- if it takes 1 second to process an object and we have `replicator.pool_size: 20` in config, the next iteration will happen in 10_000 second which is much larger than 30 second. However we still consume a lot of memory, so it makes sense to use saner default.


carpawell (approved):
Changelog?

Issue #2154, opened 2022-12-14
if a container occupies one node only, the error: could not sync trees, no nodes to fetch trees from


ikovalev1:
## Expected Behavior
if a container occupies only one node, no errors in log

## Current Behavior
Dec 14 08:07:04 buky neofs-node[17310]: 2022-12-14T08:07:04.444Z        error        tree/sync.go:284        could not sync trees        {"cid": "E27yucM4TaTwzpsdC6PZAPAKKA8RDcMRcw5puWELJ5kT", "error": "no nodes to fetch trees from"}

## Steps to Reproduce (for bugs)
create container with policy CBF 1 SELECT 1 FILTER Node=node1 IP
```
neofs-cli container create -r node1:8080 -w wallet.json --policy "REP 1 IN X CBF 1 SELECT 1 FROM N1 AS X FILTER Node EQ '172.26.160.190' AS N1 " --await
container ID: CkjVs3CnXs4ALeEn2sfqeg2NnpfXBtaoZrtRYDJXjUQw
```
tick epoch
watch errors in the neofs-storage log
```
journalctl -fu neofs-storage
Dec 14 09:07:23 az neofs-node[17341]: 2022-12-14T09:07:23.111Z        error        tree/sync.go:284        could not sync trees        {"cid": "CkjVs3CnXs4ALeEn2sfqeg2NnpfXBtaoZrtRYDJXjUQw", "error": "no nodes to fetch trees from"}
```

## Your Environment
```
# neofs-node --version
NeoFS Storage node
Version: v0.34.0-103-gcb43a6ae-dirty
GoVersion: go1.18.4
```

Issue #2146, opened 2022-12-09
Make tree synchronization more robust


fyrchik:
Discuss possible approaches for #2127 .

fyrchik:
We could remember the last height of each synchronized tree per node.
This height is used as a starting point for the next cycle.
If a node is unavailable (split-brain scenario), the height is not increased.

fyrchik:
Closed via #2165 .

PR #2161, opened 2022-12-15
Randomize tree synchronization

### LINKED ISSUE #2127: Tree-service sync after splitbrain failover
Incorrect tree-service sync after splitbrain failover.

Synchronization occurs only from one node. 
In this case, the node could synchronize data with a node from the same part of the cluster. 
The order for the container is fixed.

Need to fix this:
- get not fixed node to sync (can be selected sequentially or in random)
- sync not only after new epoch event

fyrchik:
Close #2127, #2154 .


carpawell (commented):
Does it really closes #2127? Changelog

carpawell (approved):

fyrchik:
@carpawell Yes, there is another task for more complex solution #2146.

acid-ant (approved):

PR #2162, opened 2022-12-16
services/tree: close connection after the syncronization

fyrchik:
There was a goroutine leak here.

No need to test.


carpawell (approved):
The change

acid-ant (approved):

PR #2165, opened 2022-12-19
Optimize tree synchronization

fyrchik:
1. Allow to set custom synchronization interval.
2. Remember the last successful synchronization height to better handle split-brain scenario.


pkg/services/tree/sync.go
```diff
@@ -85,15 +86,36 @@ func (s *Service) synchronizeAllTrees(ctx context.Context, cid cid.ID) error {
...
+
 	for _, tid := range treesToSync {
-		err = s.synchronizeTree(ctx, d, tid, nodes)
+		h, err := s.synchronizeTree(ctx, d, syncStatus[tid], tid, nodes)
 		if err != nil {
 			s.log.Error("could not sync tree",
 				zap.Stringer("cid", cid),
 				zap.String("treeID", tid))
 		}
+		if syncStatus[tid] < h {
```
carpawell, line 110: can we remember `syncStatus[tid]` in a local var and not read `map` twice?

pkg/services/tree/sync.go
```diff
@@ -118,18 +140,20 @@ func (s *Service) SynchronizeTree(ctx context.Context, cid cid.ID, treeID string
...
 }
 
-func (s *Service) synchronizeTree(ctx context.Context, d pilorama.CIDDescriptor,
-	treeID string, nodes []netmapSDK.NodeInfo) error {
-	lm, err := s.forest.TreeGetOpLog(d.CID, treeID, 0)
-	if err != nil && !errors.Is(err, pilorama.ErrTreeNotFound) {
-		return err
-	}
+func (s *Service) synchronizeTree(ctx context.Context, d pilorama.CIDDescriptor, from uint64,
+	treeID string, nodes []netmapSDK.NodeInfo) (uint64, error) {
```
carpawell, line 148: `error` looks useless now, i would remove it for readability

pkg/services/tree/sync.go
```diff
@@ -303,12 +335,20 @@ func (s *Service) syncLoop(ctx context.Context) {
...
+			s.cnrMapMtx.Lock()
+			var removed []cid.ID
 			for cnr := range s.cnrMap {
 				if _, ok := newMap[cnr]; ok {
 					continue
 				}
+				removed = append(removed, cnr)
+			}
+			for i := range removed {
+				delete(s.cnrMap, removed[i])
```
carpawell, line 347: we have one more `delete` without mutex above

pkg/local_object_storage/pilorama/interface.go
```diff
@@ -17,7 +17,8 @@ type Forest interface {
 	// Internal nodes in path should have exactly one attribute, otherwise a new node is created.
 	TreeAddByPath(d CIDDescriptor, treeID string, attr string, path []string, meta []KeyValue) ([]LogMove, error)
 	// TreeApply applies replicated operation from another node.
-	TreeApply(d CIDDescriptor, treeID string, m *Move) error
+	// If background is true, TreeApply will first check whether an operation exists.
+	TreeApply(d CIDDescriptor, treeID string, m *Move, background bool) error
```
carpawell, line 21: are there any pros why `background` is better than `sync`?

cmd/neofs-node/tree.go
```diff
@@ -63,12 +64,29 @@ func initTreeService(c *cfg) {
...
-	})
+	if d := treeConfig.SyncInterval(); d == 0 {
+		addNewEpochNotificationHandler(c, func(_ event.Event) {
+			err := c.treeService.SynchronizeAll()
+			if err != nil {
+				c.log.Error("could not synchronize Tree Service", zap.Error(err))
+			}
+		})
+	} else {
+		go func() {
```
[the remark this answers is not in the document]
fyrchik: We have non-blocking send there and the channel is created in `new`, so there is no problem here besides possibly logging incorrect error.

pkg/services/tree/sync.go
fyrchik: It is more complex IMO and it won't be any faster because network requests take a lot more time to execute.

pkg/services/tree/sync.go
fyrchik: Fixed

pkg/local_object_storage/pilorama/interface.go
fyrchik: `sync` is too generic, renamed to `backgroundSync`

pkg/services/tree/sync.go
fyrchik: Fixed

carpawell (approved):

PR #2166, opened 2022-12-20
go.mod: Update dependencies

fyrchik:
Mostly about vulnerabilities in dependencies:
1. Cobra now uses yaml.v3
2. neo-go now doesn't use `btcd`.


acid-ant (approved):

roman-khimov (approved):
We still have yaml.v2 via some indirect deps though.

fyrchik:
@roman-khimov it is used in `viper` only if `viper_yaml2` build flag is provided.

Issue #2099, opened 2022-11-24
Simple object with lock and expiration time not removed after locks are expired


abereziny:
Let say we have:
1. Current epoch is `100`
2. Simple object with `__NEOFS__EXPIRATION_EPOCH=101`
3. Simple object is locked with lock until epoch `103`

Steps:
Tick epochs until current epoch = 103
Check object availability

## Expected Behavior
Simple object should be removed since locks are gone and object expired

## Current Behavior
Simple object continues to present in storage even if we wait additional extra epochs

## Tests
in branch [feature-abereziny-add-lock-tests-2](https://github.com/abereziny/neofs-testcases/tree/feature-abereziny-add-lock-tests-2)

- pytest_tests/testsuites/object/test_object_lock.py#test_expired_object_should_be_deleted_after_locks_are_expired

acid-ant:
Blocked by:
- https://github.com/nspcc-dev/neofs-api/issues/253
- https://github.com/nspcc-dev/neofs-sdk-go/issues/370

acid-ant:
it works as expected. Fixed help message for parameter here #2169. I've added one extra epoch tick and test passed.

abereziny:
Currently it's not working as expected because objects got removed despite of locks
![image](https://user-images.githubusercontent.com/116558828/209701901-8001d472-ffe7-43fe-bfe8-b6b1d8bc0438.png)

acid-ant:
@abereziny it works as expected. It is ok to get `not found` error in your case, because expiration epoch in lock blocks any action with data, but do not affect presence of the object on file system. Now it is impossible to test.
In other words, it is possible create object with expiration epoch N and lock with M where M > N.

abereziny:
@acid-ant I do not agree with that because current behavior have no real practical use case.
Why do we need to lock object which will be unavailable? What's the point?

for example: we have document with expiration, court need this document in the process so it must not be deleted and must be available until process is complete, so lock is assigned. If the document is not available - it practically means that lock doesn't work from user point of view.

abereziny:
Logged and issue for API to further elaboration

roman-khimov:
@evgeniiz321, do we have any test scenario like this?

evgeniiz321:
[...]

PR #2169, opened 2022-12-21
Clarify help for `--expire-at` parameter for commands `object lock/put` and `bearer create`

### LINKED ISSUE #2097: Locked object can be deleted on one extra epoch away from lock expiration
Let say we have:
1. Current epoch is `100`
2. Object is put to the container
3. Object is locked with lock with expiration at epoch `101`

Steps:
1. Tick epoch, current epoch becomes `101`
2. Try to delete object

## Expected Behavior
Object should be removed since lock expired at current epoch

## Current Behavior
Object cannot be removed and treated as locked object until next extra epoch is ticked

## Tests
- [pytest_tests/testsuites/object/test_object_lock.py#test_after_lock_expiration_with_expire_at_user_should_be_able_to_delete_object](https://github.com/nspcc-dev/neofs-testcases/blob/master/pytest_tests/testsuites/object/test_object_lock.py#L506)
- [pytest_tests/testsuites/object/test_object_lock.py#test_after_lock_expiration_with_lifetime_user_should_be_able_to_delete_object](https://github.com/nspcc-dev/neofs-testcases/blob/master/pytest_tests/testsuites/object/test_object_lock.py#L464)

acid-ant:
Closes #2097 and #2099

### DIFF
--- a/cmd/neofs-cli/modules/object/lock.go
+++ b/cmd/neofs-cli/modules/object/lock.go
@@ -66,7 +66,7 @@ var objectLockCmd = &cobra.Command{
 			exp = currEpoch + lifetime
 		}
 
-		common.PrintVerbose("Lock object will expire at %d epoch", exp)
+		common.PrintVerbose("Lock object will expire after %d epoch", exp)
 
 		var expirationAttr objectSDK.Attribute
 		expirationAttr.SetKey(objectV2.SysAttributeExpEpoch)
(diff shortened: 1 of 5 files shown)


carpawell (commented):
Could you recheck all `expire-at` flag descriptions? It must mean the same for all the entities (objects, storage groups, locks, tokens, etc)

cmd/neofs-cli/modules/object/lock.go
```diff
@@ -104,7 +104,7 @@ func initCommandObjectLock() {
 	ff.StringSlice(commonflags.OIDFlag, nil, commonflags.OIDFlagUsage)
 	_ = objectLockCmd.MarkFlagRequired(commonflags.OIDFlag)
 
-	ff.Uint64P(commonflags.ExpireAt, "e", 0, "Lock expiration epoch")
+	ff.Uint64P(commonflags.ExpireAt, "e", 0, "Lock expires after that epoch")
```
fyrchik, line 107: Can we just write `The last active epoch of the lock`, I mean `expire-at` is somewhat self-descriptive, we could clarify further.

acid-ant: Updated.

carpawell (approved):

fyrchik (approved):

PR #2171, opened 2022-12-21
Fix/`GETRANGEHASH` with session

### LINKED ISSUE #2144: RANGEHASH with static session sometimes can't find object
Steps:
1. Have 2 wallets: owner and user
2. Create container using owner wallet
```
neofs-cli --config owner_wallet_config.yml container create --rpc-endpoint '<endpoint>' --wallet owner_wallet.json --await --policy 'REP 2 IN X CBF 1 SELECT 4 FROM * AS X'

STDOUT:
container ID: 4jxFyPEzcEAz2wNcMSz3SUDdQvcaDy7z2UeooR2Asu8M
```

4. Put object to a node to a container using owner wallet 
```
neofs-cli  --config owner_wallet_config.yml object put --rpc-endpoint '<endpoint>' --wallet owner_wallet.json --cid '4jxFyPEzcEAz2wNcMSz3SUDdQvcaDy7z2UeooR2Asu8M' --file '1000.txt' --no-progress

STDOUT:
Object successfully stored
  OID: APQPiqPdtKPZWhfs8D5W5uYkFrRcfkYZMc3nQenXCD4V
  CID: 4jxFyPEzcEAz2wNcMSz3SUDdQvcaDy7z2UeooR2Asu8M
```
5. Create and sign static session for user wallet with RANGEHASH verb for this container and this object (note base58 -> base64 conversions)
6. Do rangehash request
```
neofs-cli --config user_wallet_config.yml object hash --rpc-endpoint '<endpoint>' --wallet 'user_wallet.json' --cid '4jxFyPEzcEAz2wNcMSz3SUDdQvcaDy7z2UeooR2Asu8M' --oid 'APQPiqPdtKPZWhfs8D5W5uYkFrRcfkYZMc3nQenXCD4V' --range '0:10' --session 'signed_session_token.json'

```

## Expected Behavior
Rangehash returns hash of the object

## Current Behavior
Sometimes object is not found
```
Error:
return code: 1
output: rpc error: read payload hashes via client: status: code = 2049 message = object not found
```

## Tests
https://github.com/nspcc-dev/neofs-testcases/blob/master/pytest_tests/testsuites/session_token/test_static_object_session_token.py#L192

carpawell:
Closes #2144.

Could not come up with a better than [comment](https://github.com/nspcc-dev/neofs-node/issues/2144#issuecomment-1357759999) solution. If fact, i was surprised that `GETRANGEHASH` spawns `GETRANGE` requests.

I think we can localize all the places when a node signs a new request to minimize such unexpected bugs.

### DIFF
--- a/pkg/services/object/get/exec.go
+++ b/pkg/services/object/get/exec.go
@@ -109,6 +109,12 @@ func (exec execCtx) isChild(obj *objectSDK.Object) bool {
 }
 
 func (exec execCtx) key() (*ecdsa.PrivateKey, error) {
+	if exec.prm.signerKey != nil {
+		// the key has already been requested and
+		// cached in the previous operations
+		return exec.prm.signerKey, nil
+	}
+
 	var sessionInfo *util.SessionInfo
 
 	if tok := exec.prm.common.SessionToken(); tok != nil {
--- a/pkg/services/object/get/v2/util.go
+++ b/pkg/services/object/get/v2/util.go
@@ -362,6 +362,23 @@ func (s *Service) toHashRangePrm(req *objectV2.GetRangeHashRequest) (*getsvc.Ran
 
 	p.WithAddress(addr)
 
+	if tok := commonPrm.SessionToken(); tok != nil {
+		signerKey, err := s.keyStorage.GetKey(&util.SessionInfo{
+			ID:    tok.ID(),
+			Owner: tok.Issuer(),
+		})
+		if err != nil && errors.As(err, new(apistatus.SessionTokenNotFound)) {
+			commonPrm.ForgetTokens()
+			signerKey, err = s.keyStorage.GetKey(nil)
+		}
+
+		if err != nil {
+			return nil, fmt.Errorf("fetching session key: %w", err)
+		}
+
+		p.WithCachedSignerKey(signerKey)
+	}
+
 	rngsV2 := body.GetRanges()
 	rngs := make([]object.Range, len(rngsV2))
 
--- a/pkg/services/object/util/key.go
+++ b/pkg/services/object/util/key.go
@@ -51,10 +51,13 @@ type SessionInfo struct {
 	Owner user.ID
 }
 
-// GetKey returns private key of the node.
+// GetKey fetches private key depending on the SessionInfo.
 //
-// If token is not nil, session private key is returned.
-// Otherwise, node private key is returned.
+// If info is not `nil`, searches for dynamic session token through the
+// underlying token storage. Returns apistatus.SessionTokenNotFound if
+// token storage does not contain information about provided dynamic session.
+//
+// If info is `nil`, returns node's private key.
 func (s *KeyStorage) GetKey(info *SessionInfo) (*ecdsa.PrivateKey, error) {
 	if info != nil {
 		binID, err := info.ID.MarshalBinary()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ Changelog for NeoFS Node
 - Use `sync.Pool` in Object.PUT service (#2139)
 - Shard uses metabase for `HEAD` requests by default, not write-cache (#2167)
 - Clarify help for `--expire-at` parameter for commands `object lock/put` and `bearer create` (#2097)
+- Node spawns `GETRANGE` requests signed with the node's key if session key was not found for `RANGEHASH` (#2144) 
 
 ### Fixed
 - Open FSTree in sync mode by default (#1992)
--- a/pkg/services/object/get/prm.go
+++ b/pkg/services/object/get/prm.go
@@ -1,6 +1,7 @@
 package getsvc
 
 import (
+	"crypto/ecdsa"
 	"errors"
 	"hash"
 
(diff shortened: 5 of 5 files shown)


pkg/services/object/get/exec.go
```diff
@@ -118,7 +119,16 @@ func (exec execCtx) key() (*ecdsa.PrivateKey, error) {
...
 	}
 
-	return exec.svc.keyStore.GetKey(sessionInfo)
+	key, err := exec.svc.keyStore.GetKey(sessionInfo)
+	if err != nil && errors.As(err, new(apistatus.SessionTokenNotFound)) {
+		// a new request with session token should be
+		// spawned but no session token has been found
+		// so try to use node's key as the best effort
+		exec.prm.common.ForgetTokens()
+		key, err = exec.svc.keyStore.GetKey(nil)
```
roman-khimov, line 128: I don't see this logic being limited to some specific types of requests. And it makes the node sign the request with its own key in some cases. Can this be misused (specifically triggering the node to sign some requests)?

acid-ant (approved):

pkg/services/object/get/exec.go
fyrchik: Can we make it for range-hash only? Just move this check in `GetRangeHash`.

Yes, it looks ugly, but the problem is deep -- we have 4 types of requests being processed by the same code. And in each function we are switching and doing different things based on the request type (actually, returning `Object` for `GetRange` and making `GetRange` instead of `GetRangeHash` are two most obvious examples). I am doing some simplification refactoring currently, will create a PR after release.

fyrchik: To clarify, there is no problem for other requests because they can setup request forwarder.

roman-khimov: They probably can, but I'm more concerned about situations when they deliberately do not do that and then have (or not?) the node using its key to sign something. Can this be leveraged for some privilege escalation when the request originator doesn't have appropriate permission while the node does?

carpawell: @roman-khimov, @fyrchik, yes, the fix isn't perfect and yes, the problem is deeper than just allowing static sessions for range requests, requires a refactor and/or an arch revision.

> I don't see this logic being limited to some specific types of requests.

> Can we make it for range-hash only?

@fyrchik, @roman-khimov yes, we can but what for? if `func (exec execCtx) key()` can't get the session token, attaching tokens is useless for _any_ request _the node_ signs (ok, except the case when a requester attached a _static_ session _with a node_, could not imagine when that is possible). We either fail here or try to do our best (the test from the issue asks us to fix it now).

> Can this be misused (specifically triggering the node to sign some requests)?

@roman-khimov, well, it could be used somehow i guess but can not imagine any practical case: the changes are for "get" requests only, and new requests are spawned only when a node tries to fetch an object (that means that a node has already agreed about a request (the original one, with the original tokens and the original requester) and allows it in terms of our (e)ACL rules).

fyrchik: `SessionNotFound` error is much more descriptive than `ObjectNotFound` or `ObjectAccessDenied`. Also, we do not spawn unnecessary network requests in this case.

carpawell: ok, moved

pkg/services/object/util/key.go
```diff
@@ -51,10 +51,13 @@ type SessionInfo struct {
 	Owner user.ID
 }
 
-// GetKey returns private key of the node.
+// GetKey fetches private key.
```
fyrchik, line 54: `of the node`?

pkg/services/object/get/v2/util.go
```diff
@@ -362,6 +362,23 @@ func (s *Service) toHashRangePrm(req *objectV2.GetRangeHashRequest) (*getsvc.Ran
...
 	p.WithAddress(addr)
 
+	if tok := commonPrm.SessionToken(); tok != nil {
+		signerKey, err := s.keyStorage.GetKey(&util.SessionInfo{
+			ID:    tok.ID(),
+			Owner: tok.Issuer(),
+		})
+		if err != nil && errors.As(err, new(apistatus.SessionTokenNotFound)) {
+			commonPrm.ForgetTokens()
+			signerKey, err = s.keyStorage.GetKey(nil)
```
fyrchik, line 372: Do we need it if we use `ForgetTokens`?

pkg/services/object/util/key.go
carpawell: it is not the node's key if a dynamic session token is used (`info != nil`)

pkg/services/object/get/v2/util.go
carpawell: not necessary but more readable IMO: try to solve the signer key in the one place for `RANGEHASH` requests

pkg/services/object/util/key.go
fyrchik: It is just that `fetches private key` is obvious from the function signature, can we clarify somehow?

carpawell: changed to `// GetKey fetches private key depending on the SessionInfo.`

fyrchik (approved):

PR #2176, opened 2022-12-26
[#2152] neofs-node: Update container list cache properly

### LINKED ISSUE #2052: When a container is created, it appears in the listing before it is available
The creation of the container takes time. At the moment of creation we already get the id of the container in the listing, but command `get --cid` is not available since the container has not been created yet.

Is it ok that we already see the container in the listing, but cannot get it?

In my case, it happens in [panel-fs-neo-org](https://github.com/nspcc-dev/panel-fs-neo-org): the user creates a container, reloads the page and sees an error. It repeats every time, because creation container takes approximately 20 sec to create a container.

### LINKED ISSUE #2152: Not working static session token for delete container operation
Delete container operation returned code "2" with stderr output: code = 2048

```
neofs-cli --config wallet_config.yml container delete --rpc-endpoint 'node1:8081' --wallet 'wallet.json' --cid 'B1pGbB9XttGntTSDtnVdk7ugeepUScm8jb8Pecr3EryX' --session '99e660d6-12e8-4890-91c2-e4de9177374b'
Error:
return code: 2
output: can't search for LOCK objects: read object list: status: code = 2048 message = access to object operation denied: access to operation OBJECT_SEARCH is denied by basic ACL check
```

## Expected Behavior
Container delete operation completed successfully

## Current Behavior
Container delete operation failed

## Autotest
Validate static session with delete operation (test_static_session_token_container_delete)
See details in last Allure report

1. Generate Session Token For Container ({'container': {'verb': 'DELETE', 'wildcard': False}})
2. Sign Session Token
3. Create container
4. Other user try delete container with static session token

## Environment
neo-go 1.0.0
neofs-authmate 0.25.0-18-g92a5e092-dirty
neofs-cli 0.34.0-103-gcb43a6ae-dirty
neofs-adm 0.34.0-103-gcb43a6ae-dirty

### LINKED ISSUE #2155: on create container the "error": "status: code = 3072 message = container not found
create container
```
neofs-cli container create -r node1:8080 -w wallet.json --policy "REP 4 IN X  SELECT 4 FROM * AS X" --await
Enter password >
container ID: 8XLU5iSNayA5UNEkMysC6GUz46n9YU1cLnLyUV83tMZ4
```
watch log
```
Dec 14 09:58:13 az neofs-node[17341]: 2022-12-14T09:58:13.185Z        error        neofs-node/container.go:92        read newly created container after the notification        {"id": "8XLU5iSNayA5UNEkMysC6GUz46n9YU1cLnLyUV83tMZ4", "error": "status: code = 3072 message = container not found"}
```

## Your Environment
```
neofs-node --version
NeoFS Storage node
Version: v0.34.0-103-gcb43a6ae-dirty
GoVersion: go1.18.4
```

fyrchik:
Close #2152 .
Close #2155. (?) @acid-ant 
Close #2052. (?)

Update the list of all containers if needed.

### DIFF
--- a/cmd/neofs-node/cache.go
+++ b/cmd/neofs-node/cache.go
@@ -314,7 +314,7 @@ func (s *ttlContainerLister) List(id *user.ID) ([]cid.ID, error) {
 func (s *ttlContainerLister) update(owner user.ID, cnr cid.ID, add bool) {
 	strOwner := owner.EncodeToString()
 
-	val, ok := (*ttlNetCache)(s).cache.Get(strOwner)
+	val, ok := (*ttlNetCache)(s).cache.Peek(strOwner)
 	if !ok {
 		// we could cache the single cnr but in this case we will disperse
 		// with the Sidechain a lot
--- a/cmd/neofs-node/cache.go
+++ b/cmd/neofs-node/cache.go
@@ -321,9 +321,14 @@ func (s *ttlContainerLister) update(owner user.ID, cnr cid.ID, add bool) {
 		return
 	}
 
+	vt := val.(*valueWithTime)
+	if s.ttl <= time.Since(vt.t) {
+		return
+	}
+
 	// panic on typecast below is OK since developer must be careful,
 	// runtime can do nothing with wrong type occurrence
-	item := val.(*valueWithTime).v.(*cacheItemContainerList)
+	item := vt.v.(*cacheItemContainerList)
 
 	item.mtx.Lock()
 	{
--- a/cmd/neofs-node/cache.go
+++ b/cmd/neofs-node/cache.go
@@ -236,7 +236,10 @@ func (s *lruNetmapSource) Epoch() (uint64, error) {
 
 // wrapper over TTL cache of values read from the network
 // that implements container lister.
-type ttlContainerLister ttlNetCache
+type ttlContainerLister struct {
+	inner  ttlNetCache
+	client *cntClient.Client
+}
 
 // value type for ttlNetCache used by ttlContainerLister.
 type cacheItemContainerList struct {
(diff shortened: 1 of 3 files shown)


acid-ant (approved):

cmd/neofs-node/cache.go
```diff
@@ -312,41 +312,41 @@ func (s *ttlContainerLister) List(id *user.ID) ([]cid.ID, error) {
 // All described race cases aren't critical since cache values expire anyway, we just try
 // to increase cache actuality w/o huge overhead on synchronization.
 func (s *ttlContainerLister) update(owner user.ID, cnr cid.ID, add bool) {
-	strOwner := owner.EncodeToString()
+	for _, key := range []string{"", owner.EncodeToString()} {
```
roman-khimov, line 315: `""` for "owner-less" containers? Why in this order then? I'd expect a check for direct `owner` hit first and a fallback then.

fyrchik: Actually, we need both. `continue` inside the cycle is just for the case when we do not cache the result.

roman-khimov: Can `""` case grow a lot in some cases? What if a container is cached is owner-less, but then is added with an owner? Something's fishy here.

fyrchik: Nothing bad happens -- both lists are completely independent.
I have removed caching for the full-list because on close inspection we do not really use it that often.

roman-khimov (approved):
Looks a little more risky this way, but if it solves the problem and `List(nil)` is once-per-epoch thing then it should be OK.

cmd/neofs-node/cache.go
```diff
@@ -314,16 +315,21 @@ func (s *ttlContainerLister) List(id *user.ID) ([]cid.ID, error) {
 func (s *ttlContainerLister) update(owner user.ID, cnr cid.ID, add bool) {
 	strOwner := owner.EncodeToString()
 
-	val, ok := (*ttlNetCache)(s).cache.Get(strOwner)
+	val, ok := s.inner.cache.Peek(strOwner)
```
carpawell, line 318: what does `Get` -> `Peek` change solve?

carpawell (approved):

cmd/neofs-node/cache.go
fyrchik: `Peek` is faster (read-only) and does not update the position of an item in LRU cache -- which is what we do here, we don't _use_ anything, just passively update.

PR #2182, opened 2022-12-27
cli: `common.PrintVerbose` prints via `cobra.Command.Printf`

### LINKED ISSUE #1962: cli: `PrintVerbose` writes to stdout instead of command out
https://github.com/nspcc-dev/neofs-node/blob/41f93dcc1dea5001a5674c37847fdca4d6280ab1/cmd/neofs-cli/internal/common/verbose.go#L18
Is it expected? I think we should write all the data to commands' streams.

acid-ant:
Close #1962

### DIFF
--- a/cmd/neofs-cli/internal/client/sdk.go
+++ b/cmd/neofs-cli/internal/client/sdk.go
@@ -21,25 +21,25 @@ var errInvalidEndpoint = errors.New("provided RPC endpoint is incorrect")
 // GetSDKClientByFlag returns default neofs-sdk-go client using the specified flag for the address.
 // On error, outputs to stderr of cmd and exits with non-zero code.
 func GetSDKClientByFlag(cmd *cobra.Command, key *ecdsa.PrivateKey, endpointFlag string) *client.Client {
-	cli, err := getSDKClientByFlag(key, endpointFlag)
+	cli, err := getSDKClientByFlag(cmd, key, endpointFlag)
 	if err != nil {
 		common.ExitOnErr(cmd, "can't create API client: %w", err)
 	}
 	return cli
 }
 
-func getSDKClientByFlag(key *ecdsa.PrivateKey, endpointFlag string) (*client.Client, error) {
+func getSDKClientByFlag(cmd *cobra.Command, key *ecdsa.PrivateKey, endpointFlag string) (*client.Client, error) {
 	var addr network.Address
 
 	err := addr.FromString(viper.GetString(endpointFlag))
 	if err != nil {
 		return nil, fmt.Errorf("%v: %w", errInvalidEndpoint, err)
 	}
-	return GetSDKClient(key, addr)
+	return GetSDKClient(cmd, key, addr)
 }
 
 // GetSDKClient returns default neofs-sdk-go client.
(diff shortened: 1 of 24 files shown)


cmd/neofs-cli/internal/client/sdk.go
```diff
@@ -69,7 +69,7 @@ func GetSDKClient(key *ecdsa.PrivateKey, addr network.Address) (*client.Client,
 }
 
 // GetCurrentEpoch returns current epoch.
-func GetCurrentEpoch(ctx context.Context, endpoint string) (uint64, error) {
+func GetCurrentEpoch(cmd *cobra.Command, ctx context.Context, endpoint string) (uint64, error) {
```
roman-khimov, line 72: `revive` linter is not happy about this change.

acid-ant: Thanks, updated.

cmd/neofs-cli/modules/util/convert_eacl.go
```diff
@@ -58,7 +58,7 @@ func convertEACLTable(cmd *cobra.Command, _ []string) {
 func prettyPrintJSON(cmd *cobra.Command, data []byte) {
```
roman-khimov, line 58: BTW, can this be replaced with `common.PrettyPrintJSON()`?

cmd/neofs-cli/modules/container/get.go
```diff
@@ -79,12 +79,12 @@ func prettyPrintContainer(cmd *cobra.Command, cnr container.Container, jsonEncod
 	if jsonEncoding {
```
roman-khimov, line 79: And this one too?

roman-khimov (commented):
On one hand it's OK since it solves the problem. On the other I'm think about changing `common` to `interactive` or `cmdctx` of some kind with `interactive.Command` (or similar) type that would alias `*cobra.Command` and implement `PrettyPrintJSON`, `PrintVerbose`, `ReadEACL` and all the other methods. Maybe it'd be easier to instantiate this new type once then use it instead of adding `cmd` everywhere, because it looks to me that having a number of print calls in a function (chain of functions) is a more popular scenario than a single occasional print.

Just a suggestion though.

fyrchik:
@roman-khimov could you elaborate a bit? We cannot define methods on non-internal type _aliases_.
So do you mean something like this:
```
package interactive
type Command cobra.Command // and define all methods we need
type Command struct { // and define only additional methods
  *cobra.Command
}
```
and then refactoring all functions to make use of this new type?

roman-khimov:
> type Command cobra.Command // and define all methods we need

More like this one (yes, technically it's not an alias, but a separate type). At least until we have some need for a real `struct`.

cmd/neofs-cli/modules/util/convert_eacl.go
acid-ant: Done.

cmd/neofs-cli/modules/container/get.go
acid-ant: Yes, replaced.

carpawell (approved):
Solves the original issue, all the comments could be solved via a separate PR, IMO.

PR #2197, opened 2023-01-11
Close pilorama in degraded mode

fyrchik:

### DIFF
--- a/pkg/local_object_storage/pilorama/boltdb.go
+++ b/pkg/local_object_storage/pilorama/boltdb.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/nspcc-dev/neo-go/pkg/io"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/shard/mode"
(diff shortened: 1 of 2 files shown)


acid-ant (approved):

carpawell:
Tests+changelog.

pkg/local_object_storage/pilorama/boltdb.go
```diff
@@ -71,12 +78,9 @@ func (t *boltForest) SetMode(m mode.Mode) error {
 	if t.mode == m {
 		return nil
 	}
-	if t.mode.ReadOnly() == m.ReadOnly() {
-		return nil
-	}
 
 	err := t.Close()
```
carpawell, line 82: not related to that PR but why is it the `t` receiver for `boltForest`? i thought that i was looking at a test code for a second

carpawell (approved):

pkg/local_object_storage/pilorama/boltdb.go
fyrchik: `t` for `forest of Trees` (initially it was all about `trees`, so `t` stuck)

PR #2208, opened 2023-01-23
Improve FSTree behaviour for partially written objects

fyrchik:
1. Partially written object is deleted.
2. First write to a temporary file to avoid partially written files during hard-reset

Also, log time-consuming operations to help administrator observe system behaviour until we have a proper async operations support in `control` service.


acid-ant (approved):

roman-khimov (commented):
Good, but something should clean up `f#` files on startup or as some additional maintenance command.

PR #2211, opened 2023-01-24
Concatenate NeoFS Inner Ring node with Neo Go node

cthulhu-rider:
## Done
Neo (Go) blockchain API is fine-grained and placed into `internal/blockchain` package. This simplifies code development/support. Inner Ring specifics are implemented through internal package only (except Neo Go core types and utility packages).

Contract processors handle only "distilled" notifications from now (event parameters' types have been strengthened). Processors return processing result of the single event w/o work scheduling via worker pools (this logic is concentrated on app-side). Event handlers are divided into 2 kinds:

- with approved/denied outcome which leads to signing of the original event/request
- producing some new data

NeoFS business logic is concentrated on processors' side while blockchain mechanics - on app side,

## TBD
All other processors should be adopted to new code arch. Hacks with RPC to node itself should be eliminated.

Load balancer of event handling should returned or reconsidered. In previous implementation we had per-contract pool with finite number of workers. Each contract event was submitted to the corresponding pool and handled by dedicated worker asynchronously. Overall it sounds OK, however, async processing breaks the property of the ordering of events, which in theory can lead to unforeseen outcomes. Example: modification of the same resource such as container eACL (`SetEACL`). Lets think about it.

Mainchain connection is presented in two ways:

1. RPC client when Mainchain is separated
2. as Sidechain when Mainchain is "disabled"

It would be nice to fish out same interface of the Mainchain and implement it for both cases.

Consider typedef of notification events in [NeoFS Contracts](https://github.com/nspcc-dev/neofs-contract).
```
type ContainerCreationRequestNotification struct {
	Container []byte
	Signature []byte
	Session   []byte
}
```
and notify like this [1]:
```
var n ContainerCreationRequestNotification
// set fields

// notify with fields layout
runtime.Notify("containerPut", n.Container, n.Signature, n.Session)
```
or like this [2]:
```
var n ContainerCreationRequestNotification
// set fields

// notify with fields layout
runtime.Notify("containerPut", n)
```
[1] may lead to GAS excess (if so how mush?) while [2] isn't backward compatible.


pkg/innerring/node.go
```diff
@@ -0,0 +1,354 @@
...
+	tx.Nonce = binary.BigEndian.Uint32(withdrawTx[:4])
+	tx.ValidUntilBlock = withdrawHeight + 50
+	tx.Script = callRes.Script
+	tx.Attributes = []transaction.Attribute{{
+		Type: transaction.NotaryAssistedT,
+		Value: &transaction.NotaryAssisted{
+			NKeys: uint8(len(alphabetKeys)),
+		},
+	}}
+	tx.Scripts = []transaction.Witness{
```
roman-khimov, line 313: Unsatisfied with the notary actor interface?

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,608 @@
...
+	// 		return fmt.Errorf("init Oracle module: %w", err)
+	// 	}
+	//
+	// 	artifacts.blockchain.SetOracle(oracleService)
+	// 	artifacts.netServer.AddService(oracleService)
+	// }
+
+	var consensusService consensus.Service
+
+	if cfg.ApplicationConfiguration.Consensus.Enabled {
```
roman-khimov, line 171: Not really needed (the flag and unlock logic), you know when your IR is an alphabet node and CN uses the same key.

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,608 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("init Consensus module: %w", err)
+		}
+
+		netServer.AddConsensusService(consensusService, consensusService.OnPayload, consensusService.OnTransaction)
+	}
+
+	var notaryService *notarysvc.Notary
+
+	if cfg.ApplicationConfiguration.P2PNotary.Enabled {
```
roman-khimov, line 192: This one also doesn't need to be configured explicitly, CNs usually serve as notaries as well.

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,608 @@
...
+	res := &Blockchain{
+		base:      blockChain,
+		netServer: netServer,
+		nep17GAS:  nep17.New(_actor, gas.Hash),
+		actor:     _actor,
+		rpcActor:  rpcActor,
+		chErr:     chErrRw,
+	}
+
+	if !cfg.ApplicationConfiguration.RPC.StartWhenSynchronized {
```
roman-khimov, line 241: IIRC NeoFS instances usually use `StartWhenSynchronized` as it allows to avoid replaying old events. For us it means that we can have this set the way we want to in the node.

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,608 @@
...
+//
+// Use Stop to stop the Blockchain.
+func (x *Blockchain) Run() error {
+	go x.netServer.Start(x.chErr)
+
+	if x.rpcServerStarter != nil {
+		x.rpcServerStarter()
+	}
+
+	// TODO: consider changing to neo(fs)-morph/neofs-ir
```
roman-khimov, line 264: Not needed here at all, I think.

roman-khimov (commented):
Let's start. There are more details here, of course.

pkg/innerring/listen.go
```diff
@@ -0,0 +1,114 @@
...
+
+			x.log.Info("notification from blockchain",
+				zap.String("name", n.name),
+				zap.Stringer("contract", n.contract),
+				zap.Int("item num", len(n.items)),
+			)
+
+			switch n.contract {
+			default:
+				x.log.Info("ignore unsupported notification",
```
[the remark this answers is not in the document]
cthulhu-rider: Agree. Without specific NeoFS indicator in notifications, I don't think this is useful at all. So, debug is useless as well IMO.

pkg/innerring/listen.go
```diff
@@ -0,0 +1,114 @@
...
+
+			x.log.Info("notary request from the blockchain",
+				zap.Stringer("contract", req.contract),
+				zap.String("method", req.method),
+				zap.Int("op num", len(req.ops)),
+			)
+
+			switch req.contract {
+			default:
+				x.log.Info("ignore unsupported notary request",
```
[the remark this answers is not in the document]
cthulhu-rider: Same as in notifications or not?

pkg/innerring/internal/blockchain/actor.go
```diff
@@ -0,0 +1,284 @@
...
+	return &result.Invoke{
+		State:          ic.VM.State().String(),
+		GasConsumed:    ic.VM.GasConsumed(),
+		Script:         script,
+		Stack:          ic.VM.Estack().ToArray(),
+		FaultException: errStr,
+	}, nil
+}
+
+func (x *directRPCActor) CalculateNetworkFee(_tx *transaction.Transaction) (int64, error) {
```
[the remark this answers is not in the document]
cthulhu-rider: Then we need an issue in Neo Go.

pkg/innerring/internal/blockchain/actor.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/nspcc-dev/neo-go/pkg/smartcontract/trigger"
+	"github.com/nspcc-dev/neo-go/pkg/util"
+	"github.com/nspcc-dev/neo-go/pkg/vm"
+	"github.com/nspcc-dev/neo-go/pkg/vm/emit"
+	"github.com/nspcc-dev/neo-go/pkg/vm/stackitem"
+)
+
+// directRPCActor provides actor.RPCActor without actual RPC: it uses core
+// blockchain components of the Neo Go node directly.
+type directRPCActor struct {
```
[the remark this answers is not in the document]
cthulhu-rider: Will you create an issue for this?

pkg/innerring/node.go
```diff
@@ -0,0 +1,354 @@
...
+	}
+
+	return letter >= 0, nil
+}
+
+// EmitSidechainGAS implements alphabet.LocalNode via triggering production of
+// the sidechain GAS and its distribution among Inner Ring nodes and Proxy
+// contract by calling 'emit' method of the Alphabet contract associated with
+// the local Inner Ring node.
+func (x *node) EmitSidechainGAS() error {
```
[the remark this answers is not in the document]
cthulhu-rider: I try to make processors to perform logical handling, not a thing where events go irrevocably. Direct use of actors complicates maintainability, because the control over events' postprocessing should be handled by the node, not by a single processor IMO.

pkg/innerring/processors/common/node.go
```diff
@@ -0,0 +1,9 @@
+package common
+
+// NodeStatus represents status of the local Inner Ring node in the NeoFS network.
+type NodeStatus interface {
```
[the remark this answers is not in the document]
cthulhu-rider: Yeah, otherwise we will copy this in almost all processors. The only alternative I see is to eject Alphabet-only check from processors and perform it preliminaliry on event "distillation".

pkg/innerring/processors/netmap/processor.go
```diff
@@ -91,7 +91,7 @@ type (
 		CleanupEnabled   bool
 		CleanupThreshold uint64 // in epochs
 		ContainerWrapper *container.Client
-		SubnetContract   *util.Uint160
+		SubnetContract   util.Uint160
```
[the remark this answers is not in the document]
cthulhu-rider: Why here or why is it necessary? Here accidentally, needed because param isn't optional.

pkg/innerring/node.go
cthulhu-rider: I'll try.

pkg/innerring/processors/container/common.go
```diff
@@ -46,12 +44,14 @@ type signatureVerificationData struct {
 //   - v.binPublicKey is a public session key
 //   - session context corresponds to the container and verb in v
 //   - session is "alive"
-func (cp *Processor) verifySignature(v signatureVerificationData) error {
+func (x *Processor) verifySignature(v signatureVerificationData) error {
 	var err error
-	var key neofsecdsa.PublicKeyRFC6979
+	var key *neofsecdsa.PublicKeyRFC6979
```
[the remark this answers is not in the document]
cthulhu-rider: This allows to call `Authorize` at the bottom directly, otherwise it'd require extra `if-else` condition.

pkg/innerring/internal/blockchain/blockchain.go
cthulhu-rider: Dropped logo.

pkg/innerring/internal/blockchain/blockchain.go
cthulhu-rider: Made behavior with static `true`.

pkg/innerring/node.go
cthulhu-rider: Done in https://github.com/nspcc-dev/neofs-node/pull/2211/commits/9709c1e6ccd1c2182dbbb79676ab4d4cf2a41d09. I couldn't test changes in the NeoFS Dev Env because I didn't succeeded in withdraw ops without Mainchain. Are they even real without the Mainchain?

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,719 @@
...
+// SubscribeToNotifications performs subscription to notification events spawned
+// by Neo smart contracts and stream them into the returned channel as
+// state.ContainedNotificationEvent instances. Caller must read the channel
+// regularly in the background to prevent congestion. When notifications are no
+// longer needed, caller should cancel subscription by calling the returned
+// function.
+//
+// SubscribeToNotifications is expected to be called once since event
+// multiplexing does not make sense.
+func (x *Blockchain) SubscribeToNotifications() (<-chan *state.ContainedNotificationEvent, func()) {
```
roman-khimov, line 697: No longer needed, regular subscriptions via Internal are the same as via WSClient.

pkg/innerring/listen.go
```diff
@@ -0,0 +1,104 @@
...
+
+// listenNotaryRequests listens to notary requests received from Notary service
+// of the NeoFS blockchain network, distills them to notaryRequest type
+// instances and schedules them for handling by the local Inner Ring
+// node according to related NeoFS smart contracts.
+//
+// The listenNotaryRequests is blocking: it's aborted by breaking the flow of
+// requests from the blockchain or by external signal read from the provided
+// channel.
+func (x *node) listenNotaryRequests(chDone <-chan struct{}) {
```
roman-khimov, line 24: As well as this one. Using regular subscription APIs you get exactly what you've subscribed to in the channel, almost as before (but #2254).

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,719 @@
...
+	go func() {
+		err, ok := <-chErrRw
+		if ok {
+			chErr <- err
+		}
+	}()
+
+	rpcServer := rpcsrv.New(blockChain, cfg.ApplicationConfiguration.RPC, netServer, nil, log, chErrRw)
+
+	netServer.AddService(&rpcServer) // is it a cyclic dependency?
```
roman-khimov, line 270: Not exactly, but RPC needs a network server and then network server can control RPC as any other service (it's an interface here).

pkg/innerring/internal/blockchain/items.go
```diff
@@ -0,0 +1,47 @@
+package blockchain
+
+import (
+	"fmt"
+
+	"github.com/nspcc-dev/neo-go/pkg/vm/stackitem"
+)
+
+// TODO: think how to merge functions with ones from client package (like client.ArrayFromStackItem)
```
roman-khimov, line 9: There is an `unwrap` for it. Current code can use anything it has already, but long-term it's `unwrap` for generic items + custom (per-contract probably) functions for custom things (like some structures).

pkg/innerring/blockchain.go
```diff
@@ -0,0 +1,707 @@
...
+
+	return -1, nil
+}
+
+// alphabetContract returns address of the NeoFS Alphabet contract associated
+// with the given GlagoliticLetter.
+func (x *blockChain) alphabetContract(letter GlagoliticLetter) (res util.Uint160, err error) {
+	res = getAlphabetContract(x.contracts, letter)
+	if res == (util.Uint160{}) {
+		// TODO: request from chain again?
```
roman-khimov, line 389: It's not supposed to happen unless you're requesting some uninitialized letter, but it's not initialized because it's not used and if it's not used then why is it requested?

roman-khimov:
If we want to do some processor refactorings, let's leave it to another PR. We have a nice way to start internal chain/wait for contracts here, we can create Internal clients from that point. The rest of the logic is expected to work as is (well, except for reconnects maybe) because Internal was specifically made to provide the same API. I think we can have a much smaller changeset here (again, maybe other changes are nice, but let's evaluate them in another PR).

And we can have it side-by-side with the old RPC-based configuration.

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,719 @@
...
+// must be previously unlocked for signing, and SignTools.AccountFile must
+// correspond to it. If the account belongs to the committee of the configured
+// Neo blockchain, resulting Blockchain is run in consensus mode.
+//
+// Given zap.Logger instance is used to log Blockchain events. It should be
+// previously initialized: Blockchain does not modify the logger.
+//
+// Specified error channel is used to report internal Blockchain errors to the
+// superior context (Inner Ring application) listening to it in the background.
+func New(neoNodeConfigFilepath string, signTools SignTools, log *zap.Logger, chErr chan<- error) (*Blockchain, error) {
```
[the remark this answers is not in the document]
cthulhu-rider: Initially I decided to completely merge IR and Neo Go configs. But since there is no need to configure some neo-go parameters in IR, this parameter can be made a local structure.

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,719 @@
...
+	chErr chan error
+}
+
+// SignTools groups signing components for New constructor.
+type SignTools struct {
+	// NEO account of the node.
+	Account *wallet.Account
+
+	// File containing Account.
+	AccountFile struct {
```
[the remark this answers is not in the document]
cthulhu-rider: `config.Wallet` fits well (at least for now).

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,719 @@
...
+		// TODO: configure or reconsider the value
+		time.Sleep(3 * time.Second)
+	}
+}
+
+// CallContractMethod creates transaction of Neo smart contract method call and
+// spreads it in the blockchain network. Contract is referenced by the given
+// address, and particular call is done according to the specified method name
+// and various arguments.
+func (x *Blockchain) CallContractMethod(contract util.Uint160, method string, args ...interface{}) error {
```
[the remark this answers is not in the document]
cthulhu-rider: This is not just a desire to wrap smth into smth to produce more types. At any point in time Inner Ring uses a fixed set of Neo blockchain services that does not depend on a specific implementation. `internal.Blockchain` is just a window into the world of the Neo blockchain which highlights the minimum set of services necessary for the functioning of the Inner Ring. It answers the question _"What services Inner Ring expects from Neo blockchain?"._ Spreading general purpose entities throughout the application's codebase complicates maintenance, and mixes mechanics with bussinesslogic. On the contrary, it is convenient to use a single package for all needs for Neo services (this is what `internal/blockchain` package was intended for).

cthulhu-rider:
> let's leave it to another PR

It will be easier for me (and for reviewers I guess) to re-open PR and leave this one as a draft with working samples and design solutions.

cthulhu-rider:
re-opened in #2277, all discussions we resolved or became irrelevant

cthulhu-rider:
re-opened in #2277, all discussions we resolved or became irrelevant

PR #2224, opened 2023-01-30
adm: use native neo-go sessions in `dump-hashes`

fyrchik:
If we had lots of domains in one zone, `dump-hashes` for all others can miss some domains, because we need to restrict ourselves with _some_ number.
In this commit we use neo-go sessions by default, with a proper failback to in-script iterator unwrapping.

### DIFF
--- a/cmd/neofs-adm/internal/modules/morph/dump_hashes.go
+++ b/cmd/neofs-adm/internal/modules/morph/dump_hashes.go
@@ -2,6 +2,7 @@ package morph
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"strings"
 	"text/tabwriter"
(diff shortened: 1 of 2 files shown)


cmd/neofs-adm/internal/modules/morph/dump_hashes.go
```diff
@@ -140,6 +138,37 @@ func dumpCustomZoneHashes(cmd *cobra.Command, nnsHash util.Uint160, zone string,
 		})
 	}
 
+	sessionID, iter, err := unwrap.SessionIterator(inv.Call(nnsHash, "tokens"))
+	if err != nil {
+		if errors.Is(err, unwrap.ErrNoSessionID) {
+			cmd.PrintErrln("Provided RPC endpoint doesn't seem to support sessions, some hashes might be lost.")
```
roman-khimov, line 144: You may print it if `len(items) == nnsMaxTokens` below, otherwise we have all of them anyway.

cmd/neofs-adm/internal/modules/morph/dump_hashes.go
```diff
@@ -107,31 +108,28 @@ func dumpContractHashes(cmd *cobra.Command, _ []string) error {
...
 	}
 
 	var infos []contractDumpInfo
-	for i := range arr {
-		bs, err := arr[i].TryBytes()
+	processItem := func(item stackitem.Item) {
+		bs, err := item.TryBytes()
 		if err != nil {
-			continue
+			return
```
roman-khimov, line 122: `cmd.PrintErrf` below, but silence here?

fyrchik: FIxed

cmd/neofs-adm/internal/modules/morph/dump_hashes.go
fyrchik: Would be nice, fixed

roman-khimov (dismissed):
Nice.

roman-khimov (approved):

carpawell (approved):

acid-ant (dismissed):

PR #2231, opened 2023-02-02
pilorama: merge in-queue batches

fyrchik:
To achieve high performance we must choose proper values for both batch size and delay. For user operations we want to set low delay. However it would prevent tree synchronization operations to form big enough batches. For these operations, batching gives the most benefit not only in terms of on-CPU execution cost, but also by speeding up transaction persist (`fsync`).
In this commit we try merging batches that are already _triggered_, but not yet _started to execute_. This way we can still query batches for execution after the provided delay while also allowing multiple formed batches to execute faster.

Hardware performance tests show ~10x improvement for small objects put in _one_ bucket.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ Changelog for NeoFS Node
 - Separate batching for replicated operations over the same container in pilorama (#1621)
 
 ### Changed
+- Pilorama now can merge multiple batches into one (#2231)
+
 ### Fixed
 - Big object removal with non-local parts (#1978)
 - Disable pilorama when moving to degraded mode (#2197)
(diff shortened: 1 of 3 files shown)


acid-ant (approved):

carpawell (commented):
Commit msg not formated; deserves a CHANGELOG record.

acid-ant (approved):

roman-khimov:
> Hardware performance tests show ~10x improvement for small objects put in one bucket.

What are the numbers in ops/s?

fyrchik:
I don't have exact numbers saved (but @anatoly-bogatyrev can help here).

carpawell (approved):

carpawell:
Wrong PR number in commit msg :grimacing:.

PR #2235, opened 2023-02-03
pilorama: gracefully handle disk errors in triggered batches

fyrchik:


roman-khimov:
We no longer maintain support branches and tree service is not a priority.

PR #2238, opened 2023-02-06
engine: Add tests for initialization failure

fyrchik:
Fix bugs along the way.
Invalid mode on a directory is an approximation of missing media.
We expect any `Open`/`Init` errors to be logged and a shard to be disabled.

### DIFF
--- a/cmd/neofs-node/config.go
+++ b/cmd/neofs-node/config.go
@@ -790,13 +790,18 @@ func initLocalStorage(c *cfg) {
 		tombstone.WithTombstoneSource(tombstoneSrc),
 	)
 
+	var shardsAttached int
 	for _, optsWithMeta := range c.shardOpts() {
 		id, err := ls.AddShard(append(optsWithMeta.shOpts, shard.WithTombstoneSource(tombstoneSource))...)
-		fatalOnErr(err)
-
-		c.log.Info("shard attached to engine",
-			zap.Stringer("id", id),
-		)
+		if err != nil {
+			c.log.Error("failed to attach shard to engine", zap.Error(err))
+		} else {
+			shardsAttached++
+			c.log.Info("shard attached to engine", zap.Stringer("id", id))
+		}
+	}
+	if shardsAttached == 0 {
+		fatalOnErr(engineconfig.ErrNoShardConfigured)
 	}
 
 	c.cfgObject.cfgLocalStorage.localStorage = ls
--- a/pkg/local_object_storage/engine/control.go
+++ b/pkg/local_object_storage/engine/control.go
@@ -23,27 +23,44 @@ func (e *StorageEngine) Open() error {
 }
 
 func (e *StorageEngine) open() error {
-	e.mtx.RLock()
-	defer e.mtx.RUnlock()
+	e.mtx.Lock()
+	defer e.mtx.Unlock()
 
 	var wg sync.WaitGroup
-	var errCh = make(chan error, len(e.shards))
+	var errCh = make(chan shardInitError, len(e.shards))
 
 	for id, sh := range e.shards {
 		wg.Add(1)
 		go func(id string, sh *shard.Shard) {
 			defer wg.Done()
 			if err := sh.Open(); err != nil {
-				errCh <- fmt.Errorf("could not open shard %s: %w", id, err)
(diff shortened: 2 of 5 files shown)


cmd/neofs-node/config.go
```diff
@@ -790,13 +790,18 @@ func initLocalStorage(c *cfg) {
...
-			zap.Stringer("id", id),
-		)
+		if err != nil {
+			c.log.Error("failed to attach shard to engine", zap.Error(err))
+		} else {
+			shardsAttached++
+			c.log.Info("shard attached to engine", zap.Stringer("id", id))
+		}
+	}
+	if shardsAttached == 0 {
```
carpawell, line 803: is it discussed behavior? i mean could a node starting with just an error in the logs but without some planned shards be expected by an admin?

pkg/local_object_storage/engine/control.go
```diff
@@ -27,23 +27,40 @@ func (e *StorageEngine) open() error {
...
-		if err != nil {
-			return err
+	for res := range errCh {
+		if res.err != nil {
+			e.log.Error("could not open shard, closing and skipping",
+				zap.String("id", res.id),
+				zap.Error(res.err))
+
+			sh := e.shards[res.id]
+			delete(e.shards, res.id)
```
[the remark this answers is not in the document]
fyrchik: Fixed

cmd/neofs-node/config.go
fyrchik: Currently we are still dropping shards in `Init` , so this behaviour is not new.
In our model shard is an (almost) independent domain of failure. I am mostly thinking about automatic node restart after hardware failures. We can discuss this in future, I am not sure what the expected behaviour here is: for 11 shards repetitve config manipulations is a laborious and error-prone task.

fyrchik: Or do you mean that we could start it in a degraded mode if we cannot write ID to the metabase?

pkg/local_object_storage/engine/control.go
```diff
@@ -23,27 +23,44 @@ func (e *StorageEngine) Open() error {
 }
 
 func (e *StorageEngine) open() error {
-	e.mtx.RLock()
-	defer e.mtx.RUnlock()
+	e.mtx.Lock()
+	defer e.mtx.Unlock()
```
roman-khimov, line 27: Why do you duplicate this much code from `Init()`?

cmd/neofs-node/config.go
carpawell: > so this behaviour is not new

but we had `fatalOnErr(err)` previously if a shard has not been attached. as i understand, that was a _very different_. losing shards without some huge "you are losing shards" banner bothers me but i dont have any good ideas of how to solve that

pkg/local_object_storage/engine/control.go
fyrchik: Could you elaborate?
This line is here because we can remove shards in `Open` after this PR.

fyrchik:
> The biggest problem of this approach is that you can lose shards (and potentially data) on benign misconfiguration. This will happen.

I agree. But the alternative here is _lose all_ vs _lose some_ data.

pkg/local_object_storage/engine/control.go
roman-khimov: It's not about this particular line, I just had to attach this comment somewhere. There is quite some duplication between `open` and `Init` now it seems.

cmd/neofs-node/config.go
carpawell: > In our model shard is an (almost) independent domain of failure.

also, can't fully agree with this. one shard could contain `Lock`s, `Tombstone`s that protect/remove data from another so losing one shard is not just losing some objects. but that is our general problem

carpawell (dismissed):
I would add changes that affect node's actions to the CHANGELOG.

pkg/local_object_storage/engine/control.go
fyrchik: Oh, I see.
I've thought about it here, though, I am not sure it will be simpler with some ad-hoc parrallelizing function which accepts functions as arguments.

PR #2239, opened 2023-02-06
writecache: fix possible deadlock

fyrchik:
Here is a scenario:
1. Read from `flushed` inside of a `View` transaction.
2. Eviction handler in other goroutine deletes items from the database which can lead to allocating new storage for mmap.

### DIFF
--- a/pkg/local_object_storage/writecache/flush.go
+++ b/pkg/local_object_storage/writecache/flush.go
@@ -99,10 +99,6 @@ func (c *cache) flushDB() {
 					lastKey = slice.Copy(k)
 				}
 
-				if _, ok := c.flushed.Peek(string(k)); ok {
-					continue
-				}
-
 				m = append(m, objectInfo{
 					addr: string(k),
 					data: slice.Copy(v),
@@ -111,12 +107,18 @@ func (c *cache) flushDB() {
 			return nil
 		})
 
+		var count int
 		for i := range m {
+			if c.flushed.Contains(m[i].addr) {
+				continue
+			}
+
 			obj := object.New()
 			if err := obj.Unmarshal(m[i].data); err != nil {
 				continue
 			}
 
+			count++
 			select {
 			case c.flushCh <- obj:
 			case <-c.closeCh:
@@ -125,7 +127,7 @@ func (c *cache) flushDB() {
 			}
 		}
 
-		if len(m) == 0 {
+		if count == 0 {
 			c.modeMtx.RUnlock()
 			break
 		}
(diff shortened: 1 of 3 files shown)


acid-ant (dismissed):

pkg/local_object_storage/writecache/flush.go
```diff
@@ -99,10 +99,6 @@ func (c *cache) flushDB() {
 					lastKey = slice.Copy(k)
 				}
 
-				if _, ok := c.flushed.Peek(string(k)); ok {
```
roman-khimov, line 102: Why don't you just use `Contains()` instead?

roman-khimov: Also, `Peek` is not supposed to delete anything.

fyrchik: I didn't knew about `Contains`, fixed!

`Peek` doesn't delete but it takes read mutex _inside_ of a `View` transaction. `View` transaction itself takes `mmapLock`: https://github.com/etcd-io/bbolt/blob/8f4a7e1f92975fbd1536a569d7aad6800809ef4e/db.go#L708
which is lifted after tx finishes (in `tx.Commit()` -> `tx.close()` -> `tx.db.removeTx`)

When we evict items from LRU cache mutex order is different: first we lock cache and then execute `Batch` which usually doesn't take `mmapLock`. However it case we need to remap it _does_. Thus the deadlock.

Added this to a commit message.

carpawell (approved):

pkg/local_object_storage/writecache/flush.go
roman-khimov: IIUC, `flushed` is a _simple_ LRU as in https://github.com/hashicorp/golang-lru/blob/v0.6.0/simplelru/lru.go#L93. Where does it take any read mutex?

You're writing "eviction handler removes untracked items from the database". But neither Peek nor Contains remove items from the LRU cache. Am I missing something?

pkg/local_object_storage/writecache/flush.go
```diff
@@ -112,6 +108,10 @@ func (c *cache) flushDB() {
 		})
 
 		for i := range m {
```
roman-khimov, line 110: Notice also that down below you've got a `len(m) == 0` check, but you've got a different `m` now in some situations, you can ignore some items in this loop, but they will be accounted for later.

pkg/local_object_storage/writecache/flush.go
fyrchik: `simplelru` package is `LRU` (simple LRU) and `LRUCache` (an interface). We use the latter.
https://github.com/hashicorp/golang-lru/blob/v0.6.0/simplelru/lru_interface.go

Yes, but we _can_ execute `flushed.Add` in another goroutine in which case it _will_ evict.

pkg/local_object_storage/writecache/flush.go
fyrchik: Fixed!

roman-khimov (approved):

PR #2246, opened 2023-02-08
node: Allow to configure tombsone lifetime

fyrchik:
Currently, DELETE service sets tombstone expiration epoch to `current epoch + 5`. This works less than ideal in private networks where an epoch can be e.g. 10 minutes. In this case, after a node is unavailable for more than 1 hour, already deleted objects have a chance to reappear.

After this commit tombstone lifetime can be configured.


docs/storage-node-configuration.md
```diff
@@ -415,14 +415,15 @@ replicator:
...
     pool_size_remote: 100
 ```
 
-| Parameter              | Type  | Default value | Description                                                                                    |
-|------------------------|-------|---------------|------------------------------------------------------------------------------------------------|
-| `put.pool_size_remote` | `int` | `10`          | Max pool size for performing remote `PUT` operations. Used by Policer and Replicator services. |
\ No newline at end of file
+| Parameter                   | Type  | Default value | Description                                                                                    |
+|-----------------------------|-------|---------------|------------------------------------------------------------------------------------------------|
+| `delete.tombstone_lifetime` | `int` | `5`           | Tombstone lifetime for removed objects.                                                        |
```
anikeev-yadro, line 428: Please clarify units for this parameter

fyrchik: Fixed!

roman-khimov (commented):
If it's a network-specific parameter, then it should be a network-wide configuration, isn't it? Otherwise you can have different nodes using different configurations on the same network and it can lead to some surprises.

fyrchik:
@roman-khimov I partially agree with you. The case is that this lifetime _can_ be arbitrary, we set it only in the DELETE service for every tombstone. Like with REGULAR objects the expiration epoch is not restricted.

roman-khimov:
Or take it from another perspective, you have some local configuration now that potentially affects the network globally. Is it a good thing to have?

fyrchik:
Ok, I agree, it would probably also make it easier to scale.

This fix is kinda HOT, so I will check how invasive it is to take a network setting into account here.

realloc (approved):

realloc:
This parameters must be treated as a local node configuration option. While set network-wise it doesn't prevent any node from misbehavior nor affects interoperability.

PR #2260, opened 2023-02-17
Fixes for a failover scenario.

fyrchik:
In master:
`network/cache`-related changes will be replaced with gRPC-native keepalive mechanism.
Client cache separation should be approached more rigourously. It seems like a good idea, but we need to evaluate it further.

### DIFF
--- a/pkg/services/object/get/exec.go
+++ b/pkg/services/object/get/exec.go
@@ -128,7 +128,7 @@ func (exec execCtx) key() (*ecdsa.PrivateKey, error) {
 }
 
 func (exec *execCtx) canAssemble() bool {
-	return exec.svc.assembly && !exec.isRaw() && !exec.headOnly()
+	return exec.svc.assembly && !exec.isRaw() && !exec.headOnly() && !exec.isLocal()
 }
 
 func (exec *execCtx) splitInfo() *objectSDK.SplitInfo {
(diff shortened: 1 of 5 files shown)


acid-ant (approved):

pkg/services/object/get/exec.go
```diff
@@ -128,7 +128,7 @@ func (exec execCtx) key() (*ecdsa.PrivateKey, error) {
 }
 
 func (exec *execCtx) canAssemble() bool {
-	return exec.svc.assembly && !exec.isRaw() && !exec.headOnly()
+	return exec.svc.assembly && !exec.isRaw() && !exec.headOnly() && !exec.isLocal()
```
carpawell, line 131: just to highlight: we have `exec.svc.assembly` and it is somehow always `true`

carpawell (approved):
While that helped, do we have an issue related to failover problems? I am not totally sure that separating client caches solves 20s+ object streaming/fetching and it still requires additional investigations.

carpawell:
CHANGELOG would be also appreciated.

fyrchik:
Yes, we temporarily disabled compression. Currently, we cannot control disk load for policer (well, we can, but it is suboptimal) and it iterates over each shard in order. Failover case makes things worse because with high probability we will also read full object in memory to replicate.

PR #2264, opened 2023-02-17
Fix client transaction scopes

roman-khimov:
I think it's also relevant for 0.35.1.


cthulhu-rider (dismissed):

roman-khimov:
To be recreated for the master branch (just changing the base is not enough because of conflicts and it has '0.35.0' in the branch name, so force-pushing into it also doesn't seem to be correct).

PR #2272, opened 2023-02-22
Fix/Subscribe dead lock

carpawell:
It led to a neo-go dead-lock in the `subscriber` component.

### DIFF
--- a/pkg/morph/event/listener.go
+++ b/pkg/morph/event/listener.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"sync"
 
-	"github.com/nspcc-dev/neo-go/pkg/core/block"
 	"github.com/nspcc-dev/neo-go/pkg/core/state"
 	"github.com/nspcc-dev/neo-go/pkg/neorpc/result"
 	"github.com/nspcc-dev/neo-go/pkg/util"
@@ -158,6 +157,19 @@ func (l *listener) ListenWithError(ctx context.Context, intError chan<- error) {
 }
 
 func (l *listener) listen(ctx context.Context, intError chan<- error) error {
+	// mark listener as started
+	l.started = true
+
+	subErrCh := make(chan error)
+
+	go l.subscribe(subErrCh)
+
+	l.listenLoop(ctx, intError, subErrCh)
+
+	return nil
+}
+
+func (l *listener) subscribe(errCh chan error) {
 	// create the list of listening contract hashes
 	hashes := make([]util.Uint160, 0)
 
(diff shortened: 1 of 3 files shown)


pkg/morph/event/listener.go
```diff
@@ -181,55 +181,70 @@ func (l *listener) listen(ctx context.Context, intError chan<- error) error {
...
 
-			return
+			return err
 		}
 	}
 
+	return nil
+}
+
+func (l *listener) listenLoop(ctx context.Context, intErr chan<- error,
```
fyrchik, line 241: Do we need to change it's signature? Ideally, only startup code should be affected.

fyrchik (commented):
Could you also describe a particular deadlock scenario in the commit message?

carpawell:
Extended commit msg.

pkg/morph/event/listener.go
carpawell: it was not `listenLoop` before: it was `subscribeAndListen`. it could be done in some other way but i found it more readable

> Ideally, only startup code should be affected.

that is startup code changes in fact. just listening and subscribing routines shares some channels now (and they _have_ to do that to prevent dead-lock)

pkg/morph/subscriber/subscriber.go
```diff
@@ -91,20 +93,28 @@ func (s *subscriber) Close() {
...
-		return nil, fmt.Errorf("could not subscribe for new block events: %w", err)
+		return fmt.Errorf("could not subscribe for new block events: %w", err)
 	}
 
-	return s.blockChan, nil
+	return nil
 }
 
-func (s *subscriber) SubscribeForNotaryRequests(mainTXSigner util.Uint160) (<-chan *result.NotaryRequestEvent, error) {
+func (s *subscriber) SubscribeForNotaryRequests(rcv chan<- *result.NotaryRequestEvent, mainTXSigner util.Uint160) error {
```
fyrchik, line 108: Previously we could subscribe multiple times, but now we have a single channel. IMO this is a time-bomb.
I am also bothered about possible race-conditions  when we switch to another endpoint.
Not something we would like to do in the support branch.

Specifically: can we _prove_ that we cannot block when we read the channel, switch to another endpoint and then send a notifications?

carpawell:
Simplified changes.

pkg/morph/event/listener.go
```diff
@@ -175,70 +187,47 @@ func (l *listener) listen(ctx context.Context, intError chan<- error) error {
...
+func (l *listener) listenLoop(ctx context.Context, intErr chan<- error, subErrCh chan error) {
+	chs := l.subscriber.NotificationChannels()
 
 loop:
 	for {
 		select {
+		case err := <-subErrCh:
+			if intErr != nil {
+				intErr <- err
+			}
```
fyrchik, line 222: We had a log in the `else` branch in the previous implementation. Why do we not have it now?

fyrchik (dismissed):

pkg/morph/event/listener.go
carpawell: added `Error` (not `Debug` like was before) log

fyrchik (approved):

fyrchik (approved):

PR #2277, opened 2023-03-15
IR: Local consensus launch mode

### LINKED ISSUE #2194: Integrate NeoGo CN into IR node
I'm always frustrated when I have to update NeoFS because it has so many moving parts and links:
Screenshot_20221215_152853

One of the problems specifically is IR/CN relationship. On one hand they're different components performing somewhat different tasks and theoretically they could be deployed to different nodes, use different keys and so on. On the other we always have 1:1 relationship between them (for Alphabet), they always use the same key and they're always on the same machine.

We can simplify deployment/updates by merging them into one service (binary). IR node would get a ProtocolConfiguration section of the NeoGo config initially, will run an instance of `core.Blockchain` along with `network.Server` and `consensus.Service` inside and IR code itself can be just another service managed by `network.Server` (see https://github.com/nspcc-dev/neo-go/blob/v0.100.1/cli/server/server.go#L428 also).

IR will then subscribe to events internally via https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/core#Blockchain.SubscribeForNotifications and could rely on this channel to always work. Transactions could be pushed via https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/network#Server.RelayTxn and https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/network#Server.RelayP2PNotaryRequest. Initially I though that RPC configuration can still be useful, but in fact it can be removed, even public networks use 1:1 IR/CN setups, so we can simplify the code wrt this.

Of course we can also have non-Alphabet IR nodes, but those could just run a regular RPC NeoGo instance internally.

Pros:
 * no need to manage/configure another service
 * CN will always have the right version compatible with IR (no problems with outdated/too new CNs)
 * faster CN/IR communication
 * more reliable CN/IR communication (this link can't go down)

Cons:
 * theoretical scalability issues

The old model as is. But it's hard to manage even in presence of Ansible roles/neofs-adm tools.

## Additional context
We'll get this picture instead of the previous one:
Notes_230109_155422

Notice that M nodes can be omitted as well in many cases, leaving us with just two binaries.

cthulhu-rider:
* closes #2194 
* reborn of #2211 

## Summary
New launch mode is a pure feature activated when `morph.client.endpoint` is omitted only.

Local blockchain is configured in `morph.consensus` section.

NeoGo module is upgraded to revision with meaningful fix making consensus switch loop work.

Only meaningful Sidechain settings are configured. Some NeoGo defaults are overridden according to the specifics of NeoFS Sidechains (@roman-khimov double-check please).

## P.S.
Storage node's config action triggered to no purpose.

### DIFF
--- a/config/example/ir.yaml
+++ b/config/example/ir.yaml
@@ -19,6 +19,71 @@ morph:
   validators: # List of hex-encoded 33-byte public keys of sidechain validators to vote for at application startup
     - 0283120f4c8c1fc1d792af5063d2def9da5fddc90bc1384de7fcfdda33c3860170
   switch_interval: 2m # interval b/w RPC switch attempts if the node is not connected to the highest priority node
+  consensus: # Local consensus launch mode activated only when 'endpoint.client' is unset.
+    magic: 15405 # Network magic. Must be unsigned integer in range [1:4294967295]
+    committee: # Initial committee
+      - 02b3622bf4017bdfe317c58aed5f4c753f206b7db896046fa7d774bbc4bf7f8dc2 # Hex-encoded public key
+    storage: # Blockchain storage
+      type: boltdb # One of following storage types:
+        # boltdb (local BoltDB)
+        # leveldb (local LevelDB)
+        # inmemory (volatile storage, mostly suitable for testing)
+      #
+      # If 'path' exists, it should correspond to the selected variant
+      path: ./db/morph.bolt # File system path to the storage. Ignored for 'inmemory' storage.
+    notary_disabled: true # Optional flag to disable Notary service. Defaults to false
+    time_per_block: 1s # Optional time period (approximate) between two adjacent blocks. Defaults to 15s.
+      # Must not be negative
+    max_traceable_blocks: 2102400 # Optional length of the chain accessible to smart contracts. Defaults to 2102400.
+      # Must not be greater than 4294967295
+    seed_nodes: # Optional list of existing nodes to communicate with over Neo P2P protocol. By default, node runs as standalone
+      # Same format as 'p2p.listen'
+      - node2:20333
+      - node3:20333
+    hardforks: # Optional hard-forks
+      name: 1730000 # Maps name to chain height. Heights must not be greater than 4294967295
+    validators_history: # Optional number of consensus nodes to use after given height
+      10: 7 # Maps chain height to number of consensus nodes. Values must not be greater than 2147483647
+    rpc: # Optional RPC settings
+      listen: # Optional list of network addresses to listen Neo RPC on. By default, protocol is not served
+        # TCP addresses in 'host:port' format
+        - localhost:30333
+        - localhost:30334
+    p2p: # Optional P2P settings
+      dial_timeout: 5s # Optional maximum duration a single peer dial may take. Defaults to 5s. Must not be negative
+      proto_tick_interval: 2s # Optional time period between protocol ticks with each connected peer. Defaults to 2s.
+        # Must not be negative
+      listen: # Optional list of network addresses to listen Neo P2P on. By default, protocol is not served
+        # TCP addresses in 'host:port' format
+        - localhost:20333
+        - localhost:20334
+      peers: # Optional peer settings
+        min: 1 # Optional minimum number of peers a node needs for normal operation. Defaults to consensus minimum
+          # of 'committee' size (ceil of 2/3N-1). Must not be greater than 2147483647. Note that consensus service
+          # won't start until at least 'min' number of peers are connected
+        max: 5 # Optional limits of maximum number of peers dealing with the node. Defaults to 100. Must not be
+          # greater than 2147483647
+        attempts: 20 # How many peers node should try to dial after falling under 'min' count. Defaults to 'min'+10.
+          # Must not be greater than 2147483647
+      ping: # Optional settings of pinging mechanism
+        interval: 30s # Optional time period between pings. Defaults to 30s. Must not be negative
+        timeout: 90s # Optional time period to wait for pong. Defaults to 1m. Must not be negative
+      native_activations: # Optional chronology of native contracts updates. Maps name of Neo native contract
+        # to height list. Values must not be empty. Each array element in the value must be unsigned integer
+        # not larger than 4294967295.
+        # Attention! Keep a close eye on the list: if some contracts are specified, then the rest of them are
+        # treated as disabled and won't be deployed. Use this option with care, and if you do, the list should
+        # match the one below.
+        ContractManagement: [0]
+        LedgerContract: [0]
+        NeoToken: [0]
+        GasToken: [0]
+        PolicyContract: [0]
+        OracleContract: [0]
+        RoleManagement: [0]
+        Notary: [0]
+        CryptoLib: [0]
+        StdLib: [0]
 
 mainnet:
   dial_timeout: 5s # Timeout for RPC client connection to mainchain; ignore if mainchain is disabled
--- a/pkg/innerring/config.go
+++ b/pkg/innerring/config.go
@@ -1,0 +1,386 @@
+package innerring
+
+import (
+	"errors"
+	"fmt"
+	"math"
+	"net"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/nspcc-dev/neo-go/pkg/config/netmode"
+	"github.com/nspcc-dev/neo-go/pkg/core/native/nativenames"
+	"github.com/nspcc-dev/neo-go/pkg/core/storage/dbconfig"
+	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
+	"github.com/nspcc-dev/neofs-node/pkg/innerring/internal/blockchain"
+	"github.com/nspcc-dev/neofs-node/pkg/util/logger"
+	"github.com/spf13/cast"
+	"github.com/spf13/viper"
+)
+
+func parseBlockchainConfig(v *viper.Viper, _logger *logger.Logger) (c blockchain.Config, err error) {
+	const rootSection = "morph.consensus"
+
+	if !v.IsSet(rootSection) {
+		return c, fmt.Errorf("missing root section '%s'", rootSection)
+	}
+
+	_uint, err := parseConfigUint64Range(v, rootSection+".magic", "network magic", 1, math.MaxUint32)
+	if err != nil {
+		return c, err
+	}
+	c.NetworkMagic = netmode.Magic(_uint)
+
+	const notaryDisabledKey = rootSection + ".notary_disabled"
+	if v.IsSet(notaryDisabledKey) {
+		c.NotaryDisabled, err = cast.ToBoolE(v.Get(notaryDisabledKey))
+		if err != nil {
+			return c, fmt.Errorf("invalid notary flag '%s' (boolean): %w", notaryDisabledKey, err)
+		}
+	}
+
+	const storageSection = rootSection + ".storage"
+	if !v.IsSet(storageSection) {
+		return c, fmt.Errorf("missing storage section '%s'", storageSection)
+	}
+	const storageTypeKey = storageSection + ".type"
+	if !v.IsSet(storageTypeKey) {
+		return c, fmt.Errorf("missing storage type '%s'", storageTypeKey)
+	}
+	const storagePathKey = storageSection + ".path"
+	switch typ := v.GetString(storageTypeKey); typ {
+	default:
+		return c, fmt.Errorf("unsupported storage type '%s': '%s'", storageTypeKey, typ)
+	case dbconfig.BoltDB:
+		if !v.IsSet(storagePathKey) {
+			return c, fmt.Errorf("missing path to the BoltDB '%s'", storagePathKey)
+		}
+		c.Storage = blockchain.BoltDB(v.GetString(storagePathKey))
+	case dbconfig.LevelDB:
+		if !v.IsSet(storagePathKey) {
+			return c, fmt.Errorf("missing path to the LevelDB '%s'", storagePathKey)
+		}
+		c.Storage = blockchain.LevelDB(v.GetString(storagePathKey))
+	case dbconfig.InMemoryDB:
+		c.Storage = blockchain.InMemory()
+	}
+
+	const committeeKey = rootSection + ".committee"
+	c.Committee, err = parseConfigPublicKeys(v, committeeKey, "committee members")
+	if err != nil {
+		return c, err
+	} else if len(c.Committee) == 0 {
+		return c, fmt.Errorf("empty committee members '%s'", committeeKey)
+	}
+
+	c.BlockInterval, err = parseConfigDurationPositive(v, rootSection+".time_per_block", "block interval")
+	if err != nil && !errors.Is(err, errMissingConfig) {
+		return c, err
+	}
+
+	traceableChainLength, err := parseConfigUint64Range(v, rootSection+".max_traceable_blocks", "traceable chain length", 1, math.MaxUint32)
+	if err != nil && !errors.Is(err, errMissingConfig) {
+		return c, err
+	}
+	c.TraceableChainLength = uint32(traceableChainLength)
+
+	c.SeedNodes, err = parseConfigAddressesTCP(v, rootSection+".seed_nodes", "seed nodes")
+	if err != nil && !errors.Is(err, errMissingConfig) {
+		return c, err
+	}
+
+	const hardForksKey = rootSection + ".hardforks"
+	if v.IsSet(hardForksKey) {
+		c.HardForks, err = parseConfigMapUint32(v, hardForksKey, "hard forks", math.MaxUint32)
+		if err != nil {
+			return c, err
+		}
+	}
+
+	const validatorsHistoryKey = rootSection + ".validators_history"
+	if v.IsSet(validatorsHistoryKey) {
+		c.ValidatorsHistory = make(map[uint32]uint32)
+		err = parseConfigMap(v, validatorsHistoryKey, "validators history", func(name string, val interface{}) error {
+			height, err := strconv.ParseUint(name, 10, 32)
+			if err != nil {
+				return fmt.Errorf("parse unsigned integer: %w", err)
+			}
+			num, err := cast.ToUint32E(val)
+			if err != nil {
+				return err
+			} else if num > math.MaxInt32 {
+				return fmt.Errorf("value %d is out of allowable range", num)
+			}
+			c.ValidatorsHistory[uint32(height)] = num
+			return nil
+		})
+		if err != nil {
+			return c, err
+		}
+	}
+
+	const nativeActivationsKey = rootSection + ".native_activations"
+	if v.IsSet(nativeActivationsKey) {
+		requiredContracts := []string{
+			nativenames.Management,
+			nativenames.Ledger,
+			nativenames.Neo,
+			nativenames.Gas,
+			nativenames.Policy,
+			nativenames.Oracle,
+			nativenames.Designation,
+			nativenames.Notary,
+			nativenames.CryptoLib,
+			nativenames.StdLib,
+		}
+		c.NativeActivations = make(map[string][]uint32)
+		err = parseConfigMap(v, nativeActivationsKey, "native update histories", func(name string, val interface{}) error {
+			supported := false
+			for i := range requiredContracts {
+				// TODO: viper lowers YAML keys, so, for example, 'NativeContract' becomes 'nativecontract'
+				//  Track https://github.com/spf13/viper/issues/1014. Until then, keep track of
+				//  new contracts in nativenames package.
+				supported = name == strings.ToLower(requiredContracts[i])
+				if supported {
+					name = requiredContracts[i]
+					break
+				}
+			}
+			if !supported {
+				return fmt.Errorf("unsupported Neo native contract name '%s'", name)
+			}
+			rawHeights, err := cast.ToSliceE(val)
+			if err != nil {
+				return fmt.Errorf("cast value to array '%s': %w", name, err)
+			} else if len(rawHeights) == 0 {
+				return fmt.Errorf("empty array '%s'", name)
+			}
+			heights := make([]uint32, len(rawHeights))
+			for i := range rawHeights {
+				u64, err := cast.ToUint64E(rawHeights[i])
+				if err != nil {
+					return fmt.Errorf("cast element to unsigned integer '%s' #%d: %w", name, i, err)
+				} else if u64 > math.MaxUint32 {
+					return fmt.Errorf("value overflows limit %v", u64)
+				}
+				heights[i] = uint32(u64)
+			}
+			c.NativeActivations[name] = heights
+			return nil
+		})
+		if err != nil {
+			return c, err
+		}
+		for i := range requiredContracts {
+			if _, ok := c.NativeActivations[requiredContracts[i]]; !ok {
+				return c, fmt.Errorf("missing required Neo native contract name '%s' in '%s'", requiredContracts[i], nativeActivationsKey)
+			}
+		}
+	}
+
+	const rpcSection = rootSection + ".rpc"
+	if v.IsSet(rpcSection) {
+		c.RPCListenAddresses, err = parseConfigAddressesTCP(v, rpcSection+".listen", "network addresses to listen Neo RPC on")
+		if err != nil && !errors.Is(err, errMissingConfig) {
+			return c, err
+		}
+	}
+
+	const p2pSection = rootSection + ".p2p"
+	if v.IsSet(p2pSection) {
+		c.P2P.DialTimeout, err = parseConfigDurationPositive(v, p2pSection+".dial_timeout", "P2P dial timeout")
+		if err != nil && !errors.Is(err, errMissingConfig) {
+			return c, err
+		}
+		c.P2P.ProtoTickInterval, err = parseConfigDurationPositive(v, p2pSection+".proto_tick_interval", "P2P protocol tick interval")
+		if err != nil && !errors.Is(err, errMissingConfig) {
+			return c, err
+		}
+		c.P2P.ListenAddresses, err = parseConfigAddressesTCP(v, p2pSection+".listen", "network addresses to listen Neo P2P on")
+		if err != nil && !errors.Is(err, errMissingConfig) {
+			return c, err
+		}
+		const p2pPeersSection = p2pSection + ".peers"
+		if v.IsSet(p2pPeersSection) {
+			minPeers, err := parseConfigUint64Max(v, p2pPeersSection+".min", "minimum number of P2P peers", math.MaxInt32)
+			if err != nil {
+				if !errors.Is(err, errMissingConfig) {
+					return c, err
+				}
+				// we calculate default here since explicit 0 is also a valid setting
+				n := uint64(len(c.Committee))
+				minPeers = n - (n-1)/3 - 1
+			}
+			c.P2P.MinPeers = uint(minPeers)
+			maxPeers, err := parseConfigUint64Range(v, p2pPeersSection+".max", "maximum number of P2P peers", 1, math.MaxInt32)
+			if err != nil && !errors.Is(err, errMissingConfig) {
+				return c, err
+			}
+			c.P2P.MaxPeers = uint(maxPeers)
+			attemptConnPeers, err := parseConfigUint64Range(v, p2pPeersSection+".attempts", "number of P2P connection attempts", 1, math.MaxInt32)
+			if err != nil && !errors.Is(err, errMissingConfig) {
+				return c, err
+			}
+			c.P2P.AttemptConnPeers = uint(attemptConnPeers)
+		}
+		const pingSection = p2pSection + ".ping"
+		if v.IsSet(pingSection) {
+			c.P2P.Ping.Interval, err = parseConfigDurationPositive(v, pingSection+".interval", "P2P ping interval")
+			if err != nil {
+				return c, err
+			}
+			c.P2P.Ping.Timeout, err = parseConfigDurationPositive(v, pingSection+".timeout", "P2P ping timeout")
+			if err != nil {
+				return c, err
+			}
+		}
+	}
+
+	c.Logger = _logger.Logger
+
+	return c, nil
+}
+
+var errMissingConfig = errors.New("config value is missing")
+
+func parseConfigUint64Condition(v *viper.Viper, key, desc string, cond func(uint64) error) (uint64, error) {
+	var res uint64
+	var err error
+	if !v.IsSet(key) {
+		err = errMissingConfig
+	}
+	if err == nil {
+		res, err = cast.ToUint64E(v.Get(key))
+		if err == nil && cond != nil {
+			err = cond(res)
+		}
+	}
+	if err != nil {
+		return res, fmt.Errorf("invalid %s '%s' (unsigned integer): %w", desc, key, err)
+	}
+	return res, nil
+}
+
+func parseConfigUint64Range(v *viper.Viper, key, desc string, min, max uint64) (uint64, error) {
+	return parseConfigUint64Condition(v, key, desc, func(val uint64) error {
+		if val < min || val > max {
+			return fmt.Errorf("out of allowable range [%d:%d]", min, max)
+		}
+		return nil
+	})
+}
+
+func parseConfigUint64Max(v *viper.Viper, key, desc string, max uint64) (uint64, error) {
+	return parseConfigUint64Range(v, key, desc, 0, max)
+}
+
+func parseConfigDurationCondition(v *viper.Viper, key, desc string, cond func(time.Duration) error) (time.Duration, error) {
+	var res time.Duration
+	var err error
+	if !v.IsSet(key) {
+		err = errMissingConfig
+	}
+	if err == nil {
+		res, err = cast.ToDurationE(v.Get(key))
+		if err == nil && cond != nil {
+			err = cond(res)
+		}
+	}
+	if err != nil {
+		return res, fmt.Errorf("invalid %s '%s' (duration): %w", desc, key, err)
+	}
+	return res, nil
+}
+
+func parseConfigDurationPositive(v *viper.Viper, key, desc string) (time.Duration, error) {
+	return parseConfigDurationCondition(v, key, desc, func(d time.Duration) error {
+		if d <= 0 {
+			return errors.New("must be positive")
+		}
+		return nil
+	})
+}
+
+func parseConfigStrings(v *viper.Viper, key, desc string) ([]string, error) {
+	var res []string
+	var err error
+	if !v.IsSet(key) {
+		err = errMissingConfig
+	}
+	if err == nil {
+		res, err = cast.ToStringSliceE(v.Get(key))
+	}
+	if err != nil {
+		return res, fmt.Errorf("invalid %s '%s' (string array): %w", desc, key, err)
+	}
+	return res, nil
+}
+
+func parseConfigPublicKeys(v *viper.Viper, key, desc string) (keys.PublicKeys, error) {
+	ss, err := parseConfigStrings(v, key, desc)
+	if err != nil {
+		return nil, err
+	}
+	res, err := keys.NewPublicKeysFromStrings(ss)
+	if err != nil {
+		return res, fmt.Errorf("invalid %s '%s' (public keys): %w", desc, key, err)
+	}
+	return res, nil
+}
+
+func parseConfigAddressesTCP(v *viper.Viper, key, desc string) ([]string, error) {
+	ss, err := parseConfigStrings(v, key, desc)
+	if err != nil {
+		return nil, err
+	}
+	for i := range ss {
+		_, err = net.ResolveTCPAddr("tcp", ss[i])
+		if err != nil {
+			return ss, fmt.Errorf("invalid %s '%s' (TCP addresses): %w", desc, key, err)
+		}
+	}
+	return ss, nil
+}
+
+func parseConfigMap(v *viper.Viper, key, desc string, f func(name string, val interface{}) error) error {
+	var err error
+	if !v.IsSet(key) {
+		err = errMissingConfig
+	}
+	if err == nil {
+		var m map[string]interface{}
+		m, err = cast.ToStringMapE(v.Get(key))
+		if err == nil {
+			for name, val := range m {
+				err = f(name, val)
+				if err != nil {
+					err = fmt.Errorf("invalid element '%s': %w", name, err)
+					break
+				}
+			}
+		}
+	}
+	if err != nil {
+		return fmt.Errorf("invalid %s '%s' (dictionary): %w", desc, key, err)
+	}
+	return nil
+}
+
+func parseConfigMapUint32(v *viper.Viper, key, desc string, limit uint64) (map[string]uint32, error) {
+	res := make(map[string]uint32)
+	return res, parseConfigMap(v, key, desc, func(name string, val interface{}) error {
+		if name == "" {
+			return errors.New("empty key")
+		}
+		u64, err := cast.ToUint64E(val)
+		if err == nil {
+			if u64 > limit {
+				err = fmt.Errorf("value overflows limit %v", u64)
+			} else {
+				res[name] = uint32(u64)
+			}
+		}
+		return err
+	})
+}
--- a/pkg/innerring/contracts.go
+++ b/pkg/innerring/contracts.go
@@ -1,13 +1,18 @@
 package innerring
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/nspcc-dev/neo-go/pkg/util"
 	"github.com/nspcc-dev/neofs-node/pkg/morph/client"
+	"github.com/nspcc-dev/neofs-node/pkg/util/logger"
+	"github.com/spf13/cast"
 	"github.com/spf13/viper"
+	"go.uber.org/zap"
 )
 
 type contracts struct {
@@ -25,19 +30,21 @@ type contracts struct {
 	alphabet alphabetContracts // in morph
 }
 
-func parseContracts(cfg *viper.Viper, morph *client.Client, withoutMainNet, withoutMainNotary, withoutSideNotary bool) (*contracts, error) {
+func initContracts(ctx context.Context, _logger *logger.Logger, cfg *viper.Viper, morph *client.Client, withoutMainNet, withoutMainNotary, withoutSideNotary bool) (*contracts, error) {
 	var (
 		result = new(contracts)
 		err    error
 	)
 
 	if !withoutMainNet {
+		_logger.Debug("decoding configured NeoFS contract...")
 		result.neofs, err = util.Uint160DecodeStringLE(cfg.GetString("contracts.neofs"))
 		if err != nil {
 			return nil, fmt.Errorf("can't get neofs script hash: %w", err)
 		}
 
 		if !withoutMainNotary {
+			_logger.Debug("decoding configured Processing contract...")
 			result.processing, err = util.Uint160DecodeStringLE(cfg.GetString("contracts.processing"))
 			if err != nil {
 				return nil, fmt.Errorf("can't get processing script hash: %w", err)
@@ -45,8 +52,10 @@ func parseContracts(cfg *viper.Viper, morph *client.Client, withoutMainNet, with
 		}
 	}
 
+	nnsCtx := &nnsContext{Context: ctx}
+
 	if !withoutSideNotary {
-		result.proxy, err = parseContract(cfg, morph, "contracts.proxy", client.NNSProxyContractName)
+		result.proxy, err = parseContract(nnsCtx, _logger, cfg, morph, "contracts.proxy", client.NNSProxyContractName)
 		if err != nil {
 			return nil, fmt.Errorf("can't get proxy script hash: %w", err)
 		}
@@ -67,14 +76,14 @@ func parseContracts(cfg *viper.Viper, morph *client.Client, withoutMainNet, with
 	}
 
 	for _, t := range targets {
-		*t.dest, err = parseContract(cfg, morph, t.cfgName, t.nnsName)
+		*t.dest, err = parseContract(nnsCtx, _logger, cfg, morph, t.cfgName, t.nnsName)
 		if err != nil {
 			name := strings.TrimPrefix(t.cfgName, "contracts.")
 			return nil, fmt.Errorf("can't get %s script hash: %w", name, err)
 		}
 	}
 
-	result.alphabet, err = parseAlphabetContracts(cfg, morph)
+	result.alphabet, err = parseAlphabetContracts(nnsCtx, _logger, cfg, morph)
 	if err != nil {
 		return nil, err
 	}
@@ -82,8 +91,23 @@ func parseContracts(cfg *viper.Viper, morph *client.Client, withoutMainNet, with
 	return result, nil
 }
 
-func parseAlphabetContracts(cfg *viper.Viper, morph *client.Client) (alphabetContracts, error) {
-	num := GlagoliticLetter(cfg.GetUint("contracts.alphabet.amount"))
+func parseAlphabetContracts(ctx *nnsContext, _logger *logger.Logger, cfg *viper.Viper, morph *client.Client) (alphabetContracts, error) {
+	var num GlagoliticLetter
+	const numConfigKey = "contracts.alphabet.amount"
+	if cfg.IsSet(numConfigKey) {
+		u, err := cast.ToUintE(cfg.Get(numConfigKey))
+		if err != nil {
+			return nil, fmt.Errorf("invalid config '%s': %w", numConfigKey, err)
+		}
+		num = GlagoliticLetter(u)
+	} else {
+		committee, err := morph.Committee()
+		if err != nil {
+			return nil, fmt.Errorf("get Sidechain committee: %w", err)
+		}
+		num = GlagoliticLetter(len(committee))
+	}
+
 	alpha := newAlphabetContracts()
 
 	if num > lastLetterNum {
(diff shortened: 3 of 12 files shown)


pkg/innerring/contracts.go
```diff
@@ -67,23 +76,38 @@ func parseContracts(cfg *viper.Viper, morph *client.Client, withoutMainNet, with
...
 	}
 
 	return result, nil
 }
 
-func parseAlphabetContracts(cfg *viper.Viper, morph *client.Client) (alphabetContracts, error) {
-	num := GlagoliticLetter(cfg.GetUint("contracts.alphabet.amount"))
+func parseAlphabetContracts(ctx *nnsContext, _logger *logger.Logger, cfg *viper.Viper, morph *client.Client) (alphabetContracts, error) {
+	var num GlagoliticLetter
+	const numConfigKey = "contracts.alphabet.amount"
```
roman-khimov, line 96: Just out of curiosity, has anyone ever used this? I see it comes from #405, but I doubt it should exist, especially in the merged IR/CN scenario.

pkg/innerring/contracts.go
```diff
@@ -121,11 +145,61 @@ func parseAlphabetContracts(cfg *viper.Viper, morph *client.Client) (alphabetCon
...
-		return morph.NNSContractAddress(nnsName)
+type nnsContext struct {
+	context.Context
+
+	nnsDeployed bool
+}
+
+func parseContract(ctx *nnsContext, _logger *logger.Logger, cfg *viper.Viper, morph *client.Client, cfgName, nnsName string) (res util.Uint160, err error) {
+	if cfg.IsSet(cfgName) {
+		_logger.Info("decoding configured contract...", zap.String("config key", cfgName))
```
roman-khimov, line 156: Looks like debug, not something to log normally.

pkg/innerring/contracts.go
```diff
@@ -121,11 +145,61 @@ func parseAlphabetContracts(cfg *viper.Viper, morph *client.Client) (alphabetCon
...
+			}
+
+			_, err = morph.NNSHash()
+			if err == nil {
+				ctx.nnsDeployed = true
+				break
+			}
+
+			// FIXME: unstable approach, but currently there is no other way
+			//  Track https://github.com/neo-project/proposals/pull/156 and its support in
```
roman-khimov, line 179: I think we need some issue in neofs-node to track these places and fix them when it's fixed in some NeoGo release.

pkg/innerring/innerring.go
```diff
@@ -352,15 +348,124 @@ func New(ctx context.Context, log *logger.Logger, cfg *viper.Viper, errChan chan
...
+		cfgBlockchain.Storage.Path = cfg.GetString(storagePathKey)
+		cfgBlockchain.ErrorListener = errChan
+		cfgBlockchain.NetworkMagic = netmode.Magic(cfg.GetUint32(netMagicKey))
+		cfgBlockchain.Committee = committee
+
+		server.blockChain, err = blockchain.New(cfgBlockchain, blockchain.Options{
+			Logger:         log.Logger,
+			SeedNodes:      cfg.GetStringSlice(morphConsensusKeyPrefix + "seed_nodes"),
+			RPCEndpoints:   cfg.GetStringSlice(morphConsensusKeyPrefix + "rpc_endpoints"),
+			P2pEndpoints:   cfg.GetStringSlice(morphConsensusKeyPrefix + "p2p_endpoints"),
```
roman-khimov, line 408: s/P2pEndpoints/P2PEndpoints/?

pkg/innerring/innerring.go
```diff
@@ -352,15 +348,124 @@ func New(ctx context.Context, log *logger.Logger, cfg *viper.Viper, errChan chan
...
+			client.WithContext(ctx),
+			client.WithLogger(log),
+			client.WithSingleClient(wsClient),
+		)
+		if err != nil {
+			return nil, fmt.Errorf("init internal morph client: %w", err)
+		}
+	} else {
+		// fallback to the pure RPC architecture
+		// TODO: think how we continue to support the case
```
roman-khimov, line 452: I think it's perfectly fine to keep it. The amount of code required to support it is minimal and yet it allows for easier migration and alternative setups in case they're really really really needed for some reason.

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,383 @@
...
+		standByCommittee[i] = hex.EncodeToString(cfg.Committee[i].Bytes())
+	}
+
+	var cfgBase config.Config
+
+	cfgBaseProto := &cfgBase.ProtocolConfiguration
+	cfgBaseProto.Magic = cfg.NetworkMagic
+	cfgBaseProto.StandbyCommittee = standByCommittee
+	cfgBaseProto.ValidatorsCount = len(standByCommittee)
+	cfgBaseProto.TimePerBlock = time.Second
```
roman-khimov, line 145: Must be configurable.

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,383 @@
...
+	cfgBaseApp.RPC.Addresses = opts.RPCEndpoints
+	cfgBaseApp.RPC.StartWhenSynchronized = true
+	cfgBaseApp.RPC.MaxGasInvoke = fixedn.Fixed8FromInt64(100)
+	cfgBaseApp.P2P.Addresses = opts.P2pEndpoints
+	cfgBaseApp.P2P.DialTimeout = 3 * time.Second
+	cfgBaseApp.P2P.ProtoTickInterval = 2 * time.Second
+	cfgBaseApp.P2P.PingInterval = 30 * time.Second
+	cfgBaseApp.P2P.PingTimeout = 90 * time.Second
+	cfgBaseApp.P2P.MaxPeers = 100
+	cfgBaseApp.P2P.AttemptConnPeers = 20
```
roman-khimov, line 169: These P2P params are for public networks, but they may be different for single node or for special setups. MinPeers is also missing, but it's very important for `StartWhenSynchronized` to work correctly.

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,383 @@
...
+
+	cfgBaseProto := &cfgBase.ProtocolConfiguration
+	cfgBaseProto.Magic = cfg.NetworkMagic
+	cfgBaseProto.StandbyCommittee = standByCommittee
+	cfgBaseProto.ValidatorsCount = len(standByCommittee)
+	cfgBaseProto.TimePerBlock = time.Second
+	cfgBaseProto.SeedList = opts.SeedNodes
+	cfgBaseProto.VerifyTransactions = true
+	cfgBaseProto.P2PSigExtensions = !opts.NotaryDisabled
+
```
roman-khimov, line 149: `MaxTraceableBlocks` is also an interesting one that can differ from network to network.
`Hardforks` is important to have for future updates.

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,383 @@
...
+// If Run failed, the Blockchain should no longer be used. After Blockchain has
+// been successfully run, all internal failures are written to the configured
+// listener.
+//
+// Run should not be called more than once.
+//
+// Use Stop to stop the Blockchain.
+func (x *Blockchain) Run() error {
+	go x.netServer.Start(x.chErr)
+	go x.switchConsensusLoop()
```
roman-khimov, line 279: Maybe we can omit it at all? The consensus service will be in watch-only mode if the node is not elected, so it can always kick in, but it'll do almost nothing if it's not.

roman-khimov (commented):
Really nice this way.

pkg/innerring/contracts.go
cthulhu-rider: right now I don't see any real use and general practical applicability of this

pkg/innerring/contracts.go
cthulhu-rider: yeah I'll create one

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,398 @@
...
+// (omitted) values don't change the behavior. Launched Blockchain should be
+// stopped.
+type Options struct {
+	// Writer of the Blockchain's logs and internal errors. By default, Blockchain
+	// doesn't write logs.
+	Logger *zap.Logger
+
+	// Network addresses to listen Neo P2P on. Each element must be a valid TCP
+	// address in 'host:port' format. By default, Neo P2P is not served.
+	P2PListenAddresses []string
```
roman-khimov, line 154: Make it a part of `P2PConfig`?

pkg/innerring/contracts.go
```diff
@@ -121,11 +145,60 @@ func parseAlphabetContracts(cfg *viper.Viper, morph *client.Client) (alphabetCon
...
+	nnsDeployed bool
+}
+
+func parseContract(ctx *nnsContext, _logger *logger.Logger, cfg *viper.Viper, morph *client.Client, cfgName, nnsName string) (res util.Uint160, err error) {
+	if cfg.IsSet(cfgName) {
+		_logger.Debug("decoding configured contract...", zap.String("config key", cfgName))
+		return util.Uint160DecodeStringLE(cfg.GetString(cfgName))
+	}
+
+	const pollInterval = 5 * time.Second // no too long and not so frequently
```
roman-khimov, line 160: You can use time per block setting, you have it here (or can get from a client). It'd be more correct for a particular network (1s or 15s, there is a difference).

pkg/innerring/internal/blockchain/blockchain.go
cthulhu-rider: Here I have separated the required (`Config`) and optional (`Options`) settings even if they belong to the same section. Do you think it's better to keep a single configuration structure and explicitly prescribe the optional settings?

pkg/innerring/config.go
```diff
@@ -0,0 +1,260 @@
...
+
+	"github.com/nspcc-dev/neo-go/pkg/config/netmode"
+	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
+	"github.com/nspcc-dev/neofs-node/pkg/innerring/internal/blockchain"
+	"github.com/nspcc-dev/neofs-node/pkg/util/logger"
+	"github.com/spf13/cast"
+	"github.com/spf13/viper"
+)
+
+func parseBlockchainConfig(v *viper.Viper, _logger *logger.Logger) (c blockchain.Config, o blockchain.Options, err error) {
```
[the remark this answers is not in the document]
cthulhu-rider: yep I'll cover write unit tests for this function

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,398 @@
...
+	cfgBaseApp.RPC.MaxGasInvoke = fixedn.Fixed8FromInt64(100)
+	cfgBaseApp.P2P.Addresses = opts.P2PListenAddresses
+	cfgBaseApp.P2P.DialTimeout = cfg.P2P.DialTimeout
+	cfgBaseApp.P2P.ProtoTickInterval = cfg.P2P.ProtoTickInterval
+	cfgBaseApp.P2P.PingInterval = cfg.P2P.Ping.Interval
+	cfgBaseApp.P2P.PingTimeout = cfg.P2P.Ping.Timeout
+	cfgBaseApp.P2P.MinPeers = int(cfg.P2P.MinPeers)
+	cfgBaseApp.P2P.AttemptConnPeers = int(cfg.P2P.AttemptConnPeers)
+	cfgBaseApp.P2P.MaxPeers = int(cfg.P2P.MaxPeers)
+	// TODO: profiler, metrics
```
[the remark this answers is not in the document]
cthulhu-rider: don't you think it's worth merging the profiler/metrics server?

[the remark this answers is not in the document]
roman-khimov: What exactly to merge? I think if you're just to enable pprof/metrics in the IR node it'll work.

pkg/innerring/internal/blockchain/blockchain.go
roman-khimov: Optional ones can have some per-setting defaults, so they can be omitted if not needed.

cthulhu-rider: Currently omitted `Config` field is incorrect, and empty `Options` field defaults to prescribed value. Transferring field `P2PListenAddresses` to `P2PConfig` will mean that it is mandatory and cannot be omitted. Otherwise I will have to merge `Config` with `Options`.

pkg/innerring/config.go
```diff
@@ -0,0 +1,260 @@
...
+	return res, nil
+}
+
+func parseConfigAddressesTCP(v *viper.Viper, key, desc string) ([]string, error) {
+	ss, err := parseConfigStrings(v, key, desc)
+	if err != nil {
+		return nil, err
+	}
+	for i := range ss {
+		_, err = net.ResolveTCPAddr("tcp", ss[i])
```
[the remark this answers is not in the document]
cthulhu-rider: NeoGo node is considered completely private and we do not inherit the entity completely. Otherwise, it would be better to completely inherit the node config to avoid duplication. So I don't consider it a regression, anyway many NeoGo settings are not configurable in the current variant.

pkg/innerring/internal/blockchain/blockchain.go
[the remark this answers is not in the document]
cthulhu-rider: I mean single endpoint with all metrics/profiles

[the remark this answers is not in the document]
roman-khimov: I think if you're to just configure IR node's metrics/profiler services you'd get everything there already. You can try it out, I'd expect a whole set of `neogo_*` metrics to be available.

pkg/innerring/internal/blockchain/blockchain.go
roman-khimov: Just have a `P2P` section (yes, mandatory) and have some defaults for unconfigured fields there. It's OK for people to specify some fields, but not specify the others. Some P2P config must be present and this setting is organically a part of it.

pkg/innerring/innerring.go
```diff
@@ -63,6 +64,8 @@ type (
 	Server struct {
 		log *logger.Logger
 
+		blockChain *blockchain.Blockchain
```
AnnaShaleva, line 67: s/blockChain/blockchain?

config/example/ir.yaml
```diff
@@ -19,6 +19,45 @@ morph:
...
+        - localhost:30334
+    p2p: # P2P settings
+      dial_timeout: 5s # Maximum duration a single peer dial may take. Must be positive
+      proto_tick_interval: 2s # Time period between protocol ticks with each connected peer. Must be positive
+      listen: # Optional list of network addresses to listen Neo P2P on. By default, protocol is not served
+        # TCP addresses in 'host:port' format
+        - localhost:20333
+        - localhost:20334
+      peers:
+        min: 1 # Minimum number of peers a node needs for normal operation. Must be in range [0:2147483647]
```
AnnaShaleva, line 55: It should probably be noted that if being too large, this setting will prevent node from starting the consensus process. In other words, if the node is unable to connect to at least `min` peers, then it won't start dbft and (if `StartWhenSynchronized` is on) the RPC.

pkg/innerring/contracts.go
```diff
@@ -121,11 +145,60 @@ func parseAlphabetContracts(cfg *viper.Viper, morph *client.Client) (alphabetCon
...
+				return
+			}
+
+			time.Sleep(pollInterval)
+		}
 	}
 
-	return util.Uint160DecodeStringLE(contractStr)
+	for {
+		_logger.Info("waiting for contract registration in the NNS...", zap.String("name", nnsName))
```
AnnaShaleva, line 189: Just a question: NNS contract address is being regirtered in the NNS contract itself, right?

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,398 @@
...
+	}
+
+	blockchainStorage, err := storage.NewStore(cfgDB)
+	if err != nil {
+		return nil, fmt.Errorf("init storage for blockchain: %w", err)
+	}
+
+	blockChain, err := core.NewBlockchain(blockchainStorage, cfgBase.Blockchain(), opts.Logger)
+	if err != nil {
+		closeErr := blockchainStorage.Close()
```
AnnaShaleva, line 275: `blockChain` but `blockchainStorage`? Sorry for nitpicking, I like the latter.

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,398 @@
...
+		if closeErr != nil {
+			return nil, fmt.Errorf("init core blockchain component: %w; failed to close blockchain storage: %v", err, closeErr)
+		}
+
+		return nil, fmt.Errorf("init core blockchain component: %w", err)
+	}
+
+	go blockChain.Run()
```
AnnaShaleva, line 283: So the neo-go's blockchain is run inside the neofs-node's blockchain constructor. What happens if neofs'node's blockchain will be created but won't be run (due to some error on initialisation, probably)? Will the neo-go's blockchain be stopped in this case?

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,398 @@
...
+			}
+
+			return nil
+		})
+		if err != nil {
+			return nil, fmt.Errorf("init Notary service: %w", err)
+		}
+
+		netServer.AddService(notaryService)
+		blockChain.SetNotary(notaryService)
```
AnnaShaleva, line 310: No way, read the comment to `SetNotary`, please:
```
// SetNotary sets notary module. It doesn't protected by mutex and
// must be called before `bc.Run()` to avoid data race.
```
The same thing is for stateroot/oracle/other services.

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,398 @@
...
+// Returns any error encountered which prevented the Blockchain to be started.
+// If Run failed, the Blockchain should no longer be used. After Blockchain has
+// been successfully run, all internal failures are written to the configured
+// listener.
+//
+// Run should not be called more than once.
+//
+// Use Stop to stop the Blockchain.
+func (x *Blockchain) Run() error {
+	go x.netServer.Start(x.chErr)
```
AnnaShaleva, line 367: I think that we need to run underlying neo-go's blockchain right here.

config/example/ir.yaml
```diff
@@ -19,6 +19,45 @@ morph:
...
+    storage: # Blockchain storage
+      type: boltdb # One of following storage types:
+        # boltdb (local BoltDB)
+        # leveldb (local LevelDB)
+        # inmemory (volatile storage, mostly suitable for testing)
+        #
+        # If 'path' exists, it should correspond to the selected variant.
+      path: ./db/morph.bolt # File system path to the storage. Ignored for 'inmemory' storage.
+    rpc: # RPC settings
+      listen: # Optional list of network addresses to listen Neo RPC on. By default, protocol is not served
```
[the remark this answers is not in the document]
cthulhu-rider: Overall RPC is required by storage layer to work. Theoretically we can disable RPC in some (not all) nodes, but I don't know why. Correct me if I'm wrong @AnnaShaleva @roman-khimov.

config/example/ir.yaml
```diff
@@ -19,6 +19,45 @@ morph:
   validators: # List of hex-encoded 33-byte public keys of sidechain validators to vote for at application startup
     - 0283120f4c8c1fc1d792af5063d2def9da5fddc90bc1384de7fcfdda33c3860170
   switch_interval: 2m # interval b/w RPC switch attempts if the node is not connected to the highest priority node
+  consensus: # Local consensus launch mode. Activates only when 'endpoint.client' is unset.
```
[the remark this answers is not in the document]
cthulhu-rider: Maybe
```suggestion
  consensus: # Local consensus launch mode activated only when 'endpoint.client' is unset.
```
?

config/example/ir.yaml
```diff
@@ -19,6 +19,45 @@ morph:
...
+        # TCP addresses in 'host:port' format
+        - localhost:30333
+        - localhost:30334
+    p2p: # P2P settings
+      dial_timeout: 5s # Maximum duration a single peer dial may take. Must be positive
+      proto_tick_interval: 2s # Time period between protocol ticks with each connected peer. Must be positive
+      listen: # Optional list of network addresses to listen Neo P2P on. By default, protocol is not served
+        # TCP addresses in 'host:port' format
+        - localhost:20333
+        - localhost:20334
```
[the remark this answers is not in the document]
cthulhu-rider: it will be equal to listening port in this case, am I wrong? If so, IMO we don't need to split them right now. Anyway it can supported compatible in the future.

config/example/ir.yaml
```diff
@@ -19,6 +19,45 @@ morph:
   validators: # List of hex-encoded 33-byte public keys of sidechain validators to vote for at application startup
     - 0283120f4c8c1fc1d792af5063d2def9da5fddc90bc1384de7fcfdda33c3860170
   switch_interval: 2m # interval b/w RPC switch attempts if the node is not connected to the highest priority node
+  consensus: # Local consensus launch mode. Activates only when 'endpoint.client' is unset.
+    block_interval: 1s # Time period (approximate) between two adjacent blocks. Must be positive
+    magic: 15405 # Network magic. Must be unsigned integer in range [1:4294967295]
+    traceable_chain_length: 2102400 # Length of the chain accessible to smart contracts. Must be in range [1:4294967295]
```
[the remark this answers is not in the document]
cthulhu-rider: Just tried to clarify naming. OK I'll leave it `max_traceable_blocks`.

pkg/innerring/config.go
```diff
@@ -0,0 +1,260 @@
...
+	const storageSection = rootSection + "storage."
+	const storageTypeKey = storageSection + "type"
+	if !v.IsSet(storageTypeKey) {
+		return c, o, fmt.Errorf("missing storage type '%s'", storageTypeKey)
+	}
+	const storagePathKey = storageSection + "path"
+	switch typ := v.GetString(storageTypeKey); typ {
+	default:
+		return c, o, fmt.Errorf("unsupported storage type '%s': '%s'", storageTypeKey, typ)
+	case "boltdb":
```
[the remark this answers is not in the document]
cthulhu-rider: Now their meanings are the same, but in general they are different things. The config must be resistant to changing constants (although I understand that they are unlikely to ever be changed, but it happened once https://github.com/nspcc-dev/neo-go/issues/2931). Finally I use `dbconfig` constants on NeoGo node init.

pkg/innerring/innerring.go
cthulhu-rider: package name collision, but here legit

AnnaShaleva: Make it just ‘bc’ like everywhere in neo-go?

pkg/innerring/config.go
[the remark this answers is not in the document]
AnnaShaleva: OK, but how about moving them to separate constants then?

config/example/ir.yaml
cthulhu-rider: NeoGo doesn't note that.

> too large

Can you please give a specific (albeit approximate) value? 100, 1000, 10000...?

config/example/ir.yaml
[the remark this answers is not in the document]
AnnaShaleva: Yes, let’s do this.

pkg/innerring/contracts.go
cthulhu-rider: if I understand the question correctly: we have an unspoken prerequisite for deploying the NeoFS NNS contract with ID=1

pkg/innerring/internal/blockchain/blockchain.go
cthulhu-rider: I don't wanna collide with the package name. Is `_blockchain` OK for u?

pkg/innerring/internal/blockchain/blockchain.go
cthulhu-rider: ur right we should stop it then

pkg/innerring/internal/blockchain/blockchain.go
cthulhu-rider: relates to https://github.com/nspcc-dev/neofs-node/pull/2277#discussion_r1144677795

config/example/ir.yaml
```diff
@@ -19,6 +19,45 @@ morph:
...
+        # inmemory (volatile storage, mostly suitable for testing)
+        #
+        # If 'path' exists, it should correspond to the selected variant.
+      path: ./db/morph.bolt # File system path to the storage. Ignored for 'inmemory' storage.
+    rpc: # RPC settings
+      listen: # Optional list of network addresses to listen Neo RPC on. By default, protocol is not served
+        # TCP addresses in 'host:port' format
+        - localhost:30333
+        - localhost:30334
+    p2p: # P2P settings
```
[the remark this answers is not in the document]
cthulhu-rider: https://github.com/nspcc-dev/neofs-node/pull/2277#discussion_r1144656896

config/example/ir.yaml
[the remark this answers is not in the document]
roman-khimov: IR node can't work without RPC, so it's always enabled. Whether it's listening or not is a separate matter, in most of the cases it will, becase SNs need to connect via RPC to something as well.

pkg/innerring/config.go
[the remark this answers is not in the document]
cthulhu-rider: I don't mind

config/example/ir.yaml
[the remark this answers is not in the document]
roman-khimov: Yep, it's https://github.com/nspcc-dev/neofs-node/pull/2277#discussion_r1143639278

pkg/innerring/config.go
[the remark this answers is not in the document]
roman-khimov: In fact, they've never changed, it's just that documentation has `xDB` instead of a proper type names.

config/example/ir.yaml
roman-khimov: We mention some "optimal" numbers in https://github.com/nspcc-dev/neo-go/blob/master/docs/consensus.md#start-nodes-manually, but it all depends on the network. I bet private installations may have CNs only on the network in many cases.

config/example/ir.yaml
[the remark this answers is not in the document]
roman-khimov: NeoGo doesn't have any enabled/disabled flags for P2P. Either you have seeds/peers or not.

pkg/innerring/contracts.go
roman-khimov: @AnnaShaleva : this is one is about non-NNS contracts registering in NNS as a part of their deployment procedure.

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,398 @@
...
+	Storage StorageConfig
+
+	// Identifier of the Neo network.
+	NetworkMagic netmode.Magic
+
+	// Initial committee staff.
+	Committee keys.PublicKeys
+
+	// Time period (approximate) between two adjacent blocks. Must be positive.
+	BlockInterval time.Duration
```
roman-khimov, line 135: Default: 15s

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,398 @@
...
+
+	// Time period to wait for pong. Must be positive.
+	Timeout time.Duration
+}
+
+// P2PConfig configures communication over Neo P2P protocol.
+type P2PConfig struct {
+	// Specifies the minimum number of peers a node needs for normal operation. Must
+	// not be greater than math.MaxInt32.
+	MinPeers uint
```
roman-khimov, line 92: Can we have a dynamic default for it? Ceiling of (2/3 of committee size minus one).

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,398 @@
...
+
+	// Specifies how many peers node should try to dial when connection counter
+	// drops below the MinPeers value. Must be in range [1, math.MaxInt32].
+	AttemptConnPeers uint
```
roman-khimov, line 96: Default to MinPeers? @AnnaShaleva?

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,398 @@
...
+
+	// Limits maximum number of peers dealing with the node. Must be in range
+	// [1, math.MaxInt32].
+	MaxPeers uint
```
roman-khimov, line 100: Default to 100

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,398 @@
...
+
+	// Pinging mechanism.
+	Ping PingConfig
+
+	// Maximum duration a single dial may take. Must be positive.
+	DialTimeout time.Duration
```
roman-khimov, line 106: Default: 5s

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,398 @@
...
+
+	// Interval between protocol ticks with each connected peer. Must be positive.
+	ProtoTickInterval time.Duration
```
roman-khimov, line 109: Default: 2s

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,398 @@
...
+	Committee keys.PublicKeys
+
+	// Time period (approximate) between two adjacent blocks. Must be positive.
+	BlockInterval time.Duration
+
+	// P2P settings.
+	P2P P2PConfig
+
+	// Length of the chain accessible to smart contracts. Must be positive.
+	TraceableChainLength uint32
```
roman-khimov, line 141: Default: 2102400

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,398 @@
...
+}
```
roman-khimov, line 142: Another thing to support for the future is NativeActivations.

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,398 @@
...
+func InMemory() StorageConfig {
+	return StorageConfig{
+		typ: dbconfig.InMemoryDB,
+	}
+}
+
+// PingConfig configures P2P pinging mechanism.
+type PingConfig struct {
+	// Interval between pings. Must be positive.
+	Interval time.Duration
```
roman-khimov, line 82: Default: 30s

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,398 @@
...
+
+	// Time period to wait for pong. Must be positive.
+	Timeout time.Duration
```
roman-khimov, line 85: Default: 60s

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,398 @@
...
+	cfgBaseProto.Hardforks = opts.HardForks
+
+	cfgBaseApp := &cfgBase.ApplicationConfiguration
+	cfgBaseApp.Relay = true
+	cfgBaseApp.UnlockWallet = cfg.Wallet
+	cfgBaseApp.Consensus.Enabled = true
+	cfgBaseApp.Consensus.UnlockWallet = cfg.Wallet
+	cfgBaseApp.P2PNotary.Enabled = !opts.NotaryDisabled
+	cfgBaseApp.P2PNotary.UnlockWallet = cfg.Wallet
+	cfgBaseApp.StateRoot.UnlockWallet = cfg.Wallet
```
roman-khimov, line 240: Not needed btw, we don't enable stateroot service anyway.

config/example/ir.yaml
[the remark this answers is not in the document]
AnnaShaleva: Oh, shame on me, I've used to test our node in standalone mode by removing all its peers from config, and somehow I wrote this about `p2p`...

config/example/ir.yaml
AnnaShaleva: > Can you please give a specific (albeit approximate) value? 100, 1000, 10000...?

We're misunderstanding each other. Let the `min` be set to `2`. Then the node won't start accepting blocks until `2` peers are connected. Sometimes it may be unobvious why the node doesn't start the consensus process (it just stales in waiting for more peers to be connected), and Alex has encountered this problem. I'll add a comment about it to neo-go doc.

pkg/innerring/internal/blockchain/blockchain.go
AnnaShaleva: `bc` OK for me!

pkg/innerring/internal/blockchain/blockchain.go
AnnaShaleva: Проси слона - получишь лошадь!

It's set to be `20` in the neo-go and I vote for some value that is larger that MinPeers. These are just a potential peers, some of them may refuse to establish connection, thus, I think it's better to ask more than MinPeers.

AnnaShaleva: But at the same time I suppose that `20` is too large for neofs node, so we may choose some `10` value.

pkg/innerring/internal/blockchain/blockchain.go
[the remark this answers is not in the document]
cthulhu-rider: yeah it works

pkg/innerring/internal/blockchain/blockchain.go
cthulhu-rider: moved `Run` to `Start`

pkg/innerring/internal/blockchain/blockchain.go
roman-khimov: OK, doesn't hurt.

pkg/innerring/contracts.go
AnnaShaleva: Oh, I got it, `nnsName` is not the "Neo name service`, it's just the method's parameter and the name of some other neofs contract.

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,500 @@
...
+	}
+
+	bc, err := core.NewBlockchain(bcStorage, cfgBase.Blockchain(), cfg.Logger)
+	if err != nil {
+		closeErr := bcStorage.Close()
+		if closeErr != nil {
+			return nil, fmt.Errorf("init core blockchain component: %w; failed to close blockchain storage: %v", err, closeErr)
+		}
+
+		return nil, fmt.Errorf("init core blockchain component: %w", err)
```
AnnaShaleva, line 383: It's better to perform storage closing in `defer` statement in case if there will be some error in the code below. In windows (if someday we'll support it) it's critical to free the resources and such unclosed processes hang.

AnnaShaleva: The storage should be closed manually until the `bc.Run` is called. After that `bc.Close()` will close the storage by itself.

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,500 @@
...
+
+	// "make" channel rw to satisfy Start method
+	chErrRw := make(chan error)
+
+	go func() {
+		err, ok := <-chErrRw
+		if ok {
+			cfg.ErrorListener <- err
+		}
+	}()
```
AnnaShaleva, line 440: This function will exit immediately. It must be the loop for services errors habdling. And it's something that doesn't work properly in neo-go as far, we have https://github.com/nspcc-dev/neo-go/issues/2896.

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,500 @@
...
+		return nil, fmt.Errorf("compose NeoGo server config from the base one: %w", err)
+	}
+
+	var cfgDB dbconfig.DBConfiguration
+	cfgDB.Type = cfg.Storage.typ
+	if cfgDB.Type == dbconfig.BoltDB {
+		cfgDB.BoltDBOptions.FilePath = cfg.Storage.path
+	} else if cfgDB.Type == dbconfig.LevelDB {
+		cfgDB.LevelDBOptions.DataDirectoryPath = cfg.Storage.path
+	}
```
AnnaShaleva, line 369: You don't need these if-else, just fill both level's and bolt's paths.

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,500 @@
...
+	NativeActivations map[string][]uint32
+}
+
+// New returns new Blockchain configured by the specified Config. New panics if
+// any required Config field is zero or unset. Resulting Blockchain is ready to
+// run. Launched Blockchain should be finally stopped.
+func New(cfg Config) (*Blockchain, error) {
+	switch {
+	case cfg.Storage.typ == "":
+		panic("uninitialized storage config")
```
AnnaShaleva, line 231: Why do we panic here and below? Why not to provide some logger and do `log.Fatal(...)` or just return an error? Is there any recover from this panic? It's not very pleasent when the user sees panic on the screen.

cthulhu-rider: Because that's developer failure, not user. All requirements are strict and stated in the documentation, they are checked at the upper level. The package is internal an not intended to be widely used. For example, stdlib panics on input sometimes and you should care about it.

I don't mind to return an error, but at the end of the day this is still a programmer's mistake.

@roman-khimov thoughts?

pkg/innerring/internal/blockchain/blockchain.go
cthulhu-rider: Why? Then one of the statements will definitely be no-op, isn't it?

pkg/innerring/internal/blockchain/blockchain.go
cthulhu-rider: Added loop

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,500 @@
...
+	//
+	// Optional: defaults to 1m. Must not be negative.
+	Timeout time.Duration
+}
+
+// P2PConfig configures communication over Neo P2P protocol.
+type P2PConfig struct {
+	// Specifies the minimum number of peers a node needs for normal operation.
+	//
+	// Required. Must not be larger than math.MaxInt32.
```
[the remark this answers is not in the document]
cthulhu-rider: Here it's not optional because 0 is a valid config which is passed to the NeoGo. The default is set in superior context.

pkg/innerring/innerring.go
```diff
@@ -352,15 +346,88 @@ func New(ctx context.Context, log *logger.Logger, cfg *viper.Viper, errChan chan
...
+		if err != nil {
+			return nil, fmt.Errorf("init internal blockchain: %w", err)
+		}
+
+		// don't move this code area within current function
+		{
+			log.Info("running blockchain...")
+			// we run node here instead of Server.Start because some Server components
+			// require blockchain to be run in the current function. Later it would be nice
+			// to separate the stages of construction and launch
```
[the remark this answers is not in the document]
cthulhu-rider: Almost everyone - they accept `morph.Client` which should be instantly initialized right now.

pkg/innerring/internal/blockchain/blockchain.go
roman-khimov: This particular check is internal, not related to user data (which is checked in `parseBlockchainConfig`), so yes, it can be a panic here.

pkg/innerring/config.go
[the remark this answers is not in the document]
cthulhu-rider: ready

pkg/innerring/innerring.go
```diff
@@ -352,15 +346,88 @@ func New(ctx context.Context, log *logger.Logger, cfg *viper.Viper, errChan chan
...
+			}
+
+			defer func() {
+				if err != nil {
+					server.bc.Stop()
+				}
+			}()
+		}
+
+		wsClient, err := server.bc.BuildWSClient(ctx)
```
roman-khimov, line 396: We build WSClient here (immediately after blockchain init) and then use it everywhere. This made me wonder about `StartWhenSynchronized` option and I think we have a problem with it.

The way it was is:
 * NeoGo starts independently and synchronizes with the network
 * NeoFS IR starts independently and tries to connect to NeoGo
 * if NeoGo is not yet in sync, the connection won't succeed, IR will fail init, but it's likely to be restarted
 * so effectively it'll retry again and again until it can connect, which happens when NeoGo is in sync with the network (`StartWhenSynchronized`)
 * everything works smoothly, IR only sees current state and events

Now we've got this:
 * internal blockchain starts, we build WSClient on top of its RPC server and it's always there
 * if the node is in sync, it's OK
 * if it's not, we have two problems:
   - internal RPC server's event handler loop is not running, but subscriptions can happen via internal client -> deadlock on event processing
   - potentially we're running as if `StartWhenSynchronized` is false, old events may be received by IR causing it to send some transactions which may then lead to notary balance drain

Options to solve it:
 * failing to create a client (requires NeoGo change), which is problematic as it will lead to IR restarts and not allow blockchain to sync, but can an option if we're to handle this particular error here
 * creating a client, but failing requests to it (requires NeoGo change), I'm not sure how this will be handled by IR (it'll try to reconnect, but get the error again?)
 * checking the status in IR itself, maybe the easiest to implement now (can be done in NeoFS itself), we have https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/network#Server.IsInSync and it can be checked until it's `true`
 * add IR to the network server via `AddService` and let it be started when it's appropriate to do so

Thoughts, opinions?

cthulhu-rider: > add IR to the network server via AddService

Overall it looks better to me, but once we have #2294 it's not really easy to implement right now.

In current architecture IR depends on NeoGo node, so IR should start only after NeoGo is ready-to-go. I'd start with any working NeoGo-way to reach sync state (now it is vital for IR) and block `internal.Blockchain.Run` method until this state. This will keep the current component hierarchy up and running.

Speaking of handling events from the past: I would come up with some generic way for the IR (e.g. for alternative implementation) to avoid handling them. Maybe sync-await will fix it or at least reduce to a minimum?

AnnaShaleva: > add IR to the network server via AddService

In perfect world it's my favourite option. And doing this will require move all IR/RPC client/etc initialisation staff to the `service.Start()` callback which will make our code much cleaner.

But since this PR is just an initial implementation, I vote for the third option (polling waiting for `server.IsInSync()` to become `true`), and after that we can safely use subscriptions and to be sure that all notifications are up-to-date. First two options are harder to handle and look not natural to the neo-go itself. It's not clear to me why the client should fail the construction/requists just because the server is out of sync.

cthulhu-rider: Talking about async/await:
1. as an alternative to polling, I see a special  `network.Service` as a working option, the `Start` of which will mean the onset of synchronization (taking into account the always-on `StartWhenSynchronized` option). This seems a bit complicated, but polling is always less transparent approach overall to me.
2. if we choose polling approach: what value is better to use as a time period?

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -0,0 +1,505 @@
...
+	chErrRw := make(chan error)
+
+	go func() {
+		for {
+			err, ok := <-chErrRw
+			if !ok {
+				return
+			}
+			cfg.ErrorListener <- err
+		}
```
AnnaShaleva, line 446: How about exiting this loop on IR stop and draining/closing the channel afterwards? It's not so vital, but still.

config/example/ir.yaml
```diff
@@ -19,6 +19,59 @@ morph:
...
+      # Same format as 'p2p.listen'
+      - node2:20333
+      - node3:20333
+    hardforks: # Optional hard-forks
+      name: 1730000 # Maps name to chain height. Heights must not be greater than 4294967295
+    validators_history: # Optional number of consensus nodes to use after given height
+      10: 7 # Maps chain height to number of consensus nodes. Values must not be greater than 2147483647
+    native_activations: # Optional chronology of native contracts updates
+      GasToken: [0] # Maps name of Neo native contract to height list. Keys must be on of
+        # [].
```
AnnaShaleva, line 49: s/`Keys must be on of [].`/`Each key must be one of [NeoToken, ...].`?

config/example/ir.yaml
```diff
@@ -19,6 +19,59 @@ morph:
...
+      # Must not be greater than 4294967295
+    seed_nodes: # Optional list of existing nodes to communicate with over Neo P2P protocol. By default, node runs as standalone
+      # Same format as 'p2p.listen'
+      - node2:20333
+      - node3:20333
+    hardforks: # Optional hard-forks
+      name: 1730000 # Maps name to chain height. Heights must not be greater than 4294967295
+    validators_history: # Optional number of consensus nodes to use after given height
+      10: 7 # Maps chain height to number of consensus nodes. Values must not be greater than 2147483647
+    native_activations: # Optional chronology of native contracts updates
```
AnnaShaleva, line 47: This setting is treated as follows: if no contracts are specified, then the default value `[0]` will be applied for all known native contracts. If *some* of the contracts are specified, then the rest of them are treated as disabled and won't be deployed. So in the example config we'd better specify the whole list of native update histories with proper `[0]` value and add a note about it. Technically, current example config says: "deploy the only GasToken contract at the genezis " which is not the desired behaviour.

pkg/innerring/internal/blockchain/blockchain.go
cthulhu-rider: I thought about it. We provide writable channel to `rpcsrv.Server`, so technically we shouldn't close it to prevent potential panic (which is not critical problem, but not a good practice). Relying on the current design of the components, it is expected that when `Stop` is called, the channel will be closed inside the components based on this channel, and when it is closed, we exit the routine.

What do you think?

config/example/ir.yaml
cthulhu-rider: > treated as disabled and won't be deployed

Is it useful for Inner Ring? IMO it's pretty easy to forget some contract. What do you think? @AnnaShaleva @roman-khimov

At first glance, it seems that it is more convenient to configure custom heights partially, and default to `[0]` for the rest. But maybe I don't fully feel this mechanism.

pkg/innerring/internal/blockchain/blockchain.go
AnnaShaleva: > technically we shouldn't close it to prevent potential panic

It's a subject for discussion, but technically, we should firstly wait for the service to finish its operation (we have `rpcServer.Shutdown()` for this, it returns after service is completely closed) and then close the channel.

> the channel will be closed inside the components based on this channel

How about several services based on a single error channel? One service may be shutted down, but the others should still be able to use the provided channel. We have it in neo-go, BTW. 

In general, I think that the caller is resposible for resources release, but in some urgent and special cases the callee may be allowed to close the channel (like RPC client does with the subscriber's channel when it receives MissedEvent). I don't think it's a spacial case.

@roman-khimov, who is responsible for channel's closing, the caller or the callee?

config/example/ir.yaml
AnnaShaleva: > Is it useful for Inner Ring

It is not used anywhere even inside the Neo networks (at least as far as I know). Even with Notary contract there's a separate `P2PSigExtensions` that perfectly handles the default (missing) NativeActivations.

> it's pretty easy to forget some contract

Just leave this section completely emty then.

> it is more convenient to configure custom heights partially

Or do this logic in ir's config handler. We can't do this in neo-go because it's just the way how we specify disable contract.

pkg/innerring/internal/blockchain/blockchain.go
AnnaShaleva: It's all about `chErrRw`.

pkg/innerring/innerring.go
```diff
@@ -352,15 +346,88 @@ func New(ctx context.Context, log *logger.Logger, cfg *viper.Viper, errChan chan
...
+		// don't move this code area within current function
+		{
+			log.Info("running blockchain...")
+			// we run node here instead of Server.Start because some Server components
+			// require blockchain to be run in the current function. Later it would be nice
+			// to separate the stages of construction and launch
+			err = server.bc.Run()
+			if err != nil {
+				return nil, fmt.Errorf("run internal blockchain: %w", err)
+			}
```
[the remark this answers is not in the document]
cthulhu-rider: Agree, I'll add this code to `Run` regardless of current implementation.

pkg/innerring/internal/blockchain/blockchain.go
cthulhu-rider: > How about several services based on a single error channel?

IMO external channel should never be closed, otherwise `panic: send on closed channel` occurs. We should close only channels which are completely controlled by the single layer of code.

config/example/ir.yaml
cthulhu-rider: > Just leave this section completely emty then.

Sad, but I can't. Completely hide or completely clarify, no other options.

> Or do this logic in ir's config handler.

This replaces user error to IR dev's error, so finally I decided to just clarify your initial comment with a loud caution.

pkg/innerring/innerring.go
roman-khimov: 2 --- block time, as usual. I think it's OK for us to poll for now, blockchain threads will do their job in the background and then IR will kick in. Let's leave more complex approaches to #2294.

pkg/innerring/innerring.go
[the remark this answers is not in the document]
cthulhu-rider: Done

pkg/innerring/innerring.go
cthulhu-rider: Done

pkg/innerring/internal/blockchain/blockchain.go
cthulhu-rider: @roman-khimov thoughts?

AnnaShaleva (approved):
Good work!

I'll keep an eye on https://github.com/nspcc-dev/neofs-node/issues/2294 and we'll refactor neo-go's services start as far.

PR #2286, opened 2023-03-17
Merge 0.35.x support fixes into master

roman-khimov:
This one will be handled manually (because there is no other way really), the PR is mostly for informative purposes.


roman-khimov:
Huh. Pretty fun that the only real problem is CHANGELOG. Frankly, I've expected more conflicts than that.

cthulhu-rider (approved):

Issue #2226, opened 2023-01-31
Ban (?) tokens issued to self


cthulhu-rider:
Currently tokens (session and bearer) issued to self are not prohibited. Semantically this is _I am me, but not forever_. There are no statements in the NeoFS API on the incorrectness of such tokens, but they lead to redundant computations.

Possible treatments:
1. refuse to service such requests  (mb too harsh)
2. write a warning to the client (OK with warning) and hope he will fix this

At least, we can prevent such phenomena in the NeoFS SDK.

cthulhu-rider:
> How exactly can this be done?

This won't be an error: just a warn message attached to the response. A caring client will be able to diagnose this misbehavior. I would still try to highlight this kind of "strange" cases. Of course, it's better through a server error, but compatibility can really suffer. It's a dilemma for me right now.

> Do we have to do anything at all here?

Such case can be considered "too rare to care". I just suddenly encountered it in one project.

roman-khimov:
Can be related to #2228.

roman-khimov:
Warnings will be ignored and we can't error, so there is nothing to do here to me.

PR #2287, opened 2023-03-20
object: Allow self-signed session tokens

### LINKED ISSUE #2288: TestNet replication fails
TestNet nodes couldn't replicate following object from public container:
```
`error replicator/process.go:62 could not replicate object {"component": "Object Replicator", "node": "03000a222431891c481f2fbce61297547e816e164750f5ea94e2831c219bbe3504", "object": "GqNR1XtkucMcxE86FtALjVUSFVREa9ZMF2fbmeXEffJV/JDbZcmiZWCXRocjHU6i9ZHqJZzTDrhk2HEET33vxJJ6s", "error": "(*putsvc.RemoteSender) could not send object: (*putsvc.remoteTarget) could not put object to [/dns4/st4.t5.fs.neo.org/tcp/8080]: write object via client: status: code = 1024 message = (*object.FormatValidator) different owner identifiers NfFA5g92e32YvraVmipA9oY8AjvbdgFdjY/NTBbJxp2PrPXS7XyhKVqroifvsp1QCttif"}`
```

* related to #2226
* related to #2228

cthulhu-rider:
* related #2226
* closes #2288

This should fix TestNet replication problem.

### DIFF
--- a/pkg/core/object/fmt.go
+++ b/pkg/core/object/fmt.go
@@ -1,7 +1,6 @@
 package object
 
 import (
-	"crypto/ecdsa"
 	"errors"
 	"fmt"
 	"strconv"
@@ -14,7 +13,6 @@ import (
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"github.com/nspcc-dev/neofs-sdk-go/storagegroup"
-	"github.com/nspcc-dev/neofs-sdk-go/user"
 )
 
 // FormatValidator represents an object format validator.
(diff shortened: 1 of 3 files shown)


pkg/core/object/fmt.go
```diff
@@ -147,28 +145,11 @@ func (v *FormatValidator) validateSignatureKey(obj *object.Object) error {
 		return fmt.Errorf("decode public key: %w", err)
 	}
 
-	token := obj.SessionToken()
-
-	if token == nil || !token.AssertAuthKey(&key) {
-		return v.checkOwnerKey(*obj.OwnerID(), key)
```
roman-khimov, line 153: This raises two questions:
 * when objects in a public container are replicated, we routinely have signatures from different keys (not container owner key), how this is controlled wrt EACL? One object can be created when this is allowed to happen (EACL rule allowing PUT for a key exists), then rules might change. We want to replicate the first object, but if a malicious node creates an object signed by some random keys and tries to replicate it this shouldn't be possible. Granted, this is not exactly a session token problem, but it's related.
 * won't this change break `token == nil` case? random keys signing objects are not good, maybe we have a private container.

roman-khimov:
CHANGELOG conflict, rebase pls.

pkg/core/object/fmt.go
cthulhu-rider: One big pile of problems is that NeoFS doesn't distinguish object creation and regular transmission like replication. Such a division of operations would clearly improve access management and prevent potential vulnerabilities in the form of malicious data generation by storage nodes.

`token == nil` condition removal doesn't break IAM since it is contolled on the other layer. Technically, object owner may be any according to the protocol and specification.

roman-khimov (approved):

PR #2296, opened 2023-04-07
adm: fix wrong nnsResolveKey() code

roman-khimov:
I have no idea how it was tested/used, but NeoFS NNS contracts returns an array from resolve(), therefore this code just can't work correctly (assuming some byte string to returned directly).


cthulhu-rider (approved):

roman-khimov:
Neofs_Adm_Tests is broken, no green runs for the past ~six months.

PR #2297, opened 2023-04-07
adm: use network settings on upgrade, fix #2191

### LINKED ISSUE #2191: neofs-adm should take network settings into account during contract update
## Expected Behavior
`neofs-adm morph update-contracts` taking wallets/contracts/RPC parameters and doing its job.

## Current Behavior
On netmap contract upgrade any network settings are overwritten with some default ones unless they're given as a part of environment/config.

## Possible Solution
If we're upgrading we already have some network and it already has some parameters, so the tool should just take (via RPC) and use them, except for parameters that are new and where default values are to be used.

1. `neofs-adm morph update-contracts --alphabet-wallets ... --contracts ... -r`

## Context
Mainnet upgrade.

## Your Environment

v0.34.0

roman-khimov:
Not the most elegant piece of code, but it works for now, reducing potential for operator errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Changelog for NeoFS Node
 - Pilorama now can merge multiple batches into one (#2231)
 - Storage engine now can start even when some shard components are unavailable (#2238)
 - `neofs-cli` buffer for object put increased from 4 KiB to 3 MiB (#2243)
+- `neofs-adm` now reuses network's setting during contract update if they're not overriden by configuration (#2191)
 
 ### Fixed
 - Pretty printer of basic ACL in the NeoFS CLI (#2259)
--- a/cmd/neofs-adm/internal/modules/morph/initialize_deploy.go
+++ b/cmd/neofs-adm/internal/modules/morph/initialize_deploy.go
@@ -19,6 +19,7 @@ import (
 	"github.com/nspcc-dev/neo-go/pkg/rpcclient"
 	"github.com/nspcc-dev/neo-go/pkg/rpcclient/actor"
 	"github.com/nspcc-dev/neo-go/pkg/rpcclient/management"
+	"github.com/nspcc-dev/neo-go/pkg/rpcclient/unwrap"
 	"github.com/nspcc-dev/neo-go/pkg/smartcontract"
 	"github.com/nspcc-dev/neo-go/pkg/smartcontract/callflag"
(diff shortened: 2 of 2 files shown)


cmd/neofs-adm/internal/modules/morph/initialize_deploy.go
```diff
@@ -551,20 +552,43 @@ func (c *initializeContext) getContractDeployData(ctrName string, keysParam []in
...
+			{netmapAuditFeeKey, auditFeeInitFlag},
+			{netmapContainerFeeKey, containerFeeInitFlag},
+			{netmapContainerAliasFeeKey, containerAliasFeeInitFlag},
+			{netmapBasicIncomeRateKey, incomeRateInitFlag},
+			{netmapInnerRingCandidateFeeKey, candidateFeeInitFlag},
+			{netmapWithdrawFeeKey, withdrawFeeInitFlag},
+		} {
+			i64, err := unwrap.Int64(c.ReadOnlyInvoker.Call(ctrHash, "config", kf.key))
+			if err != nil {
+				i64 = viper.GetInt64(kf.flag)
```
cthulhu-rider, line 571: AFAIK netmap contract doesn't touch already set values. Maybe we can just check config setting via `viper.IsSet` and pass explicitly specified values only?

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Changelog for NeoFS Node
 - Pilorama now can merge multiple batches into one (#2231)
 - Storage engine now can start even when some shard components are unavailable (#2238)
 - `neofs-cli` buffer for object put increased from 4 KiB to 3 MiB (#2243)
+- `neofs-adm` now reuses network's setting during contract update if they're not overrided by configuration (#2191)
```
cthulhu-rider, line 20: overridden?

cmd/neofs-adm/internal/modules/morph/initialize_deploy.go
cthulhu-rider: Thanks, I think I got your point. 

Right now we must set all values explicitly since defaults are not defined. It's worth to think about them to move towards easier deployment

CHANGELOG.md
roman-khimov: Fixed.

cthulhu-rider (approved):

PR #2301, opened 2023-04-11
IR: Support embedded contracts with automatic deployment/updating

### LINKED ISSUE #2195: Bundle neofs-contracts into IR nodes
I'm always frustrated when I have to update NeoFS because we have neofs-contracts that are completely separate thing that no Ansible can help with. neofs-adm can help somewhat, but it still is an additional entity to think about (especially wrt compatibilty, is neofs-contracts X.Y.Z compatible with IR A.B.C?), while it's a completely internal NeoFS thing (implementation detail). It's too easy to forget to update them, it's too easy to make some mistake, it's hard to track these versions and relations to IR/SN versions (what version of the neofs-contracts the network N uses? Ansible doesn't help with that).

IR should have a proper version of neofs-contracts stored internally. It can be used for deployment, but we need to deal with updates first, IR should check contract versions during startup and if it has something newer initialize an update procedure (send appropriate transactions, or rather notary requests). So the end-user experience would be "update IR node binaries, done" (given #2194 is to be implemented).

Leaving things as is. Doesn't seem to be viable long-term.

cthulhu-rider:
* closes #2195

Currently rough draft, code will be improved later.

### DIFF
--- a/contracts/contracts.go
+++ b/contracts/contracts.go
@@ -1,0 +1,122 @@
+package contracts
+
+import (
+	"embed"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/fs"
+	"strconv"
+	"strings"
+
+	"github.com/nspcc-dev/neo-go/pkg/io"
+	"github.com/nspcc-dev/neo-go/pkg/smartcontract/manifest"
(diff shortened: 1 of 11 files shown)


cthulhu-rider:
everything has changed a lot

PR #2312, opened 2023-04-20
Add automated system tests run 

vvarg229:
This is my first experience with Gihub Actions.
I ask you to watch this PR carefully.
It's important that we do not use bad practices using Gihub Actions from the beginning, so I would welcome any feedback.


vvarg229:
I moved the branch from the fork to this repo:
https://github.com/nspcc-dev/neofs-node/pull/2313

PR #2335, opened 2023-05-11
Fix/NPE panic on tree services connection close

notimetoname:
Error can appear during new connection establishing and the `nil` connection is stored in LRU cache. Not try to close such connections.


roman-khimov:
Changelog, please.

roman-khimov (approved):

PR #2339, opened 2023-05-12
Fix/Iterating over just removed objects in FSTree

notimetoname:
I tried to write tests for it (i really did) and they do pass for `FSTree` and did not pass for blobovniczas cause it caches every object during the iteration and `LazyHandler` means nothing for them (in other words that fix just does not relate them). So i decided to keep it as is for now since it works and _the only one_ unit test for `FSTree` looks strange IMO, cause we use a general test "framework" for every storage.

### DIFF
--- a/pkg/local_object_storage/blobstor/fstree/fstree.go
+++ b/pkg/local_object_storage/blobstor/fstree/fstree.go
@@ -137,11 +137,19 @@ func (t *FSTree) iterate(depth uint64, curPath []string, prm common.IteratePrm) 
 
 		if prm.LazyHandler != nil {
 			err = prm.LazyHandler(*addr, func() ([]byte, error) {
-				return os.ReadFile(filepath.Join(curPath...))
+				data, err := os.ReadFile(filepath.Join(curPath...))
+				if err != nil && errors.Is(err, fs.ErrNotExist) {
+					return nil, logicerr.Wrap(apistatus.ObjectNotFound{})
+				}
+
+				return data, err
 			})
 		} else {
 			var data []byte
 			data, err = os.ReadFile(filepath.Join(curPath...))
+			if err != nil && errors.Is(err, fs.ErrNotExist) {
+				continue
+			}
 			if err == nil {
 				data, err = t.Decompress(data)
 			}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Changelog for NeoFS Node
 - Inability to restore RPC connection after the second disconnect (#2325)
 - Tree service panics when cleaning up failed connections (#2335)
 - Dropping small objects on any error on write-cache side (#2336)
+- Iterating over just removed files by FSTree (#98)
 
 ### Removed
 - Non-notary mode support for sidechain (#2321)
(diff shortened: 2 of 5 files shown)


pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -137,11 +137,19 @@ func (t *FSTree) iterate(depth uint64, curPath []string, prm common.IteratePrm)
 
 		if prm.LazyHandler != nil {
 			err = prm.LazyHandler(*addr, func() ([]byte, error) {
-				return os.ReadFile(filepath.Join(curPath...))
+				data, err := os.ReadFile(filepath.Join(curPath...))
+				if err != nil && os.IsNotExist(err) {
```
roman-khimov, line 141: https://pkg.go.dev/os#IsNotExist

> This function predates errors.Is. It only supports errors returned by the os package. New code should use errors.Is(err, fs.ErrNotExist). 

Not that it's critical for our usage pattern, but this patter might be reused by other code, so we better use proper APIs.

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -137,11 +137,19 @@ func (t *FSTree) iterate(depth uint64, curPath []string, prm common.IteratePrm)
...
+					return nil, logicerr.Wrap(apistatus.ObjectNotFound{})
```
roman-khimov, line 142: I'm really not sure this `ObjectNotFound` wrapping is good. What other `LazyHandler`s we have, can we have another source of `apistatus.ObjectNotFound` that won't be related to files missing between listing and action?

cthulhu-rider: currently all callers just skip such cases, so IMO it'd be easier to skip them here

pkg/local_object_storage/blobstor/fstree/fstree.go
notimetoname: sure, fixed

pkg/local_object_storage/blobstor/fstree/fstree.go
notimetoname: > I'm really not sure this ObjectNotFound wrapping is good.

@roman-khimov, why not? that is the idea of the `logicerr` (at least how i remembered it): every err that can be caused by non-disk errors should be _logic_ (and that line will be executed in the iterate-remove-read scenarios most likely). if drop it will be easy to make a bug and increment the internal disk errors counter and move a shard to the degraded mode

> What other LazyHandlers we have

WC only (blobovnicza can be called with that func too but it caches objects so it is impossible to `stat`+`read`+"oopps")

> can we have another source of apistatus.ObjectNotFound that won't be related to files missing between listing and action?

`Get`/etc only. not in the `Iterate`

roman-khimov: To me `apistatus.ObjectNotFound` is something that we return for networked requests, not something that we can have locally. At the same time if we're to ever have any other non-FS backends, there may be no `fs.ErrNotExist`, something else would be returned and then we'll have to unify them with some common error anyway.

notimetoname: > To me apistatus.ObjectNotFound is something that we return for networked requests, not something that we can have locally.

i do agree that it can be revised but now that is what we have: SE returns `apistatus` errors. do not have a strict opinion about that RN

Issue #2319, opened 2023-04-25
Update Go version (drop 1.17, add 1.20)


roman-khimov:
* drop 1.17 support
* fix workflows to use 1.20
* replace interface{} with any

Like https://github.com/nspcc-dev/neo-go/pull/2957 or https://github.com/nspcc-dev/neofs-sdk-go/pull/394.

roman-khimov:
All regular builds should use 1.20 now, pay attention to Docker files and GO_VERSION.

PR #2340, opened 2023-05-12
Upd/Generics have landed

### LINKED ISSUE #2181: Use generics for morph caches
We cache containers, container ids, eACLs and IR nodes. Casting all of them to `ttlNetCache` is a bit ugly, we could use generics after min go version is bumped.

notimetoname:
Closes #2181. Blocked by #2342.

### DIFF
--- a/cmd/neofs-node/cache.go
+++ b/cmd/neofs-node/cache.go
@@ -4,7 +4,7 @@ import (
 	"sync"
 	"time"
 
-	lru "github.com/hashicorp/golang-lru"
+	lru "github.com/hashicorp/golang-lru/v2"
 	"github.com/nspcc-dev/neofs-node/pkg/core/container"
 	"github.com/nspcc-dev/neofs-node/pkg/core/netmap"
 	cntClient "github.com/nspcc-dev/neofs-node/pkg/morph/client/container"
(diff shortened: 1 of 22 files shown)


roman-khimov:
See #2319 for things to fix when dropping 1.17 support. There is #2181, but I'm not so immediately sure about that.

roman-khimov:
Can we split it to #2319 and #2181 parts?

notimetoname:
#2342 is done. Lets have generic fun.

cmd/neofs-node/cache.go
```diff
@@ -53,47 +53,45 @@ func newNetworkTTLCache(sz int, ttl time.Duration, netRdr netValueReader) *ttlNe
...
 
-func (c *ttlNetCache) remove(key interface{}) {
+func (c *ttlNetCache[K, V]) remove(key K) {
 	c.cache.Remove(key)
 }
 
 // entity that provides LRU cache interface.
 type lruNetCache struct {
-	cache *lru.Cache
+	cache *lru.Cache[uint64, *netmapSDK.NetMap]
```
roman-khimov, line 87: As much as I'm skeptical about generics, this probably makes things more readable.

cmd/neofs-node/cache.go
```diff
@@ -15,32 +15,32 @@ import (
...
 
 // entity that provides TTL cache interface.
-type ttlNetCache struct {
+type ttlNetCache[K comparable, V any] struct {
 	ttl time.Duration
 
 	sz int
 
-	cache *lru.Cache
+	cache *lru.Cache[K, *valueWithTime[V]]
```
roman-khimov, line 33: While this maybe not so much.

cmd/neofs-node/cache.go
```diff
@@ -125,73 +123,63 @@ func (c *lruNetCache) get(key interface{}) (interface{}, error) {
...
 // that implements container storage.
-type ttlContainerStorage ttlNetCache
+type ttlContainerStorage struct {
+	tc ttlNetCache[cid.ID, *container.Container]
+}
 
 func newCachedContainerStorage(v container.Source, ttl time.Duration) *ttlContainerStorage {
 	const containerCacheSize = 100
 
-	lruCnrCache := newNetworkTTLCache(containerCacheSize, ttl, func(key interface{}) (interface{}, error) {
```
roman-khimov, line 133: Changes like this are good as well.

roman-khimov (commented):
Can't say I'm impressed with the diffstat though (`+137 −159`), my expectations were somewhat higher. Something tells me we'll need to revise these caches anyway.

Please add #2181 fix/close somewhere.

notimetoname:
> Please add https://github.com/nspcc-dev/neofs-node/issues/2181 fix/close somewhere.

@roman-khimov, added to PR's description. It is 8 commits now, not sure which of them should have that link (or all?).

> Can't say I'm impressed with the diffstat though (+137 −159), my expectations were somewhat higher.

In fact, mine was too. But i like that we do not have casts (ok, "assertions") in that code more, i find that parentheses dance (4+ in one expression) hard to read. Also, i would say that they just do their work, and if it can be done with a small amount of code it should be done. In other words, i would say that our caches are too complex, so generics do not make it shorter where nothing relates to "generic code".

Anyway, i think, LRU caches will be updated to `v2` somehow, so the issue says to update it, that PR updates it. Any other refactor should be considered in a separate way (can create an issue about that).

roman-khimov (approved):

PR #2342, opened 2023-05-15
Upd/Go version to `1.18`

### LINKED ISSUE #2319: Update Go version (drop 1.17, add 1.20)

notimetoname:
Closes #2319.

@roman-khimov, @cthulhu-rider, @AnnaShaleva, @smallhive, when generics came out i found some interesting conv about it and now im recalling some arg about that: https://github.com/golang/go/issues/33232#issuecomment-915333205 (especially, `4.`). Do you have any thoughts about that?

Well, i think most of the projs just move `interface{}` to `any` so i do not mind just doing the same. Just want to share a little bit of concern about it.


roman-khimov:
I don't see any problem with `any` in non-generic code, it's just a syntactic sugar and we have waited long enough to say bye-bye to 1.17 and be able to use it. The question of generic-related refactorings is orthogonal to adding some sugar to the syntax style we have. It may be nice in some aspects, it be not so much in others (where regular, but more specific interfaces can still be useful).

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ Changelog for NeoFS Node
 - Non-notary mode support for sidechain (#2321)
 
 ### Updated
+- Update go version up to v1.18 (#2340)
```
roman-khimov, line 19: `minimal supported Go version`

roman-khimov (approved):

notimetoname:
> It may be nice in some aspects, it be not so much in others (where regular, but more specific interfaces can still be useful).

@roman-khimov, yes, i meant that revising process (and a refactor itself) could be more complex if we just move to `any` now, and only after that will think (and will search for) every case where they can be applied.
Once again, just merging that PR is OK for me.

roman-khimov (approved):

PR #2344, opened 2023-05-15
Feat/Add version metrics

### LINKED ISSUE #2326: add version metric
like this one:
```
vm_app_version{version="vmagent-20230407-010133-tags-v1.90.0-0-gb5d18c0d2", short_version="v1.90.0"} 1
```

notimetoname:
Closes #2326. Gonna look smth like that:

```
# HELP neofs_node_build_info 
# TYPE neofs_node_build_info gauge
neofs_node_build_info{version="v0.36.1-32-gb20018c6-dirty"} 1
```

@roman-khimov, @cthulhu-rider, @532910, @aprasolova, @anyone_looking_at_metrics, please, look at the first commit. Something was certainly wrong, pay attention to it and make me know if some SW depends on it and we are not able to make such changes OR such changes must be synced before the merge.

### DIFF
--- a/pkg/metrics/innerring.go
+++ b/pkg/metrics/innerring.go
@@ -1,33 +1,49 @@
 package metrics
 
-import "github.com/prometheus/client_golang/prometheus"
+import (
+	"fmt"
 
-const innerRingSubsystem = "object"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const innerRingNameSpace = "neofs_ir"
+
+// FIXME: drop after v0.38.0 release: #2347.
+const innerRingNameSpaceDeprecated = storageNodeNameSpace
+const innerRingSubsystemDeprecated = objectSubsystem
 
 // InnerRingServiceMetrics contains metrics collected by inner ring.
 type InnerRingServiceMetrics struct {
-	epoch prometheus.Gauge
+	epoch           prometheus.Gauge
+	epochDeprecated prometheus.Gauge
 }
 
 // NewInnerRingMetrics returns new instance of metrics collectors for inner ring.
 func NewInnerRingMetrics() InnerRingServiceMetrics {
-	var (
-		epoch = prometheus.NewGauge(prometheus.GaugeOpts{
-			Namespace: namespace,
-			Subsystem: innerRingSubsystem,
-			Name:      "epoch",
-			Help:      "Current epoch as seen by inner-ring node.",
-		})
-	)
-
+	epoch := prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: innerRingNameSpace,
+		Subsystem: stateSubsystem,
+		Name:      "epoch",
+		Help:      "Current epoch as seen by inner-ring node.",
+	})
 	prometheus.MustRegister(epoch)
 
+	epochDeprecated := prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: innerRingNameSpaceDeprecated,
+		Subsystem: innerRingSubsystemDeprecated,
+		Name:      "epoch",
+		Help:      fmt.Sprintf("Current epoch as seen by inner-ring node. DEPRECATED: use [%s_%s_epoch] instead.", innerRingNameSpace, stateSubsystem),
+	})
+	prometheus.MustRegister(epochDeprecated)
+
 	return InnerRingServiceMetrics{
-		epoch: epoch,
+		epoch:           epoch,
+		epochDeprecated: epochDeprecated,
 	}
 }
 
 // SetEpoch updates epoch metrics.
 func (m InnerRingServiceMetrics) SetEpoch(epoch uint64) {
 	m.epoch.Set(float64(epoch))
+	m.epochDeprecated.Set(float64(epoch))
 }
--- a/pkg/metrics/innerring.go
+++ b/pkg/metrics/innerring.go
@@ -19,7 +19,9 @@ type InnerRingServiceMetrics struct {
 }
 
 // NewInnerRingMetrics returns new instance of metrics collectors for inner ring.
-func NewInnerRingMetrics() InnerRingServiceMetrics {
+func NewInnerRingMetrics(version string) InnerRingServiceMetrics {
+	registerVersionMetric(innerRingNameSpace, version)
+
 	epoch := prometheus.NewGauge(prometheus.GaugeOpts{
 		Namespace: innerRingNameSpace,
 		Subsystem: stateSubsystem,
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -16,7 +16,9 @@ type NodeMetrics struct {
 	epochDeprecated prometheus.Gauge
 }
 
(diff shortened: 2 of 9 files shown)


pkg/metrics/innerring.go
```diff
@@ -13,8 +13,8 @@ type InnerRingServiceMetrics struct {
 func NewInnerRingMetrics() InnerRingServiceMetrics {
 	var (
 		epoch = prometheus.NewGauge(prometheus.GaugeOpts{
-			Namespace: namespace,
-			Subsystem: innerRingSubsystem,
+			Namespace: innerRingNameSpace,
+			Subsystem: stateSubsystem,
```
roman-khimov, line 19: Whoa. Not that easy. That's an external node interface, we need to keep backwards compatibility or _all_ of the monitors/alerts will break with the new release.

pkg/metrics/build_info.go
```diff
@@ -0,0 +1,16 @@
+package metrics
+
+import "github.com/prometheus/client_golang/prometheus"
+
+func registerVersionMetric(namespace string, version string) {
+	g := prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: namespace,
+		Name:      "build_info",
```
roman-khimov, line 8: Why not just `version`? Many other components use this name and I think we better have some common one.

pkg/metrics/innerring.go
notimetoname: yes but:

1. we still tag out releases with as v**0**.X.Y;
2. we have `Updating from` sections for such things (at least we _had_ for such things);
3. we already did CLI args, config changes (`metrics` -> `prometheus`, `profile` -> `pprof`);
4. i have no explanations for a user why i give him such metrics:
```
# HELP neofs_node_object_epoch Current epoch as seen by inner-ring node.
# TYPE neofs_node_object_epoch gauge
neofs_node_object_epoch 3
```
(`object`)
```
# HELP neofs_node_object_epoch Current epoch as seen by inner-ring node.
# TYPE neofs_node_object_epoch gauge
neofs_node_object_epoch 6
```
(that is IR metrics but still `object` and also `neofs_node` (i remember no place where i can read "neofs node" and think about IR))

so i would classify that rather as a bug not a breaking change (commit that adds that const does not look like a decision, more like a mistake)

pkg/metrics/build_info.go
notimetoname: i had nothing to be expired by so i decided to go to the origins directly: https://github.com/prometheus/common/blob/2f0f5c4b6d958963f2a8d5447ad27ca7a931dcda/version/info.go#L38-L61 and https://github.com/prometheus/client_golang/blob/297fea317b926e4bb8d3c72022e172def2f68ed2/prometheus/metric.go#L89-L92

in other words, that is how prometheus dev would do such things _in my interpretation_. more strict args i would say sound like: const "build_info" is the only metric that is presented here (in the PR) and will be presented forever, it contains any needed build info but also it will be easy to expand it with any new build info cause build attrs are just labels and not smth that encoded in a metric name (comparing with `neofs_ir_version: 1`).
what additional info? well, even the issue that PR closes has two suggestion (`version` and `short_version`). i would also think about `go_version` (ok, we have another metric for that but i can imagine that one metric for everything may be more convenient), `build_time`, etc

pkg/metrics/innerring.go
roman-khimov: 1. Sure, but this doesn't really matter. When something is used, you can't easily break users assumptions about how it works unless you absolutely have to. NeoGo is 0.smth as well, but there is a https://github.com/nspcc-dev/neo-go/blob/master/ROADMAP.md#deprecated-functionality
2. Sure, we still do. And what I'd expect there is "please convert to this new smth before we drop it" and in some later version "btw, if you're still using that old smth, sorry, it's gone now".
3. And we've had a number of problems with these transitions with lots of broken configurations. infra-repo remembers a number of related commits fixing things months after the upgrade.
4. Yeah, that's a problem, but irrespective of how these metrics are called, people already use them. And they'd be somewhat disappointed to hear they need to change everything immediately after the upgrade.

It's OK to clean these things up, but give everyone some transition period. Most of the time it doesn't cost a lot, but people would be able to upgrade their networks with old configs/monitoring setups, ensure they work fine and then change whatever is needed to change. Otherwise it's too many things to handle at once during the upgrade.

pkg/metrics/build_info.go
roman-khimov: `GaugeVec` is OK and it can be extended if need be, it's just that we already use `version` for it in
 * nspcc-dev/neo-go#3009
 * nspcc-dev/neofs-http-gw#251
 * nspcc-dev/neofs-s3-gw#779

So I'd rather continue using it instead of introducing `build_info`. Nothing bad about the name itself except that we're already using a different one. And some third-party apps do that too, so it's not something surprisingly original here.

notimetoname: really sad :(
ok, i'll unify

pkg/metrics/innerring.go
notimetoname: oh, so you were talking about deprecating them? i might get you wrong

notimetoname: ok, i got you wrong: i thought that you did not want to remove anything. fixed.

cthulhu-rider (approved):
out-of-context, code ok

pkg/metrics/innerring.go
```diff
@@ -2,32 +2,44 @@ package metrics
...
+		Name:      "epoch",
+		Help:      "Current epoch as seen by inner-ring node.",
+	})
 	prometheus.MustRegister(epoch)
 
+	epochDeprecated := prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: innerRingNameSpaceDeprecated,
+		Subsystem: innerRingSubsystemDeprecated,
+		Name:      "epoch",
+		Help:      "Current epoch as seen by inner-ring node.",
```
roman-khimov, line 33: Mark as deprecated in the help?

pkg/metrics/metrics.go
```diff
@@ -22,13 +23,23 @@ func NewNodeMetrics() *NodeMetrics {
...
 	})
 	prometheus.MustRegister(epoch)
 
+	// FIXME: drop after v0.38.0 release: #2347.
+	const stateSubsystemDeprecated = objectSubsystem
+	epochDeprecated := prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: storageNodeNameSpace,
+		Subsystem: stateSubsystemDeprecated,
+		Name:      "epoch",
+		Help:      "Current epoch as seen by inner-ring node.",
```
roman-khimov, line 41: Ditto.

pkg/metrics/innerring.go
notimetoname: sure (the line is too long intentionally if you dont mind)

pkg/metrics/metrics.go
notimetoname: sure surely

roman-khimov (approved):

PR #2345, opened 2023-05-16
Feat/Allow RPC over TLS serving for IR with local consensus

### LINKED ISSUE #2322: TLS for local consensus in Inner Ring
NeoGo node is able to serve TLS endpoint while Inner Ring node with local consensus isn't.

## Expected Behavior
I can configure the TLS endpoint for blockchain in IR.

## Current Behavior
Consesnsus endpoints in IR node serve without TLS.

notimetoname:
Closes #2322.

### DIFF
--- a/config/example/ir.yaml
+++ b/config/example/ir.yaml
@@ -48,6 +48,14 @@ morph:
         # TCP addresses in 'host:port' format
         - localhost:30333
         - localhost:30334
+      tls: # Additional addresses to listen to using TLS setup; must not overlap with `listen` section
+        enabled: false # Additional TLS serving switcher
+        listen: # Addresses to listen to; required to be at least 1-length if 'enabled' is 'true'
+          # TCP addresses in 'host:port' format
+          - localhost:30335
+          - localhost:30336
+        cert_file: serv.crt # TLS certificate file path; required if 'enabled' is 'true'
+        key_file: serv.key # TLS private key file path; required if 'enabled' is 'true'
     p2p: # Optional P2P settings
       dial_timeout: 5s # Optional maximum duration a single peer dial may take. Defaults to 5s. Must not be negative
       proto_tick_interval: 2s # Optional time period between protocol ticks with each connected peer. Defaults to 2s.
--- a/pkg/innerring/config.go
+++ b/pkg/innerring/config.go
@@ -181,9 +181,31 @@ func parseBlockchainConfig(v *viper.Viper, _logger *logger.Logger) (c blockchain
 
 	const rpcSection = rootSection + ".rpc"
 	if v.IsSet(rpcSection) {
-		c.RPCListenAddresses, err = parseConfigAddressesTCP(v, rpcSection+".listen", "network addresses to listen Neo RPC on")
+		c.RPC.Addresses, err = parseConfigAddressesTCP(v, rpcSection+".listen", "network addresses to listen insecure Neo RPC on")
 		if err != nil && !errors.Is(err, errMissingConfig) {
 			return c, err
+		}
+
+		const rpcTLSSection = rpcSection + ".tls"
+		if v.GetBool(rpcTLSSection + ".enabled") {
+			c.RPC.TLSConfig.Enabled = true
+
+			c.RPC.TLSConfig.Addresses, err = parseConfigAddressesTCP(v, rpcTLSSection+".listen", "network addresses to listen to Neo RPC over TLS")
+			if err != nil {
+				return c, err
+			}
+
+			const certCfgKey = rpcTLSSection + ".cert_file"
+			c.RPC.TLSConfig.CertFile = v.GetString(certCfgKey)
+			if strings.TrimSpace(c.RPC.TLSConfig.CertFile) == "" {
+				return c, fmt.Errorf("RPC TLS setup is enabled but no certificate ('%s') is provided", certCfgKey)
+			}
+
+			const keyCfgKey = rpcTLSSection + ".key_file"
+			c.RPC.TLSConfig.KeyFile = v.GetString(keyCfgKey)
+			if strings.TrimSpace(c.RPC.TLSConfig.KeyFile) == "" {
+				return c, fmt.Errorf("RPC TLS setup is enabled but no key ('%s') is provided", keyCfgKey)
+			}
 		}
 	}
 
--- a/pkg/innerring/internal/blockchain/blockchain.go
+++ b/pkg/innerring/internal/blockchain/blockchain.go
@@ -91,6 +91,42 @@ type PingConfig struct {
 	Timeout time.Duration
 }
 
+// RPCConfig configures RPC serving.
+type RPCConfig struct {
+	// Network addresses to listen Neo RPC on. Each element must be a valid TCP
+	// address in 'host:port' format.
+	//
+	// Optional: by default, insecure Neo RPC is not served.
+	Addresses []string
+
+	// Additional addresses that use TLS.
+	//
+	// Optional.
+	TLSConfig
+}
+
+// TLSConfig configures additional RPC serving over TLS.
+type TLSConfig struct {
+	// Additional TLS serving switcher.
+	//
+	// Optional: by default TLS is switched off.
+	Enabled bool
+
+	// Network addresses to listen Neo RPC on if Enabled. Each element must be a valid TCP
+	// address in 'host:port' format.
+	Addresses []string
+
+	// TLS certificate file path.
+	//
+	// Required if Enabled and one or more addresses are provided.
+	CertFile string
+
+	// TLS private key file path.
+	//
+	// Required if Enabled and one or more addresses are provided.
+	KeyFile string
+}
+
 // P2PConfig configures communication over Neo P2P protocol.
 type P2PConfig struct {
 	// Specifies the minimum number of peers a node needs for normal operation.
@@ -165,11 +201,10 @@ type Config struct {
 	// Optional: defaults to 15s. Must not be negative.
 	BlockInterval time.Duration
 
-	// Network addresses to listen Neo RPC on. Each element must be a valid TCP
-	// address in 'host:port' format.
+	// Neo RPC service configuration.
 	//
-	// Optional: by default, Neo RPC is not served.
-	RPCListenAddresses []string
+	// Optional: see RPCConfig defaults.
+	RPC RPCConfig
 
 	// Length of the chain accessible to smart contracts.
 	//
@@ -333,8 +368,6 @@ func New(cfg Config) (res *Blockchain, err error) {
 	cfgBaseApp.Consensus.UnlockWallet = cfg.Wallet
 	cfgBaseApp.P2PNotary.Enabled = true
 	cfgBaseApp.P2PNotary.UnlockWallet = cfg.Wallet
-	cfgBaseApp.RPC.Enabled = true
-	cfgBaseApp.RPC.Addresses = cfg.RPCListenAddresses
 	cfgBaseApp.RPC.StartWhenSynchronized = true
 	cfgBaseApp.RPC.MaxGasInvoke = fixedn.Fixed8FromInt64(100)
 	cfgBaseApp.P2P.Addresses = cfg.P2P.ListenAddresses
@@ -345,6 +378,15 @@ func New(cfg Config) (res *Blockchain, err error) {
 	cfgBaseApp.P2P.MinPeers = int(cfg.P2P.MinPeers)
 	cfgBaseApp.P2P.AttemptConnPeers = int(cfg.P2P.AttemptConnPeers)
 	cfgBaseApp.P2P.MaxPeers = int(cfg.P2P.MaxPeers)
+
+	cfgBaseApp.RPC.Enabled = true
+	cfgBaseApp.RPC.Addresses = cfg.RPC.Addresses
+	if tlsCfg := cfg.RPC.TLSConfig; tlsCfg.Enabled {
+		cfgBaseApp.RPC.TLSConfig.Enabled = true
+		cfgBaseApp.RPC.TLSConfig.Addresses = tlsCfg.Addresses
+		cfgBaseApp.RPC.TLSConfig.CertFile = tlsCfg.CertFile
+		cfgBaseApp.RPC.TLSConfig.KeyFile = tlsCfg.KeyFile
+	}
 
 	err = cfgBase.ProtocolConfiguration.Validate()
 	if err != nil {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Changelog for NeoFS Node
 
 ### Added
 - `neofs_[node|ir]_version` with `version` label (#2326)
+- TLS RPC support for IR consensus mode (#2322)
 
 ### Changed
 
@@ -28,6 +29,8 @@ Changelog for NeoFS Node
 - `neofs_node_object_epoch` metric for IR and SN (the same for both) has been deprecated and will be removed with the
   next minor release. Use `neofs_node_state_epoch` for SN and `neofs_ir_state_epoch` for IR instead.
 - Storage and Inner-ring nodes exposes their version via the `neofs_[node|ir]_version` metric now.
+- In the local consensus mode (IR) it is allowed to provide additional TLS setup addresses now, see
+  `morph.consensus.rpc.tls` section.
 
 ## [0.36.1] - 2023-04-26
 
--- a/pkg/innerring/config_test.go
+++ b/pkg/innerring/config_test.go
@@ -57,6 +57,13 @@ const validConfigOptions = `
       listen:
         - localhost:30000
         - localhost:30001
+      tls:
+        enabled: true
+        listen:
+          - localhost:30002
+          - localhost:30003
+        cert_file: /path/to/cert
+        key_file: /path/to/key
     p2p:
       dial_timeout: 111s
       proto_tick_interval: 222s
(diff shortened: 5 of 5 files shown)


pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -91,6 +91,42 @@ type PingConfig struct {
...
+	// Optional.
+	TLSConfig
+}
+
+// TLSConfig configures additional RPC serving over TLS.
+type TLSConfig struct {
+	// Additional TLS serving switcher.
+	//
+	// Optional: by default TLS is switched off.
+	Enabled bool
```
cthulhu-rider, line 113: even more stange to see it here, I'd determine extension by `Addresses` field emptiness

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -91,6 +91,42 @@ type PingConfig struct {
...
+
+	// Network addresses to listen Neo RPC on if Enabled. Each element must be a valid TCP
+	// address in 'host:port' format.
+	Addresses []string
+
+	// TLS certificate file path.
+	//
+	// Required if Enabled and more than one address is provided.
```
cthulhu-rider, line 121: what if single address is provided? I don't see this statement in `config/example/ir.yaml`

if it's required then `New` must assert this

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -91,6 +91,42 @@ type PingConfig struct {
 	Timeout time.Duration
 }
 
+// RPCConfig configures RPC serving.
+type RPCConfig struct {
+	// Network addresses to listen Neo RPC on. Each element must be a valid TCP
+	// address in 'host:port' format.
+	//
+	// Optional: by default, Neo RPC is not served.
```
cthulhu-rider, line 99: can we provide TLS interface only? If no, then correct, otherwise
```suggestion
	// Optional: by default, insecure Neo RPC is not served.
```

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -165,11 +201,10 @@ type Config struct {
 	// Optional: defaults to 15s. Must not be negative.
 	BlockInterval time.Duration
 
-	// Network addresses to listen Neo RPC on. Each element must be a valid TCP
-	// address in 'host:port' format.
+	// Network addresses to listen Neo RPC on.
```
cthulhu-rider, line 204:
```suggestion
	// Neo RPC service configuration.
	//
	// Optional: see RPCConfig defaults.
```

cthulhu-rider (commented):
overall correct, lets discuss some cosmetics

config/example/ir.yaml
```diff
@@ -48,6 +48,14 @@ morph:
         # TCP addresses in 'host:port' format
         - localhost:30333
         - localhost:30334
+      tls_config: # Additional addresses to listen to using TLS setup; must not overlap with `listen` section
+        enabled: false # Additional TLS serving switcher
```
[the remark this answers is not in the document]
roman-khimov: We have the same flag for GRPC TLS and Prometheus service, for example, so we better be consistent and allow it everywhere appropriate.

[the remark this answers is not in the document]
cthulhu-rider: agree with consistency, but the point remains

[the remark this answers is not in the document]
roman-khimov: Admins find it useful, it requires less editing to quickly temporarily enable/disable some component when needed.

pkg/innerring/internal/blockchain/blockchain.go
AnnaShaleva: IMO, it's better to have an option here and to keep insecure connections availability.

pkg/innerring/internal/blockchain/blockchain.go
AnnaShaleva: It should be required even if only one address is provided, isn't it?

config/example/ir.yaml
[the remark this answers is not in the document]
cthulhu-rider: toggling is more rare case (not sure if it's actually used at all) than configuring from scratch. As node admin, I'd like to write endpoints/filepaths and that's it, writing additional `enabled: true` is an extra routine to me. If admin forgets to write `enabled`, then TLS isn't served at all even with specified endpoints and certificate - aso pretty strange.

i don't mind to have `enabled` flag, but still for the most dry config without dualities

pkg/innerring/internal/blockchain/blockchain.go
notimetoname: no problemo

pkg/innerring/internal/blockchain/blockchain.go
notimetoname: > can we provide TLS interface only?

i think yes, why not? (in fact, i expect it to be either TLS only or non-TLS only)

pkg/innerring/internal/blockchain/blockchain.go
notimetoname: it is bad wording in fact; that is not smth that i meant, fixed

> if it's required then New must assert this

well, do not agree with you totally cause then it even can go deeper (check if the file exists, check if it _is a key/cert_, etc) but fixed

pkg/innerring/internal/blockchain/blockchain.go
notimetoname: we have `neo-go` [struct](https://github.com/nspcc-dev/neo-go/blob/d463bad96fa4dd6216f7609455d77d47ea4e4494/pkg/config/basic_service.go#L10) that should be filled. usually, i try to go 1:1 correspondence for better readability IMO (anyway we must fill the `Enabled` field on `neo-go` side)

config/example/ir.yaml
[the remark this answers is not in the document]
notimetoname: i took it from `neo-go` and also took part in the same discussion with the same agrs once

the same as in the thread above: i agree with both of you so i am OK with any case @roman-khimov, @AnnaShaleva @cthulhu-rider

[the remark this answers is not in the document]
notimetoname: or even @aprasolova, @532910

config/example/ir.yaml
```diff
@@ -48,6 +48,14 @@ morph:
         # TCP addresses in 'host:port' format
         - localhost:30333
         - localhost:30334
+      tls_config: # Additional addresses to listen to using TLS setup; must not overlap with `listen` section
```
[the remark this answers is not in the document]
roman-khimov: `tls` is OK, we have it this way in NeoGo since like forever (https://github.com/nspcc-dev/neo-go/commit/dfbb84ef381d9abf554ec60746c08747c7518170) but `tls` is easily recognizable (unlike SomeNeoProtocolParameters). Maybe it can be improved along with nspcc-dev/neo-go#2901.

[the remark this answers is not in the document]
notimetoname: fixed

pkg/innerring/config.go
```diff
@@ -181,10 +181,34 @@ func parseBlockchainConfig(v *viper.Viper, _logger *logger.Logger) (c blockchain
...
+			c.RPC.TLSConfig.Enabled = true
+
+			c.RPC.TLSConfig.Addresses, err = parseConfigAddressesTCP(v, rpcTLSSection+".listen", "network addresses to listen to Neo RPC over TLS")
+			if err != nil {
+				return c, err
+			}
+
+			const certCfgKey = rpcTLSSection + ".cert_file"
+			if !v.IsSet(certCfgKey) {
+				return c, fmt.Errorf("RPC TLS setup is enabled but no certificate ('%s') is provided", certCfgKey)
```
roman-khimov, line 200: Do you have tests for these things?

pkg/innerring/internal/blockchain/blockchain.go
roman-khimov: Can be any way: TLS/non-TLS/both.

roman-khimov (approved):

config/example/ir.yaml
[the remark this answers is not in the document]
532910: Port config looks like a trash overall, why tls is inside non tls? What is reason to listen multiple ports with the same cert? I believe my opinion doesn't matter, please don't mention me.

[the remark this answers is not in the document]
roman-khimov: 1. It's a service in the first place, it can have some service-specific settings applicable to _any_ communication channel, that's why `service.tls` scheme.
2. I know about nspcc-dev/neo-go#2901, but that's not the way RPC is configured now, IR can't provide this functionality until NeoGo has it.
3. I don't care much about beliefs, but I care about valid feedback from product users. You're one of them so you'll be asked anyway.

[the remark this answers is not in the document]
532910: I do not use this product, sorry ):

TLS may have a toggle, if it's an option, like in synapse `listeners` or ejabberd `listen` section.

If I'm a user, please, make the config clear to users, like synapse or ejabberd for example.

pkg/innerring/internal/blockchain/blockchain.go
cthulhu-rider: readability comes with duality when `Enabled == false` and `len(Addresses) == 0`? I guess no, lets just remove the flag and determine behavior by the state of one field - `Addresses`

pkg/innerring/internal/blockchain/blockchain.go
cthulhu-rider: yes, check all this stuff. Docs are written not to be nice but have strict behavior

pkg/innerring/internal/blockchain/blockchain.go
cthulhu-rider: then my change suggestion is valid, Neo RPC can be served with TLS even if `Addresses` is empty

pkg/innerring/config.go
```diff
@@ -181,10 +181,34 @@ func parseBlockchainConfig(v *viper.Viper, _logger *logger.Logger) (c blockchain
...
 	if v.IsSet(rpcSection) {
-		c.RPCListenAddresses, err = parseConfigAddressesTCP(v, rpcSection+".listen", "network addresses to listen Neo RPC on")
+		c.RPC.Addresses, err = parseConfigAddressesTCP(v, rpcSection+".listen", "network addresses to listen Neo RPC on")
 		if err != nil && !errors.Is(err, errMissingConfig) {
 			return c, err
 		}
+
+		const rpcTLSSection = rpcSection + ".tls"
+		if v.GetBool(rpcTLSSection + ".enabled") {
+			c.RPC.TLSConfig.Enabled = true
```
cthulhu-rider, line 191: pretty useless, `c.RPC.TLSConfig.Addresses` is required to be non-empty and that's enough to determine the behavior

pkg/innerring/config.go
```diff
@@ -181,10 +181,34 @@ func parseBlockchainConfig(v *viper.Viper, _logger *logger.Logger) (c blockchain
...
+			if err != nil {
+				return c, err
+			}
+
+			const certCfgKey = rpcTLSSection + ".cert_file"
+			if !v.IsSet(certCfgKey) {
+				return c, fmt.Errorf("RPC TLS setup is enabled but no certificate ('%s') is provided", certCfgKey)
+			}
+
+			c.RPC.TLSConfig.CertFile = v.GetString(certCfgKey)
```
cthulhu-rider, line 203: required if `c.RPC.TLSConfig.Addresses`, so check emptiness, `c.RPC.TLSConfig.KeyFile` too

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -346,6 +379,15 @@ func New(cfg Config) (res *Blockchain, err error) {
 	cfgBaseApp.P2P.AttemptConnPeers = int(cfg.P2P.AttemptConnPeers)
 	cfgBaseApp.P2P.MaxPeers = int(cfg.P2P.MaxPeers)
 
+	cfgBaseApp.RPC.Enabled = true
```
cthulhu-rider, line 382: it's not always enabled

cthulhu-rider (changes requested):

pkg/innerring/config.go
notimetoname: > pretty useless, `c.RPC.TLSConfig.Addresses` is required to be non-empty

why?

pkg/innerring/config.go
notimetoname: added blank string check

pkg/innerring/internal/blockchain/blockchain.go
notimetoname: fixed

pkg/innerring/internal/blockchain/blockchain.go
notimetoname: i meant what if it _is set_ and its value is `""`? what if `" "`? `blockchain` either can use provided config, or returns `error` from its methods, why perform checks that will be performed anyway?

pkg/innerring/internal/blockchain/blockchain.go
notimetoname: > that's where my own brain suffers

can you expand your suffering, please?

> readability comes with duality when Enabled == false and len(Addresses) == 0?

what duality is here? external lib (`neo-go`) requires `Enable` flag to be set, we have to use it and we agreed that we keep it in config files. if the prev state is correct, then we have `Enabled` in `yaml`, lose it in `Config` struct and dedicate it back for `neo-go` structs, what for?

> lets just remove the flag and determine behavior by the state of one field - Addresses

what if `enabled: false` and `listen: [*non-empty-list*]`? what `Addresses` field is? empty because of `enabled: false`? additional `if` checks? by the readability i meant: read the flag, put it in a middle struct if it is needed, take it to the `neo-go` lib. i do not like implicit transformations (one struct to another, to another, etc; one field to another (with another naming and another meaning), to another, etc) usually

pkg/innerring/internal/blockchain/blockchain.go
cthulhu-rider: then why `Blockchain` declares it optional?

roman-khimov: It's absolutely always has to be enabled. It's just that it may have no addresses to listen (neither http, nor https), but the service itself has to be enabled.

pkg/innerring/config.go
cthulhu-rider: cause IIUC if `enabled` flag is set then `addresses` are required, isn't it?

pkg/innerring/internal/blockchain/blockchain.go
cthulhu-rider: what the service does w/o listening?

roman-khimov: Provides local client that is used by IR.

pkg/innerring/config.go
notimetoname: > cause IIUC if enabled flag is set then addresses are required, isn't it?

i would say "no". while sound useless, it is not as strange as `tls.enabled` but no `tls.cert_file` for me (which is required with the last push). it is enabled but listens to no addresses. like an empty `for` loop. why not?

pkg/innerring/internal/blockchain/blockchain.go
notimetoname: > lets resolve it first

sure

> overall, ur trying to get stuck in how Neo Go stuff is configured structurally

in fact, i am trying to make internal code as simple, as possible. i doubt our `blockchain` wrapper over the `neo-go` `blockchain` will be imported by anyone. i wanted to treat `Config` struct as an internal interpretation of `yml`/`json` config file: the one has `enabled` flag -- the other has too

> focus on docs and behavior

yes, i want the behavior to be the same as `neo-go` internals are. we import an incredibly complex thing from `neo-go` and it has the same config as we will provide (well, ok, that thread is not solved yet anyway). so why not just move it from `yml` to `blockchain` 1:1, the same way `neo-go` does in their own app?

roman-khimov: Gentlemen, the simpler the code for this stuff is the better. We do have `enabled` exposed externally in the node config. We do have it in NeoGo. Likely it'd be easier to avoid skipping/inferring it in the middle. I'd think that this way we get less transformations and less potential for errors. Or not?

pkg/innerring/internal/blockchain/blockchain.go
cthulhu-rider: reasonable questions, I expect `New` to fail. Could you pls try this?

pkg/innerring/config.go
cthulhu-rider: ok lets stick to proposed approach, will change later if needed

pkg/innerring/internal/blockchain/blockchain.go
cthulhu-rider: one field less anywhere and there is a simplification of the code in my understanding, I try not to produce dual conditions, but everyone has their own perception

let it remain as suggested while it works, we can always change

config/example/ir.yaml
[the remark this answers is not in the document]
cthulhu-rider: i guess conversation is stuck, therefore, we stick to the current version, for the rest it will be possible to create an issue

pkg/innerring/internal/blockchain/blockchain.go
notimetoname: if we speak about our `New`, it fails [now](https://github.com/nspcc-dev/neofs-node/pull/2345#discussion_r1197936419). if `neo-go` is our goal, than it does not touch cert and key until `Start` is called (but errors are returned via `errChan`)

cthulhu-rider: i expect `New` to immediately fail on empty key/cert filepath, looks like it doesn't with proposed changes. Not critical for now since app checks that on config parsing, but still not very good to me

cthulhu-rider (approved):

PR #2352, opened 2023-05-19
Rm/Drop RPC priorities in morph

### LINKED ISSUE #2306: RPC node priorities are broken by design
## Expected Behavior
Once a NeoFS node is connected to an RPC node it must stick to this connection for as long as it can. A node can only use one RPC safely (because two RPCs can have different state).

## Current Behavior
Node can "switch to the most prioritized" RPC even if it has a properly functioning connection. This then leads to #1337. 

Priorities themselves are just a way to order RPCs even though they're perfectly ordered already in a config.

## Possible Solution
Drop `switch_interval` and drop priorities, use the order we get in config. This is an incompatible change, thus tracked separately from #2304, even though it's related a bit.

notimetoname:
@roman-khimov, @cthulhu-rider, @532910, @aprasolova, @MaxGelbakhiani, please, look at the second commit. It exists cause I find it strange to keep
```yml
  rpc_endpoint:
    - address: wss://rpc1.morph.fs.neo.org:40341/ws
    - address: wss://rpc2.morph.fs.neo.org:40341/ws
```
config just to make it compatible (it [was not like that before the priorities](https://github.com/nspcc-dev/neofs-node/commit/7410827db8354e296078f8be9c8d0ff26675bc47)). Also it is a good opportunity to unify config values across that repo apps (it was `rpc_endpoint` not "enpointS" (some time ago it was impossible to provide more that one endpoint), it was `endpoint.client` for... what?). If deprecation is a perfect idea, you can provide your own examples of how it can be named (mb `addresses` like `neogo` does? will be fully unified then) and we can discuss.

Closes #2306.


config/example/node.yaml
```diff
@@ -86,12 +86,9 @@ morph:
   cache_ttl: 15s  # Sidechain cache TTL value (min interval between similar calls). Negative value disables caching.
                   # Default value: block time. It is recommended to have this value less or equal to block time.
                   # Cached entities: containers, container lists, eACL tables.
-  switch_interval: 3m # interval b/w RPC switch attempts if the node is connected not to the highest priority node
```
roman-khimov, line 89: Should had been removed in the first commit.

config/example/ir.yaml
```diff
@@ -12,13 +12,11 @@ without_mainnet: false # Run application in single chain environment without mai
...
-  endpoint:
-    client: # List of websocket RPC endpoints in sidechain
-      - address: wss://sidechain1.fs.neo.org:30333/ws
-      - address: wss://sidechain2.fs.neo.org:30333/ws
+  endpoints: # List of websocket RPC endpoints in sidechain
+      - wss://sidechain1.fs.neo.org:30333/ws
+      - wss://sidechain2.fs.neo.org:30333/ws
   validators: # List of hex-encoded 33-byte public keys of sidechain validators to vote for at application startup
     - 0283120f4c8c1fc1d792af5063d2def9da5fddc90bc1384de7fcfdda33c3860170
-  switch_interval: 2m # interval b/w RPC switch attempts if the node is not connected to the highest priority node
```
roman-khimov, line 21: And this one too.

cmd/neofs-node/config/morph/config.go
```diff
@@ -16,47 +15,36 @@ const (
...
 // from "morph" section.
 //
 // Throws panic if list is empty.
-func RPCEndpoint(c *config.Config) []client.Endpoint {
-	var es []client.Endpoint
+func RPCEndpoint(c *config.Config) []string {
+	var endpointsDeprecated []string
 
-	sub := c.Sub(subsection).Sub("rpc_endpoint")
+	sub := c.Sub(subsection).Sub("endpoint")
```
roman-khimov, line 30: But how is it compatible with the old config? I think you have to keep `rpc_endpoint` here the way it was. Of course it'll be deleted afterwards, but for now it should be 1:1 as in the previous config version.

cmd/neofs-node/config/morph/config.go
```diff
@@ -16,47 +15,36 @@ const (
...
-	SwitchIntervalDefault = 2 * time.Minute
 )
 
 // RPCEndpoint returns list of the values of "rpc_endpoint" config parameter
 // from "morph" section.
 //
 // Throws panic if list is empty.
-func RPCEndpoint(c *config.Config) []client.Endpoint {
-	var es []client.Endpoint
+func RPCEndpoint(c *config.Config) []string {
```
cthulhu-rider, line 27: currently correct, but imo `[]client.Endpoint` is clearer and easier-to-support than `[]string` in code

cmd/neofs-node/config/morph/config.go
notimetoname: my bad, i had ~3 ideas about how that PR should look and that is how multi ideas affected it. fixed

cmd/neofs-node/config/morph/config.go
notimetoname: but `client.Endpoint` has been removed. do you mean keeping that string in `client` package with the only `string` field?

in fact, i do not like parsing internal structs from a config. that package should just read built-in types and return them IMO

cthulhu-rider: yeah, it's probably more accurate to say "maybe let's not purge `client.EndpointInfo` if it already exists?" I leave u to decide

about config package: i agree that it would make more sense to determine the type in place, because in fact they may well be different. Will be taken into account in future

cthulhu-rider (approved):

roman-khimov (approved):

PR #2357, opened 2023-05-24
cli: drop support for binary keys, use wallets

roman-khimov:
Wallets were introduced way back when in 0.27.5 and that's the way keys should be stored. We've had enough transition period for people to adjust to using them. Binary keys are dangerous and inconvenient (NEP-6 is common for all of the Neo tooling, everyone knows how to use it), therefore we shouldn't provide support for them.

Related to #2136.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Changelog for NeoFS Node
 ### Removed
 - Non-notary mode support for sidechain (#2321)
 - Priority switching b/w RPC endpoints in the morph client (#2306)
+- Support for binary keys in the CLI (#2357)
 
 ### Updated
 - Update minimal supported Go version up to v1.18 (#2340)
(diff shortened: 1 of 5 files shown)


notimetoname (commented):
How about filling the "Updating..." section? CLIs can be used in scripts.

notimetoname (approved):

CHANGELOG.md
```diff
@@ -35,6 +36,12 @@ Changelog for NeoFS Node
 - `morph.switch_interval` IR and SN config value is not used anymore.
 - `morph.rpc_endpoint` SN config value and `morph.endpoint.client` IR config value has been deprecated and will be 
   removed with the next minor release. Use `morph.endpoints` for both instead (NOTE: it does not have priorities now).
+- If you're using binary keys with neofs-cli (`-w`), convert them to proper
```
aprasolova, line 39: Please, could you duplicate this manual to README or any place where it'd be easy to find?

roman-khimov: I think it's only relevant for this upgrade and careful admins read this documentation when they perform upgrade. And frankly I don't expect to see anyone still using these binaries keys, they're not even easy to create now (everyone just goes `neo-go wallet create` route).

PR #170, opened 2020-11-13
Invocatons of alphabet vote method

alexvanin:
Inner ring can invoke `Vote` method on corresponding alphabet contracts. This way it controls side chain. Now inner ring node invokes voting on startup if it is presented in config. This way alphabet contracts are able to produce more GAS for inner ring and storage nodes.

You can setup list of predefined validators in `config.yml`
```yml
morph:
  validators:
    - 02b3622bf4017bdfe317c58aed5f4c753f206b7db896046fa7d774bbc4bf7f8dc2
    - 0228b3b86d632a755da1e1f226bd782038acbe48e1f14e256b0afa58e32a45695e
```

or with environment variable (whitespace separated list)
```bash
NEOFS_IR_MORPH_VALIDATORS="02b3622bf4017bdfe317c58aed5f4c753f206b7db896046fa7d774bbc4bf7f8dc2 0228b3b86d632a755da1e1f226bd782038acbe48e1f14e256b0afa58e32a45695e"
```

If list is empty, then `vote` won't be invoked on startup.

In runtime node also can vote for validator from the comma-separated list.
```bash
./neofs-ir --config config.yml \
--vote 02b3622bf4017bdfe317c58aed5f4c753f206b7db896046fa7d774bbc4bf7f8dc2,0228b3b86d632a755da1e1f226bd782038acbe48e1f14e256b0afa58e32a45695e
```

PR #2363, opened 2023-05-31
feat/Use parsed committee as validator list

### LINKED ISSUE #2311: Take validators from standby committee when using internal CN
We've got `validators` section in the `morph` config and currently that's required for IR. But IR can have an internal CN that has a standby committee list that is essentially the same.

Make morph.validators optional if internal CN is used, take the keys from network's committee.

## Additional context
Related to #2194.

notimetoname:
Closes #2311.

@roman-khimov, check, please. I am not sure that i got the issue correctly. In fact, it is not clear to me why we need some voting on the application start. I found "This way alphabet contracts are able to produce more GAS for inner ring and storage nodes." from https://github.com/nspcc-dev/neofs-node/pull/170 but not sure i can understand it.

Also, you mentioned that "currently that's required for IR" (meaning validators) but i don't get how it is _required_ and what exactly the difference is.

### DIFF
--- a/config/example/ir.yaml
+++ b/config/example/ir.yaml
@@ -17,7 +17,7 @@ morph:
   endpoints: # List of websocket RPC endpoints in sidechain
       - wss://sidechain1.fs.neo.org:30333/ws
       - wss://sidechain2.fs.neo.org:30333/ws
-  validators: # List of hex-encoded 33-byte public keys of sidechain validators to vote for at application startup
+  validators: # List of hex-encoded 33-byte public keys of sidechain validators to vote for at application startup; can be omitted if equals `consensus.committee`
     - 0283120f4c8c1fc1d792af5063d2def9da5fddc90bc1384de7fcfdda33c3860170
   consensus: # Local consensus launch mode activated only when 'endpoint.client' is unset.
     magic: 15405 # Network magic. Must be unsigned integer in range [1:4294967295]
--- a/pkg/innerring/innerring.go
+++ b/pkg/innerring/innerring.go
@@ -358,12 +358,22 @@ func New(ctx context.Context, log *logger.Logger, cfg *viper.Viper, errChan chan
 	walletPath := cfg.GetString(walletPathKey)
 	walletPass := cfg.GetString("wallet.password")
 
+	// parse default validators
+	server.predefinedValidators, err = parsePredefinedValidators(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("can't parse predefined validators list: %w", err)
+	}
+
 	// create morph client
 	if isLocalConsensusMode(cfg) {
 		// go on a local blockchain
 		cfgBlockchain, err := parseBlockchainConfig(cfg, log)
 		if err != nil {
 			return nil, fmt.Errorf("invalid blockchain configuration: %w", err)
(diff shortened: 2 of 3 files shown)


roman-khimov (changes requested):
1. This needs to be documented.
2. `len(server.predefinedValidators) == 0` when `!isLocalConsensusMode()` is a configuration error that should be reported as error. Voting is important to generate proper amount of GAS and it's important for the network to be operable (being able to change alphabet keys, at least theoretically).

notimetoname:
> This needs to be documented.

Are comments in the config files enough?

roman-khimov:
IIRC we don't have any documentation for IR node (unlike `./docs/storage-node-configuration.md`), so yeah, at least some mentioning in config samples/CHANGELOG/update/etc.

pkg/innerring/innerring.go
```diff
@@ -358,6 +358,12 @@ func New(ctx context.Context, log *logger.Logger, cfg *viper.Viper, errChan chan
 	walletPath := cfg.GetString(walletPathKey)
 	walletPass := cfg.GetString("wallet.password")
 
+	// parse default validators
+	server.predefinedValidators, err = parsePredefinedValidators(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("ir: can't parse predefined validators list: %w", err)
```
cthulhu-rider, line 364:
```suggestion
		return nil, fmt.Errorf("can't parse predefined validators list: %w", err)
```
upper context knows that this is ir

notimetoname: fixed

config/example/ir.yaml
```diff
@@ -17,7 +17,7 @@ morph:
   endpoints: # List of websocket RPC endpoints in sidechain
       - wss://sidechain1.fs.neo.org:30333/ws
       - wss://sidechain2.fs.neo.org:30333/ws
-  validators: # List of hex-encoded 33-byte public keys of sidechain validators to vote for at application startup
+  validators: # List of hex-encoded 33-byte public keys of sidechain validators to vote for at application startup; ignored in local consensus mode
```
roman-khimov, line 20: It shouldn't be ignored. Just can be omitted if we're voting for the same set of nodes (most of the time we are). Otherwise we won't be able to vote for a different set (blockchain standby validators list can't change).

I think it was a little different in the previous patch version.

notimetoname: > I think it was a little different in the previous patch version.

Yes, it was.

> It shouldn't be ignored.

Can you, please, describe the desired flow? I start thinking that I'm just guessing what should be the true way.

```go
validators = parsedFromCfg
if localMode {
   if len(validators) == 0 {
        validators = committee
    }
} else {
    if len(validators) == 0 {
        panic
    }
}
```

?

roman-khimov: Yeah, like that.

notimetoname: fixed

roman-khimov (approved):

roman-khimov (commented):
Changelog, though.

cthulhu-rider (approved):

roman-khimov (approved):

PR #2371, opened 2023-06-03
workflow: disable all capturing

vvarg229:
Disabling all captures when running full system tests, as when running sanity tests after PR.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -212,7 +212,7 @@ jobs:
         timeout-minutes: 480
         if: github.event_name != 'pull_request'
         run: |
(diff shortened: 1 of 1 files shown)


vvarg229:
Verifying this workflow file when running ALL of the tests here:
https://github.com/vvarg229/neofs-node/actions/runs/5164038341/jobs/9302700755

roman-khimov (commented):
I guess we'll just see logs in reports then?

roman-khimov:
Why closing?

vvarg229:
> Why closing?

I ran tests on the local repo and lots of failed tests.

This change does not reduce the number of logs and only complicates the code.
I don't think it's necessary.

PR #2378, opened 2023-06-12
Fix/Fill split info for the first child correctly

notimetoname:
SDK RC8 does not attach a split ID to the first child in a split chain.

See https://github.com/nspcc-dev/neofs-sdk-go/issues/448 and https://nspcc-dev.github.io/neofs-node/191/#suites/beb587d64f4c3e0a7b76bb9b5b406a00/57ccae6fb58af59b.


roman-khimov (approved):
I guess with this one you've almost reimplemented the slicer in the node anyway...

cthulhu-rider (approved):

PR #2391, opened 2023-06-18
Provide embedded compiled NeoFS contracts

cthulhu-rider:
isolated changes from https://github.com/nspcc-dev/neofs-node/pull/2301. Starting from this, I'll try introduce changes partially to lighten review process

note that these changes do not yet present the script for compiling and copying contract files into `contracts` dir, suggestions are welcome

in contrast to the proposed option for naming files in #2195, i decided not to name files with serial numbers because to me it's clearer to see what contract is represented by particular files. Anyway, this pattern can be changed at any time


roman-khimov:
You can name them like `01-alphabet`, `20-netmap`, etc. But it's important since you can easily see the deployment sequence, you can make the code more generic (especially with reduced number of deployment parameters) and you can more easily stuff some new contract into the mix.

roman-khimov:
The only reason for not doing so, btw, is storing contracts in `neofs-contract` and getting them right from that repo. This in fact can be done with a simple wrapper that would make a proper Go package out of these files.

cthulhu-rider:
@roman-khimov moved to `NN.nef + NN.manifest.json` naming pattern, check

for now added NNS contract only just to prevent `go:embed` directive failure

contracts/contracts.go
```diff
@@ -0,0 +1,145 @@
...
+	NEF      nef.File
+	Manifest manifest.Manifest
+}
+
+//go:embed *.nef *.manifest.json
+var _fs embed.FS
+
+// Read reads compiled contracts stored in the package sorted numerically.
+// File schema:
+//   - compiled executables (NEF) are named by pattern 'N.nef'
```
roman-khimov, line 32: But why not `N-name.nef`?

cthulhu-rider: i don't mind to add names, required or optional (allow `N.nef` for any custom contract)?

roman-khimov: Either way, I'd go with required just to be consistent and have a bit more visibility into the mix.

cthulhu-rider: done, require `a-z` name for now

roman-khimov (approved):

PR #2402, opened 2023-06-22
Introduce procedure for the NeoFS Sidechain deployment

cthulhu-rider:
* one more iteration within #2195

i plan to add contracts' deployment in the next PR to reduce the amount of changes at a time (although there are a lot of current ones)

i've tested changes manually with NNS contract built from https://github.com/nspcc-dev/neofs-contract/tree/6e9b183b219353bd07a5d985f2ee38473f4f6a28 on x4 node cluster from [neo-go ](https://github.com/nspcc-dev/neo-go/blob/master/.docker/docker-compose.yml) project modified with Notary service

i guess it's possible to write unit tests using `neotest` framework, but it'll take too much time i bet

i also plan to describe the procedure in detail in `docs/` after approval (next PR i think)

### DIFF
--- a/pkg/morph/deploy/deploy.go
+++ b/pkg/morph/deploy/deploy.go
@@ -1,0 +1,205 @@
+// Package deploy provides NeoFS Sidechain deployment functionality.
+package deploy
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"sort"
+
+	"github.com/nspcc-dev/neo-go/pkg/core/block"
+	"github.com/nspcc-dev/neo-go/pkg/core/state"
+	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
+	"github.com/nspcc-dev/neo-go/pkg/neorpc"
+	"github.com/nspcc-dev/neo-go/pkg/rpcclient/actor"
+	"github.com/nspcc-dev/neo-go/pkg/smartcontract/manifest"
+	"github.com/nspcc-dev/neo-go/pkg/smartcontract/nef"
+	"github.com/nspcc-dev/neo-go/pkg/wallet"
+	"go.uber.org/zap"
+)
+
+// Blockchain groups services provided by particular Neo blockchain network
+// representing NeoFS Sidechain that are required for its deployment.
+type Blockchain interface {
+	// RPCActor groups functions needed to compose and send transactions to the
+	// blockchain.
+	actor.RPCActor
+
+	// GetCommittee returns list of public keys owned by Neo blockchain committee
+	// members. Resulting list is non-empty, unique and unsorted.
+	GetCommittee() (keys.PublicKeys, error)
+
+	// GetContractStateByID returns network state of the smart contract by its ID.
+	// GetContractStateByID returns error with 'Unknown contract' substring if
+	// requested contract is missing.
+	GetContractStateByID(id int32) (*state.Contract, error)
+
+	// ReceiveBlocks starts background process that forwards new blocks of the
+	// blockchain to the provided channel. The process handles all new blocks when
+	// ReceiveBlocks is called with nil filter. Returns unique identifier to be used
+	// to stop the process via Unsubscribe.
+	ReceiveBlocks(*neorpc.BlockFilter, chan<- *block.Block) (id string, err error)
+
+	// Unsubscribe stops background process started by ReceiveBlocks by ID.
+	Unsubscribe(id string) error
+}
+
+// KeyStorage represents storage of the private keys.
+type KeyStorage interface {
+	// GetPersistedPrivateKey returns singleton private key persisted in the
+	// storage. GetPersistedPrivateKey randomizes the key initially. All subsequent
+	// successful calls return the same key.
+	GetPersistedPrivateKey() (*keys.PrivateKey, error)
+}
+
+// CommonDeployPrm groups common deployment parameters of the smart contract.
+type CommonDeployPrm struct {
+	NEF      nef.File
+	Manifest manifest.Manifest
+}
+
+// NNSPrm groups deployment parameters of the NeoFS NNS contract.
+type NNSPrm struct {
+	Common      CommonDeployPrm
+	SystemEmail string
+}
+
+// Prm groups all parameters of the NeoFS Sidechain deployment procedure.
+type Prm struct {
+	// Writes progress into the log.
+	Logger *zap.Logger
+
+	// Particular Neo blockchain instance to be used as NeoFS Sidechain.
+	Blockchain Blockchain
+
+	// Local process account used for transaction signing (must be unlocked).
+	LocalAccount *wallet.Account
+
+	// Storage for single committee group key.
+	KeyStorage KeyStorage
+
+	NNS NNSPrm
+}
+
+// Deploy initializes Neo network represented by given Prm.Blockchain as NeoFS
+// Sidechain and makes it full-featured for NeoFS storage system operation.
+//
+// Deploy aborts only by context or when a fatal error occurs. Deployment
+// progress is logged in detail. It is expected that some situations can be
+// changed/fixed on the chain from the outside, so Deploy adapts flexibly and
+// does not stop at the moment.
+//
+// Deployment process is detailed in NeoFS docs. Summary of stages:
+//  1. NNS contract deployment
+//  2. launch of a notary service for the committee
+//  3. committee group initialization
+//  4. deployment of the NeoFS system contracts (currently not done)
+//  5. deployment of custom contracts
+//
+// See project documentation for details.
+func Deploy(ctx context.Context, prm Prm) error {
+	committee, err := prm.Blockchain.GetCommittee()
+	if err != nil {
+		return fmt.Errorf("get Neo committee of the network: %w", err)
+	}
+
+	sort.Sort(committee)
+
+	// determine a leader
+	localPrivateKey := prm.LocalAccount.PrivateKey()
+	localPublicKey := localPrivateKey.PublicKey()
+	localAccCommitteeIndex := -1
+
+	for i := range committee {
+		if committee[i].Equal(localPublicKey) {
+			localAccCommitteeIndex = i
+			break
+		}
+	}
+
+	if localAccCommitteeIndex < 0 {
+		return errors.New("local account does not belong to any Neo committee member")
+	}
+
+	deployNNSPrm := deployNNSContractPrm{
+		logger:                prm.Logger,
+		blockchain:            prm.Blockchain,
+		localAcc:              prm.LocalAccount,
+		localNEF:              prm.NNS.Common.NEF,
+		localManifest:         prm.NNS.Common.Manifest,
+		systemEmail:           prm.NNS.SystemEmail,
+		initCommitteeGroupKey: nil, // set below
+	}
+
+	// if local node is the first committee member (Az) => deploy NNS contract,
+	// otherwise just wait
+	if localAccCommitteeIndex == 0 {
+		// Why such a centralized approach? There is a need to initialize committee
+		// contract group and share its private key between all committee members (the
+		// latter is done in the current procedure next). Currently, there is no
+		// convenient Neo service for this, and we don't want to use anything but
+		// blockchain, so the key is distributed through domain NNS records. However,
+		// then the chicken-and-egg problem pops up: committee group must be also set
+		// for the NNS contract. To set the group, you need to know the contract hash in
+		// advance, and it is a function from the sender of the deployment transaction.
+		// Summing up all these statements, we come to the conclusion that the one who
+		// deploys the contract creates the group key, and he shares it among the other
+		// members. Technically any committee member could deploy NNS contract, but for
+		// the sake of simplicity, this is a fixed node. This makes the procedure even
+		// more centralized, however, in practice, at the start of the network, all
+		// members are expected to be healthy and active.
+		//
+		// Note that manifest can't be changed w/o NEF change, so it's impossible to set
+		// committee group dynamically right after deployment. See
+		// https://github.com/nspcc-dev/neofs-contract/issues/340
+		deployNNSPrm.initCommitteeGroupKey = prm.KeyStorage.GetPersistedPrivateKey
+	}
+
+	prm.Logger.Info("initializing NNS contract on the chain...")
+
+	nnsOnChainAddress, err := initNNSContract(ctx, deployNNSPrm)
+	if err != nil {
+		return fmt.Errorf("init NNS contract on the chain: %w", err)
+	}
+
+	prm.Logger.Info("NNS contract successfully initialized on the chain", zap.Stringer("address", nnsOnChainAddress))
+
+	prm.Logger.Info("enable Notary service for the committee...")
+
+	err = enableNotary(ctx, enableNotaryPrm{
+		logger:                 prm.Logger,
+		blockchain:             prm.Blockchain,
+		nnsOnChainAddress:      nnsOnChainAddress,
+		systemEmail:            prm.NNS.SystemEmail,
+		committee:              committee,
+		localAcc:               prm.LocalAccount,
+		localAccCommitteeIndex: localAccCommitteeIndex,
+	})
+	if err != nil {
+		return fmt.Errorf("enable Notary service for the committee: %w", err)
+	}
+
+	prm.Logger.Info("Notary service successfully enabled for the committee")
+
+	prm.Logger.Info("initializing committee group for contract management...")
+
+	committeeGroupKey, err := initCommitteeGroup(ctx, initCommitteeGroupPrm{
+		logger:                 prm.Logger,
+		blockchain:             prm.Blockchain,
+		nnsOnChainAddress:      nnsOnChainAddress,
+		systemEmail:            prm.NNS.SystemEmail,
+		committee:              committee,
+		localAcc:               prm.LocalAccount,
+		localAccCommitteeIndex: localAccCommitteeIndex,
+		keyStorage:             prm.KeyStorage,
+	})
+	if err != nil {
+		return fmt.Errorf("init committee group: %w", err)
+	}
+
+	prm.Logger.Info("committee group successfully initialized", zap.Stringer("public key", committeeGroupKey.PublicKey()))
+
+	// TODO: deploy contracts
+
+	return nil
+}
--- a/pkg/morph/deploy/group.go
+++ b/pkg/morph/deploy/group.go
@@ -1,0 +1,341 @@
+package deploy
+
+import (
+	"context"
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/rand"
+	"encoding/base64"
+	"errors"
+	"fmt"
+
+	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
+	"github.com/nspcc-dev/neo-go/pkg/rpcclient/actor"
+	"github.com/nspcc-dev/neo-go/pkg/rpcclient/invoker"
+	"github.com/nspcc-dev/neo-go/pkg/rpcclient/nns"
+	"github.com/nspcc-dev/neo-go/pkg/util"
+	"github.com/nspcc-dev/neo-go/pkg/wallet"
+	"go.uber.org/zap"
+)
+
+// initCommitteeGroupPrm groups parameters of committee group initialization.
+type initCommitteeGroupPrm struct {
+	logger *zap.Logger
+
+	blockchain Blockchain
+
+	nnsOnChainAddress util.Uint160
+	systemEmail       string
+
+	committee              keys.PublicKeys
+	localAcc               *wallet.Account
+	localAccCommitteeIndex int
+
+	keyStorage KeyStorage
+}
+
+// initCommitteeGroup initializes committee group and returns corresponding private key.
+func initCommitteeGroup(ctx context.Context, prm initCommitteeGroupPrm) (*keys.PrivateKey, error) {
+	monitor, err := newBlockchainMonitor(prm.logger, prm.blockchain)
+	if err != nil {
+		return nil, fmt.Errorf("init blockchain monitor: %w", err)
+	}
+	defer monitor.stop()
+
+	inv := invoker.New(prm.blockchain, nil)
+	const leaderCommitteeIndex = 0
+	var committeeGroupKey *keys.PrivateKey
+	var leaderTick func()
+
+upperLoop:
+	for ; ; monitor.waitForNextBlock(ctx) {
+		select {
+		case <-ctx.Done():
+			return nil, fmt.Errorf("wait for committee group key to be distributed: %w", ctx.Err())
+		default:
+		}
+
+		prm.logger.Info("checking domains with shared committee group key...")
+
+		nShared := 0
+
+		for i := range prm.committee {
+			domain := committeeGroupDomainForMember(i)
+
+			rec, err := lookupNNSDomainRecord(inv, prm.nnsOnChainAddress, domain)
+			if err != nil {
+				if errors.Is(err, errMissingDomain) || errors.Is(err, errMissingDomainRecord) {
+					prm.logger.Info("NNS record with committee group key shared with the committee member is still missing, waiting...",
+						zap.String("domain", domain))
+				} else {
+					prm.logger.Error("failed to lookup NNS domain record, will try again later",
+						zap.String("domain", domain), zap.Error(err))
+				}
+
+				continue
+			}
+
+			if committeeGroupKey == nil && i == prm.localAccCommitteeIndex {
+				committeeGroupKey, err = decryptSharedPrivateKey(rec, prm.committee[leaderCommitteeIndex], prm.localAcc.PrivateKey())
+				if err != nil {
+					prm.logger.Error("failed to decrypt shared committee group key, will wait for a background fix",
+						zap.String("domain", domain), zap.Error(err))
+					continue upperLoop
+				}
+			}
+
+			nShared++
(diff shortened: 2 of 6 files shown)


pkg/morph/deploy/notary.go
```diff
@@ -0,0 +1,872 @@
...
+	}
+
+	x.validUntilBlock = binary.BigEndian.Uint32(b[util.Uint160Size:])
+	x.nonce = binary.BigEndian.Uint32(b[util.Uint160Size+4:])
+
+	return nil
+}
+
+const (
+	sharedTransactionDataLen         = util.Uint160Size + 4 + 4
```
roman-khimov, line 806: But then it won't be as convenient doing this (`io.GetVarSize` can help, but it requires implementing `io.Serializable`).

pkg/morph/deploy/notary.go
```diff
@@ -0,0 +1,872 @@
...
+	tx.ValidUntilBlock = sharedTxData.validUntilBlock
+	tx.Nonce = sharedTxData.nonce
+	tx.Signers = []transaction.Signer{
+		{
+			Account: sharedTxData.sender,
+			Scopes:  transaction.None,
+		},
+		{
+			Account: committeeMultiSigAcc.ScriptHash(),
+			Scopes:  transaction.CustomGroups | transaction.CalledByEntry,
```
roman-khimov, line 863: Again, `CustomGroups` is a no-op without an actual group.

pkg/morph/deploy/notary.go
```diff
@@ -0,0 +1,872 @@
...
+	tx, err := roleContract.DesignateAsRoleUnsigned(noderoles.P2PNotary, committee)
+	if err != nil {
+		return nil, err
+	}
+
+	tx.ValidUntilBlock = sharedTxData.validUntilBlock
+	tx.Nonce = sharedTxData.nonce
+	tx.Signers = []transaction.Signer{
+		{
+			Account: sharedTxData.sender,
```
roman-khimov, line 858: Isn't it the only field to be changed in `signers`?

pkg/morph/deploy/notary.go
```diff
@@ -0,0 +1,872 @@
...
+			Scopes:  transaction.None,
+		},
+		{
+			Account: committeeMultiSigAcc.ScriptHash(),
+			Scopes:  transaction.CustomGroups | transaction.CalledByEntry,
+		},
+	}
+
+	// FIXME: w/o this transaction fails with 'invalid signature' error due to lack of GAS for some period
+	// (later succeeds when there is more GAS)
```
roman-khimov, line 868: `invalid signature` is exactly that, an invalid signature. Insufficient network fee should be `vm execution has failed`. And you network fee should be precise since you have all signers known and not changed between estimating/verifying (sender hash doesn't count, he has the same script cost anyway).

pkg/morph/deploy/group.go
```diff
@@ -0,0 +1,345 @@
...
+
+			l.Info("committee group key is shared with the committee member in NNS domain record")
+		}
+	}, nil
+}
+
+// encryptSharedPrivateKey encrypts private key using provided coder's private
+// key to be decrypted using decoder's private key. Inverse operation to
+// decryptSharedPrivateKey.
+func encryptSharedPrivateKey(sharedPrivKey, coderPrivKey *keys.PrivateKey, decoderPubKey *keys.PublicKey) (string, error) {
```
roman-khimov, line 250: I'd probably encrypt the private key with a symmetric cipher and then share this cipher's key via ECDH encryption, rather than encrypting the key directly with per-node symmetric key. But probably we can leave it as is until neo-project/proposals#161 makes it much simpler.

roman-khimov (commented):
The main logic seems to be correct though.

pkg/morph/deploy/deploy.go
```diff
@@ -0,0 +1,210 @@
...
+// See project documentation for details.
+func Deploy(ctx context.Context, prm Prm) error {
+	committee, err := prm.Blockchain.GetCommittee()
+	if err != nil {
+		return fmt.Errorf("get Neo committee of the network: %w", err)
+	} else if len(committee) == 0 {
+		return errors.New("empty Neo committee of the network")
+	}
+
+	// just in case: docs don't deny, but we are tied to the number
```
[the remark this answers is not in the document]
cthulhu-rider: added requirement to `Blockchain`

pkg/morph/deploy/deploy.go
```diff
@@ -0,0 +1,210 @@
...
+//  3. committee group initialization
+//  4. deployment of the NeoFS system contracts (currently not done)
+//  5. deployment of custom contracts
+//
+// See project documentation for details.
+func Deploy(ctx context.Context, prm Prm) error {
+	committee, err := prm.Blockchain.GetCommittee()
+	if err != nil {
+		return fmt.Errorf("get Neo committee of the network: %w", err)
+	} else if len(committee) == 0 {
```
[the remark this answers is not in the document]
cthulhu-rider: added requirement to `Blockchain`

pkg/morph/deploy/notary.go
```diff
@@ -0,0 +1,872 @@
...
+	monitor, err := newBlockchainMonitor(prm.logger, prm.blockchain)
+	if err != nil {
+		return fmt.Errorf("init blockchain monitor: %w", err)
+	}
+	defer monitor.stop()
+
+	var tact func()
+	roleContract := rolemgmt.NewReader(invoker.New(prm.blockchain, nil))
+
+	for recycle := false; ; recycle = true {
```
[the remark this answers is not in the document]
cthulhu-rider: yep, didn't mention this can be simplified now, done

pkg/morph/deploy/notary.go
```diff
@@ -0,0 +1,872 @@
...
+
+// enableNotary makes Notary service ready-to-go for the committee members.
+func enableNotary(ctx context.Context, prm enableNotaryPrm) error {
+	monitor, err := newBlockchainMonitor(prm.logger, prm.blockchain)
+	if err != nil {
+		return fmt.Errorf("init blockchain monitor: %w", err)
+	}
+	defer monitor.stop()
+
+	var tact func()
```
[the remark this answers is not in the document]
cthulhu-rider: renamed to `tick`

pkg/morph/deploy/notary.go
```diff
@@ -0,0 +1,872 @@
...
+				if registerDomainTxValidUntilBlock > 0 {
+					l.Info("transaction registering NNS domain was sent earlier, checking relevance...")
+
+					if cur := monitor.currentHeight(); cur <= registerDomainTxValidUntilBlock {
+						l.Info("previously sent transaction registering NNS domain may still be relevant, will wait for the outcome",
+							zap.Uint32("current height", cur), zap.Uint32("retry after height", registerDomainTxValidUntilBlock))
+						return
+					}
+
+					l.Info("previously sent transaction registering NNS domain expired without side-effect")
```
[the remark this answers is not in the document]
cthulhu-rider: i don't remember exact transaction (maybe not this particular one), but I met this case at least once, so decided to protect from it everywhere (pretty cheap for us). Not expected to really happen in "healthy" network, but IMO check still worth it and we wanna see when it happens

pkg/morph/deploy/notary.go
```diff
@@ -0,0 +1,872 @@
...
+type sharedTransactionData struct {
+	sender          util.Uint160
+	validUntilBlock uint32
+	nonce           uint32
+}
+
+// bytes serializes sharedTransactionData.
+func (x sharedTransactionData) bytes() []byte {
+	b := make([]byte, sharedTransactionDataLen)
+	// fixed size is more convenient for potential format changes in the future
```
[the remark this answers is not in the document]
cthulhu-rider: so change or leave? I just wanted to stick to a fixed length with it it is easier to change the format (if needed) in the future in a backwards compatible style

pkg/morph/deploy/notary.go
cthulhu-rider: dropped

pkg/morph/deploy/group.go
cthulhu-rider: lets wait for the proposal to be accepted, created https://github.com/nspcc-dev/neofs-node/issues/2405

pkg/morph/deploy/notary.go
```diff
@@ -0,0 +1,872 @@
...
+			Signer: transaction.Signer{
+				Account: prm.localAcc.ScriptHash(),
+				Scopes:  transaction.CalledByEntry,
+			},
+			Account: prm.localAcc,
+		},
+		{
+			Signer: transaction.Signer{
+				Account: committeeMultiSigAcc.ScriptHash(),
+				Scopes:  transaction.CustomGroups | transaction.CalledByEntry,
```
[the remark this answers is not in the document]
cthulhu-rider: dropped

pkg/morph/deploy/notary.go
```diff
@@ -0,0 +1,872 @@
...
+				needReset = true
+			}
+
+			recordExists = true
+		}
+
+		if needReset {
+			prm.logger.Info("generating shared data for the transaction designating Notary role to the committee...")
+
+			txVub, err := _actor.CalculateValidUntilBlock()
```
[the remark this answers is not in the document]
cthulhu-rider: nice catch, done

pkg/morph/deploy/notary.go
cthulhu-rider: checked, ur right, modified

pkg/morph/deploy/notary.go
cthulhu-rider: double-checked, now there is no problem, dropped

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
roman-khimov: OK, let's not touch this for now.

pkg/morph/deploy/notary.go
```diff
@@ -0,0 +1,872 @@
...
+
+	committeeActor, err := actor.New(prm.blockchain, committeeSigners)
+	if err != nil {
+		return nil, fmt.Errorf("init transaction sender with committee signers: %w", err)
+	}
+
+	_invoker := invoker.New(prm.blockchain, nil)
+	roleContract := rolemgmt.New(committeeActor)
+
+	// multi-tact context
```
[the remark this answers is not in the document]
cthulhu-rider: renamed to `tick`

pkg/morph/deploy/notary.go
```diff
@@ -0,0 +1,872 @@
...
+	_actor, err := actor.NewSimple(prm.blockchain, prm.localAcc)
+	if err != nil {
+		return nil, fmt.Errorf("init transaction sender from local account: %w", err)
+	}
+
+	committeeSigners := []actor.SignerAccount{
+		{
+			Signer: transaction.Signer{
+				Account: prm.localAcc.ScriptHash(),
+				Scopes:  transaction.CalledByEntry,
```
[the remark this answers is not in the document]
cthulhu-rider: changed

cthulhu-rider:
tested after bunch of changes, everything still works ok

contracts/00-nns.manifest.json
```diff
@@ -1 +1 @@
...
...ffset":619,"parameters":[],"returntype":"Integer","safe":true},{"name":"transfer","offset":965,"parameters":[{"name":"to","type":"Hash160"},{"name":"tokenID","type":"ByteArray"},{"name":"data","type":"Any"}],"returntype":"Boolean","safe":false},{"name":"update","offset":520,"parameters":[{"name":"nef","type":"ByteArray"},{"name":"manifest","type":"String"},{"name":"data","type":"Any"}],"returntype":"Void","safe":false},{"name":"updateSOA","offset":2541,"parameters":[{"name":"name","type":"String"},{"name":"email","type":"String"},{"name":"refresh","type":"Integer"},{"name":"retry","type":"Integer"},{"name":"expire","type":"Integer"},{"name":"ttl","type":"Integer"}],"returntype":"Void","safe":false},{"name":"version","offset":615,"parameters":[],"returntype":"Integer","safe":true}],"events":[{"name":"Transfer","parameters":[{"name":"from","type":"Hash160"},{"name":"to","type":"Hash160"},{"name":"amount","type":"Integer"},{"name":"tokenId","type":"ByteArray"}]}]},"features":{},"groups":[],"permissions":[{"contract":"0xfffdc93764dbaddd97c48f252a53ea4643faa3fd","methods":["update"]},{"contract":"*","methods":["onNEP11Payment"]}],"supportedstandards":["NEP-11"],"trusts":[],"extra":null}
```
roman-khimov, line 1: This update should be a separate commit.

roman-khimov (commented):
I've only looked at the commented changes and GH can't easily compare old/new PR state, but it should be OK otherwise.

contracts/00-nns.manifest.json
cthulhu-rider: added accidentally, dropped, will be updated later

roman-khimov (approved):

pkg/morph/deploy/notary.go
```diff
@@ -0,0 +1,880 @@
...
+
+		prm.logger.Info("transaction designating Notary role to the local account has been successfully sent, will wait for the outcome")
+	}, nil
+}
+
+// initDesignateNotaryRoleAsLeaderTick returns a function that preserves context
+// of the Notary role designation to the multi-acc committee between calls. The
+// operation is performed by the leading committee member which is assigned to
+// collect signatures for the corresponding transaction.
+func initDesignateNotaryRoleAsLeaderTick(prm enableNotaryPrm, monitor *blockchainMonitor) (func(), error) {
```
notimetoname, line 175: i think what func would not pass `funlen` linter

pkg/morph/deploy/notary.go
```diff
@@ -0,0 +1,880 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("compose committee multi-signature account: %w", err)
+	}
+
+	_actor, err := actor.NewSimple(prm.blockchain, prm.localAcc)
+	if err != nil {
+		return nil, fmt.Errorf("init transaction sender from local account: %w", err)
+	}
+
+	committeeSigners := []actor.SignerAccount{
```
notimetoname, line 569: @cthulhu-rider, @roman-khimov can you, please, explain me in short why do we have two signers here?

pkg/morph/deploy/deploy.go
```diff
@@ -0,0 +1,205 @@
...
+	// ReceiveBlocks is called with nil filter. Returns unique identifier to be used
+	// to stop the process via Unsubscribe.
+	ReceiveBlocks(*neorpc.BlockFilter, chan<- *block.Block) (id string, err error)
+
+	// Unsubscribe stops background process started by ReceiveBlocks by ID.
+	Unsubscribe(id string) error
+}
+
+// KeyStorage represents storage of the private keys.
+type KeyStorage interface {
```
notimetoname, line 48: why is it an interface? you have literally described its (the only) implementation in a few lines

notimetoname (commented):
IMO, sometimes it is too `info`.

pkg/morph/deploy/notary.go
roman-khimov: Because the first one pays for the transaction, while the second one witnesses a role change.

pkg/morph/deploy/deploy.go
cthulhu-rider: cuz package doesn't care how exactly key is stored (in file, in db, in memory, remotely)

pkg/morph/deploy/notary.go
cthulhu-rider: afaik we don't use `funlen` linter now. I don't think splitting into several functions will improve readability, at least for now

pkg/morph/deploy/notary.go
```diff
@@ -0,0 +1,880 @@
...
+
+// enableNotary makes Notary service ready-to-go for the committee members.
+func enableNotary(ctx context.Context, prm enableNotaryPrm) error {
+	monitor, err := newBlockchainMonitor(prm.logger, prm.blockchain)
+	if err != nil {
+		return fmt.Errorf("init blockchain monitor: %w", err)
+	}
+	defer monitor.stop()
+
+	var tick func()
```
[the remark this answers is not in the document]
cthulhu-rider: because it's defined conditionally (may be not needed)

pkg/morph/deploy/nns.go
```diff
@@ -0,0 +1,229 @@
...
+		default:
+		}
+
+		prm.logger.Info("reading on-chain state of the NNS contract by ID=1")
+
+		stateOnChain, err := readNNSOnChainState(prm.blockchain)
+		if err != nil {
+			prm.logger.Error("failed to read on-chain state of the NNS contract, will try again later", zap.Error(err))
+			continue
+		} else if stateOnChain != nil {
```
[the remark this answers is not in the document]
cthulhu-rider: that's how the function behaves, or u suggest to make `if` without `else`?

cthulhu-rider:
> IMO, sometimes it is too info.

what exactly do u mean? iiuc u suggest to write more messages with `debug` severity?

pkg/morph/deploy/nns.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+
+		sentTxValidUntilBlock = vub
+
+		prm.logger.Info("transaction deploying NNS contract has been successfully sent, will wait for the outcome")
+	}
+}
+
+// lookupNNSDomainRecord looks up for the 1st record of the NNS domain with
+// given name. Returns errMissingDomain is domain doesn't exist. Returns
```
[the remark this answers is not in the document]
cthulhu-rider: fixed

notimetoname:
> iiuc u suggest to write more messages with debug severity?

Yes, I've found 43 `Info` logs. Are all of them necessary for admins? Many of them contain specific words ("committee", "NNS", "transaction").

pkg/morph/deploy/nns.go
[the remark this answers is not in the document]
notimetoname: > or u suggest to make if without else

if it is possible. it seems so. i think the linter failed to understand that it can be simplified

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
notimetoname: IMO it will be easier to read if we have definition logic separately (the `for` below is already complex)

pkg/morph/deploy/notary.go
notimetoname: > afaik we don't use funlen linter now

yes, i will create an issue for that. i would not find it if the question was about ~60-120 but it is already 400 lines long. also, i did not measure that and did not _want_ to find smth long; i just was reading and understood that the func was not going to end and i had lost the initial context of the code i was reading

pkg/morph/deploy/deploy.go
notimetoname: why do we have to store it? i understand your point but it is not about storing only but about generating too. why?

pkg/morph/deploy/nns.go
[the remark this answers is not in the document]
cthulhu-rider: i don't know how
```go
if {
  // ...
  // continue/return
}

if {
  // ...
  // continue/return
}
```
is better than
if {
  // ...
  // continue/return
} else if {
  // ...
  // continue/return
}
```

but ok, i'll do it. Approach to never use `else` statement is the most strange thing to me popularized by smbd in Go

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
cthulhu-rider: definition or initialization? u suggest to compose multi-tick context before loop?

pkg/morph/deploy/notary.go
cthulhu-rider: function length is an eternal topic of discussion. Same as functions which are created to just decrease length of some function and called once

> lost the initial context of the code i was reading
i believe, but log messages tell u what's goin on, dont think separate functions would ease readability: u then need to jump into them and read docs and go back

pkg/morph/deploy/deploy.go
cthulhu-rider: we have to store it because we must not loose it between stages (which is really possible). This also relates to https://github.com/neo-project/proposals/issues/161 that will simplify this stuff with distributed single private key

cthulhu-rider:
> Yes, I've found 43 Info logs. Are all of them necessary for admins? Many of them contain specific words ("committee", "NNS", "transaction").

yes, i'd say it's necessary cuz this is an automatic decentralized scheme which also awaits for background fixes which could be necessary in some situations. Normally there will be not so many messages, but if smth will stuck u'll need to realize what's goin on and how we've come to this: here log will help u

pkg/morph/deploy/nns.go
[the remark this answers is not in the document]
notimetoname: usually, i'm open to any code rule and dont mind considering any of them, but once they are applied, they should be treated as a rule, or cancelation should be discussed separately. omitting `else` is described in https://go.dev/doc/effective_go; i am also almost sure that `gosimple` should have found such code but (i think) it just failed here

> i don't know how ... is better than

you dont need to keep the context of the first statement to be sure that the second _is being checked_. also `go` has a lot of `if err != nil { return }` so a gopher should read a separate small `if { return }` relatively fast but more complex `if {} else if {}` (i think) slows a reader down

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
notimetoname: > definition or initialization?

i see no difference in that context. initialization, i think

> u suggest to compose multi-tick context before loop?

i saw that `var` and the `for` below looked more complex than they could be: you declared a var and defined it in the place where it should be used (also defined it more than one time in some cases cause it is a `for` loop). its definition (initialization) does not need to be flexible (it only returns an error if `actor` can't be created, quite a static thing). so my question is: can it be initialized once (or two vars created (or an `actor` and just a static func that is a closure over just created `actor`)) instead of `var v func()` and `v = someStaticThingThatAlwayWillBeTheSame()` in every iteration?

pkg/morph/deploy/deploy.go
notimetoname: > we have to store it because we must not loose it between stages

do you mean that initialization should be idempotent? if no, can just a local variable save us?

cthulhu-rider: yep, as much as possible ofc. File can dissapear and there is nothing else we could do with this. But in 97% cases this will work

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
cthulhu-rider: logic is simple: `tick` can be not needed at all, so it's initialized once under certain condition in for loop. Do u suggest to insta initialize it outside loop?

pkg/morph/deploy/nns.go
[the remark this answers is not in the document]
cthulhu-rider: nope, `gosimple` worked correctly. And i follow with effetive Go here. Next piece of code is not a regular statement but condition (which also continues/breaks/returns), so `else` is not redundant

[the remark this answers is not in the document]
notimetoname: > In the Go libraries, you'll find that when an if statement doesn't flow into the next statement—that is, the body ends in break, continue, goto, or return—the unnecessary else is omitted.

> so `else` is not redundant

that is how i interpret go effective. any `else` that can be omited (previous `if` ends with `return`), it should be ommited. i do not think that it is smth that **I** came up with: https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html. even if in go it does not look like that, that is how my brain nests every `if` while is reading them

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
notimetoname: > Do u suggest to insta initialize it outside loop?

yes, that is the only thing i suggest. im not sure it is possible cause func is quiete complex and i do not keep it in mind completely.
OR i think it can be called in every `if` or `else`. not sure we need them (`if` and `else`) to instantiate smth that will be called anyway in the end (outside of every `if` and `else`) of every iteration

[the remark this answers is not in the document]
cthulhu-rider: done

pkg/morph/deploy/notary.go
notimetoname: https://github.com/nspcc-dev/neofs-node/issues/2407

pkg/morph/deploy/deploy.go
notimetoname: > there is nothing else we can do but init new key

is it a problem? if no key has persisted, i would just init the next one? but if a key has persisted successfully, is it possible to just read it? mb i got the idea wrong

@roman-khimov

notimetoname (approved):
> I've only looked at the commented changes and GH can't easily compare old/new PR state, but it should be OK otherwise.

Too.

pkg/morph/deploy/deploy.go
roman-khimov: If the key is on the chain, that's where the key is. If it's not, it doesn't exist and we don't care which one is gonna be used for the next attempt. The only problematic case probably is if you have an in-flight transaction that sets the key, but you restart before it's processed. This can be solved by always waiting for a new block initially (so that it'll bring this transaction with it), when the leader goes into this routine. It's not precisely accurate, but a sufficiently good way.

Issue #1337, opened 2022-04-25
Restart services after switching to another neo-go RPC node 


carpawell:
After #1170 node connects to the Morph via WS connection only and it supports node switching without application restart. However, during the switching process, it could lose some notifications that may change the node's state. We can implement some syncing callback [call](https://github.com/nspcc-dev/neofs-node/blob/v0.28.0-rc.3/pkg/morph/client/multi.go#L142) but it is not fully clear what that callback should do and do we really need that sync (may the node be placed in a totally incorrect state because of losing some morph notifications?).

alexvanin:
> it is not fully clear what that callback should do

As the simplest robust solution, node needs to update internal state of remote contracts as it does at application start. In InnerRing application you may look at [initConfigFromBlockchain()](https://github.com/nspcc-dev/neofs-node/blob/b74fb2b932cf3cfc2a9b3e652ca2e032f331d635/pkg/innerring/innerring.go#L1002-L1040) method. Something similar happens in Storage as well, when it reads current network map.

But we need to do that carefully to avoid some parallel execution of state synchronization and new event processing.

roman-khimov:
What should be done in fact is service restart for everything tied to blockchain state. Consider storage node, it has a local store that doesn't care about blockchain state, it can be reused, but any other data is potentially outdated by the time we reconnect, so a complete service restart is needed.

PR #2406, opened 2023-06-26
feat/Restart IR services after neo-go is down

notimetoname:
Relates #1337. IR only. Tested with https://github.com/notimetoname/neofs-dev-env/tree/local/add-rpc-morph-node.

### DIFF
--- a/pkg/innerring/innerring.go
+++ b/pkg/innerring/innerring.go
@@ -81,7 +81,7 @@ type (
 		epochCounter  atomic.Uint64
 		epochDuration atomic.Uint64
 		statusIndex   *innerRingIndexer
-		precision     precision.Fixed8Converter
+		precision     uint32 // not changeable
 		healthStatus  atomic.Value
 		persistate    *state.PersistentStorage
 
@@ -551,6 +551,11 @@ func New(ctx context.Context, log *logger.Logger, cfg *viper.Viper, errChan chan
 		return nil, err
 	}
 
+	server.precision, err = server.balanceClient.Decimals()
+	if err != nil {
+		return nil, fmt.Errorf("can't read balance contract precision: %w", err)
+	}
+
 	repClient, err := repClient.NewFromMorph(server.morphClient, server.contracts.reputation, 0, repClient.AsAlphabet())
 	if err != nil {
 		return nil, err
@@ -771,6 +776,8 @@ func New(ctx context.Context, log *logger.Logger, cfg *viper.Viper, errChan chan
 		return nil, err
 	}
 
+	precisionConverter := precision.NewConverter(server.precision)
+
 	// create balance processor
 	balanceProcessor, err := balance.New(&balance.Params{
 		Log:           log,
@@ -778,7 +785,7 @@ func New(ctx context.Context, log *logger.Logger, cfg *viper.Viper, errChan chan
 		NeoFSClient:   neofsCli,
 		BalanceSC:     server.contracts.balance,
 		AlphabetState: server,
-		Converter:     &server.precision,
+		Converter:     precisionConverter,
 	})
 	if err != nil {
 		return nil, err
@@ -801,7 +808,7 @@ func New(ctx context.Context, log *logger.Logger, cfg *viper.Viper, errChan chan
 			MorphClient:         server.morphClient,
 			EpochState:          server,
 			AlphabetState:       server,
-			Converter:           &server.precision,
+			Converter:           precisionConverter,
 			MintEmitCacheSize:   cfg.GetInt("emit.mint.cache_size"),
 			MintEmitThreshold:   cfg.GetUint64("emit.mint.threshold"),
 			MintEmitValue:       fixedn.Fixed8(cfg.GetInt64("emit.mint.value")),
@@ -1036,15 +1043,8 @@ func (s *Server) initConfigFromBlockchain() error {
 		return fmt.Errorf("can't read epoch duration: %w", err)
 	}
 
-	// get balance precision
-	balancePrecision, err := s.balanceClient.Decimals()
-	if err != nil {
-		return fmt.Errorf("can't read balance contract precision: %w", err)
-	}
-
 	s.epochCounter.Store(epoch)
 	s.epochDuration.Store(epochDuration)
-	s.precision.SetBalancePrecision(balancePrecision)
 
 	// get next epoch delta tick
 	s.initialEpochTickDelta, err = s.nextEpochBlockDelta()
@@ -1056,7 +1056,7 @@ func (s *Server) initConfigFromBlockchain() error {
 		zap.Bool("active", s.IsActive()),
 		zap.Bool("alphabet", s.IsAlphabet()),
 		zap.Uint64("epoch", epoch),
-		zap.Uint32("precision", balancePrecision),
+		zap.Uint32("precision", s.precision),
 		zap.Uint32("init_epoch_tick_delta", s.initialEpochTickDelta),
 	)
 
--- a/pkg/innerring/innerring.go
+++ b/pkg/innerring/innerring.go
@@ -535,13 +535,7 @@ func New(ctx context.Context, log *logger.Logger, cfg *viper.Viper, errChan chan
 		return nil, err
 	}
 
-	// form morph container client's options
-	morphCnrOpts := make([]cntClient.Option, 0, 3)
-	morphCnrOpts = append(morphCnrOpts,
-		cntClient.AsAlphabet(),
-	)
-
-	cnrClient, err := cntClient.NewFromMorph(server.morphClient, server.contracts.container, 0, morphCnrOpts...)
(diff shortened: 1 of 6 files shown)


pkg/morph/client/multi.go
```diff
@@ -15,21 +15,28 @@ type Endpoint struct {
...
 		select {
 		case <-time.After(c.cfg.reconnectionDelay):
 		case <-c.closeChan:
+			c.switchLock.Unlock()
 			return false
 		}
 	}
 
 	c.inactive = true
+	c.switchLock.Unlock()
```
cthulhu-rider, line 39: placing this before statemenet below changes method behavior, i'd do it in a separate commit with explanation (i realize why u've done it). And then autonomous commit regarding the new callback

cthulhu-rider (commented):
429e23a4734df9017fe529aeda09a3eacb23762d is useless to me, why don't we add this method along with the first use case?

notimetoname:
> https://github.com/nspcc-dev/neofs-node/commit/429e23a4734df9017fe529aeda09a3eacb23762d is useless to me, why don't we add this method along with the first use case?

Different files, different structs (the first is about `innerRingIndexer`, the second about `Server`). But well, my args are not that strong, fixed them up.

pkg/innerring/innerring.go
```diff
@@ -1011,8 +1015,9 @@ func createListener(ctx context.Context, cli *client.Client, p *chainParams) (ev
 	return listener, err
 }
 
-func createClient(ctx context.Context, p *chainParams, errChan chan<- error) (*client.Client, error) {
-	endpoints := p.cfg.GetStringSlice(p.name + ".endpoints")
+func (s *Server) createClient(ctx context.Context, p *chainParams, errChan chan<- error) (*client.Client, error) {
+	name := p.name
```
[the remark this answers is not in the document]
notimetoname: > p is mutated on caller's side?

yes, spent some time fixing a bug that relates to the pointer. ok lets drop the pointer

pkg/innerring/innerring.go
```diff
@@ -73,17 +73,19 @@ type (
 		blockTimers     []*timer.BlockTimer
 		epochTimer      *timer.BlockTimer
 
+		morphClient      *client.Client
```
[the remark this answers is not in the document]
notimetoname: the problem was: 90% of the issue fixing time i spend on code reading and realizing what is _the state_ and what is not _the state_. after the n-th time reading "// global state" block with the clients, i flipped out and decided to split them. so:
1. a separate commit OR
2. you are against that move at all?

pkg/innerring/innerring.go
```diff
@@ -1145,3 +1163,30 @@ func (s *Server) newEpochTickHandlers() []newEpochHandler {
 
 	return newEpochHandlers
 }
+
+func (s *Server) restartMorph() error {
+	s.log.Info("restarting internals because of RPC connection loss...")
```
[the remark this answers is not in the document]
notimetoname: > what exactly've been restarted

i would not do that. their number can grow up and i doubt a dev would fix that message

> or at least services|internal services

fixed

pkg/util/precision/converter.go
```diff
@@ -22,7 +22,7 @@ type (
 	// This is a JSON bound that uses neo node. Neo-go has int64 limit for
 	// `smartcontract.Parameter` of integer type.
 	Fixed8Converter struct {
-		m *sync.RWMutex
+		m sync.RWMutex
```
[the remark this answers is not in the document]
notimetoname: no idea how it happened, it was a mistake, fixed

pkg/innerring/innerring.go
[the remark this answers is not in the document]
cthulhu-rider: 1 ofc: the change itself can be really useful for readability/maintainability, especially since u stumble

cthulhu-rider (approved):

cthulhu-rider (approved):

pkg/util/precision/converter.go
```diff
@@ -21,6 +22,7 @@ type (
 	// This is a JSON bound that uses neo node. Neo-go has int64 limit for
 	// `smartcontract.Parameter` of integer type.
 	Fixed8Converter struct {
+		m sync.RWMutex
```
roman-khimov, line 25: Is it really worth the trouble? Why not recreating the converter where needed?

pkg/util/precision/converter.go
```diff
@@ -71,6 +79,9 @@ func (c *Fixed8Converter) SetBalancePrecision(precision uint32) {
 		exp = -exp
 	}
 
+	c.m.Lock()
+	defer c.m.Unlock()
```
roman-khimov, line 83: Frankly, I'd just internalize this `SetBalancePrecision` and leave it all to `NewConverter`.

pkg/innerring/innerring.go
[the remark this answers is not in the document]
roman-khimov: Option 1, yes, it looks suspicious now.

pkg/innerring/innerring.go
```diff
@@ -1145,3 +1163,30 @@ func (s *Server) newEpochTickHandlers() []newEpochHandler {
 
 	return newEpochHandlers
 }
+
+func (s *Server) restartMorph() error {
```
roman-khimov, line 1166: I've been expecting some `innerring.New()` instead of this, but looking at the way connections are managed now maybe it's not realistic...

pkg/util/precision/converter.go
notimetoname: dont you like how every blockchain config is updated [the same way](https://github.com/notimetoname/neofs-node/blob/92c391dda95e151bc96ca55d9242b20c8406d6f4/pkg/innerring/innerring.go#L1101-L1104) (i even have not created any func for that, `initConfigFromBlockchain` already exists!)?

pkg/innerring/innerring.go
notimetoname: so that is why issue fixing is so slow. i can change it but i dont know what way is the best (for now), so after trying some ways i came up with the shortest one (111 lines only and it works)

> I've been expecting some innerring.New()

do you mean the full [New](https://github.com/notimetoname/neofs-node/blob/92c391dda95e151bc96ca55d9242b20c8406d6f4/pkg/innerring/innerring.go#L329) every time a connections is lost?

also, left https://github.com/nspcc-dev/neofs-node/issues/2410

pkg/util/precision/converter.go
roman-khimov: I mean using `NewConverter()` when needed (on reinit). Once initialized it never changes. If connection is broken then all users of it terminate anyway. Once it's up again we'll have a new converter and they will use this new one. Something like that (I'm not entirely sure it fits the current code well, but it theory it should).

pkg/util/precision/converter.go
roman-khimov: Epoch-related things _can_ change without reconnecting, balance precision can't.

pkg/innerring/innerring.go
roman-khimov: The problem is that we have dependencies inverted here. There can't be any IR unless there is a chain (well, two chains) and there is a connection to it. But we have this IR struct that pretends to be independent of its own state (that is inherently tied to chains) and handles connections internally. It can accept current solution if refactoring it now requires too much effort.

pkg/util/precision/converter.go
notimetoname: > Once initialized it never changes.

oh, you mean that we can not take into account that [precision](https://github.com/notimetoname/neofs-node/blob/0aee98732d75f67446df2167a294e560bad74039/pkg/util/precision/converter.go#L76) may change while we were down? or i got you wrong one more time?

pkg/util/precision/converter.go
notimetoname: > balance precision can't

oh, really? it simplifies things a little.

pkg/util/precision/converter.go
notimetoname: made the precision constant

pkg/innerring/innerring.go
notimetoname: > It can accept current solution if refactoring it now requires too much effort.

refactor will be done [separately](https://github.com/nspcc-dev/neofs-node/issues/2410#issuecomment-1613554057)

pkg/util/precision/converter.go
roman-khimov: I take as [NEP-17 `decimals()`](https://github.com/neo-project/proposals/blob/06a6b550987a74025009cf6841757f2426462697/nep-17.mediawiki#decimals). It can't change for a contract. And it's the same here, no one should even of changing it on a real network. It can only happen if you're to create another contract, another token and migrate to it. But this is a completely different story.

roman-khimov (approved):

PR #2408, opened 2023-06-27
Update NNS contract in Sidechain deployment procedure

cthulhu-rider:
* continue #2195 

i decided to start with one contract since:
* we start to work with notary deposit and i want it to be reviewed
* other contracts will behave almost the same (with NNS `register` and without deployment precondition)

i've commented some places with other possible solutions, pls take a  look

tested with locally upgraded and downgraded NNS contract, works well

### DIFF
--- a/pkg/morph/deploy/contracts.go
+++ b/pkg/morph/deploy/contracts.go
@@ -1,0 +1,7 @@
+package deploy
+
+// various common methods of the NeoFS contracts.
+const (
+	methodUpdate  = "update"
+	methodVersion = "version"
+)
--- a/pkg/morph/deploy/deploy.go
+++ b/pkg/morph/deploy/deploy.go
@@ -121,9 +121,17 @@ func Deploy(ctx context.Context, prm Prm) error {
 		return errors.New("local account does not belong to any Neo committee member")
 	}
 
+	monitor, err := newBlockchainMonitor(prm.Logger, prm.Blockchain)
+	if err != nil {
+		return fmt.Errorf("init blockchain monitor: %w", err)
+	}
+
+	defer monitor.stop()
+
 	deployNNSPrm := deployNNSContractPrm{
 		logger:                prm.Logger,
 		blockchain:            prm.Blockchain,
+		monitor:               monitor,
 		localAcc:              prm.LocalAccount,
 		localNEF:              prm.NNS.Common.NEF,
 		localManifest:         prm.NNS.Common.Manifest,
@@ -169,6 +177,7 @@ func Deploy(ctx context.Context, prm Prm) error {
 	err = enableNotary(ctx, enableNotaryPrm{
 		logger:                 prm.Logger,
 		blockchain:             prm.Blockchain,
+		monitor:                monitor,
 		nnsOnChainAddress:      nnsOnChainAddress,
 		systemEmail:            prm.NNS.SystemEmail,
 		committee:              committee,
@@ -186,6 +195,7 @@ func Deploy(ctx context.Context, prm Prm) error {
 	committeeGroupKey, err := initCommitteeGroup(ctx, initCommitteeGroupPrm{
 		logger:                 prm.Logger,
 		blockchain:             prm.Blockchain,
+		monitor:                monitor,
 		nnsOnChainAddress:      nnsOnChainAddress,
 		systemEmail:            prm.NNS.SystemEmail,
 		committee:              committee,
--- a/pkg/morph/deploy/deploy.go
+++ b/pkg/morph/deploy/deploy.go
@@ -106,6 +106,11 @@ type Prm struct {
 //
 // See project documentation for details.
 func Deploy(ctx context.Context, prm Prm) error {
+	// wrap the parent context into the context of the current function so that
+	// transaction wait routines do not leak
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
 	committee, err := prm.Blockchain.GetCommittee()
 	if err != nil {
 		return fmt.Errorf("get Neo committee of the network: %w", err)
@@ -198,7 +203,7 @@ func Deploy(ctx context.Context, prm Prm) error {
 
 	prm.Logger.Info("Notary service successfully enabled for the committee")
 
-	onNotaryDepositDeficiency, err := initNotaryDepositDeficiencyHandler(prm.Logger, prm.Blockchain, monitor, prm.LocalAccount)
+	onNotaryDepositDeficiency, err := initNotaryDepositDeficiencyHandler(ctx, prm.Logger, prm.Blockchain, prm.LocalAccount)
 	if err != nil {
 		return fmt.Errorf("construct action depositing funds to the local account's Notary balance: %w", err)
 	}
--- a/pkg/morph/deploy/deploy.go
+++ b/pkg/morph/deploy/deploy.go
@@ -11,7 +11,8 @@ import (
 	"github.com/nspcc-dev/neo-go/pkg/core/state"
 	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
 	"github.com/nspcc-dev/neo-go/pkg/neorpc"
-	"github.com/nspcc-dev/neo-go/pkg/rpcclient/actor"
+	"github.com/nspcc-dev/neo-go/pkg/neorpc/result"
+	"github.com/nspcc-dev/neo-go/pkg/rpcclient/notary"
 	"github.com/nspcc-dev/neo-go/pkg/smartcontract/manifest"
 	"github.com/nspcc-dev/neo-go/pkg/smartcontract/nef"
 	"github.com/nspcc-dev/neo-go/pkg/wallet"
@@ -21,9 +22,9 @@ import (
 // Blockchain groups services provided by particular Neo blockchain network
 // representing NeoFS Sidechain that are required for its deployment.
 type Blockchain interface {
-	// RPCActor groups functions needed to compose and send transactions to the
-	// blockchain.
-	actor.RPCActor
+	// RPCActor groups functions needed to compose and send transactions (incl.
+	// Notary service requests) to the blockchain.
+	notary.RPCActor
 
 	// GetCommittee returns list of public keys owned by Neo blockchain committee
 	// members. Resulting list is non-empty, unique and unsorted.
@@ -40,7 +41,14 @@ type Blockchain interface {
 	// to stop the process via Unsubscribe.
 	ReceiveBlocks(*neorpc.BlockFilter, chan<- *block.Block) (id string, err error)
 
-	// Unsubscribe stops background process started by ReceiveBlocks by ID.
+	// ReceiveNotaryRequests starts background process that forwards new notary
+	// requests of the blockchain to the provided channel. The process skips
+	// requests that don't match specified filter. Returns unique identifier to be
+	// used to stop the process via Unsubscribe.
+	ReceiveNotaryRequests(*neorpc.TxFilter, chan<- *result.NotaryRequestEvent) (string, error)
+
+	// Unsubscribe stops background process started by ReceiveBlocks or
+	// ReceiveNotaryRequests by ID.
 	Unsubscribe(id string) error
 }
 
@@ -93,7 +101,7 @@ type Prm struct {
 //  1. NNS contract deployment
 //  2. launch of a notary service for the committee
 //  3. committee group initialization
-//  4. deployment of the NeoFS system contracts (currently not done)
+//  4. deployment/update of the NeoFS system contracts (currently only NNS)
 //  5. deployment of custom contracts
 //
 // See project documentation for details.
@@ -190,6 +198,22 @@ func Deploy(ctx context.Context, prm Prm) error {
 
 	prm.Logger.Info("Notary service successfully enabled for the committee")
 
+	onNotaryDepositDeficiency, err := initNotaryDepositDeficiencyHandler(prm.Logger, prm.Blockchain, monitor, prm.LocalAccount)
+	if err != nil {
+		return fmt.Errorf("construct action depositing funds to the local account's Notary balance: %w", err)
+	}
+
+	err = listenCommitteeNotaryRequests(ctx, listenCommitteeNotaryRequestsPrm{
+		logger:                    prm.Logger,
+		blockchain:                prm.Blockchain,
+		localAcc:                  prm.LocalAccount,
+		committee:                 committee,
+		onNotaryDepositDeficiency: onNotaryDepositDeficiency,
+	})
+	if err != nil {
+		return fmt.Errorf("start listener of committee notary requests: %w", err)
+	}
+
 	prm.Logger.Info("initializing committee group for contract management...")
 
 	committeeGroupKey, err := initCommitteeGroup(ctx, initCommitteeGroupPrm{
@@ -209,7 +233,30 @@ func Deploy(ctx context.Context, prm Prm) error {
 
 	prm.Logger.Info("committee group successfully initialized", zap.Stringer("public key", committeeGroupKey.PublicKey()))
 
-	// TODO: deploy contracts
+	prm.Logger.Info("updating on-chain NNS contract...")
+
+	err = updateNNSContract(ctx, updateNNSContractPrm{
+		logger:                        prm.Logger,
+		blockchain:                    prm.Blockchain,
+		monitor:                       monitor,
+		localAcc:                      prm.LocalAccount,
+		localNEF:                      prm.NNS.Common.NEF,
+		localManifest:                 prm.NNS.Common.Manifest,
+		systemEmail:                   prm.NNS.SystemEmail,
+		committee:                     committee,
+		committeeGroupKey:             committeeGroupKey,
+		buildVersionedExtraUpdateArgs: noExtraUpdateArgs,
+		onNotaryDepositDeficiency:     onNotaryDepositDeficiency,
+	})
+	if err != nil {
+		return fmt.Errorf("update NNS contract on the chain: %w", err)
+	}
+
+	prm.Logger.Info("on-chain NNS contract successfully updated")
+
+	// TODO: deploy/update other contracts
 
 	return nil
 }
+
+func noExtraUpdateArgs(contractVersion) ([]interface{}, error) { return nil, nil }
--- a/pkg/morph/deploy/deploy.go
+++ b/pkg/morph/deploy/deploy.go
@@ -134,7 +134,9 @@ func Deploy(ctx context.Context, prm Prm) error {
 		return errors.New("local account does not belong to any Neo committee member")
 	}
 
-	monitor, err := newBlockchainMonitor(prm.Logger, prm.Blockchain)
+	chNewBlock := make(chan struct{}, 1)
+
+	monitor, err := newBlockchainMonitor(prm.Logger, prm.Blockchain, chNewBlock)
 	if err != nil {
 		return fmt.Errorf("init blockchain monitor: %w", err)
 	}
@@ -203,17 +205,13 @@ func Deploy(ctx context.Context, prm Prm) error {
 
 	prm.Logger.Info("Notary service successfully enabled for the committee")
 
-	onNotaryDepositDeficiency, err := initNotaryDepositDeficiencyHandler(ctx, prm.Logger, prm.Blockchain, prm.LocalAccount)
-	if err != nil {
-		return fmt.Errorf("construct action depositing funds to the local account's Notary balance: %w", err)
-	}
+	go autoReplenishNotaryBalance(ctx, prm.Logger, prm.Blockchain, prm.LocalAccount, chNewBlock)
 
 	err = listenCommitteeNotaryRequests(ctx, listenCommitteeNotaryRequestsPrm{
-		logger:                    prm.Logger,
-		blockchain:                prm.Blockchain,
-		localAcc:                  prm.LocalAccount,
-		committee:                 committee,
-		onNotaryDepositDeficiency: onNotaryDepositDeficiency,
+		logger:     prm.Logger,
+		blockchain: prm.Blockchain,
+		localAcc:   prm.LocalAccount,
+		committee:  committee,
 	})
 	if err != nil {
 		return fmt.Errorf("start listener of committee notary requests: %w", err)
@@ -251,7 +249,6 @@ func Deploy(ctx context.Context, prm Prm) error {
 		committee:                     committee,
 		committeeGroupKey:             committeeGroupKey,
 		buildVersionedExtraUpdateArgs: noExtraUpdateArgs,
-		onNotaryDepositDeficiency:     onNotaryDepositDeficiency,
 	})
 	if err != nil {
 		return fmt.Errorf("update NNS contract on the chain: %w", err)
--- a/pkg/morph/deploy/nns.go
+++ b/pkg/morph/deploy/nns.go
@@ -2,6 +2,7 @@ package deploy
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
@@ -158,13 +159,13 @@ func initNNSContract(ctx context.Context, prm deployNNSContractPrm) (res util.Ui
 		prm.logger.Info("sending new transaction deploying NNS contract...")
 
 		if managementContract == nil {
-			_actor, err := actor.NewSimple(prm.blockchain, prm.localAcc)
+			localActor, err := actor.NewSimple(prm.blockchain, prm.localAcc)
 			if err != nil {
 				prm.logger.Warn("NNS contract is missing on the chain but attempts to deploy are disabled, will try again later")
 				continue
 			}
 
-			managementContract = management.New(_actor)
+			managementContract = management.New(localActor)
 
 			setGroupInManifest(&prm.localManifest, prm.localNEF, committeeGroupKey, prm.localAcc.ScriptHash())
 		}
@@ -225,4 +226,175 @@ func lookupNNSDomainRecord(inv *invoker.Invoker, nnsContract util.Uint160, domai
 	}
 
 	return "", errMissingDomainRecord
+}
+
+// updateNNSContractPrm groups parameters of NeoFS NNS contract update.
+type updateNNSContractPrm struct {
+	logger *zap.Logger
+
+	blockchain Blockchain
+
+	// based on blockchain
+	monitor *blockchainMonitor
+
+	localAcc *wallet.Account
+
+	localNEF      nef.File
+	localManifest manifest.Manifest
+	systemEmail   string
+
+	committee         keys.PublicKeys
+	committeeGroupKey *keys.PrivateKey
+
+	// constructor of extra arguments to be passed into method updating the
+	// contract. If returns both nil, no data is passed (noExtraUpdateArgs may be
+	// used).
+	buildVersionedExtraUpdateArgs func(versionOnChain contractVersion) ([]interface{}, error)
+
+	onNotaryDepositDeficiency notaryDepositDeficiencyHandler
+}
+
+// updateNNSContract synchronizes on-chain NNS contract (its presence is a
+// precondition) with the local one represented by compiled executables. If
+// on-chain version is greater or equal to the local one, nothing happens.
+// Otherwise, transaction calling 'update' method is sent.
+//
+// Local manifest is extended with committee group represented by the
+// parameterized private key.
+//
+// Function behaves similar to initNNSContract in terms of context.
+func updateNNSContract(ctx context.Context, prm updateNNSContractPrm) error {
+	bLocalNEF, err := prm.localNEF.Bytes()
+	if err != nil {
+		// not really expected
+		return fmt.Errorf("encode local NEF of the NNS contract into binary: %w", err)
+	}
+
+	jLocalManifest, err := json.Marshal(prm.localManifest)
+	if err != nil {
+		// not really expected
+		return fmt.Errorf("encode local manifest of the NNS contract into JSON: %w", err)
+	}
+
+	committeeActor, err := newCommitteeNotaryActor(prm.blockchain, prm.localAcc, prm.committee)
+	if err != nil {
+		return fmt.Errorf("create Notary service client sending transactions to be signed by the committee: %w", err)
+	}
+
+	localVersion, err := readContractLocalVersion(prm.blockchain, prm.localNEF, prm.localManifest)
+	if err != nil {
+		return fmt.Errorf("read version of the local NNS contract: %w", err)
+	}
+
+	var updateTxValidUntilBlock uint32
+
+	for ; ; prm.monitor.waitForNextBlock(ctx) {
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("wait for NNS contract synchronization: %w", ctx.Err())
+		default:
+		}
+
+		prm.logger.Info("reading on-chain state of the NNS contract...")
+
+		nnsOnChainState, err := readNNSOnChainState(prm.blockchain)
+		if err != nil {
+			prm.logger.Error("failed to read on-chain state of the NNS contract, will try again later", zap.Error(err))
+			continue
+		} else if nnsOnChainState == nil {
+			// NNS contract must be deployed at this stage
+			return errors.New("missing required NNS contract on the chain")
+		}
+
+		if nnsOnChainState.NEF.Checksum == prm.localNEF.Checksum {
+			// manifests may differ, but currently we should bump internal contract version
+			// (i.e. change NEF) to make such updates. Right now they are not supported due
+			// to dubious practical need
+			// Track https://github.com/nspcc-dev/neofs-contract/issues/340
+			prm.logger.Info("same local and on-chain checksums of the NNS contract NEF, update is not needed")
+			return nil
+		}
+
+		prm.logger.Info("NEF checksums of the on-chain and local NNS contracts differ, need an update")
+
+		versionOnChain, err := readContractOnChainVersion(prm.blockchain, nnsOnChainState.Hash)
+		if err != nil {
+			prm.logger.Error("failed to read on-chain version of the NNS contract, will try again later", zap.Error(err))
+			continue
+		}
+
+		if v := localVersion.cmp(versionOnChain); v == -1 {
+			prm.logger.Info("local contract version is < than the on-chain one, update is not needed",
+				zap.Stringer("local", localVersion), zap.Stringer("on-chain", versionOnChain))
+			return nil
+		} else if v == 0 {
+			return fmt.Errorf("local and on-chain contracts have different NEF checksums but same version '%s'", versionOnChain)
+		}
+
+		extraUpdateArgs, err := prm.buildVersionedExtraUpdateArgs(versionOnChain)
+		if err != nil {
+			prm.logger.Error("failed to prepare build extra arguments for NNS contract update, will try again later",
+				zap.Stringer("on-chain version", versionOnChain), zap.Error(err))
+			continue
+		}
+
+		setGroupInManifest(&prm.localManifest, prm.localNEF, prm.committeeGroupKey, prm.localAcc.ScriptHash())
+
+		var vub uint32
+
+		// we pre-check 'already updated' case via MakeCall in order to not potentially
+		// wait for previously sent transaction to be expired (condition below) and
+		// immediately succeed
+		tx, err := committeeActor.MakeCall(nnsOnChainState.Hash, methodUpdate,
+			bLocalNEF, jLocalManifest, extraUpdateArgs)
+		if err != nil {
+			if isErrContractAlreadyUpdated(err) {
+				// note that we can come here only if local version is > than the on-chain one
+				// (compared above)
+				prm.logger.Info("NNS contract has already been updated, skip")
+				return nil
+			}
+		} else {
+			if updateTxValidUntilBlock > 0 {
+				prm.logger.Info("transaction updating NNS contract was sent earlier, checking relevance...")
+
+				if cur := prm.monitor.currentHeight(); cur <= updateTxValidUntilBlock {
+					prm.logger.Info("previously sent transaction updating NNS contract may still be relevant, will wait for the outcome",
+						zap.Uint32("current height", cur), zap.Uint32("retry after height", updateTxValidUntilBlock))
+					continue
+				}
+
+				prm.logger.Info("previously sent transaction updating NNS contract expired without side-effect")
+			}
+
+			prm.logger.Info("sending new transaction updating NNS contract...")
+
+			_, _, vub, err = committeeActor.Notarize(tx, nil)
+		}
+		if err != nil {
+			lackOfGAS := isErrNotEnoughGAS(err)
+			// here lackOfGAS=true always means lack of Notary balance and not related to
+			// the main transaction itself
+			if !lackOfGAS {
+				if !isErrNotaryDepositExpires(err) {
+					prm.logger.Error("failed to send transaction deploying NNS contract, will try again later", zap.Error(err))
+					continue
+				}
+			}
+
+			// same approach with in-place deposit is going to be used in other functions.
+			// Consider replacement with background process (e.g. blockchainMonitor
+			// internal) which periodically checks Notary balance and updates it when, for
+			// example, balance goes lower than 20% of desired amount or expires soon. With
+			// this approach functions like current will not try to make a deposit, but
+			// simply wait until it becomes enough.
+			prm.onNotaryDepositDeficiency(lackOfGAS)
+
+			continue
+		}
+
+		updateTxValidUntilBlock = vub
+
+		prm.logger.Info("transaction updating NNS contract has been successfully sent, will wait for the outcome")
+	}
 }
--- a/pkg/morph/deploy/nns.go
+++ b/pkg/morph/deploy/nns.go
@@ -95,9 +95,19 @@ type deployNNSContractPrm struct {
 // If contract is missing and deployNNSContractPrm.initCommitteeGroupKey is provided,
 // initNNSContract attempts to deploy local contract.
 func initNNSContract(ctx context.Context, prm deployNNSContractPrm) (res util.Uint160, err error) {
-	var managementContract *management.Contract
-	var sentTxValidUntilBlock uint32
+	localActor, err := actor.NewSimple(prm.blockchain, prm.localAcc)
+	if err != nil {
+		return res, fmt.Errorf("init transaction sender from local account: %w", err)
+	}
+
+	// wrap the parent context into the context of the current function so that
+	// transaction wait routines do not leak
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
 	var committeeGroupKey *keys.PrivateKey
+	txMonitor := newTransactionGroupMonitor(localActor)
+	managementContract := management.New(localActor)
 
 	for ; ; prm.monitor.waitForNextBlock(ctx) {
 		select {
@@ -141,47 +151,29 @@ func initNNSContract(ctx context.Context, prm deployNNSContractPrm) (res util.Ui
 				continue
 			}
 
+			setGroupInManifest(&prm.localManifest, prm.localNEF, committeeGroupKey, prm.localAcc.ScriptHash())
+
 			prm.logger.Info("private key of the committee group has been initialized", zap.Stringer("public key", committeeGroupKey.PublicKey()))
 		}
 
-		if sentTxValidUntilBlock > 0 {
-			prm.logger.Info("transaction deploying NNS contract was sent earlier, checking relevance...")
-
-			if cur := prm.monitor.currentHeight(); cur <= sentTxValidUntilBlock {
-				prm.logger.Info("previously sent transaction deploying NNS contract may still be relevant, will wait for the outcome",
-					zap.Uint32("current height", cur), zap.Uint32("retry after height", sentTxValidUntilBlock))
-				continue
-			}
-
-			prm.logger.Info("previously sent transaction deploying NNS contract expired without side-effect")
+		if txMonitor.isPending() {
+			prm.logger.Info("previously sent transaction updating NNS contract is still pending, will wait for the outcome")
+			continue
 		}
 
 		prm.logger.Info("sending new transaction deploying NNS contract...")
-
-		if managementContract == nil {
-			localActor, err := actor.NewSimple(prm.blockchain, prm.localAcc)
-			if err != nil {
-				prm.logger.Warn("NNS contract is missing on the chain but attempts to deploy are disabled, will try again later")
-				continue
-			}
-
-			managementContract = management.New(localActor)
-
-			setGroupInManifest(&prm.localManifest, prm.localNEF, committeeGroupKey, prm.localAcc.ScriptHash())
-		}
 
 		// just to definitely avoid mutation
 		nefCp := prm.localNEF
 		manifestCp := prm.localManifest
 
-		_, vub, err := managementContract.Deploy(&nefCp, &manifestCp, []interface{}{
+		txID, vub, err := managementContract.Deploy(&nefCp, &manifestCp, []interface{}{
 			[]interface{}{
 				[]interface{}{domainBootstrap, prm.systemEmail},
 				[]interface{}{domainContractAddresses, prm.systemEmail},
 			},
 		})
 		if err != nil {
-			sentTxValidUntilBlock = 0
 			if isErrNotEnoughGAS(err) {
 				prm.logger.Info("not enough GAS to deploy NNS contract, will try again later")
 			} else {
@@ -190,9 +182,11 @@ func initNNSContract(ctx context.Context, prm deployNNSContractPrm) (res util.Ui
 			continue
 		}
 
-		sentTxValidUntilBlock = vub
+		prm.logger.Info("transaction deploying NNS contract has been successfully sent, will wait for the outcome",
+			zap.Stringer("tx", txID), zap.Uint32("vub", vub),
+		)
 
-		prm.logger.Info("transaction deploying NNS contract has been successfully sent, will wait for the outcome")
+		txMonitor.trackPendingTransactionsAsync(ctx, vub, txID)
 	}
 }
 
@@ -286,7 +280,12 @@ func updateNNSContract(ctx context.Context, prm updateNNSContractPrm) error {
 		return fmt.Errorf("read version of the local NNS contract: %w", err)
 	}
 
-	var updateTxValidUntilBlock uint32
+	// wrap the parent context into the context of the current function so that
+	// transaction wait routines do not leak
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	txMonitor := newTransactionGroupMonitor(committeeActor)
 
 	for ; ; prm.monitor.waitForNextBlock(ctx) {
 		select {
@@ -340,8 +339,6 @@ func updateNNSContract(ctx context.Context, prm updateNNSContractPrm) error {
 
 		setGroupInManifest(&prm.localManifest, prm.localNEF, prm.committeeGroupKey, prm.localAcc.ScriptHash())
 
-		var vub uint32
-
 		// we pre-check 'already updated' case via MakeCall in order to not potentially
 		// wait for previously sent transaction to be expired (condition below) and
 		// immediately succeed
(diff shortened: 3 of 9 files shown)


pkg/morph/deploy/notary.go
```diff
@@ -869,3 +878,357 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
 
 	return tx, nil
 }
+
+// newCommitteeNotaryActor returns notary.Actor builds and sends Notary service
+// requests witnessed by the specified committee members to the provided
+// Blockchain. Given local account pays for transactions.
+func newCommitteeNotaryActor(b Blockchain, localAcc *wallet.Account, committee keys.PublicKeys) (*notary.Actor, error) {
+	committeeMultiSigM := smartcontract.GetMajorityHonestNodeCount(len(committee))
```
notimetoname, line 886: you make it twice (one more in the `listenCommitteeNotaryRequests`). can we do it once?

pkg/morph/deploy/deploy.go
```diff
@@ -209,7 +233,32 @@ func Deploy(ctx context.Context, prm Prm) error {
...
+		logger:            prm.Logger,
+		blockchain:        prm.Blockchain,
+		monitor:           monitor,
+		localAcc:          prm.LocalAccount,
+		localNEF:          prm.NNS.Common.NEF,
+		localManifest:     prm.NNS.Common.Manifest,
+		systemEmail:       prm.NNS.SystemEmail,
+		committee:         committee,
+		committeeGroupKey: committeeGroupKey,
+		buildVersionedExtraUpdateArgs: func(versionOnChain contractVersion) ([]interface{}, error) {
```
notimetoname, line 248: but why is it defined here? it does not close anything. why not just a separate func?

pkg/morph/deploy/contracts.go
```diff
@@ -0,0 +1,7 @@
+package deploy
+
+// various common methods of the NNS contracts.
+const (
+	methodUpdate  = "update"
```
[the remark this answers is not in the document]
cthulhu-rider: this file is for common stuff of the NeoFS contracts. And u caught a mistake `NNS` -> `NeoFS`. In future deployment/update functionality will be added here

pkg/morph/deploy/notary.go
```diff
@@ -869,3 +878,357 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
...
+					prm.logger.Info("unsupported type of the notary request event, skip",
+						zap.Stringer("got", notaryEvent.Type), zap.Stringer("expect", mempoolevent.TransactionAdded))
+					continue
+				case len(mainTx.Signers) != expectedSigners:
+					prm.logger.Info("unsupported number of signers of main transaction from the received notary request, skip",
+						zap.Int("expected", expectedSigners), zap.Int("got", len(mainTx.Signers)))
+					continue
+				case !mainTx.HasSigner(committeeMultiSigAccID):
+					// in theory, there can be any notary requests besides those sent by the current
+					// auto-deploy procedure. However, it's better to log with 'info' severity since
```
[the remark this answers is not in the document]
cthulhu-rider: i dont have strong opinion about severity of these messages now, so `info` fits better when u dont have administrator's feedback. Ok lets make it `warn`, it's better to watch than to miss

pkg/morph/deploy/notary.go
cthulhu-rider: as a micro optimization - yes. For me it's not worth it, because the calculations are not so complicated, and you will have to write a lot of preconditions (initialize this account in no other way), and now the signature of this function looks quite concise

pkg/morph/deploy/deploy.go
cthulhu-rider: to be clear, function body may change over time. Guess u suggest to make a global function `func noExtraUpdateArgs()` for this? ok i dont mind

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
cthulhu-rider: done

>  almost all logs in that switch are strange

nope, some of them are pretty natural for Notary service

pkg/morph/deploy/contracts.go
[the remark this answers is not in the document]
notimetoname: and what about prefixes?

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
notimetoname: > nope, some of them are pretty natural for Notary service

i meant that if it is a startup procedure, i would not expect some other notary TXs. but ok

pkg/morph/deploy/notary.go
notimetoname: > you will have to write a lot of preconditions

or duplicate the code. but ok

pkg/morph/deploy/deploy.go
notimetoname: > Guess u suggest to make a global function func noExtraUpdateArgs() for this? ok i dont mind

right you are

pkg/morph/deploy/nns.go
```diff
@@ -226,3 +227,174 @@ func lookupNNSDomainRecord(inv *invoker.Invoker, nnsContract util.Uint160, domai
...
+			_, _, vub, err = committeeActor.Notarize(tx, nil)
+		}
+		if err != nil {
+			lackOfGAS := isErrNotEnoughGAS(err)
+			// here lackOfGAS can become true even if error corresponds to insufficient
+			// funds for the main transaction, not Notary balance problem. Right now we
+			// cannot distinguish between these two cases. So, we naively believe that the
+			// problem is in the Notary balance, because according to used procedure we make
+			// Notary deposit when at least 50GAS is available (most likely enough for all
+			// transactions).
```
AnnaShaleva, line 377: This statement is not true. Notarizer sends notary request to the notary pool, the pool cares only about sender's notary deposit amount (see [the link](https://github.com/nspcc-dev/neo-go/blob/abb58fb7de6053dc9734b5acfa334870481b6110/pkg/network/server.go#L1190)), thus the "insufficient funds" error can be returned only when sender doesn't have enough funds deposited to Notary contract.

And I saw your conversation with @roman-khimov in the NeoGo element chat, there probably was a misunderstanding.

pkg/morph/deploy/nns.go
```diff
@@ -226,3 +227,174 @@ func lookupNNSDomainRecord(inv *invoker.Invoker, nnsContract util.Uint160, domai
...
+						zap.Uint32("current height", cur), zap.Uint32("retry after height", updateTxValidUntilBlock))
+					continue
+				}
+
+				prm.logger.Info("previously sent transaction updating NNS contract expired without side-effect")
+			}
+
+			prm.logger.Info("sending new transaction updating NNS contract...")
+
+			_, _, vub, err = committeeActor.Notarize(tx, nil)
```
AnnaShaleva, line 368: Just use the Waiter here and wait one of main/fallback to be persisted. Insufficient funds error can still be handled in this case, and you don't have to use monitor's `waitForNextBlock` to wait for every new block manually.

pkg/morph/deploy/notary.go
```diff
@@ -869,3 +878,353 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
 
 	return tx, nil
 }
+
+// newCommitteeNotaryActor returns notary.Actor builds and sends Notary service
```
AnnaShaleva, line 882: s/`returns notary.Actor builds`/`returns notary.Actor that builds`

pkg/morph/deploy/notary.go
```diff
@@ -869,3 +878,353 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
...
+	err := committeeMultiSigAcc.ConvertMultisig(committeeMultiSigM, committee)
+	if err != nil {
+		return nil, fmt.Errorf("compose committee multi-signature account: %w", err)
+	}
+
+	return notary.NewActor(b, []actor.SignerAccount{
+		{
+			Signer: transaction.Signer{
+				Account: localAcc.ScriptHash(),
+				Scopes:  transaction.CalledByEntry,
```
AnnaShaleva, line 898: Why do you need `CalledByEntry` for this signer? It pays the bills, that's all, isn't it?

pkg/morph/deploy/notary.go
```diff
@@ -869,3 +878,353 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
...
+// waiting for success. Caller should by default wait for the problem to be
+// fixed, and if not, retry.
+//
+// notaryDepositDeficiencyHandler must not be called from multiple routines in
+// parallel.
+type notaryDepositDeficiencyHandler = func(lackOfGAS bool)
+
+// Amount of GAS for the single local account's GAS->Notary transfer. Relatively
+// small value for fallback transactions' fees.
+var singleNotaryDepositAmount = big.NewInt(int64(fixedn.Fixed8FromInt64(1))) // 1 GAS
```
AnnaShaleva, line 928: You don't need to use fixedn here, just provide the number.

pkg/morph/deploy/notary.go
```diff
@@ -869,3 +878,353 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
...
+
+// constructs notaryDepositDeficiencyHandler working with the specified
+// Blockchain and GAS/Notary balance of the given account.
+func initNotaryDepositDeficiencyHandler(l *zap.Logger, b Blockchain, monitor *blockchainMonitor, localAcc *wallet.Account) (notaryDepositDeficiencyHandler, error) {
+	_actor, err := actor.NewSimple(b, localAcc)
```
AnnaShaleva, line 933: s/`_actor`/`act`

Or choose any other nice name, it's not the C#.

pkg/morph/deploy/notary.go
```diff
@@ -869,3 +878,353 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
...
+		// nep17.TokenWriter.Transfer doesn't support notary.OnNEP17PaymentData
+		// directly, so split the args
+		// Track release of https://github.com/nspcc-dev/neo-go/issues/2987
+		_, vub, err := gasContract.Transfer(localAccID, notary.Hash, singleNotaryDepositAmount, []interface{}{transferData.Account, transferData.Till})
+		if err != nil {
+			l.Error("failed to send transaction transferring local account's GAS to the Notary contract, will try again later", zap.Error(err))
+			return
+		}
+
+		transferTxValidUntilBlock = vub
```
AnnaShaleva, line 1049: Just wait for it and pass the context if you need an early stop. You will update contracts one-by-one, thus no concurrency issues are expected.

AnnaShaleva (commented):
OK, I need more time to carefully review the notary code. Heres the initial set of comments.

pkg/morph/deploy/notary.go
cthulhu-rider: to me `_actor` is a nice name, and it's a valid Go identifier. Are underscored ids not recommended in Go? ref me to spec or effective go if so

pkg/morph/deploy/nns.go
cthulhu-rider: we don't wanna miss successful background update from other member, that's why we `continue`. Waiter doesn't support custom conditions stopping wait loop (maybe that's an idea btw)

pkg/morph/deploy/notary.go
cthulhu-rider: that's a mistake, `transaction.None` is enough here

pkg/morph/deploy/notary.go
cthulhu-rider: same https://github.com/nspcc-dev/neofs-node/pull/2408#discussion_r1250888297

pkg/morph/deploy/contracts.go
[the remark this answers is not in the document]
cthulhu-rider: pls explain what u suggest? `update` method is common for all contracts, it's not needed to have `nnsUpdateMethod` and `netmapUpdateMethod`

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
cthulhu-rider: it's an application startup procedure, but may be not first run of the Sidechain, so other requests are pretty normal here

pkg/morph/deploy/nns.go
```diff
@@ -226,3 +227,174 @@ func lookupNNSDomainRecord(inv *invoker.Invoker, nnsContract util.Uint160, domai
...
+		versionOnChain, err := readContractOnChainVersion(prm.blockchain, nnsOnChainState.Hash)
+		if err != nil {
+			prm.logger.Error("failed to read on-chain version of the NNS contract, will try again later", zap.Error(err))
+			continue
+		}
+
+		// we could also try to compare on-chain version with the local one and tune
+		// update strategy: don't try to update with earlier version and catch updated
+		// contracts with unchanged version (blunder, we should react to it). For
+		// simplicity, we naively rely on the version change when the contract changes
```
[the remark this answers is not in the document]
cthulhu-rider: we can't catch this because one contract on the chain, and the other is local - I can build application with any contract. So, in practice we will TRY to upgrade versions properly, but in theory everything is possible

pkg/morph/deploy/nns.go
cthulhu-rider: so, let me double-check, if `Notarize` returns `insufficient funds`, then its 100% Notary deposit problem?

pkg/morph/deploy/notary.go
cthulhu-rider: u mean number `1_0000_0000` or what? I guess `big.NewInt(1)` != 1GAS

pkg/morph/deploy/nns.go
AnnaShaleva: Exactly. You can get `insufficient funds` problem for main transaction only after it's completed by Notary service and Notary service is trying to add main transaction into standard mempool and relay. And even in this case you'll only see the following log message: `new transaction callback finished with error` warning `insufficient funds` error field.

pkg/morph/deploy/nns.go
AnnaShaleva: > we don't wanna miss successful background update from other member

Who is responsible for NNS update? Are we supposing that several clients with different wallets will start NNS update simultaneously?

> Waiter doesn't support custom conditions stopping wait loop

It supports better solution: context that can be cancelled. See https://github.com/nspcc-dev/neo-go/blob/abb58fb7de6053dc9734b5acfa334870481b6110/pkg/rpcclient/actor/waiter.go#L57 and documentation to it.

pkg/morph/deploy/notary.go
AnnaShaleva: > u mean number 1_0000_0000 or what?

Sure, use `var singleNotaryDepositAmount = big.NewInt(1_0000_0000)`, no need to allocate new fixed8.

> I guess big.NewInt(1) != 1GAS

Of course not.

pkg/morph/deploy/notary.go
```diff
@@ -869,3 +878,353 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
...
+
+				senderKey, err := keys.NewPublicKeyFromBytes(bSenderKey, elliptic.P256())
+				if err != nil {
+					prm.logger.Info("failed to decode sender's public key from first script of main transaction from the received notary request, skip", zap.Error(err))
+					continue
+				}
+
+				// copy transaction to avoid pointer mutation. For simplicity, make a shallow
+				// copy: we change only Scripts field below. Overall, this approach is very
+				// risky, and it'd better to make a deep copy
```
AnnaShaleva, line 1170: It's OK here to use this shallow copy, you can remove this comment.

pkg/morph/deploy/notary.go
```diff
@@ -869,3 +878,353 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
...
+				mainTxCp := *mainTx
+				mainTxCp.Scripts = nil
+
+				mainTx = &mainTxCp // source one isn't needed anymore
+
+				// it'd be safer to get into the transaction and analyze what it is trying to do.
+				// For simplicity, now we blindly sign it
```
AnnaShaleva, line 1177: Either fix this in the current PR or create an issue for that.

pkg/morph/deploy/notary.go
```diff
@@ -869,3 +878,353 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
...
+					continue
+				case !mainTx.HasSigner(committeeMultiSigAccID):
+					// in theory, there can be any notary requests besides those sent by the current
+					// auto-deploy procedure. However, it's better to log with 'info' severity since
+					// this isn't really expected in practice.
+					prm.logger.Warn("committee is not a signer of main transaction from the received notary request, skip")
+					continue
+				case mainTx.HasSigner(prm.localAcc.ScriptHash()):
+					prm.logger.Info("main transaction from the received notary request is signed by a local account, skip")
+					continue
```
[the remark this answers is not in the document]
AnnaShaleva: It's the most significant question here, I'm not sure we have to create all these notary requests.

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
cthulhu-rider: all of them suit my needs because that's how i (admin) track what's goin on right now

pkg/morph/deploy/nns.go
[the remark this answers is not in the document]
cthulhu-rider: we can encounter case when NEF changed but version not bumped. Strictly speaking, it's a failure. And it's worth to catch this case. That's what the comment about

pkg/morph/deploy/nns.go
cthulhu-rider: > It supports better solution: context that can be cancelled.

it's good, but then i should cancel context in a separate routine (which checks background side-effect). Technically i can do this, but i don't think it will improve anything really much

what do u think?

cthulhu-rider: > Who is responsible for NNS update? Are we supposing that several clients with different wallets will start NNS update simultaneously?

any committee member

cthulhu-rider: > Which way do you like?

it seems pretty good, but i will be an async notification listener, too, so `Waiter` looks easier to me if we decide to wait for side-effect asynchronously

pkg/morph/deploy/notary.go
cthulhu-rider: > no need to allocate new fixed8

ok that's reasonable

pkg/morph/deploy/notary.go
cthulhu-rider: ok, lets replace it. But i don't think `act` is good since it is a verb and more appropriate for function. How about `actr`/`simpleActor`/`localActor`?

pkg/morph/deploy/notary.go
cthulhu-rider: yeah, i've written a comment about separate routine which controls only by itself. Do u agree it's worth to implement it? And if so, how do u see its behavior? My thoughts were:
1. just async track Notary balance (e.g. on new block receival) and make sure it doesn't go below the threshold (e.g. 1GAS)
2. wait for external signals from other routines (any notary action during procedure) and refresh deposit only by signal
3. maybe some other alternative?

@AnnaShaleva im not strong enought in Notary system, so full trust to u :grinning:

pkg/morph/deploy/nns.go
```diff
@@ -226,3 +227,174 @@ func lookupNNSDomainRecord(inv *invoker.Invoker, nnsContract util.Uint160, domai
...
+					continue
+				}
+			}
+
+			// same approach with in-place deposit is going to be used in other functions.
+			// Consider replacement with background process (e.g. blockchainMonitor
+			// internal) which periodically checks Notary balance and updates it when, for
+			// example, balance goes lower than 20% of desired amount or expires soon. With
+			// this approach functions like current will not try to make a deposit, but
+			// simply wait until it becomes enough.
```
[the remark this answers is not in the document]
cthulhu-rider: lets discuss a bit there https://github.com/nspcc-dev/neofs-node/pull/2408#discussion_r1252269316
if implementation will be estimated pretty easy, we'll do it here. Otherwise separate issue

pkg/morph/deploy/notary.go
```diff
@@ -869,3 +878,353 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
...
+
+		// simple deposit scheme: transfer 1GAS (at most 2% of GAS token balance) for
+		// 100 blocks after the latest deposit's expiration height (if first, then from
+		// current height).
+		//
+		// If we encounter deposit expiration and current Notary balance >=20% of single
+		// transfer, we just increase the expiration time of the deposit, otherwise, we
+		// make transfer.
+
+		till := currentDepositExpirationHeight + 100
```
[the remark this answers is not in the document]
cthulhu-rider: that's not a problem, but do u agree with the exact value? It also relates to our discussion about async Notary balance routine though

pkg/morph/deploy/notary.go
```diff
@@ -869,3 +878,353 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
...
+
+		if !lackOfGAS { // deposit expired
+			if new(big.Int).Mul(notaryBalance, big.NewInt(5)).Cmp(singleNotaryDepositAmount) >= 0 {
```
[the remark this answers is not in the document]
cthulhu-rider: ok i'll make them separate constants when we'll resolve https://github.com/nspcc-dev/neofs-node/pull/2408#discussion_r1252269316

pkg/morph/deploy/notary.go
```diff
@@ -869,3 +878,353 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
...
+		var transferData notary.OnNEP17PaymentData
+		transferData.Account = &localAccID
+		transferData.Till = till
+
+		l.Info("sending new transaction transferring local account's GAS to the Notary contract...",
+			zap.Stringer("amount", singleNotaryDepositAmount), zap.Uint32("till", transferData.Till))
+
+		// nep17.TokenWriter.Transfer doesn't support notary.OnNEP17PaymentData
+		// directly, so split the args
+		// Track release of https://github.com/nspcc-dev/neo-go/issues/2987
```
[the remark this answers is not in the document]
cthulhu-rider: ok ill create one

pkg/morph/deploy/notary.go
```diff
@@ -869,3 +878,353 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
...
+				// note: instruction above can throw NPE and it's ok to panic: we confidently
+				// expect that only non-nil pointers will come from the channel (NeoGo
+				// guarantees)
+
+				srcMainTxHash := mainTx.Hash()
+				_, processed := mProcessedMainTxs[srcMainTxHash]
+
+				switch {
+				case processed:
+					prm.logger.Info("main transaction of the notary request has already been processed, skip",
```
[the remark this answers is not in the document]
cthulhu-rider: we can make it later after first attempt of full-deploy in the public, dont u think so? maybe they'll be useful in practice. Right now i agree that we can make almost all debuggy

pkg/morph/deploy/notary.go
```diff
@@ -869,3 +878,353 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
...
+						zap.Stringer("ID", srcMainTxHash))
+					continue
+				case notaryEvent.Type != mempoolevent.TransactionAdded:
+					prm.logger.Info("unsupported type of the notary request event, skip",
+						zap.Stringer("got", notaryEvent.Type), zap.Stringer("expect", mempoolevent.TransactionAdded))
```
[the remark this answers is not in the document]
cthulhu-rider: it's logged with `info` severity, or what do u mean by `not an error`?

pkg/morph/deploy/notary.go
```diff
@@ -869,3 +878,353 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
...
+
+	committeeMultiSigAccID := committeeMultiSigAcc.ScriptHash()
+	chNotaryRequests := make(chan *result.NotaryRequestEvent, 100) // secure from blocking
+	// cache processed operations: when main transaction from received notary
+	// request is signed and sent by the local account, we receive the request from
+	// the channel again
+	mProcessedMainTxs := make(map[util.Uint256]struct{})
+
+	subID, err := prm.blockchain.ReceiveNotaryRequests(&neorpc.TxFilter{
+		Signer: &committeeMultiSigAccID,
```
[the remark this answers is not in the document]
cthulhu-rider: we won't recognize this. Changed committee will crash eveything before this. I suggest to create a separate issue `make deploy procedure support changing committee`. Right now our NNS hacks I tied to the fixed committee really much, and I guess adoption will bring a lot of changes that I would do in the future

pkg/morph/deploy/notary.go
```diff
@@ -869,3 +878,353 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
...
+				srcMainTxHash := mainTx.Hash()
+				_, processed := mProcessedMainTxs[srcMainTxHash]
+
+				switch {
+				case processed:
+					prm.logger.Info("main transaction of the notary request has already been processed, skip",
+						zap.Stringer("ID", srcMainTxHash))
+					continue
+				case notaryEvent.Type != mempoolevent.TransactionAdded:
+					prm.logger.Info("unsupported type of the notary request event, skip",
```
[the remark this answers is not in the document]
cthulhu-rider: it is not supported by this logic, so it's skipped, I don't see lies in this phrase. Ok, lets debug

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
cthulhu-rider: > simultaneousely

in general, definitely not.

yep, we send different requests because we can't predict which actor will succeed, that's why i've implemented this logic as it is

> proxy+committee 

i didn't really understood how Proxy contract will help us. Who sends notary request to update the NNS contract in ur suggestion?

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
AnnaShaleva: Right now it's you, and after this it may be another person who may wonder, why WARN log is printing, whether it's OK and should he do something about it. I doubd that for an abstract admin most these INFO/WARN logs are useful.

pkg/morph/deploy/nns.go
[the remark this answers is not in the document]
AnnaShaleva: > we can encounter case when NEF changed but version not bumped

Right now we can't, and that's what I'm talking about. Doany of our deployment/update procedurs assume that this can happen? I doubt so.

pkg/morph/deploy/nns.go
cthulhu-rider: then i will split `MakeCall` and `Notatize` calls to distinguish GAS and Notary balance problems

cthulhu-rider: actually even simple actor returns `insufficient funds` only when we `Send`, not `Make` transaction

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
AnnaShaleva: Discussed, agreed on max value.

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
AnnaShaleva: Let it be unresolved for now, and after PR finaliztion we'll create an issue for that.

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
AnnaShaleva: Usually when you have "got" and "expected" keywords you assume that something went wrong. And here everything goes OK, that's why I'm not sure whether "got" and "expected" suit here.

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
AnnaShaleva: OK, then we just assume that committee is constant during update, i think this assumption is quite safe in the current context.

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
AnnaShaleva: Discussed, moved to a separate issue (need to add the link).

pkg/morph/deploy/notary.go
AnnaShaleva: Discussed, neet to try out waiter in a separate commit and create an issue afterwards.

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
cthulhu-rider: #2416

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
cthulhu-rider: or even better to re-consider such cases separately from behavior implementation
if u don't mind, lets resolve later within #2419

pkg/morph/deploy/notary.go
```diff
@@ -869,3 +878,353 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
...
+						zap.Stringer("got", notaryEvent.Type), zap.Stringer("expect", mempoolevent.TransactionAdded))
+					continue
+				case len(mainTx.Signers) != expectedSigners:
+					prm.logger.Warn("unsupported number of signers of main transaction from the received notary request, skip",
```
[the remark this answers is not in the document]
cthulhu-rider: if u don't mind, lets resolve later within #2419

pkg/morph/deploy/notary.go
```diff
@@ -869,3 +878,353 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
...
+						zap.Int("expected", expectedSigners), zap.Int("got", len(mainTx.Signers)))
+					continue
+				case !mainTx.HasSigner(committeeMultiSigAccID):
+					// in theory, there can be any notary requests besides those sent by the current
+					// auto-deploy procedure. However, it's better to log with 'info' severity since
+					// this isn't really expected in practice.
+					prm.logger.Warn("committee is not a signer of main transaction from the received notary request, skip")
```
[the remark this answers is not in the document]
cthulhu-rider: if u don't mind, lets resolve later within #2419

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
cthulhu-rider: #2419

pkg/morph/deploy/nns.go
cthulhu-rider: fixed

pkg/morph/deploy/notary.go
cthulhu-rider: fixed

pkg/morph/deploy/notary.go
cthulhu-rider: fixed

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
cthulhu-rider: changed to max uint32

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
cthulhu-rider: done

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
cthulhu-rider: #2419

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
cthulhu-rider: #2419

pkg/morph/deploy/notary.go
cthulhu-rider: done

pkg/morph/deploy/notary.go
cthulhu-rider: created #2430

pkg/morph/deploy/notary.go
cthulhu-rider: added commit, check it out pls

AnnaShaleva:
OK, here's the testing code that demonstrates how version ov the contract can be retrieved from NEF with the help of test invocation. Use this test to retrieve local NNS version. To convert this test to real code just use invoker.Run on the provided script, provide proper invoker signers and proper native Management hash:
```
package main

import (
	"encoding/json"
	"strconv"
	"strings"
	"testing"

	"github.com/nspcc-dev/neo-go/pkg/compiler"
	"github.com/nspcc-dev/neo-go/pkg/core/interop/interopnames"
	"github.com/nspcc-dev/neo-go/pkg/core/native/nativenames"
	"github.com/nspcc-dev/neo-go/pkg/io"
	"github.com/nspcc-dev/neo-go/pkg/neotest"
	"github.com/nspcc-dev/neo-go/pkg/neotest/chain"
	"github.com/nspcc-dev/neo-go/pkg/smartcontract/callflag"
	"github.com/nspcc-dev/neo-go/pkg/vm/emit"
	"github.com/nspcc-dev/neo-go/pkg/vm/opcode"
	"github.com/stretchr/testify/require"
)

func TestCheckContractVersion(t *testing.T) {
	const version = 123

	bc, acc := chain.NewSingle(t)
	e := neotest.NewExecutor(t, bc, acc, acc)
	src := `package foo
		var version = ` + strconv.Itoa(version) + `
		func Version() int {
			return version
		}`
	ctr := neotest.CompileSource(t, e.CommitteeHash, strings.NewReader(src), &compiler.Options{Name: "Helper"})

	// Create transaction script.
	mgmtH := e.NativeHash(t, nativenames.Management)
	rawManifest, err := json.Marshal(ctr.Manifest)
	require.NoError(t, err)
	neb, err := ctr.NEF.Bytes()
	require.NoError(t, err)
	script := io.NewBufBinWriter()
	emit.Opcodes(script.BinWriter, opcode.NEWARRAY0)
	emit.Int(script.BinWriter, int64(callflag.All))
	emit.String(script.BinWriter, "version")
	emit.AppCall(script.BinWriter, mgmtH, "deploy", callflag.All, neb, rawManifest)
	emit.Opcodes(script.BinWriter, opcode.PUSH2, opcode.PICKITEM)
	emit.Syscall(script.BinWriter, interopnames.SystemContractCall)
	c := e.CommitteeInvoker(ctr.Hash)
	stack, err := c.TestInvokeScript(t, script.Bytes(), []neotest.Signer{e.Committee})
	require.NoError(t, err)
	require.Equal(t, 1, stack.Len())
	require.Equal(t, int64(version), stack.Pop().BigInt().Int64())
}
```

pkg/morph/deploy/deploy.go
```diff
@@ -203,17 +205,13 @@ func Deploy(ctx context.Context, prm Prm) error {
 
 	prm.Logger.Info("Notary service successfully enabled for the committee")
 
-	onNotaryDepositDeficiency, err := initNotaryDepositDeficiencyHandler(ctx, prm.Logger, prm.Blockchain, prm.LocalAccount)
-	if err != nil {
-		return fmt.Errorf("construct action depositing funds to the local account's Notary balance: %w", err)
-	}
+	go autoReplenishNotaryBalance(ctx, prm.Logger, prm.Blockchain, prm.LocalAccount, chNewBlock)
```
AnnaShaleva, line 208: Looks clearer now, let's keep it in a separate commit, I'll do some testing of this code.

pkg/morph/deploy/notary.go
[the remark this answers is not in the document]
AnnaShaleva: https://github.com/nspcc-dev/neofs-node/issues/2429.

pkg/morph/deploy/nns.go
```diff
@@ -141,47 +151,29 @@ func initNNSContract(ctx context.Context, prm deployNNSContractPrm) (res util.Ui
...
-			if cur := prm.monitor.currentHeight(); cur <= sentTxValidUntilBlock {
-				prm.logger.Info("previously sent transaction deploying NNS contract may still be relevant, will wait for the outcome",
-					zap.Uint32("current height", cur), zap.Uint32("retry after height", sentTxValidUntilBlock))
-				continue
-			}
-
-			prm.logger.Info("previously sent transaction deploying NNS contract expired without side-effect")
+		if txMonitor.isPending() {
+			prm.logger.Info("previously sent transaction updating NNS contract is still pending, will wait for the outcome")
+			continue
```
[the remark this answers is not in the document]
cthulhu-rider: ok, lets discuss it deeper in a separate issue. But i still think changes with Waiter are cleaner than simple ValidUntilBlock comparison. For example, if transaction fails in next block (so no side-effect), and vub=100, then we won't wait 100 blocks to retry. So, i suggest to keep this commit for now, and refactor this later within #2432 if needed. Okay?

cthulhu-rider:
> here's the testing code

@AnnaShaleva i tried and it works great! will resolve FIXME in-place

pkg/morph/deploy/nns.go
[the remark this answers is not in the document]
AnnaShaleva: Good, but then I'll take a look at it one more time and try to do some changes after today's workshop. I'll write back with the comments.

AnnaShaleva:
> > here's the testing code
> 
> @AnnaShaleva i tried and it works great! will resolve FIXME in-place

I reccomend you to use dummy zero-hash simple signer for invoker. This allows to avoid collisions on version retrieval of the local contract that has already been deployed to the chain.

cthulhu-rider:
> dummy zero-hash simple signer

not sure i got ur point right. Just check out new code

pkg/morph/deploy/notary.go
```diff
@@ -869,3 +878,349 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
...
+			// GAS:Notary proportion, see scheme above
+			gasProportion = 50
+			// even there is no lack of GAS at the moment, when the balance falls below 1/5
+			// of the supported value - replenish
+			refillProportion = 5
+			// for simplicity, we just make Notary deposit "infinite" not to prolong
+			till = math.MaxUint32
+		)
+
+		if !lackOfGAS { // deposit expired
```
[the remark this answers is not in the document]
cthulhu-rider: we can get here if we started with existing limited deposit. IMO this is ok for this commit, and we change behaviour in 2f1ed93b17ca3b565130418180672588f7b9b486 anyway

pkg/morph/deploy/nns.go
```diff
@@ -226,3 +227,174 @@ func lookupNNSDomainRecord(inv *invoker.Invoker, nnsContract util.Uint160, domai
...
+			if isErrContractAlreadyUpdated(err) {
+				// FIXME: we get here when on-chain version is >= than the local one. If equals,
+				// this case must be considered as failure since version must be changed if code
+				// changes (NEF checksum differs here according to condition above). If greater,
+				// we should try to update at all. For simplicity, such condition is considered
+				// as success by current procedure now. To catch it, we need to pre-read version
+				// of the local contract using local NEF.
+				prm.logger.Info("NNS contract has already been updated, skip")
+				return nil
+			}
```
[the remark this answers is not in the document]
cthulhu-rider: done

pkg/morph/deploy/nns.go
```diff
@@ -226,3 +227,170 @@ func lookupNNSDomainRecord(inv *invoker.Invoker, nnsContract util.Uint160, domai
...
+			lackOfGAS := isErrNotEnoughGAS(err)
+			// here lackOfGAS=true always means lack of Notary balance and not related to
+			// the main transaction itself
+			if !lackOfGAS {
+				if !isErrNotaryDepositExpires(err) {
+					prm.logger.Error("failed to send transaction deploying NNS contract, will try again later", zap.Error(err))
+					continue
+				}
+			}
+
```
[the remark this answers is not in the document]
cthulhu-rider: to me insta `err != nil` check is the best practice regardless of code block size. We can shrink `err != nil` or grow in `err == nil` code and ur suggestion will become irrelevant and will require re-refactor.
```go
err := f()
err != nil {
 // handle error
}

// OK code
```

but i dont mind, lets do this

pkg/morph/deploy/nns.go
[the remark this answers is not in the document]
cthulhu-rider: we already have an issue for that, so i suggest to not stuck while this funcionally works

pkg/morph/deploy/notary.go
```diff
@@ -869,3 +878,349 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
...
+				case processed:
+					prm.logger.Info("main transaction of the notary request has already been processed, skip",
+						zap.Stringer("ID", srcMainTxHash))
+					continue
+				case notaryEvent.Type != mempoolevent.TransactionAdded:
+					prm.logger.Info("unsupported type of the notary request event, skip",
+						zap.Stringer("got", notaryEvent.Type), zap.Stringer("expect", mempoolevent.TransactionAdded))
+					continue
+				case len(mainTx.Signers) != expectedSigners:
+					prm.logger.Warn("unsupported number of signers of main transaction from the received notary request, skip",
```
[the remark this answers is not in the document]
cthulhu-rider: done

pkg/morph/deploy/notary.go
```diff
@@ -871,3 +841,286 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
...
+						zap.Stringer("got", notaryEvent.Type), zap.Stringer("expect", mempoolevent.TransactionAdded))
+					continue
+				case len(mainTx.Signers) != expectedSigners:
+					prm.logger.Info("unsupported number of signers of main transaction from the received notary request, skip",
+						zap.Int("expected", expectedSigners), zap.Int("got", len(mainTx.Signers)))
+					continue
+				case !mainTx.HasSigner(committeeMultiSigAccID):
+					// in theory, there can be any notary requests besides those sent by the current
+					// auto-deploy procedure. However, it's better to log with 'info' severity since
+					// this isn't really expected in practice.
```
AnnaShaleva, line 1049: > this isn't really expected in practice

It's not true, change this comment. You can (and you likely will) start the auto-update procedure on the working network. Working network assumes that notary requests can appear from time to time. E.g. notary request on container put - this notary request is also signed by the committee and can also have three signers. Is it possible that some user will deside to cal `container put` during this integrated contract update procedure? Yes.

pkg/morph/deploy/notary.go
```diff
@@ -871,3 +841,286 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
...
+					prm.logger.Info("committee is not a signer of main transaction from the received notary request, skip")
+					continue
+				case mainTx.HasSigner(prm.localAcc.ScriptHash()):
+					prm.logger.Info("main transaction from the received notary request is signed by a local account, skip")
+					continue
+				case len(mainTx.Scripts) == 0:
+					prm.logger.Warn("missing scripts of main transaction from the received notary request, skip")
```
AnnaShaleva, line 1056: INFO, please. And https://github.com/nspcc-dev/neofs-node/issues/2419 of course.

AnnaShaleva (commented):
Good, the code became much cleaner and the overall logic looks OK to me (with an eye for created issues). We need to fix a couple of minor issues.

pkg/morph/deploy/nns.go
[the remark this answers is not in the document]
AnnaShaleva: > insta err != nil check is the best practice

It's always preferable, but not for the nested IFs, e.g. https://stackoverflow.com/questions/337037/replacing-nested-if-statements and https://stackoverflow.com/questions/268132/invert-if-statement-to-reduce-nesting.

It was kind of a corporate culture in EPAM...

pkg/morph/deploy/nns.go
```diff
@@ -229,3 +221,162 @@ func lookupNNSDomainRecord(inv *invoker.Invoker, nnsContract util.Uint160, domai
...
+		}
+
+		if v := localVersion.cmp(versionOnChain); v == -1 {
+			prm.logger.Info("local contract version is < than the on-chain one, update is not needed",
+				zap.Stringer("local", localVersion), zap.Stringer("on-chain", versionOnChain))
+			return nil
+		} else if v == 0 {
+			if nnsOnChainState.NEF.Checksum != prm.localNEF.Checksum {
+				return fmt.Errorf("local and on-chain contracts have different NEF checksums but same version '%s'", versionOnChain)
+			}
```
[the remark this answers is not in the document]
cthulhu-rider: fixed

pkg/morph/deploy/util.go
```diff
@@ -166,3 +187,136 @@ func readNNSOnChainState(b Blockchain) (*state.Contract, error) {
...
+	if err != nil {
+		return contractVersion{}, fmt.Errorf("run test script deploying contract and calling its '%s' method: %w", methodVersion, err)
+	}
+
+	return parseContractVersionFromInvocationResult(res)
+}
+
+type transactionGroupWaiter interface {
+	WaitAny(ctx context.Context, vub uint32, hashes ...util.Uint256) (*state.AppExecResult, error)
+}
```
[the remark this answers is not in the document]
cthulhu-rider: my goal was to make it possible to pass `notary.Actor` as waiter. Notary actor doesn't implement `actor.Waiter` (another `Wait` signature). So we need either make impossible to use `notary.Actor` as waiter or keep this narrowed in-place interface.

pkg/morph/deploy/notary.go
```diff
@@ -871,3 +841,286 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
...
+			case notaryEvent, ok := <-chNotaryRequests:
+				if !ok {
+					prm.logger.Info("stop listening to committee notary requests (subscription channel closed)")
+					return
+				}
+
+				// for simplicity, requests are handled one-by one. We could process them in parallel
+				// using worker pool, but actions seem to be relatively lightweight
+
+				const expectedSigners = 3 // sender + committee + Notary
```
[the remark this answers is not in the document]
cthulhu-rider: done

pkg/morph/deploy/util.go
```diff
@@ -166,3 +187,136 @@ func readNNSOnChainState(b Blockchain) (*state.Contract, error) {
...
+	return 1
+}
+
+func (x contractVersion) String() string {
+	const sep = "."
+	return fmt.Sprintf("%d%s%d%s%d", x.major, sep, x.minor, sep, x.patch)
+}
+
+// parses contractVersion from the invocation result of methodVersion method.
+func parseContractVersionFromInvocationResult(res *result.Invoke) (contractVersion, error) {
```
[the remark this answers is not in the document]
cthulhu-rider: added

pkg/morph/deploy/util.go
```diff
@@ -166,3 +187,136 @@ func readNNSOnChainState(b Blockchain) (*state.Contract, error) {
...
+	if err != nil {
+		return contractVersion{}, fmt.Errorf("call '%s' contract method: %w", methodVersion, err)
+	}
+
+	return parseContractVersionFromInvocationResult(res)
+}
+
+// readContractLocalVersion returns current version of the smart contract
+// presented in given Blockchain with specified address.
+func readContractLocalVersion(rpc invoker.RPCInvoke, localNEF nef.File, localManifest manifest.Manifest) (contractVersion, error) {
```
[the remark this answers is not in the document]
cthulhu-rider: added

pkg/morph/deploy/util.go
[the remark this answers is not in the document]
AnnaShaleva: Reasonable, and since we'd like to keep notary Actor's `Wait` as is (because it's needed for the users), then I doubt we can make Notary Actor implement Waiter. So let's keep it as is.

AnnaShaleva (approved):
Add a note to CHANGELOG, please.

cthulhu-rider:
> Add a note to CHANGELOG, please.

i haven't added changelog note intentionally since these changes are intermediate. Will edit changelog when the whole task will be completed

pkg/morph/deploy/deploy.go
```diff
@@ -134,7 +134,9 @@ func Deploy(ctx context.Context, prm Prm) error {
 		return errors.New("local account does not belong to any Neo committee member")
 	}
 
-	monitor, err := newBlockchainMonitor(prm.Logger, prm.Blockchain)
+	chNewBlock := make(chan struct{}, 1)
```
carpawell, line 137: why `1`? what does it solve?

pkg/morph/deploy/notary.go
```diff
@@ -871,51 +872,56 @@ func newCommitteeNotaryActor(b Blockchain, localAcc *wallet.Account, committee k
...
-	transferTxMonitor := newTransactionGroupMonitor(localActor)
-	expirationTxMonitor := newTransactionGroupMonitor(localActor)
+	for {
+		select {
+		case <-ctx.Done():
+			l.Info("Notary balance tracker stopped by context", zap.Error(ctx.Err()))
+			return
+		case _, ok := <-chTrigger:
+			if !ok {
+				l.Info("Notary balance tracker stopped by closed block channel", zap.Error(ctx.Err()))
```
carpawell, line 896: why `ctx.Err()` in than case?

pkg/morph/deploy/notary.go
```diff
@@ -871,51 +872,56 @@ func newCommitteeNotaryActor(b Blockchain, localAcc *wallet.Account, committee k
...
 
-		gasBalance, err := gasContract.BalanceOf(localAccID)
-		if err != nil {
-			l.Error("failed to read GAS token balance of the local account, will try again later", zap.Error(err))
-			return
+		// deposit when balance falls below 1/5 of the single deposit amount
+		const refillProportion = 5
+
+		if new(big.Int).Mul(notaryBalance, big.NewInt(refillProportion)).Cmp(singleNotaryDepositAmount) >= 0 {
+			l.Info("enough funds on the notary balance, deposit is not needed", zap.Stringer("balance", notaryBalance))
```
carpawell, line 923: would not go `INFO` too

pkg/morph/deploy/notary.go
```diff
@@ -925,56 +931,14 @@ func initNotaryDepositDeficiencyHandler(ctx context.Context, l *zap.Logger, b Bl
...
-			return
+		if txMonitor.isPending() {
+			l.Info("previously sent transaction transferring local account's GAS to the Notary contract is still pending, will wait for the outcome")
+			continue
 		}
 
 		var transferData notary.OnNEP17PaymentData
 		transferData.Account = &localAccID
-		transferData.Till = till
+		transferData.Till = math.MaxUint32 // deposit "forever" so we don't have to renew
```
carpawell, line 941: @AnnaShaleva, can we have problems in future work after such a deposit? e.g. when that acc will try to deposit the lower GAS amount?

pkg/morph/deploy/deploy.go
cthulhu-rider: non-blocking write with `select`

pkg/morph/deploy/notary.go
cthulhu-rider: will be resolved within #2419. For now it just helps to monitor current state

pkg/morph/deploy/notary.go
cthulhu-rider: > lower GAS amount

lower that _what_ do u mean? afaik any additional deposits will add GAS to the balance and that's it

i moved to `math.MaxUint32` by @AnnaShaleva advice btw https://github.com/nspcc-dev/neofs-node/pull/2408#discussion_r1252029371

pkg/morph/deploy/notary.go
cthulhu-rider: mistaken, fixed

pkg/morph/deploy/notary.go
carpawell: yes, i just needed some reference to that line

pkg/morph/deploy/notary.go
carpawell: i meant that we may have 50GAS deposit until maxInt an we do 1GAS deposit until maxInt, does it make the actual deposit 51GAS? i think yes but i always forget some details about notary so i wanted to be sure

if it is ok, then it is ok

pkg/morph/deploy/deploy.go
carpawell: IMO we either think that block handling is good enough and never have buffered channels or make its buffer bigger. if we are handling one block longer than the block lasts it is a possible deadlock (we saw and fixed it so many times).

`1` confuses me, so my question is what situation exactly you tried to solve here?

cthulhu-rider: we receive new block and try to send signal balance routine:
* if it's not busy, it receives signal from channel and makes balance actions
* if it's busy (currently making actions), we skip action for current block and will try again on future ones

carpawell (approved):

pkg/morph/deploy/notary.go
AnnaShaleva: Yes, it works as you described, and it's OK.

roman-khimov: The only problem is that this GAS is locked forever. But who cares about this GAS anyway...

PR #2434, opened 2023-07-12
fix/Restart SN services after `neo-go` is down

### LINKED ISSUE #1337: Restart services after switching to another neo-go RPC node

carpawell:
Closes #1337. The https://github.com/nspcc-dev/neofs-node/pull/2406 but for SN. Also tested with https://github.com/notimetoname/neofs-dev-env/tree/local/add-rpc-morph-node.

### DIFF
--- a/cmd/neofs-node/main.go
+++ b/cmd/neofs-node/main.go
@@ -154,6 +154,8 @@ func (c *cfg) onShutdown(f func()) {
 }
 
 func (c *cfg) restartMorph() error {
+	c.log.Info("restarting internal services because of RPC connection loss...")
+
 	c.shared.resetCaches()
 
 	epoch, ni, err := getNetworkState(c)
@@ -179,6 +181,8 @@ func (c *cfg) restartMorph() error {
 	if err != nil {
 		c.log.Warn("failed to re-bootstrap", zap.Error(err))
 	}
+
+	c.log.Info("internal services have been restarted after RPC connection loss")
 
 	return nil
 }
--- a/cmd/neofs-node/main.go
+++ b/cmd/neofs-node/main.go
@@ -152,3 +152,7 @@ func shutdown(c *cfg) {
 func (c *cfg) onShutdown(f func()) {
 	c.closers = append(c.closers, f)
 }
+
+func (c *cfg) restartMorph() error {
+	return nil
+}
--- a/cmd/neofs-node/main.go
+++ b/cmd/neofs-node/main.go
@@ -154,5 +154,7 @@ func (c *cfg) onShutdown(f func()) {
 }
 
 func (c *cfg) restartMorph() error {
+	c.shared.resetCaches()
+
 	return nil
 }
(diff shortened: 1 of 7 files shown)


cmd/neofs-node/main.go
```diff
@@ -152,3 +152,37 @@ func shutdown(c *cfg) {
...
+		return nil
+	}
+
+	err = c.bootstrap()
+	if err != nil {
+		c.log.Warn("can't send re-bootstrap tx", zap.Error(err))
+	}
+
+	// drop expired sessions if any has appeared while node was sleeping
+	c.shared.privateTokenStore.RemoveOld(epoch)
```
cthulhu-rider, line 183: looks very specific, are you sure that this is and will be the only part that needs to be updated? to me `start` and `restart` aren't so different: `start` starts from "undefined" epoch while `restart` from the cached one.

cmd/neofs-node/main.go
```diff
@@ -152,3 +152,37 @@ func shutdown(c *cfg) {
...
+	// way to make a heartbeat so nothing is wrong in the making sure a
+	// node is online (if it should be)
+
+	if !c.needBootstrap() || c.cfgNetmap.reBoostrapTurnedOff.Load() {
+		return nil
+	}
+
+	err = c.bootstrap()
+	if err != nil {
+		c.log.Warn("can't send re-bootstrap tx", zap.Error(err))
```
cthulhu-rider, line 179: `failed to re-bootstrap`, specific reason is in the `err` context i hope

cmd/neofs-node/main.go
carpawell: can you elaborate on your thoughts?

i look through the whole code and there is not that much code that caches smth. every other client user (in the node, i mean) just calls the methods and does everything it needs without caching

there are not that many "services" in the node in fact. storage is the main but [we cannot stop](https://github.com/nspcc-dev/neofs-node/issues/1337#issuecomment-1514791288) it, cause it would be like restarting the whole node

cthulhu-rider: i imagine any component (e.g. not yet added) which reacts the epoch change. Don't we forget to pass `epoch` to this component here? 

talking about `privateTokenStore`: we have this logical binding
https://github.com/nspcc-dev/neofs-node/blob/e5ac376c7081669fbb59c79e68cfd8eb208ccc2a/cmd/neofs-node/session.go#L49-L51

and u add new case when we handle exact epoch. Don't u think it'd be nice to have single binding of `privateTokenStore`  and new epochs?

roman-khimov (approved):

cmd/neofs-node/main.go
cthulhu-rider: okay, lets leave it as it is for now

and i'd think about single epoch controller for the node as a possible enhancement (smart controller orchestrating ordinary components)

cthulhu-rider (approved):

cthulhu-rider (approved):

PR #2437, opened 2023-07-13
feat/Add metrics per container and per shard

carpawell:


CHANGELOG.md
```diff
@@ -115,10 +116,6 @@ data storage scheme is compatible with 0.35.0.
 
 ### Added
 - Doc for extended headers (#2128)
-- Separate batching for replicated operations over the same container in pilorama (#1621)
```
cthulhu-rider, line 118: removed accidentally?

pkg/local_object_storage/metabase/delete.go
```diff
@@ -35,6 +36,11 @@ func (d DeleteRes) RawObjectsRemoved() uint64 {
 	return d.rawRemoved
 }
 
+// RemovedObjectSizes returns the sizes of removed objects.
+func (d DeleteRes) RemovedObjectSizes() []uint64 {
```
cthulhu-rider, line 40: i'd clarify the slice semantics in docs. IIUC it corresponds to parameterized address list. Zero values are also not so obvious

pkg/metrics/engine.go
```diff
@@ -102,6 +104,20 @@ func newEngineMetrics() engineMetrics {
...
+			Namespace: storageNodeNameSpace,
+			Subsystem: engineSubsystem,
+			Name:      "container_size",
+			Help:      "Accumulated size of all objects in a container",
+		}, []string{containerIDLabelKey})
+
+		payloadSize = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Namespace: storageNodeNameSpace,
+			Subsystem: engineSubsystem,
+			Name:      "payload_size",
```
cthulhu-rider, line 118: this name seems to generic to me, how about `shard_payload_size/volume`?

pkg/local_object_storage/shard/delete.go
```diff
@@ -78,8 +78,16 @@ func (s *Shard) delete(prm DeletePrm) (DeleteRes, error) {
 		return DeleteRes{}, err // stop on metabase error ?
 	}
 
+	var totalRemovedPayload uint64
+
 	s.decObjectCounterBy(physical, res.RawObjectsRemoved())
 	s.decObjectCounterBy(logical, res.AvailableObjectsRemoved())
+	for i := range prm.addr {
+		removedPayload := res.RemovedObjectSizes()[i]
```
cthulhu-rider, line 86: lets not call the method again and again

pkg/local_object_storage/shard/shard.go
```diff
@@ -336,6 +342,28 @@ func (s *Shard) updateObjectCounter() {
...
+		cnrList, err := s.metaBase.Containers()
+		if err != nil {
+			s.log.Warn("meta: can't read container list", zap.Error(err))
+			return
+		}
+
+		var totalPayload uint64
+
+		for i := range cnrList {
+			size, err := s.metaBase.ContainerSize(cnrList[i])
```
cthulhu-rider, line 355: first of all i wanted to write my concern that these ops seems pretty heavyweight for `updateMetrics`. Then i looked up where the method is called - only in `Init`. I suggest to rename to `initMetrics`

pkg/local_object_storage/metabase/delete.go
```diff
@@ -66,9 +72,11 @@ func (db *DB) Delete(prm DeletePrm) (DeleteRes, error) {
 	var rawRemoved uint64
 	var availableRemoved uint64
 	var err error
+	var sizes = make([]uint64, len(prm.addrs))
```
cthulhu-rider, line 75: do we always process these values on caller side? if not, as a slight optimization, we can make size collection optional

carpawell: > do we always process these values on caller side?

if metrics are turned off, that would be useless memuse, yes (but usually a stack mem, i guess). but i do not like the idea to provide an additional parameter (more complex code) for that. @roman-khimov

pkg/metrics/engine.go
carpawell: it will always have a `"shard"` label valued with its ID. do you still think it is required (i dont mind just confirm it, please)?

pkg/local_object_storage/metabase/delete.go
roman-khimov: We rarely run without metrics and I doubt it'll make any noticeable difference. So unless we have some real performance metrics suggesting a change, I'd stick to simpler code.

pkg/metrics/engine.go
cthulhu-rider: if `shard` will be mentioned, then it's ok

cthulhu-rider (approved):

pkg/local_object_storage/shard/control.go
```diff
@@ -172,7 +172,7 @@ func (s *Shard) refillMetabase() error {
 			return nil
 		}
 
-		//nolint: exhaustive
+		// nolint: exhaustive
```
roman-khimov, line 175: Uh-oh

pkg/local_object_storage/shard/metrics_test.go
```diff
@@ -17,35 +17,36 @@ import (
 )
 
 type metricsStore struct {
-	objectCounters map[string]uint64
-	readOnly       bool
+	objCounters map[string]uint64
```
roman-khimov, line 20: Naming in one commit and renaming in another is not the best practice.

roman-khimov (approved):

pkg/local_object_storage/shard/control.go
carpawell: fixed

pkg/local_object_storage/shard/metrics_test.go
carpawell: fixed

roman-khimov (approved):

PR #2443, opened 2023-07-17
fix/Do not vote for the same validators list twice

### LINKED ISSUE #2365: Do not vote for the same node on IR start
## Expected Behavior
Voting should be done when it can change something.

## Current Behavior
IR node will vote for whatever key is written into its config unconditionally.

## Possible Solution
Check NEO state before voting, likely there is already a vote and likely it's for the same key.

Restart IR node.

## Context
Related to #2363.

carpawell:
Closes #2365.

### DIFF
--- a/pkg/innerring/state.go
+++ b/pkg/innerring/state.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"sort"
 
+	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
 	"github.com/nspcc-dev/neo-go/pkg/util"
 	"github.com/nspcc-dev/neofs-node/pkg/innerring/processors/governance"
 	auditClient "github.com/nspcc-dev/neofs-node/pkg/morph/client/audit"
@@ -108,13 +109,21 @@ func (s *Server) voteForSidechainValidator(prm governance.VoteValidatorPrm) erro
 		return nil
 	}
 
+	voted, err := s.alreadyVoted(validators)
+	if err != nil {
+		return fmt.Errorf("could not check validators state: %w", err)
+	}
+
+	if voted {
+		return nil
+	}
+
 	epoch := s.EpochCounter()
 
 	var (
 		nonce uint32 = 1
 		vub   uint32
 		vubP  *uint32
-		err   error
 	)
 
 	if prm.Hash != nil {
@@ -136,6 +145,30 @@ func (s *Server) voteForSidechainValidator(prm governance.VoteValidatorPrm) erro
 	})
 
 	return nil
+}
+
+func (s *Server) alreadyVoted(validatorsToVote keys.PublicKeys) (bool, error) {
+	currentValidators := make(map[keys.PublicKey]struct{}, len(s.contracts.alphabet))
+	for letter, contract := range s.contracts.alphabet {
+		validator, err := s.morphClient.AccountVote(contract)
+		if err != nil {
+			return false, fmt.Errorf("receiving %s's vote: %w", letter, err)
+		}
+
+		if validator == nil {
+			continue
+		}
+
+		currentValidators[*validator] = struct{}{}
+	}
+
+	for _, v := range validatorsToVote {
+		if _, voted := currentValidators[*v]; !voted {
+			return false, nil
+		}
+	}
+
+	return true, nil
 }
 
 // VoteForSidechainValidator calls vote method on alphabet contracts with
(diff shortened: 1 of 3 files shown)


carpawell:
@AnnaShaleva, @roman-khimov, ping me if you think that we need to make it more accurate with `GetNextBlockValidators` or even with `GetAccountState` (and check whether an alphabet contract has voted for the required validator).

pkg/innerring/state.go
```diff
@@ -138,6 +147,28 @@ func (s *Server) voteForSidechainValidator(prm governance.VoteValidatorPrm) erro
 	return nil
 }
 
+func (s *Server) alreadyVoted(validatorsToVote keys.PublicKeys) (bool, error) {
+	// almost every part of NeoFS treats committee as validators list
+	validators, err := s.morphClient.Committee()
```
AnnaShaleva, line 152: Are you sure that it's safe to treat commettee as validators here? If these lists differ, then you may end up in a situation when you have to vote for some validator, but `alreadyVoted` method can return true for him (e.g. because he's a committee member, but not the validator yet).

roman-khimov: `len(committee) == len(validators)` is a wrong assumption which can be fixed easily. And you _must_ do `GetAccountState()`, you're just checking for a different thing otherwise.

Frankly, I was expecting a simple `GetAccountState()` (ok, a number of them) and comparison with the configured set. Then you don't even need to care about committee/validators list. You only need to care about votes here.

pkg/morph/client/client.go
```diff
@@ -419,6 +420,30 @@ func (c *Client) IsValidScript(script []byte, signers []transaction.Signer) (res
...
+// tokens for. Nil key with no error it returned if the account has no NEO.
+func (c *Client) VotesFor(addr util.Uint160) (*keys.PublicKey, error) {
+	c.switchLock.RLock()
+	defer c.switchLock.RUnlock()
+
+	if c.inactive {
+		return nil, ErrConnectionLost
+	}
+
+	neoReader := neo.New(c.rpcActor)
```
roman-khimov, line 433: Can have `nil` signers, doesn't really matter here (but a little less things to encode/decode for the request).

pkg/morph/client/client.go
```diff
@@ -419,6 +420,30 @@ func (c *Client) IsValidScript(script []byte, signers []transaction.Signer) (res
 	return result.State == vmstate.Halt.String(), nil
 }
 
+// VotesFor returns a key the provided account has voted with its NEO
+// tokens for. Nil key with no error it returned if the account has no NEO.
+func (c *Client) VotesFor(addr util.Uint160) (*keys.PublicKey, error) {
```
roman-khimov, line 425: `Votes` sound plural here, while we only have a single vote for a single account. Maybe `AccountVote`.

roman-khimov (approved):

pkg/morph/client/client.go
carpawell: a meant "`addr` votes for". was it hard to understand?

pkg/morph/client/client.go
carpawell: you meant `New` -> `NewReader`? fixed

roman-khimov: Yeah, `NewReader` is faster and again, you can omit `c.rpcActor`.

pkg/morph/client/client.go
roman-khimov: I see what was meant, but it doesn't sound right anyway. It's not `addr.VotesFor()` also, it's `client.VotesFor`, but client doesn't vote. In general functions are verbs, so some `GetAccountVote()` is more appropriate. At the same time we often omit `get` for data accesses.

carpawell: ok, `AccountVote`

pkg/morph/client/client.go
carpawell: oh, you meant `neoReader := neo.NewReader(invoker.New(c.client, nil))`?

carpawell: the last push already has the `invoker` version. dont mind keeping it like that

pkg/morph/client/client.go
```diff
@@ -419,6 +421,30 @@ func (c *Client) IsValidScript(script []byte, signers []transaction.Signer) (res
 	return result.State == vmstate.Halt.String(), nil
 }
 
+// AccountVote returns a key the provided account has voted with its NEO
+// tokens for. Nil key with no error is returned if the account has no NEO.
```
AnnaShaleva, line 425: `Nil key with no error is returned if the account has no NEO` or when account has some NEO and didn't voted for anyone.

AnnaShaleva (approved):

PR #2445, opened 2023-07-18
Disable aws ec2 metadata

vvarg229:
Fixed bug https://github.com/nspcc-dev/neofs-s3-gw/issues/798 This bug is only reproducible on ubuntu-latest in github actions. This is a bug specifically of github acions, not the neofs-s3-gw: https://github.com/actions/runner-images/issues/2791 https://github.com/aws/aws-cli/issues/5623

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -27,6 +27,8 @@ jobs:
   run_system_tests:
     runs-on: ubuntu-latest
     timeout-minutes: 500
+    env:
+      AWS_EC2_METADATA_DISABLED: true
     steps:
       - name: Checkout neofs-testcases repository
         uses: actions/checkout@v3


carpawell (commented):
Have you considered setting a region like someone [fixed](https://github.com/go-sari/sari-web/commit/df16e3c48d9066349320f46e538a1dd30b604889) it?

roman-khimov:
I'd like to have more specific description here, what exactly is the relationship between this var and tests. It also looks like this better be fixed in testcases, if there is anything influencing the result it should be controlled there.

vvarg229:
> I'd like to have more specific description here, what exactly is the relationship between this var and tests. It also looks like this better be fixed in testcases, if there is anything influencing the result it should be controlled there.

Yes, but this is a test environment configuration bug. It's not even a bug, it's a configuration requirement.
On other runners (local laptop, server inside infrastructure, etc) this bug is not present.
So I think the correct way to approach this would be to fix it on the GitHub runner side.

vvarg229:
> Have you considered setting a region like someone [fixed](https://github.com/go-sari/sari-web/commit/df16e3c48d9066349320f46e538a1dd30b604889) it?

https://github.com/nspcc-dev/neofs-node/pull/2445#issuecomment-1641706101

smallhive (approved):

carpawell:
> On other runners (local laptop, server inside infrastructure, etc) this bug is not present.

Cause their environment is outdated now _OR_ the "bug" _will never_ come to them?

> it's a configuration requirement

I briefly read the original issue description and I think that it is more about the way `aws` CLI is called not the way neofs-node is tested. That repo does not call `aws` at all, right? The neofs-testcases [does](https://github.com/nspcc-dev/neofs-testcases/blob/79bf0710906c8f380a682ff15cc94196afaa8f67/pytest_tests/helpers/aws_cli_client.py#L42). That is why i asked about the "call step" [solution](https://github.com/go-sari/sari-web/commit/df16e3c48d9066349320f46e538a1dd30b604889).

vvarg229:
@roman-khimov @carpawell , I agree with you, the fix is better done in the https://github.com/nspcc-dev/neofs-testcases repository.
I'm closing this PR.

PR #2449, opened 2023-07-20
Upgrade NeoFS SDK module to the latest revision

cthulhu-rider:
decided to have up-to-date SDK changes here to simplify debug and prepare for https://github.com/nspcc-dev/neofs-sdk-go/milestone/5

* also based on https://github.com/nspcc-dev/neofs-sdk-go/pull/479


cthulhu-rider:
protocol details are deeply rooted in the metabase implementation. Because this is a root and change-sensitive component, it needs to be edited more carefully

cmd/neofs-cli/modules/accounting/balance.go
```diff
@@ -34,13 +34,12 @@ var accountingBalanceCmd = &cobra.Command{
 
 		balanceOwner, _ := cmd.Flags().GetString(ownerFlag)
 		if balanceOwner == "" {
-			err := user.IDFromSigner(&idUser, neofsecdsa.SignerRFC6979(*pk))
-			common.ExitOnErr(cmd, "decoding user from key", err)
+			idUser.SetScriptHash((*keys.PublicKey)(&pk.PublicKey).GetScriptHash())
```
smallhive, line 37: This row is a good candidate to be moved to a separate function inside the `user` package because there are few places in the code, which look too similar

cthulhu-rider: yeah but we dropped "user from key" function, so, looks like it needs to be resurrected

cthulhu-rider: was added by https://github.com/nspcc-dev/neofs-sdk-go/commit/6ab5d39ba0bf62b31e9423091f508d35cb890ec5, so now i can adopt it

Issue #2382, opened 2023-06-14
Object PUT, No space left


smallhive:
Working with [AIO](https://github.com/nspcc-dev/neofs-aio) got an error in docker container
```bash
2023-06-14T05:28:40.431Z	debug	blobovniczatree/put.go:54	blobovnicza overflowed	{"path": "0/0"}
2023-06-14T05:28:40.431Z	debug	blobovniczatree/blobovnicza.go:145	updating active blobovnicza...	{"path": "0"}
2023-06-14T05:28:40.431Z	debug	blobovniczatree/blobovnicza.go:149	active blobovnicza successfully updated	{"path": "0"}
2023-06-14T05:28:40.431Z	debug	blobovniczatree/put.go:62	could not update active blobovnicza	{"level": "0", "error": "no more Blobovniczas"}
2023-06-14T05:28:40.431Z	warn	engine/put.go:137	could not put object to shard	{"shard_id": "Egr2dzCvRk1iCFRqVdEp8d", "error": "could not put object to BLOB storage: no free space"}
2023-06-14T05:28:40.431Z	error	util/log.go:11	object service error	{"node": "/ip4/0.0.0.0/tcp/8080", "request": "PUT", "error": "could not close object stream: (*putsvc.localTarget) could not put object to local storage: could not put object to any shard"}
2023-06-14T05:28:40.920Z	INFO	persisted to disk	{"blocks": 1, "keys": 49, "headerHeight": 2562, "blockHeight": 2562, "took": "6.153443ms"}
2023-06-14T05:28:41.385Z	INFO	sending PrepareRequest	{"height": 2563, "view": 0}
2023-06-14T05:28:41.385Z	INFO	sending Commit	{"height": 2563, "view": 0}
2023-06-14T05:28:41.386Z	INFO	approving block	{"height": 2563, "hash": "932e7028c0af3d8415ebf7fa9ea5fd5cc24a3abeef2f2bf15a0ffc99b722b8de", "tx_count": 0, "merkle": "0000000000000000000000000000000000000000000000000000000000000000", "prev": "be7d4accab8c62c8c5f6416a8eaf8b8d5c29fa1b8a5703f043bc4c76c29d09dc"}
```
If fact, there was enough space on disk, more then 100GB. Tried to save 17 bytes file

## Expected Behavior
Successful object put

## Current Behavior
Object put failed

## Steps to Reproduce (for bugs)
Not regular bug. It happens sometimes

## Your Environment
[...]

PR #2450, opened 2023-07-21
blobstor: Try another sub-storage when the current one is full

cthulhu-rider:
general improvement mentioned https://github.com/nspcc-dev/neofs-node/issues/2382#issuecomment-1645390685

### DIFF
--- a/pkg/local_object_storage/blobstor/put.go
+++ b/pkg/local_object_storage/blobstor/put.go
@@ -1,12 +1,14 @@
 package blobstor
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/nspcc-dev/neofs-node/pkg/core/object"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor/common"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/util/logicerr"
 	objectSDK "github.com/nspcc-dev/neofs-sdk-go/object"
+	"go.uber.org/zap"
 )
 
 // ErrNoPlaceFound is returned when object can't be saved to any sub-storage component
@@ -42,6 +44,10 @@ func (b *BlobStor) Put(prm common.PutPrm) (common.PutRes, error) {
 			res, err := b.storage[i].Storage.Put(prm)
 			if err == nil {
 				logOp(b.log, putOp, prm.Address, b.storage[i].Storage.Type(), res.StorageID)
+			} else if errors.Is(err, common.ErrNoSpace) {
+				b.log.Debug("blobstor sub-storage overflowed, will try another one",
+					zap.String("type", b.storage[i].Storage.Type()))
+				continue
 			}
 			return res, err
 		}
--- a/pkg/local_object_storage/blobstor/put_test.go
+++ b/pkg/local_object_storage/blobstor/put_test.go
@@ -1,0 +1,43 @@
+package blobstor_test
+
+import (
+	"testing"
+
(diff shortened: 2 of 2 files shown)


pkg/local_object_storage/blobstor/put.go
```diff
@@ -42,6 +44,10 @@ func (b *BlobStor) Put(prm common.PutPrm) (common.PutRes, error) {
 			res, err := b.storage[i].Storage.Put(prm)
 			if err == nil {
 				logOp(b.log, putOp, prm.Address, b.storage[i].Storage.Type(), res.StorageID)
+			} else if errors.Is(err, common.ErrNoSpace) {
```
roman-khimov, line 47: Do we want to get `ErrNoPlaceFound` if all of them fail this way or `ErrNoSpace`?

cthulhu-rider: ofc `ErrNoSpace` is clearer, i was confused by `ErrNoPlaceFound` name (look like no space)

done

pkg/local_object_storage/blobstor/put.go
```diff
@@ -44,7 +44,7 @@ func (b *BlobStor) Put(prm common.PutPrm) (common.PutRes, error) {
 			res, err := b.storage[i].Storage.Put(prm)
 			if err == nil {
 				logOp(b.log, putOp, prm.Address, b.storage[i].Storage.Type(), res.StorageID)
-			} else if errors.Is(err, common.ErrNoSpace) {
+			} else if errors.Is(err, common.ErrNoSpace) && i < len(b.storage)-1 {
```
roman-khimov, line 47: I don't think you can do it this way. The next one can have a policy that doesn't match.

cthulhu-rider: nice catch. In theory, policies may be any with any order, so we should be ready

roman-khimov (approved):

carpawell (commented):
Is it a good solution? How temporary is it? Without any (backward) migration, it seems strange to me. Don't we want to just [simplify](https://github.com/nspcc-dev/neofs-node/issues/2453) blobovniczas and then see if #2382 is done?

cthulhu-rider:
> Without any (backward) migration

u can create an issue for this, imo such routine would complicate things rather than optimize smth (data removal isn't so frequent like write load spikes are)

 > Don't we want to just https://github.com/nspcc-dev/neofs-node/issues/2453 blobovniczas

simplification doesn't prevent no space cases, so these changes are independent

carpawell (approved):
> simplification doesn't prevent no space cases, so these changes are independent

But we are not sure that it is a "no space left" bug. Simplification would solve that misunderstanding.

> imo such routine would complicate things rather than optimize smth

Without any error logs it may store dozen of gigabytes on the wrong (nonoptimal) storage.

cthulhu-rider:
> But we are not sure that it is a "no space left" bug. Simplification would solve that misunderstanding.

this PR fixes particular problem: when blobovnicza says "i'm full" we must not deny the request and must fallback to any free sub-storage

> Without any error logs it may store dozen of gigabytes on the wrong (nonoptimal) storage.

in practice, when blobovnicza is full, FS tree is unlikely to save us if it is on the same drive. In theory, blobstor doesn't operate with "optimality", it has list of sub-storages which can or can not accept the object. Right now, when blobovnicza overflowed, there is no clear event when we should overtake data from FS tree to it. If u can suggest any particular behavioral model, pls fill an issue. Right now service downtime is worse than optimality anyway

PR #2452, opened 2023-07-21
*: Use go.uber.org/zap.Logger everywhere

cthulhu-rider:


cmd/neofs-node/config.go
```diff
@@ -394,7 +395,6 @@ func (s shared) resetCaches() {
 // dynamicConfiguration stores parameters of the
 // components that supports runtime reconfigurations.
 type dynamicConfiguration struct {
```
roman-khimov, line 397: Not needed now?

cmd/neofs-node/config.go
```diff
@@ -989,15 +966,6 @@ func (c *cfg) configWatcher(ctx context.Context) {
 				continue
 			}
 
-			for _, component := range components {
-				err = component.cfg.Reload()
```
roman-khimov, line 993: I'm not sure what components are affected by this change. Are we gonna be able to reconfigure log level the way it was possible before?

cmd/neofs-node/config.go
cthulhu-rider: dropped

cmd/neofs-node/config.go
cthulhu-rider: yep, `configWatcher` still reconfigures logger but in a different way (decoding `zap.AtomicLevel`)

Issue #2453, opened 2023-07-24
Rework blobovnicza tree


roman-khimov:
## Is your feature request related to a problem? Please describe.
#2382, current bbcz tree is needlessly complex and fragile.

## Describe the solution you'd like
One DB per shard is sufficient.

## Describe alternatives you've considered
Keeping some number of them doesn't seem to be helpful.

## Additional context
We need migration.

carpawell:
> One DB per shard is sufficient.

How about one layer of DBs? Can it be more optimal for storing small objects? No open DBs caches and no complex logic at all but just one layer.

roman-khimov:
BTW, speaking of names, blobovnicza is not really appropriate here as these are not BLOBs. They're more of a BSOB or BMOB.

cthulhu-rider:
> speaking of names

blobovnicza should be definitely replaced. I thought about new one a bit: "bolt with tiny objects" - boltin/bolbin.

roman-khimov:
`boltiny`: :+1: 
`peapod`: :tada:
`pisum`: :rocket:

carpawell:
IMO, naming should not depend on a certain DB name.

cthulhu-rider:
`pisum` is ok to me, but ceteris paribus i'd select other one

roman-khimov:
Fixed in #2462, but not closed.

cthulhu-rider:
> Fixed in #2462, but not closed.

do we need anything within this issue?

roman-khimov:
It's done to me. We will drop bbcz eventually and we will optimize peapod additionally, we may simplify configurations, but all of those are different things.

PR #2454, opened 2023-07-25
blobovnicza: Fix write error at zero depth

cthulhu-rider:
well, we're willing to get rid of complex blobovnicza tree. But this fix is simple, and it's needed to benchmark flat list of DBs within #2453

### DIFF
--- a/pkg/local_object_storage/blobstor/blobovniczatree/put_test.go
+++ b/pkg/local_object_storage/blobstor/blobovniczatree/put_test.go
@@ -1,0 +1,39 @@
+package blobovniczatree_test
+
+import (
+	"testing"
+
+	. "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor/blobovniczatree"
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor/common"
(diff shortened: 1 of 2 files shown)


roman-khimov (approved):

pkg/local_object_storage/blobstor/blobovniczatree/put_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"testing"
+
+	. "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor/blobovniczatree"
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor/common"
+	oidtest "github.com/nspcc-dev/neofs-sdk-go/object/id/test"
+	objecttest "github.com/nspcc-dev/neofs-sdk-go/object/test"
+	"github.com/stretchr/testify/require"
+)
+
+func tempDir(tb testing.TB) string {
```
carpawell, line 14: how about `t.TempDir()` instead of that func?

cthulhu-rider: replaced

PR #2459, opened 2023-07-25
fix/Notary request tracking

### LINKED ISSUE #2315: Notary request tracking is incorrect
## Expected Behavior
Notary requests should be tracked by their actions and main transaction hash.

## Current Behavior
Looking at signatures to determine if it's a new notary request to handle is just plain wrong (see notary preparator code in #2310).

## Possible Solution
Rework this mechanism to improve reliability.

carpawell:
Closes #2315

### DIFF
--- a/pkg/morph/event/notary_preparator.go
+++ b/pkg/morph/event/notary_preparator.go
@@ -5,6 +5,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"sync"
 
 	lru "github.com/hashicorp/golang-lru/v2"
 	"github.com/nspcc-dev/neo-go/pkg/core/interop/interopnames"
@@ -43,6 +44,9 @@ var (
 
 	// ErrMainTXExpired is returned if received fallback TX is already valid.
 	ErrMainTXExpired = errors.New("received main tx has expired")
+
+	// ErrUnknownEvent is returned if main transaction is not expected.
+	ErrUnknownEvent = errors.New("received notary request is not allowed")
 )
 
 // BlockCounter must return block count of the network
@@ -76,6 +80,9 @@ type preparator struct {
 	// just an optimization, already handled TX are not gonna
 	// be signed once more anyway
 	alreadyHandledTXs *lru.Cache[util.Uint256, struct{}]
+
+	m             *sync.RWMutex
+	allowedEvents map[notaryScriptWithHash]struct{}
 }
 
 // notaryPreparator inits and returns preparator.
@@ -104,6 +111,8 @@ func notaryPreparator(prm PreparatorPrm) preparator {
 		alphaKeys:             prm.AlphaKeys,
 		blockCounter:          prm.BlockCounter,
 		alreadyHandledTXs:     cache,
+		m:                     &sync.RWMutex{},
+		allowedEvents:         make(map[notaryScriptWithHash]struct{}),
 	}
 }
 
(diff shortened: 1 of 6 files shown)


carpawell:
@roman-khimov, TBH, yesterday I figured out that it was not a problem at all. Yes, we parse every notary request and are ready to sign each of them **but** not every method has a handler and a parser for it: e.g. our beloved `NewEpoch` event is listened as a simple notification cause netmap processor has a [parser](https://github.com/nspcc-dev/neofs-node/blob/0168dc5df9dff0a2a253b7dd66a81319f192cad1/pkg/innerring/processors/netmap/processor.go#L169-L172) and a [handler](https://github.com/nspcc-dev/neofs-node/blob/0168dc5df9dff0a2a253b7dd66a81319f192cad1/pkg/innerring/processors/netmap/processor.go#L184-L187) for it. It is ready to update its internal state and be sure that everyone thinks that a `NewEpoch` has happened cause it believes the contract. But if another IR sends notary requests it won't be handled cause IR does not have `NewEpoch` _notary_ [parser](https://github.com/nspcc-dev/neofs-node/blob/0168dc5df9dff0a2a253b7dd66a81319f192cad1/pkg/innerring/processors/netmap/processor.go#L203-L211) and _notary_ [handler](https://github.com/nspcc-dev/neofs-node/blob/0168dc5df9dff0a2a253b7dd66a81319f192cad1/pkg/innerring/processors/netmap/processor.go#L227-L235) for it.

But why have I added the last commit then? I still think that `morph` refactor is somewhere near and we will drop (simplify) `Listener`. Also, I think notary events are missing contract side checks (the `netmap` contract will not allow ticking epoch if there are less than 5 IR requested a new epoch; **but** IR would tick epoch if there are a hander for it and a corresponding notary request has been parsed) so I think it should be a notary preparator (`preparator` now) who must skip unnecessary requests not a missing handler.

roman-khimov (approved):

pkg/morph/event/notary_preparator.go
```diff
@@ -43,6 +44,9 @@ var (
 
 	// ErrMainTXExpired is returned if received fallback TX is already valid.
 	ErrMainTXExpired = errors.New("received main tx has expired")
+
+	// ErrUnknownEvent is returned if main transaction is not expected.
+	ErrUnknownEvent = errors.New("received notary request is not allowed")
```
AnnaShaleva, line 49: Don't we need to add a special case for this event to `func (l *listener) parseAndHandleNotary(nr *result.NotaryRequestEvent) {`? We may probably safely skip logging for these events as they are unexpected and unwanted (currently `could not prepare and validate notary event` warning will be logged).

AnnaShaleva (commented):
Otherwise LGTM.

AnnaShaleva (changes requested):
We've discussed and came to the solution that transctions cache isn't needed and can be removed. The only purpose of this cache is to filter out notary requests that were sent by the alphabet node itself. Instead of this we can add a check for notary request: the second signer of fallback transaction shouldn't be the committee member.

Disadvantage of this approach is that we have to always keep the list of committee members up-to-date. It can be reached with RPC-based cache (but then a lot of `getcommittee` RPC requests are expected) or with cache that is subscribed to the committee updates (we need to add some code to our cache than).

@roman-khimov, what do you think?

roman-khimov:
Do you mean `alreadyHandledTXs`?

> sent by the alphabet node itself
> the second signer of fallback transaction shouldn't be the committee member

So is it a node key that is a problem of is it a whole committee that is a problem? What if SN uses the same key as some IR node?

Hi, neo-project/neo#2763, anyway.

AnnaShaleva:
> Do you mean alreadyHandledTXs?

Yes.

> What if SN uses the same key as some IR node?

But is it possible? I thought that SN and IR are always different antities with different accounts. If not, then the described way isn't workable.

roman-khimov:
Nothing technically prevents SN from using the same key as IR.

pkg/morph/event/notary_preparator.go
AnnaShaleva: @carpawell, could you push the update?

carpawell: sure, that was an error, i thout about extending `switch` but forgot about it

AnnaShaleva (approved):

Issue #2251, opened 2023-02-09
bytes.Clone for slice copying


roman-khimov:
Go 1.20 finally brought to us a killer feature --- bytes.Clone library function that can... copy byte slices! We can finally replace `slice.Copy` (but not a `slice.CopyReverse, ha-ha) with something standard.

PR #2462, opened 2023-07-29
Introduce Peapod

cthulhu-rider:
P.S.: we haven't finally decided name yet, but currently `peapod` wins https://github.com/nspcc-dev/neofs-node/issues/2453#issuecomment-1650344027

TODO:
- [x] `Delete`
- [x] `Exists`
- [x] `Iterate`

i didn't precisely try to optimize introduced implementation cuz tested against existing alternatives only. Current benchmark results against single BoltDB with `Update` and `Batch` methods:
```
BenchmarkPut/update-size=1-thread=1-8                     100          13390404 ns/op           18830 B/op        107 allocs/op
BenchmarkPut/batch-size=1-thread=1-8                       54          21120972 ns/op           25333 B/op        148 allocs/op
BenchmarkPut/smart-size=1-thread=1-8                      100          11687289 ns/op           13838 B/op         83 allocs/op
BenchmarkPut/update-size=1-thread=20-8                      4         268907087 ns/op          404892 B/op       2166 allocs/op
BenchmarkPut/batch-size=1-thread=20-8                      42          26277377 ns/op          147620 B/op        768 allocs/op
BenchmarkPut/smart-size=1-thread=20-8                      70          17733068 ns/op          114877 B/op        649 allocs/op
BenchmarkPut/update-size=1-thread=100-8                     1        1357502126 ns/op         2270376 B/op      11766 allocs/op
BenchmarkPut/batch-size=1-thread=100-8                     39          31524376 ns/op          681342 B/op       4010 allocs/op
BenchmarkPut/smart-size=1-thread=100-8                     55          21419636 ns/op          659725 B/op       3746 allocs/op
BenchmarkPut/update-size=1024-thread=1-8                  100          13624444 ns/op           24353 B/op        152 allocs/op
BenchmarkPut/batch-size=1024-thread=1-8                    38          29367535 ns/op           36609 B/op        233 allocs/op
BenchmarkPut/smart-size=1024-thread=1-8                   100          13456244 ns/op           21111 B/op        135 allocs/op
BenchmarkPut/update-size=1024-thread=20-8                   3         442465860 ns/op          529232 B/op       3637 allocs/op
BenchmarkPut/batch-size=1024-thread=20-8                   34          35019898 ns/op          199009 B/op       1226 allocs/op
BenchmarkPut/smart-size=1024-thread=20-8                   67          22541568 ns/op          189971 B/op       1108 allocs/op
BenchmarkPut/update-size=1024-thread=100-8                  1        1692022295 ns/op         2679576 B/op      18032 allocs/op
BenchmarkPut/batch-size=1024-thread=100-8                  36          33703418 ns/op          840072 B/op       4796 allocs/op
BenchmarkPut/smart-size=1024-thread=100-8                  51          21807732 ns/op          838428 B/op       4623 allocs/op
BenchmarkPut/update-size=102400-thread=1-8                 62          18693342 ns/op          141131 B/op        192 allocs/op
BenchmarkPut/batch-size=102400-thread=1-8                  48          27474011 ns/op          146738 B/op        234 allocs/op
BenchmarkPut/smart-size=102400-thread=1-8                  64          19075746 ns/op          133554 B/op        167 allocs/op
BenchmarkPut/update-size=102400-thread=20-8                 3         351594258 ns/op         2770773 B/op       3706 allocs/op
BenchmarkPut/batch-size=102400-thread=20-8                 33          37157903 ns/op         2482699 B/op       1340 allocs/op
BenchmarkPut/smart-size=102400-thread=20-8                 50          33805834 ns/op         2481669 B/op       1263 allocs/op
BenchmarkPut/update-size=102400-thread=100-8                1        1999296188 ns/op        13860160 B/op      18681 allocs/op
BenchmarkPut/batch-size=102400-thread=100-8                12         116099877 ns/op        11770452 B/op       5031 allocs/op
BenchmarkPut/smart-size=102400-thread=100-8                12          85920773 ns/op        12766625 B/op       5355 allocs/op

```


pkg/local_object_storage/peapod/put.go
```diff
@@ -0,0 +1,60 @@
...
+	// currently single routine writes object into the root bucket at the moment. In
+	// theory, we may make root bucket mutex r-w and have per-container locks, but
+	// number of containers may be huge. To reduce number of per-container mutexes
+	// we can add bucket levels and group containers (e.g. with similar prefix).
+	// This also changes data structure and may affect performance/space usage.
+	//
+	// For simplicity, now we acquire root lock. Described optimization worth future
+	// research.
+
+	// TODO: should pre-check existence? Btw Blobovnicza don't do that
```
roman-khimov, line 44: Doesn't matter for immutable objects.

pkg/local_object_storage/peapod/peapod.go
```diff
@@ -0,0 +1,144 @@
...
+		case <-ticker.C:
+			x.flushCurrentBatch(true)
+		}
+	}
+}
+
+func (x *Peapod) flushCurrentBatch(beginNew bool) {
+	x.currentBatchMtx.Lock()
+
+	err := x.currentBatch.tx.Commit()
```
roman-khimov, line 100: In many cases there is nothing to commit once per 10ms.

pkg/local_object_storage/peapod/peapod.go
```diff
@@ -0,0 +1,144 @@
...
+	defer close(x.chFlushDone)
+
+	ticker := time.NewTicker(10 * time.Millisecond)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-x.chClose:
+			// commit current transaction to prevent bbolt.DB.Close blocking
+			x.flushCurrentBatch(false)
```
roman-khimov, line 89: The next tick should be adjusted by the flush time.

pkg/local_object_storage/peapod/put.go
```diff
@@ -0,0 +1,60 @@
...
+
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+)
+
+// Put saves given data in the underlying database by specified object address.
+// The data can be anything, but in practice a binary NeoFS object is expected.
+// Operation is executed within provided context: if the context is done, Put
+// returns its error (in this case data may be saved).
+func (x *Peapod) Put(ctx context.Context, addr oid.Address, data []byte) error {
+	x.currentBatchMtx.RLock()
```
roman-khimov, line 15: This means that either (with the current code) we have a tick-tock pattern (10ms wating then writing, then waiting/writing, etc) and don't completely utilize the disk, or (if we're to use adaptive ticker) we don't have enough changes to flush in many cases (unlock/lock happens too fast for a meaningful rlocks to happen). This will be smoothed out somewhat by the fact that the next changeset will be committed quickly, but it still may be suboptimal.

pkg/local_object_storage/peapod/get.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Get reads data from the underlying database by the given object address.
+// Returns apistatus.ErrObjectNotFound if object is missing in the Peapod.
+func (x *Peapod) Get(addr oid.Address) ([]byte, error) {
+	var data []byte
+
+	err := x.bolt.View(func(tx *bbolt.Tx) error {
+		bktRoot := tx.Bucket(rootBucket)
+		if bktRoot == nil {
+			return fmt.Errorf("%w: missing root bucket", apistatus.ErrObjectNotFound)
```
smallhive, line 20: We usually use `fmt.Errorf("some text: %w", err)` pattern. In this function exists two different patterns for errors, with `%w` in the end and in the beginning.

Is it an exceptional situation to change this format here?

pkg/local_object_storage/peapod/peapod.go
```diff
@@ -0,0 +1,144 @@
...
+func (x *Peapod) Close() error {
+	close(x.chClose)
+	<-x.chFlushDone
+	return x.bolt.Close()
+}
+
+func (x *Peapod) flushLoop() {
+	defer close(x.chFlushDone)
+
+	ticker := time.NewTicker(10 * time.Millisecond)
```
smallhive, line 82: Do we plan to adjust the value from the external config? I think even no, this is a `magic const` here, it would be better to put it in the `const` section above

pkg/local_object_storage/peapod/put.go
cthulhu-rider: we could try to manage batch size and/or analyze flush performance in runtime and/or analyze the input load rate, but such model is more probabilistic and requires in-depth study. 

over a long period of time, the input load should smooth out and rest against the disk, and the granular flow will rest against the 10ms interval, which is negligible. Note that right now we don't deny operations with "overflow" status.

what we can definitely add right now is the transaction reset skip in the complete absence of changes for the current interval, although this is a micro-optimization and will require additional resource (bool) requiring synchronization between threads.

pkg/local_object_storage/peapod/peapod.go
cthulhu-rider: right now it doesn't make sense to expose this interval as config because we don't have strict model of how it changes the performance (i reached the current value experimentally). Having dedicated const is not bad but it'll be used in single place, i don't mind

pkg/local_object_storage/peapod/get.go
cthulhu-rider: usually reason goes after colon, and here `missing root bucket` is a particular reason of `object not found` (there may be other reasons). Here we could just `return apistatus.ErrObjectNotFound`, but additional context is good.

u may encounter similar cases in the code base

pkg/local_object_storage/peapod/peapod.go
cthulhu-rider: what exactly do u mean? when we `Close` the component it should flush all pending ops and stop, isn't it?

pkg/local_object_storage/peapod/peapod.go
cthulhu-rider: yep, mentioned this in https://github.com/nspcc-dev/neofs-node/pull/2462#discussion_r1278913801

i'll add the check

pkg/local_object_storage/peapod/put.go
```diff
@@ -0,0 +1,60 @@
...
+	if currentBatch.initErr != nil {
+		x.currentBatchMtx.RUnlock()
+		return currentBatch.initErr
+	}
+
+	// bbolt.Bucket.Put MUST NOT be called concurrently. This is not obvious from
+	// the docs, but panic occurs in practice
+	currentBatch.bktRootMtx.Lock()
+
+	bktContainer, err := currentBatch.bktRoot.CreateBucketIfNotExists(keyForContainer(addr.Container()))
```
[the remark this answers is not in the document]
cthulhu-rider: i'll try. This may also affect DB size

pkg/local_object_storage/peapod/peapod.go
roman-khimov: I mean changing the timeout value based on the real flushing delay, like in https://github.com/nspcc-dev/neo-go/blob/3bbb21bf8f388dc291fd8c02a996231b6be9fa20/pkg/core/blockchain.go#L1057-L1069

cthulhu-rider: i didn't get how it's related to `Close`. When component is closed, already collected batch is flushed and subsequent `Put` calls will fail

roman-khimov: Wrong line chosen on my side, should've been the `x.flushCurrentBatch(true)` one.

cthulhu-rider:
added all ops and made `Peapod` Blobstor-compatible

- [ ] support in storage node app
- [ ] implement util to migrate blobovnicza tree into peapod

pkg/local_object_storage/peapod/peapod.go
```diff
@@ -0,0 +1,174 @@
...
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-x.chClose:
+			// commit current transaction to prevent bbolt.DB.Close blocking
+			x.flushCurrentBatch(false)
+			return
+		case <-ticker.C:
```
carpawell, line 96: do we want to finish batches earlier if there are "too many" operations?

pkg/local_object_storage/peapod/write.go
```diff
@@ -0,0 +1,70 @@
+package peapod
```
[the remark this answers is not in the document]
cthulhu-rider: united methods using similar code in a separate source file, imo this is easier to maintain then per-method files

pkg/local_object_storage/peapod/peapod.go
cthulhu-rider: in current model: yes cuz we don't wanna block earlier operations due to load. Like i mentioned here https://github.com/nspcc-dev/neofs-node/pull/2462#discussion_r1278913801 external load and flush performance isn't analyzed for now (benchmarks show perf improvement even with this simple model). 

we'll leave an optimization issue to test peapod against itself

pkg/local_object_storage/peapod/read.go
```diff
@@ -0,0 +1,89 @@
...
+		if bktRoot == nil {
+			return fmt.Errorf("%w: missing root bucket", apistatus.ErrObjectNotFound)
+		}
+
+		val := bktRoot.Get(keyForObject(addr))
+		if val == nil {
+			return apistatus.ErrObjectNotFound
+		}
+
+		data = slice.Copy(val)
```
carpawell, line 29: i have already asked somewhere but did not get the answer: why do we keep importing that non-neo-go util from `neo-go`?

@roman-khimov

pkg/local_object_storage/peapod/read.go
```diff
@@ -0,0 +1,89 @@
...
+
+// Get reads data from the underlying database by the given object address.
+// Returns apistatus.ErrObjectNotFound if object is missing in the Peapod.
+func (x *Peapod) Get(addr oid.Address) ([]byte, error) {
+	var data []byte
+
+	err := x.bolt.View(func(tx *bbolt.Tx) error {
+		bktRoot := tx.Bucket(rootBucket)
+		if bktRoot == nil {
+			return fmt.Errorf("%w: missing root bucket", apistatus.ErrObjectNotFound)
```
carpawell, line 21: looks like false positive `errors.Is(err, apistatus.ErrObjectNotFound)` on the caller side then while missing that bucket is more like a bug according to the current `New`

pkg/local_object_storage/peapod/read.go
roman-khimov: Maybe because it's so nice and solves the problem?

pkg/local_object_storage/blobstor/peapod/peapod.go
```diff
@@ -0,0 +1,207 @@
...
+type storage struct {
+	path string
+	perm fs.FileMode
+
+	ppd *peapod.Peapod
+
+	compress *compression.Config
+}
+
+func New(path string, perm fs.FileMode) common.Storage {
```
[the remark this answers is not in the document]
cthulhu-rider: this package provide `common.Storage` over base `peapod.Peapod` to fit `blobstor` needs and not overdoze peapod interface

pkg/local_object_storage/peapod/read.go
cthulhu-rider: in next go upgrades we'll use `bytes.Clone`

pkg/local_object_storage/peapod/write_test.go
```diff
@@ -0,0 +1,88 @@
...
+		var wg sync.WaitGroup
+
+		for i := 0; i < nThreads; i++ {
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+
+				addr := oidtest.Address()
+
+				err := ppd.Put(ctx, addr, data)
```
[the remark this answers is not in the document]
cthulhu-rider: i'll try to test with some multi-threading code snippet, benchmarks are not for this

pkg/local_object_storage/blobstor/peapod/peapod.go
```diff
@@ -0,0 +1,207 @@
...
+	return err
+}
+
+func (x *storage) Init() error {
+	// no-op because peapod.New initializes everything
+	return nil
+}
+
+func (x *storage) Close() error {
+	var err error
```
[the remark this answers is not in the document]
cthulhu-rider: done

pkg/local_object_storage/blobstor/peapod/peapod.go
```diff
@@ -0,0 +1,202 @@
...
+	ctx, cancel := context.WithTimeout(context.TODO(), 5*time.Second)
+	defer cancel()
+
+	err := x.ppd.Put(ctx, prm.Address, prm.RawData)
+	if err != nil && errors.Is(err, bbolt.ErrDatabaseReadOnly) {
+		return common.PutRes{}, common.ErrReadOnly
+	}
+
+	return common.PutRes{
+		StorageID: storageID,
```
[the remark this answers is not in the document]
cthulhu-rider: u mean drop storage IDs from `common` pkg as we discussed? if so, then nope, not in this PR

pkg/local_object_storage/peapod/write.go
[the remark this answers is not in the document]
carpawell: as always, i am for unification, so i would like it to be the same for every `common.Storage` implementation

talking about naming: not that critical but i still think that method per file is preferable. files' number is not that big but more details are always welcome (once again, it there are not _a lot of details_, num of methods not gonna exceed 20, i think)

pkg/local_object_storage/blobstor/peapod/peapod.go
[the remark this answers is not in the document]
carpawell: why not copy `FSTree` approach? one simple (relatively) package in the `blobstor` dir. it is not gonna be used outside of blobstor almost always

pkg/local_object_storage/blobstor/peapod/peapod.go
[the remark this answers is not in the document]
carpawell: > not in this PR

expected to make it there but ok. do we have an issue for that?

pkg/local_object_storage/peapod/write.go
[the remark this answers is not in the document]
cthulhu-rider: then i'll need to add more files to place shared reusable methods like `batch` into and we'll come to file:func 1:1, i'd prefer to group related code

pkg/local_object_storage/peapod/read.go
carpawell: > but the fact remains - the object was not found

sure, but potential problems scare me: we even are impossible to find a potential bug: read operations are OK, write operations are redirected to another storage

pkg/local_object_storage/peapod/write_test.go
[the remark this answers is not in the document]
carpawell: ok, i would like to see the results. as i understand reading _may_ suffer with this approach: we are blocking too frequently on write operations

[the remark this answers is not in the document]
cthulhu-rider: 100/100 read/write routines, duration = `1m`, data size = `32B`

current peapod:
```
OPS READ: 310_786
OPS WRITE: 6_815
```

direct bolt `Update` method:
```
RPS READ: 387_357
RPS WRITE: 172
```

direct `Batch` method:
```
RPS READ: 291_323
RPS WRITE: 4_118
```

pkg/local_object_storage/peapod/read.go
cthulhu-rider: if bucket is missing for any reason, both read and writes ops will fail, so i didn't got ur concern point. Error message contains detailed reason

pkg/local_object_storage/peapod/read.go
cthulhu-rider: so iiuc we either duplicate this in `neofs-node` or write 2 lines in-place or ..?

pkg/local_object_storage/blobstor/peapod/peapod.go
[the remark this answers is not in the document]
cthulhu-rider: * #2480

pkg/local_object_storage/peapod/peapod.go
cthulhu-rider: got it, and fixed

pkg/local_object_storage/peapod/read.go
roman-khimov: It's already written, can be used like that. It'd also be easier to identify and fix instances of it in #2251 if we don't have 100500 variations with local copies and open-coded versions.

pkg/local_object_storage/peapod/read.go
roman-khimov: Likely you can panic here right away. Or declare this shard as dead one. Root bucket must be initialized along with the DB, if it's missing we're in big trouble.

pkg/local_object_storage/peapod/write.go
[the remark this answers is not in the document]
roman-khimov: Files should group some sets of code logically, one file per function doesn't help much, you get too many files quickly and then duplicate imports in all of them, don't remember which of them has some constant defined, switch between them in editor. Likely this is not the biggest problem we have.

pkg/local_object_storage/peapod/read.go
cthulhu-rider: yeah, root bucket is vital one, my mind is still on root-cnr-obj reality. So both of u right. I guess panic isn't an option since this ain't a programmer mistake, so i'll return some error which will be interpreted as common non-status runtime error

pkg/local_object_storage/peapod/read.go
cthulhu-rider: i really don't understand what exactly we discuss here, `neo-go` is our base lib along with std one, so, if no one minds, i'll leave this instruction as it is

pkg/local_object_storage/peapod/write.go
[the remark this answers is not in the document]
cthulhu-rider: then lets leave introduced file structure as it is (it's shorter and unites similar code at least), we'll always be able to change things later in off season (i know we dont have one :smiley_cat:). This wasnt a proposal "lets do like this everywhere from now", just new independent autonomous package

pkg/local_object_storage/blobstor/peapod/peapod.go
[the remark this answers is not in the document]
carpawell: so what is your point? i am against creating one more wrapper and not sure i understand what it solves

pkg/local_object_storage/blobstor/common/storage.go
```diff
@@ -23,3 +27,53 @@ type Storage interface {
 	Delete(DeletePrm) (DeleteRes, error)
 	Iterate(IteratePrm) (IterateRes, error)
 }
+
+// Copy copies all objects from source Storage into the destination one. If any
+// object cannot be stored, Copy immediately fails.
+func Copy(from, to Storage) error {
```
carpawell, line 33: how about `Copy(dst, src)` param order? (like the standard one)

pkg/local_object_storage/blobstor/common/storage.go
```diff
@@ -23,3 +27,53 @@ type Storage interface {
...
+	defer func() { _ = from.Close() }()
+
+	err = to.Open(false)
+	if err != nil {
+		return fmt.Errorf("open destination sub-storage: %w", err)
+	}
+
+	defer func() { _ = to.Close() }()
+
+	err = to.Init()
```
carpawell, line 48: why don't you init `from`?

carpawell (commented):
Conflicts.

pkg/local_object_storage/peapod/write.go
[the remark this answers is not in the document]
carpawell: > one file per function doesn't help much

i wasn't talking about func per file. i was more about a single approach: we had 7 "storage for object data" packages that did `Put` in a file named `put.go` and now we are going to have one (and the only one) pkg that does `Put` in a file named `write.go`. also, i am ok with both approaches but we have a little number of "API" (not so many operations with an object can be done) calls and not going to "get too many files quickly"

pkg/local_object_storage/blobstor/common/storage.go
roman-khimov, line 33: Don't we have something similar in evacuation logic?

pkg/local_object_storage/blobstor/peapod/peapod.go
```diff
@@ -178,6 +178,11 @@ func (x *storage) Delete(prm common.DeletePrm) (common.DeleteRes, error) {
 func (x *storage) Iterate(prm common.IteratePrm) (common.IterateRes, error) {
 	var e error
 	err := x.ppd.Iterate(func(addr oid.Address, data []byte) bool {
+		data, e = x.compress.Decompress(data)
```
roman-khimov, line 181: Doesn't seem to be related to migration.

cmd/neofs-node/config.go
```diff
@@ -705,16 +708,54 @@ func (c *cfg) shardOpts() []shardOptsWithID {
...
+					blobovniczatree.WithBlobovniczaSize(sRead.size),
+					blobovniczatree.WithBlobovniczaShallowDepth(sRead.depth),
+					blobovniczatree.WithBlobovniczaShallowWidth(sRead.width),
+					blobovniczatree.WithOpenedCacheSize(sRead.openedCacheSize),
+					blobovniczatree.WithLogger(c.log))
+				bbcz.SetCompressor(&compressCfg)
+
+				ppd.SetCompressor(&compressCfg)
+
+				c.log.Info("configured Blobovnicza tree is deprecated, overtaking data into Peapod...",
```
roman-khimov, line 749: So we're going to have the same configuration with `blobovnicza`, but using peapod? It seems a bit too magic to me. How about keeping bbcz as is, but creating a tool
```
blobovnicza_to_peapod [config]
```
that would:
 * do all of this logic inside
 * rewrite config to have proper data for peapod (drop bbcz, add ppd)

This would allow to:
 * work with bbcz as is for a release, nothing will break for sure
 * migrate config along with data
 * check that everything is OK on one node and then decide on subsequent migration
 * continue using bbcz if migration fails for some reason

Then we could move bbcz code into the tool itself and keep it for some time (with no support in the node).

pkg/local_object_storage/blobstor/peapod/peapod.go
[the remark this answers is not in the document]
roman-khimov: Frankly, I don't get the distinction between `los/blobstor/peapod` and `los/peapod` as well, I doubt having two packages helps anyone.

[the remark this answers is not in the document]
cthulhu-rider: it helps to keep base `Peapod` as simple as possible apart from oversaturated `common.Storage` interface. Easier to test, easier to reuse. But i dont mind to squash them into one, press :+1: and ill do this meaningful job

pkg/local_object_storage/peapod/read.go
cthulhu-rider: @carpawell so you don't mind using this feature?

pkg/local_object_storage/blobstor/common/storage.go
cthulhu-rider: changes nothing since `Storage` doesn't provide `io` pritmitives

pkg/local_object_storage/blobstor/common/storage.go
cthulhu-rider: usually not needed for read-only ops, but I agree it won't hurt

pkg/local_object_storage/blobstor/peapod/peapod.go
```diff
@@ -0,0 +1,200 @@
...
+
+func (x *storage) Open(readOnly bool) error {
+	if x.ppd != nil {
+		err := x.ppd.Close()
+		if err != nil {
+			return fmt.Errorf("close Peapod: %w", err)
+		}
+	}
+
+	err := util.MkdirAllX(filepath.Dir(x.path), x.perm)
```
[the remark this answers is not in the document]
cthulhu-rider: to place db into the configured path obviously. It's done in, for example, in FSTree as well https://github.com/nspcc-dev/neofs-node/blob/21e9d2acf2f23fd9132b4d3994000d4efca16c9a/pkg/local_object_storage/blobstor/fstree/control.go#L15
so that's a separate issue. I dont see any problem with mkdir, it should be done anyway, right now - easier for admin

pkg/local_object_storage/peapod/peapod.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+var rootBucket = []byte("root")
+
+// New returns initialized Peapod instance located at the given path. Resulting Peapod
+// must be finally closed.
+func New(path string, perm fs.FileMode, readOnly bool) (*Peapod, error) {
+	db, err := bbolt.Open(path, perm, &bbolt.Options{
+		ReadOnly: readOnly,
+		Timeout:  100 * time.Millisecond, // to handle flock
```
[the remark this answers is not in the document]
cthulhu-rider: and why not `2s`? and we use this value in other BoltDB-based components, so this is also a separate issue

pkg/local_object_storage/peapod/peapod.go
```diff
@@ -0,0 +1,184 @@
...
+func (x *Peapod) Close() error {
+	close(x.chClose)
+	<-x.chFlushDone
+	return x.bolt.Close()
+}
+
+func (x *Peapod) flushLoop() {
+	defer close(x.chFlushDone)
+
+	const flushInterval = 10 * time.Millisecond
```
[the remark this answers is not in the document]
cthulhu-rider: just for test or public prm?

pkg/local_object_storage/peapod/peapod.go
```diff
@@ -0,0 +1,184 @@
...
+		}
+	}
+}
+
+func (x *Peapod) flushCurrentBatch(beginNew bool) {
+	x.currentBatchMtx.Lock()
+
+	if !x.currentBatch.nonIdle {
+		if !beginNew && x.currentBatch.tx != nil {
+			_ = x.currentBatch.tx.Rollback()
```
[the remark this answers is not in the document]
cthulhu-rider: u suggest `Commit`?

pkg/local_object_storage/peapod/write_test.go
```diff
@@ -0,0 +1,88 @@
...
+		{100 << 10, 1},
+		{100 << 10, 20},
+		{100 << 10, 100},
+	} {
+		b.Run(fmt.Sprintf("size=%d,thread=%d", tc.objSize, tc.nThreads), func(b *testing.B) {
+			benchmark(b, ppd, tc.objSize, tc.nThreads)
+		})
+	}
+}
+
```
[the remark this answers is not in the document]
cthulhu-rider: it's tested in `Delete`, but ok lets add more

pkg/local_object_storage/blobstor/common/storage.go
cthulhu-rider: evacuation overtakes data from one shard to another not dealing with `common.Storage`

cmd/neofs-node/config.go
cthulhu-rider: nah, to complex to me. Imo we should migrate bolder. As an altenative I suggest:
1. fallback to bbcz tree automatically here if `Copy` failed
2. leave blobovnicza config flag `no_peapod` to force the use of bbcz (for example, we migrated successfully, but the data is not available from peapod)

so, if migration fails, we see the reason and node works with bbcz. If migration succeeds, node works with peapod. And if we later see that some data disappeared after migration, we set config flag manually and work with bccz until peapod is fixed.  

thoughts?

pkg/local_object_storage/blobstor/peapod/peapod.go
cthulhu-rider: fact, `Iterate` is just a method. What exactly do you mean by that?

pkg/local_object_storage/peapod/peapod.go
```diff
@@ -0,0 +1,184 @@
...
+func (x *Peapod) beginNewBatch() {
+	x.currentBatch = new(batch)
+
+	x.currentBatch.tx, x.currentBatch.initErr = x.bolt.Begin(true)
+	if x.currentBatch.initErr != nil {
+		x.currentBatch.initErr = fmt.Errorf("begin new BoltDB writable transaction: %w", x.currentBatch.initErr)
+		return
+	}
+
+	x.currentBatch.bktRoot, x.currentBatch.initErr = x.currentBatch.tx.CreateBucketIfNotExists(rootBucket)
```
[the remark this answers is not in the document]
cthulhu-rider: done

pkg/local_object_storage/peapod/peapod.go
```diff
@@ -0,0 +1,184 @@
...
+
+	if !readOnly {
+		err = db.Update(func(tx *bbolt.Tx) error {
+			_, err = tx.CreateBucketIfNotExists(rootBucket)
+			return err
+		})
+		if err != nil {
+			return nil, fmt.Errorf("create root bucket in BoltDB instance: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
cthulhu-rider: done

pkg/local_object_storage/peapod/read.go
carpawell: it is not a problem for me, i just do not get it. it won't stop me from pressing "Approve"

cthulhu-rider: > it won't stop me from pressing "Approve"

i mean when `bytes.Clone` (or even `slices.Clone`) will become available with newer Go version - u'll be still against using those utilities?

pkg/local_object_storage/peapod/peapod.go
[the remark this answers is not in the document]
cthulhu-rider: @carpawell where is the link between `Timeout` and transactions? this is a timeout for flock

pkg/local_object_storage/blobstor/common/storage.go
carpawell: not critical. i was just reading its usage and thought that it is a mistake cause i get used to `copy(dst, src)` but it is not a mistake so i decided to just describe my short confusion here

pkg/local_object_storage/peapod/read.go
roman-khimov: It's not that you need `neo-go`, it's that you already have it. Introducing it as a dependency just for this function would be inappropriate, but we're already dependent on it for different reasons, so reusing some functions is just a natural thing to do. And yes, it'll all be changed with newer Go.

pkg/local_object_storage/blobstor/peapod/peapod.go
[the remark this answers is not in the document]
roman-khimov: Well, they're both wrong. The only reason they do this is #2497, no one wants to manage things in `/srv/neofs/data0` (so _that_ is what should be configured, not `/srv/neofs/data0/bbcz`). It's kinda OK for this PR, but not OK long-term.

pkg/local_object_storage/peapod/peapod.go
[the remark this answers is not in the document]
roman-khimov: It's a balance between safety and failure delay. Failures are rare, so the value should be more safe. 100ms delay is easily reachable in many cases (leading to node start failure without any real reason), 1s or 2s less so.

pkg/local_object_storage/peapod/peapod.go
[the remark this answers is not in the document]
roman-khimov: Public one to try some tweaks on a live system.

pkg/local_object_storage/peapod/peapod.go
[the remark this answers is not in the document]
roman-khimov: `Commit` seems to be more appropriate, finish the last batch, release everyone and go.

pkg/local_object_storage/blobstor/peapod/peapod.go
roman-khimov: I mean it's placed in a commit that deals with migration, while it's not related to migration, this is the way `peapod` should operate.

pkg/local_object_storage/blobstor/common/storage.go
roman-khimov: `Copy` is well-known, so semantically following its order of arguments makes things better.

pkg/local_object_storage/peapod/peapod.go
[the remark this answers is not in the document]
cthulhu-rider: r u talking about app config?

pkg/local_object_storage/peapod/peapod.go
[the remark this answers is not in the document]
cthulhu-rider: then i'd spread to all usecases https://github.com/nspcc-dev/neofs-node/issues/2499

pkg/local_object_storage/blobstor/peapod/peapod.go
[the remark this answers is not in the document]
cthulhu-rider: agree with u, lets resolve within that issue then (as well as flock timeouts)

pkg/local_object_storage/peapod/peapod.go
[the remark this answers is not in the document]
carpawell: > where is the link between Timeout and transactions?

oh, that is for flock only. by bad

pkg/local_object_storage/peapod/peapod.go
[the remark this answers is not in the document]
cthulhu-rider: done

pkg/local_object_storage/blobstor/peapod/peapod.go
cthulhu-rider: fixed

cmd/neofs-node/config.go
roman-khimov: It's too much magic for an admin to handle, I think. It should be more predictable:
 * I have a config and I run a new node version -> I expect it to work according to the config
 * I'm doing migration with a tool, it succeeds -> I have a new config and data, all good
 * I'm doing migration with a tool, it fails -> I can use an old config and data, all good
 * I'm doing migration with a tool, it succeeds, but the node fails to start afterwards -> I take the old config and use it

Compare that to:
 * I run a new node version and pray for it to work
 * I have no idea what's going on with my data unless I'm constantly looking at the log
 * I have some config which specifically says 'blobovnicza', but I need to look at the log to see what's really used internally

In general, automigration is cool, but it'd be more possible if we were to have a configuration like in #2497 (with all internal fstree/bbcz relations hidden). While we have some `/dir/bbcz` config and will get `/dir/peapod.db` file (that technically can be in some wrong (not intended to be used) directory.

cthulhu-rider: where to place this migration tool? new `cmd/`?

roman-khimov: Yeah, it'd be a simple new tool.

pkg/local_object_storage/blobstor/peapod/peapod.go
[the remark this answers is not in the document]
cthulhu-rider: done

cmd/blobovnicza-to-peapod/main.go
```diff
@@ -0,0 +1,95 @@
...
+
+		log.Println("data successfully migrated in the current shard, going to the next one...")
+
+		return nil
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	log.Println("data successfully migrated in all shards, you may now re-configure node to work with Peapod")
```
carpawell, line 94: @roman-khimov, @cthulhu-rider, @532910 as i understood the [comment](https://github.com/nspcc-dev/neofs-node/pull/2462#discussion_r1291231421) suggested migrating config file too (remove useless lines, adjust types and so on). so we want it?

roman-khimov: Yeah, that'd be helpful in general and it's OK to do since config is explicitly given as a parameter.

pkg/local_object_storage/blobstor/peapod/peapod.go
```diff
@@ -0,0 +1,484 @@
...
+
+	x.readOnly = readOnly
+
+	return nil
+}
+
+// Init initializes internal structure of the underlying database.
+func (x *Peapod) Init() error {
+	if x.readOnly {
+		// no extra actions needed in read-only mode
```
roman-khimov, line 213: Why don't you just check for the root bucket? It's a kind of a sanity check for the DB in our case.

pkg/local_object_storage/peapod/peapod.go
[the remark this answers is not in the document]
roman-khimov: Yeah, so that we could try it in various environments. I'd bet some 30-50ms could be a better choice.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Changelog for NeoFS Node
 - Stored payload metric per shard (#2023)
 - Histogram metrics for RPC and engine operations (#2351) 
 - SN's version is announced via the attributes automatically but can be overwritten explicitly (#2455)
+- New `blobovnicza-to-peapod` tool providing blobovnicza-to-peapod data migration (#2453)
```
roman-khimov, line 14: I don't see an entry introducing `peapod` itself and warning that this is the last version to support bbcz.

pkg/local_object_storage/blobstor/peapod/peapod.go
cthulhu-rider: checked in `Open`

CHANGELOG.md
cthulhu-rider: added

pkg/local_object_storage/peapod/peapod.go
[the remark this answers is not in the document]
cthulhu-rider: added

cmd/blobovnicza-to-peapod/main.go
cthulhu-rider: added, looks pretty messy, but should help

carpawell (approved):

roman-khimov (approved):

PR #2464, opened 2023-08-01
engine: Do not ignore+detach shards with failed Open/Init ops

cthulhu-rider:
Previously, storage engine ignored configured shards if they failed to open/init (with at least one successful). This behavior is unclear: if engine is configured to work with N shards, then all N shards must work, otherwise, the whole engine is broken and needs to be reconfigured.

Immediately fail engine's  `Open`/`Init` methods on any shard failure. With this, it's redundant to open/init shard concurrently, so they are processed one-by-one now.

In the future, possibly, optional config flag will be added to ignore problem shards, but default behavior will be kept strict.

* in the vast of #2453


pkg/local_object_storage/engine/control.go
```diff
@@ -26,41 +19,9 @@ func (e *StorageEngine) open() error {
 	e.mtx.Lock()
 	defer e.mtx.Unlock()
 
-	var wg sync.WaitGroup
-	var errCh = make(chan shardInitError, len(e.shards))
```
roman-khimov, line 30: This can affect node startup time, parallel init still can be useful.

cthulhu-rider: parallelism becomes not so obviously useful when any failure should break execution. With blobovnicza->peapod expected init time will be pretty short. So, i don't think parallelism will be really visible, at the same time, i don't mind to leave it

:heart:   parallel
:+1: leave synchronous until real problems

roman-khimov: If it's just "open and go" then it's OK to init consecutively, but if there is some meta reads (or DB/dir traversals), it's a different story.

cthulhu-rider: yep, blobovnicza tree initialized slowly, so parallelism was added along with shard ignoring. Now we consider failure of any shard as complete abort, so parallelism would need some sort of context. I think this makes code and logic more complex while we try to keep components (as well as their init) simple

overall, i vote for conseq but don't mind to parallel

roman-khimov (approved):

pkg/local_object_storage/engine/control.go
cthulhu-rider: > pass ctx in every disk/network operation

network - makes sense, disk - i don't think so, `io` / `os` package don't deal with `context.Context`

carpawell (approved):

Issue #1506, opened 2022-06-09
Investigate container size estimation


alexvanin:
Container size estimation allows to roughly estimate container size in bytes. This is useful for basic income distribution and for network statistics in general. 

I think there is an issue with that at least in N3 Testnet (T4). I invoked [container.ListContainerSizes](https://github.com/nspcc-dev/neofs-contract/blob/a92ad8b8f0e7f5265bced249949ea7cbc9fdad75/container/container_contract.go#L555) for couple of previous epochs (6643, 6642) and then invoked [container.GetContainerSize](https://github.com/nspcc-dev/neofs-contract/blob/a92ad8b8f0e7f5265bced249949ea7cbc9fdad75/container/container_contract.go#L541). 

It seems like many containers are actually missing. It is also not quite consistent, some container IDs may disappear in one epoch and appear in another. I didn't find anything interesting in logs using `estimation` keyword.

I guess we can increase log level for estimations or log more details during estimation synchronization. We did something similar for debugging https://github.com/nspcc-dev/neofs-node/issues/1037

carpawell:
Have not found any missing estimations in the test net near the issue creation time.

roman-khimov:
```
$ ./bin/neo-go contract testinvokefunction -r https://rpc3.morph.t5.fs.neo.org:51331 0x70cf38deb1ff7a4f64cb1127eafcd75cd38083d4 listContainerSizes 4042
{
  "state": "HALT",
  "gasconsumed": "640472",
  "script": "AcoPEcAfDBJsaXN0Q29udGFpbmVyU2l6ZXMMFNSDgNNc1/zqJxHLZE96/7HeOM9wQWJ9W1I=",
  "stack": [
    {
      "type": "Array",
      "value": [
        {
          "type": "Buffer",
          "value": "Y25yyg8F4F+ibOI1AqOxWb5ArAsmB1EskLi5x6Pa7E7lMuZzDw=="
        },
        {
          "type": "Buffer",
          "value": "Y25yyg8+/waCGkINMIdzAENKnbqXN13T9/Yvm7sSPwE/AHVMWw=="
        },
        {
          "type": "Buffer",
          "value": "Y25yyg+LyqAoEQPpICrqHAUAFFcqd6fLGqGb3g/4iaNijVUlrw=="
        },
        {
          "type": "Buffer",
[...]

PR #2467, opened 2023-08-02
fix/Container estimations contract call

carpawell:
Relates #1506. Does not fix (of does fix, not sure yet but that is the first time I [can see](https://github.com/nspcc-dev/neofs-node/issues/1506#issuecomment-1418845811) at least one metric in the contract locally).

### DIFF
--- a/pkg/morph/client/container/client.go
+++ b/pkg/morph/client/container/client.go
@@ -59,6 +59,10 @@ func NewFromMorph(cli *client.Client, contract util.Uint160, fee fixedn.Fixed8, 
 		o.staticOpts = append(o.staticOpts, client.WithCustomFee(putNamedMethod, o.feePutNamed))
 	}
 
+	if !o.disableNotarySigning {
+		o.staticOpts = append(o.staticOpts, client.TryNotary())
+	}
+
 	sc, err := client.NewStatic(cli, contract, fee, o.staticOpts...)
 	if err != nil {
 		return nil, fmt.Errorf("can't create container static client: %w", err)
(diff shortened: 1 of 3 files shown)


pkg/morph/client/container/client.go
```diff
@@ -89,9 +89,15 @@ type opts struct {
 }
 
 func defaultOpts() *opts {
-	var o = new(opts)
-	o.staticOpts = append(o.staticOpts, client.TryNotary())
-	return o
+	return new(opts)
```
roman-khimov, line 92: Seems too dangerous to me. Can we have `DisableNotary()` specifically to use in those cases where its' needed?

roman-khimov: And to be fair it's not about enabling/disabling notary, it's more about using appropriate signers. This `TryNotary()` name always looked wrong to me.

pkg/morph/client/container/client.go
```diff
@@ -85,13 +89,24 @@ type opts struct {
...
+}
+
+// DisableNotarySinging returns option to disable
+// notary request signing. With that option, every
+// call will be created, signed with provided key
+// (a single regular sign) and sent to the side
+// chain.
+func DisableNotarySinging() Option {
```
roman-khimov, line 106: Some notary can sing probably, but they more often sign.

pkg/morph/client/container/client.go
roman-khimov: So we still change default?

pkg/morph/client/container/client.go
```diff
@@ -85,13 +89,24 @@ type opts struct {
...
+	return func(o *opts) {
+		o.staticOpts = append(o.staticOpts, client.TryNotary())
```
roman-khimov, line 108: Is it "disable" or "enable"?

carpawell: wow, hasn't committed this line changed

pkg/morph/client/container/client.go
carpawell: i pushed smth strange, check one more time, please

roman-khimov (approved):

PR #2470, opened 2023-08-04
Upgrade NeoFS SDK module to the latest revision

cthulhu-rider:
* reopens #2449 for testing


roman-khimov:
We can close #2449 then, it'd be better to have testing results in the PR anyway.

PR #2484, opened 2023-08-09
feat/Announce version via the attributes

### LINKED ISSUE #2455: Announce node version in SN attributes automatically
I'm always frustrated when I can't tell what node versions are being used on the network. We even have nspcc-dev/neofs-contract#299.

We can have version in an attribute for now. Node can set it automatically, but allow to override via config (much like #602). These attributes should be documented appropriately.

carpawell:
It is done automatically but can be overwritten if `Version` attribute is provided via the application configuration.
Closes #2455.


roman-khimov (approved):

cthulhu-rider (approved):
checked, works

* btw came to https://github.com/nspcc-dev/neofs-dev-env/pull/276

PR #2485, opened 2023-08-09
upd/Bump minimal go version up to 1.19

carpawell:


CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ Changelog for NeoFS Node
 
 ### Updated
 - `neofs-sdk-go` to `v1.0.0-rc.10`
+- Update minimal supported Go version up to v1.11 (#2485)
```
roman-khimov, line 37: What version?

roman-khimov (changes requested):
Let's not do this until 0.38.0 is out.

cthulhu-rider (commented):
lets also replace https://pkg.go.dev/go.uber.org/atomic with https://pkg.go.dev/sync/atomic@go1.19.0

roman-khimov:
Technically, all of nspcc-dev/neo-go#2626, nspcc-dev/neo-go#2627, nspcc-dev/neo-go#2628, nspcc-dev/neo-go#2629 can be relevant here as well.

cthulhu-rider:
lets convert to draft for now

@carpawell pls check all places for improvement carefully

cmd/neofs-node/config.go
```diff
@@ -553,8 +557,8 @@ func initCfg(appCfg *config.Config) *cfg {
 		scriptHash:          contractsconfig.Netmap(appCfg),
 		state:               netState,
 		workerPool:          netmapWorkerPool,
-		needBootstrap:       !relayOnly,
-		reBoostrapTurnedOff: atomic.NewBool(relayOnly),
+		needBootstrap:       !atomicRelayOnly.Load(),
```
cthulhu-rider, line 560: previous code with `relayOnly` looked much better to me than with `atomicRelayOnly.Load()`. U could change 2 strings instead of 6

cmd/neofs-node/config.go
```diff
@@ -553,8 +557,8 @@ func initCfg(appCfg *config.Config) *cfg {
...
+		reBoostrapTurnedOff: atomicRelayOnly,
```
cthulhu-rider, line 561: suggest to drop pointer from this field, afaik it doesn't make any sense right now

cmd/neofs-node/netmap.go
```diff
@@ -37,7 +37,7 @@ func newNetworkState() *networkState {
 	nmStatus.Store(control.NetmapStatus_STATUS_UNDEFINED)
 
 	return &networkState{
-		epoch:            atomic.NewUint64(0),
+		epoch:            new(atomic.Uint64),
```
cthulhu-rider, line 40: pointer is not needed too, we could skip explicit initialization of this field

cmd/neofs-node/config.go
carpawell: hope i got you right (not sure how that could be 2 vs 6) and fixed

cmd/neofs-node/config.go
carpawell: they [ask](https://cs.opensource.google/go/go/+/refs/tags/go1.21.1:src/sync/atomic/type.go;l=139) not to copy it, it wont pass `go vet` linter

cthulhu-rider:
```go
c.internals.healthStatus.Store(int32(control.HealthStatus_HEALTH_STATUS_UNDEFINED))
c.cfgNetmap.reBoostrapTurnedOff.Store(relayOnly)
```
below

carpawell: NPE then? can you elaborate on your thoughts?

carpawell: or you meant to rely on the fact that `cfg` struct is used via the pointer always?

roman-khimov: We usually embed structures, not pointers if we don't copy the structure anyway (always accessing via pointer). And yes, this allows to skip init in many cases (like below).

pkg/local_object_storage/pilorama/boltdb.go
```diff
@@ -845,19 +845,19 @@ func (t *boltForest) logFromBytes(lm *LogMove, data []byte) error {
 func (t *boltForest) logToBytes(lm *LogMove) []byte {
 	w := io.NewBufBinWriter()
 	size := 8 + 8 + lm.Meta.Size() + 1
-	//if lm.HasOld {
+	// if lm.HasOld {
 	//	size += 8 + lm.Old.Meta.Size()
-	//}
+	// }
```
roman-khimov, line 850: Beep

pkg/local_object_storage/pilorama/boltdb.go
```diff
@@ -883,10 +883,10 @@ func stateKey(key []byte, child Node) []byte {
...
 
 func (t *boltForest) putState(b *bbolt.Bucket, key []byte, parent Node, timestamp Timestamp, meta []byte) error {
-	data := make([]byte, len(meta)+8+8)
-	binary.LittleEndian.PutUint64(data, parent)
-	binary.LittleEndian.PutUint64(data[8:], timestamp)
-	copy(data[16:], meta)
+	data := make([]byte, 0, len(meta)+8+8)
+	data = binary.LittleEndian.AppendUint64(data, parent)
+	data = binary.LittleEndian.AppendUint64(data, timestamp)
+	data = append(data, meta...)
```
roman-khimov, line 889: Not sure it's an improvement here, the old one is probably more performant.

pkg/local_object_storage/pilorama/boltdb.go
carpawell: fixed, and fixed IDE settings

pkg/local_object_storage/pilorama/boltdb.go
carpawell: you mean exactly this line with `copy` before? (reverted `copy`)

roman-khimov: I mean whole chunk of course. It was pushing things into a known preallocated slice. Now it tries to `append()` at every step. It doesn't need to reallocate for obvious reasons, but it could've avoided these attempts.

carpawell: ok, reverted

roman-khimov (approved):

PR #2486, opened 2023-08-09
Add parallel test running

vvarg229:
Both sanity tests and a full test run are run in parallel. This will allow to get the result and a link to the report in NeoFS after 30 minutes after creating a PR or a commit in the repo and get the result of all tests in a few hours.
This will be useful for developers - now they will be able to get test results before changes are pushed to the master.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -26,16 +26,22 @@ permissions: write-all
 jobs:
   run_system_tests:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


vvarg229:
For example, here in my fork is the operation of multiple PRs.
There is no long wait for runners on that many jobs at the same time.

image

roman-khimov (changes requested):
Let's try splitting into gw/non-gw parts (roughly equal now, each takes ~1h to complete).

roman-khimov:
Not relevant for now, we're OK as is.

Issue #2488, opened 2023-08-09
Put Storagegroup failed


vvarg229:
neofs-cli storagegroup put failed:
```
COMMAND: neofs-cli --config /home/runner/work/neofs-node/neofs-node/neofs-testcases/wallet_config.yml storagegroup put --rpc-endpoint 's01.neofs.devenv:8080' --wallet '/home/runner/work/neofs-node/neofs-node/neofs-testcases/TemporaryDir/7744cda6-e7da-4f7e-84e0-670fcc23cda7.json' --cid '7mpsMFCYq69qVyYsbvFFDGEbhqbf5axyV8p5BmAxqhcY' --members 'ALPamBLkRXvMV8M94TRxvADtJKceGKLZQ6wyJJz2UNbp' --lifetime 10
RETCODE: 1

STDOUT:
could not collect storage group members: data must be 64-bytes long

STDERR:

Start / End / Elapsed	 18:15:12.153743 / 18:15:12.575976 / 0:00:00.422233
```

See in the report:
https://http.fs.neo.org/HXSaMJXk2g8C14ht8HSi7BBaiYZ1HeWh2xnWPGQCg4H6/424-1691517028/index.html#suites/69b447004f3c682f67955ec900a09cf0/e7df9c93693e0c5/

## Expected Behavior
neofs-cli storagegroup put failed

## Current Behavior
neofs-cli storagegroup put should not fall

## Steps to Reproduce (for bugs)
Run any TestStorageGroup test

## Regression
Yes
PR# that caused this regression:
https://github.com/nspcc-dev/neofs-node/pull/2470

## Your Environment
neo-go 0.101.1
neofs-s3-authmate 0.27.1
neofs-cli d67d5d39
neofs-adm d67d5d39
AWS aws-cli/2.13.5 Python/3.11.4 Linux/5.15.0-1042-azure exe/x86_64.ubuntu.22 prompt/off

cthulhu-rider:
homomorphic checksums of group members are empty or inaccessible (rather the first)

```
$ neofs-cli -c bin/cli.yml storagegroup put --cid 9oVtcNVCHquor6Qz2xMQy2RFGmx4jC9n3jvv32yRFNLw --lifetime 100 --members 4Qu97DCjz1HnwLdzETA2VQoQQ9nwzSGrjqwq5v6nKnfk
[...]

PR #2489, opened 2023-08-10
go.mod: Upgrade to latest github.com/nspcc-dev/neofs-sdk-go

cthulhu-rider:
untagged, but still

i'm in progress with #2488, maybe addtional SDK fixes will be needed

* based on and blocked by https://github.com/nspcc-dev/neofs-sdk-go/pull/508


carpawell:
Lets keep it in sync with https://github.com/nspcc-dev/neofs-node/pull/2547. If it is merged first, then #2547 should `Close #2531`. If #2547 is merged first this PR should `Close #2531` (because of https://github.com/nspcc-dev/neofs-sdk-go/pull/515).

pkg/local_object_storage/engine/delete.go
```diff
@@ -121,7 +121,11 @@ func (e *StorageEngine) delete(prm DeletePrm) (DeleteRes, error) {
 
 func (e *StorageEngine) deleteChildren(addr oid.Address, force bool, splitID *objectSDK.SplitID) {
 	var fs objectSDK.SearchFilters
-	fs.AddSplitIDFilter(objectSDK.MatchStringEqual, splitID)
+	if splitID != nil {
```
roman-khimov, line 124: Is it legitimate to to call `deleteChildren()` with `nil` `splitID`?

cthulhu-rider: i left previous behavior untouched. Overall, in current implementation it's safe cuz objects w/o parent ("small") don't appear in `CID_splitid` bucket so they are unprocessed. But it'd be more safe to not call `deleteChildren` at all in this case

cthulhu-rider:
> Lets keep it in sync with https://github.com/nspcc-dev/neofs-node/pull/2547.

u mean they both fix #2531 independetly? if so, then it's ok to have `Closes` notation in both of them imo

pkg/local_object_storage/engine/delete.go
roman-khimov: I'd make it `splitID objectSDK.SplitID` as well and let the upper layer handle it (not a lot of these layers here).

roman-khimov:
Conflicts

pkg/local_object_storage/engine/delete.go
cthulhu-rider: done

roman-khimov (approved):

carpawell (approved):

PR #2495, opened 2023-08-10
feat/Announce capacity via the attributes

### LINKED ISSUE #602: Capacity attribute in testnet config
Testnet config ignores capacity attribute right now, so everyone has the same zero capacity and it does not affect placement function. However in the main network this is gonna be different. Do we need to provide some capacity value in testnet? How do we motivate user to set realistic available capacity value in config?

/cc @realloc @anatoly-bogatyrev

carpawell:
It is done automatically (as a sum of existing space on every file system available on the configured shards) but can be overwritten if `Capacity` attribute is provided via the application configuration. Closes #602.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,9 @@ minor release, the component will be purged, so be prepared (see `Updating` sect
 - Stored payload metric per container (#2116)
 - Stored payload metric per shard (#2023)
 - Histogram metrics for RPC and engine operations (#2351) 
-- SN's version is announced via the attributes automatically but can be overwritten explicitly (#2455)
 - New storage component for small objects named Peapod (#2453)
 - New `blobovnicza-to-peapod` tool providing blobovnicza-to-peapod data migration (#2453)
+- SN's version and capacity is announced via the attributes automatically but can be overwritten explicitly (#2455, #602)
 
 ### Fixed
 - `neo-go` RPC connection loss handling (#1337)
--- a/cmd/neofs-node/config.go
+++ b/cmd/neofs-node/config.go
@@ -542,15 +542,16 @@ var persistateSideChainLastBlockKey = []byte("side_chain_last_processed_block")
 func initCfg(appCfg *config.Config) *cfg {
 	c := &cfg{}
 
-	// attaching version to the node's attributes; do not
-	// move it anywhere below reading the other attributes
-	// since a user should be able to overwrite it.
-	writeAppVersion(c)
-
 	err := c.readConfig(appCfg)
 	if err != nil {
 		panic(fmt.Errorf("config reading: %w", err))
 	}
+
+	// filling system attributes; do not move it anywhere
+	// below applying the other attributes since a user
+	// should be able to overwrite it.
+	err = writeSystemAttributes(c)
+	fatalOnErr(err)
 
 	key := nodeconfig.Key(appCfg)
 
@@ -986,8 +987,35 @@ func (c *cfg) configWatcher(ctx context.Context) {
 	}
 }
 
-// writeAppVersion writes app version as defined at compilation
-// step to the node's attributes.
-func writeAppVersion(c *cfg) {
+// writeSystemAttributes writes app version as defined at compilation
+// step to the node's attributes and an aggregated disk capacity
+// according to the all space on the all configured disks.
+func writeSystemAttributes(c *cfg) error {
+	// `Version` attribute
+
 	c.cfgNodeInfo.localInfo.SetAttribute("Version", misc.Version)
+
+	// `Capacity` attribute
+
+	var paths []string
+	for _, sh := range c.applicationConfiguration.EngineCfg.shards {
+		for _, storage := range sh.subStorages {
+			path := storage.path
+			paths = append(paths, path)
+
+			err := util.MkdirAllX(path, storage.perm)
+			if err != nil {
+				return fmt.Errorf("can not create (ensure it exists) dir by '%s' path: %w", path, err)
+			}
+		}
+	}
+
+	total, err := totalBytes(paths)
+	if err != nil {
+		return fmt.Errorf("calculating capacity on every shard: %w", err)
+	}
+
+	c.cfgNodeInfo.localInfo.SetCapacity(total / (1 << 30))
+
+	return nil
 }
--- a/cmd/neofs-node/fs_windows.go
+++ b/cmd/neofs-node/fs_windows.go
@@ -1,0 +1,37 @@
+//go:build windows
+
+package main
+
+import (
+	"fmt"
+	"path/filepath"
+
+	"golang.org/x/sys/windows"
+)
+
+func totalBytes(paths []string) (uint64, error) {
+	var total uint64 // bytes
+	diskMap := make(map[string]struct{}, len(paths))
+
+	for _, path := range paths {
+		disk := filepath.VolumeName(path)
+		if _, handled := diskMap[disk]; handled {
+			continue
+		}
+
+		var availB uint64
+		var totalB uint64
+		var freeTotalB uint64
+		var pathP = windows.StringToUTF16Ptr(path)
+
+		err := windows.GetDiskFreeSpaceEx(pathP, &availB, &totalB, &freeTotalB)
+		if err != nil {
+			return 0, fmt.Errorf("get disk stat by '%s' path: %w", path, err)
+		}
+
+		total += totalB
+		diskMap[disk] = struct{}{}
+	}
+
+	return total, nil
+}
--- a/cmd/neofs-node/fs_unix.go
+++ b/cmd/neofs-node/fs_unix.go
@@ -1,0 +1,31 @@
+//go:build !windows
+
+package main
+
+import (
+	"fmt"
+
+	"golang.org/x/sys/unix"
+)
+
+func totalBytes(paths []string) (uint64, error) {
+	var stat unix.Statfs_t
+	var total uint64 // bytes
+	fsIDMap := make(map[unix.Fsid]struct{})
+
+	for _, path := range paths {
+		err := unix.Statfs(path, &stat)
+		if err != nil {
+			return 0, fmt.Errorf("get FS stat by '%s' path: %w", path, err)
+		}
+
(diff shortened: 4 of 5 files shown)


carpawell:
Let's briefly review it and then decide what we are doing with:
1. windows;
2. the fact that announcing space will never be used fully in practice.

cmd/neofs-node/config.go
```diff
@@ -969,8 +971,45 @@ func (c *cfg) configWatcher(ctx context.Context) {
...
+
+	var stat syscall.Statfs_t
+	var total uint64 // bytes
+	fsIDMap := make(map[syscall.Fsid]struct{})
+
+	for _, sh := range c.applicationConfiguration.EngineCfg.shards {
+		for _, storage := range sh.subStorages {
+			path := storage.path
+
+			err := util.MkdirAllX(path, storage.perm)
```
roman-khimov, line 992: I almost hate this function even though I roughly understand why it's here.

cmd/neofs-node/config.go
```diff
@@ -969,8 +971,45 @@ func (c *cfg) configWatcher(ctx context.Context) {
...
+			if _, handled := fsIDMap[stat.Fsid]; handled {
+				continue
+			}
+
+			total += stat.Blocks * uint64(stat.Bsize)
+			fsIDMap[stat.Fsid] = struct{}{}
+		}
+	}
+
+	// according to API description, in GB, not GiB
```
roman-khimov, line 1011: Then it's obviously 1024^3. We do not use `iB` notation and likely won't be.

cmd/neofs-node/config.go
```diff
@@ -532,16 +533,17 @@ var persistateSideChainLastBlockKey = []byte("side_chain_last_processed_block")
...
-	// move it anywhere below reading the other attributes
-	// since a user should be able to overwrite it.
-	writeAppVersion(c)
-
 	err := c.readConfig(appCfg)
 	if err != nil {
 		panic(fmt.Errorf("config reading: %w", err))
 	}
 
+	// filling system attributes; do not move it anywhere
```
roman-khimov, line 541: So we're not moving it, but we're moving it? IIUC it overwrites now.

roman-khimov (commented):
We can win-compatible code as well: https://stackoverflow.com/questions/20108520/get-amount-of-free-disk-space-using-go

cmd/neofs-node/config.go
carpawell: after [the refactor](https://github.com/nspcc-dev/neofs-node/pull/1830) `readConfig` reads config _literally_ (read numbers/strings and store them in a temporary struct and do not apply them to any component). at least that is what i meant in that PR

applying attributes to a struct that are gonna be used later in `initNetmapService` [func](https://github.com/nspcc-dev/neofs-node/blob/21e9d2acf2f23fd9132b4d3994000d4efca16c9a/cmd/neofs-node/netmap.go#L136)

that is why i changed "do not move it anywhere below reading" to "do not move it anywhere below applying" in the comment

cmd/neofs-node/config.go
carpawell: so make `1 << 30` instead of `100*1000*1000`?

cmd/neofs-node/config.go
carpawell: a dir may be missing, and we will create it anyway later (or fail to create it). what exactly bothers you? `MkdirAllX` in general or using it here?

carpawell:
> We can add win-compatible code as well: https://stackoverflow.com/questions/20108520/get-amount-of-free-disk-space-using-go

Sure we can, but do we need it? We have not announced `Windows` support yet.

carpawell:
@roman-khimov, also, please, see the [comment](https://github.com/nspcc-dev/neofs-node/pull/2495#issuecomment-1673973963) (in case you missed it).

roman-khimov:
> @roman-khimov, also, please, see the [comment](https://github.com/nspcc-dev/neofs-node/pull/2495#issuecomment-1673973963) (in case you missed it).

That's exactly why Windows was mentioned at all, it's not that hard to add support here. The second question is just not really a problem.

cmd/neofs-node/config.go
roman-khimov: What bothers me is that node must not create these configured directories. But it does and it relies on this behavior, that's why we have to replicate this nonsense here. #2497.

cmd/neofs-node/config.go
roman-khimov: Sure.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Changelog for NeoFS Node
 - Stored payload metric per shard (#2023)
 - Histogram metrics for RPC and engine operations (#2351) 
 - SN's version is announced via the attributes automatically but can be overwritten explicitly (#2455)
+- SN's capacity is announced via the attributes automatically but can be overwritten explicitly (#602)
```
cthulhu-rider, line 14: correct, but i'd unite with previous bulletpoint

cmd/neofs-node/config.go
```diff
@@ -969,8 +971,45 @@ func (c *cfg) configWatcher(ctx context.Context) {
...
+			err = syscall.Statfs(path, &stat)
+			if err != nil {
+				return fmt.Errorf("get FS stat by '%s' path: %w", path, err)
+			}
+
+			if _, handled := fsIDMap[stat.Fsid]; handled {
+				continue
+			}
+
+			total += stat.Blocks * uint64(stat.Bsize)
```
cthulhu-rider, line 1006: definitely not
```suggestion
			total += stat.Bavail * uint64(stat.Bsize)
```
?

carpawell: tbh, i am not sure what [that attr](https://github.com/nspcc-dev/neofs-api/blob/953b4c5a9df1409da2a795b34bf9af66699689c6/netmap/types.proto#L160-L161) is expected to show. so you mean that every restart should change its attr value? i think that is about the total disk capacity to make a prediction about how huge a node is

/cc @roman-khimov

roman-khimov: It's going to be different with every node start. But at the same time you're correct in that if I'm a lazy mainnet admin, I don't care about shards, disks, whatever and just want my single FS on `/` to work then `blocks` will be somewhat inaccurate. Likely it's still good enough.

cthulhu-rider: i didn't think about restart. Attr may be changed anyway (e.g. node changed drives)

just confused by `Total available space ...` in protocol. Node has total space and node has available space. So `Capacity` may be misinterpreted, i'd clarify meaning in protocol first

and i didn't dive deeply in capacity issue, just drew attention. May be in such cases having 2 attrs (total and avail) is better than 1

carpawell: @roman-khimov, do not get it. after all, we need to announce available space (`stat.Bavail`) or the full disk capacity (`stat.Blocks`)?

that line was not a mistake, i printed disk capacity cause i thought that we need to print it

roman-khimov: `stat.Blocks`.

roman-khimov (approved):

cmd/neofs-node/config.go
```diff
@@ -969,8 +971,35 @@ func (c *cfg) configWatcher(ctx context.Context) {
...
+			}
+		}
+	}
+
+	total, err := totalBytes(paths)
+	if err != nil {
+		return fmt.Errorf("calculating capacity on every shard: %w", err)
+	}
+
+	c.cfgNodeInfo.localInfo.SetAttribute("Capacity", strconv.FormatUint(total/(1<<30), 10))
```
cthulhu-rider, line 1002:
```suggestion
	c.cfgNodeInfo.localInfo.SetCapacity(total/(1<<30))
```

cmd/neofs-node/fs_windows.go
```diff
@@ -0,0 +1,37 @@
...
+		var totalB uint64
+		var freeTotalB uint64
+		var pathP = windows.StringToUTF16Ptr(path)
+
+		err := windows.GetDiskFreeSpaceEx(pathP, &availB, &totalB, &freeTotalB)
+		if err != nil {
+			return 0, fmt.Errorf("get disk stat by '%s' path: %w", path, err)
+		}
+
+		disk := filepath.VolumeName(path)
```
cthulhu-rider, line 27: may be checked before syscall, or u wanna ensure that path is "ok"?

carpawell: it makes sense for unix (and is the only way) but senseless for windows, moved above the syscall

cthulhu-rider (approved):

cmd/neofs-node/config.go
```diff
@@ -969,8 +970,35 @@ func (c *cfg) configWatcher(ctx context.Context) {
...
+			}
+		}
+	}
+
+	total, err := totalBytes(paths)
+	if err != nil {
+		return fmt.Errorf("calculating capacity on every shard: %w", err)
+	}
+
+	c.cfgNodeInfo.localInfo.SetCapacity(total / (1 << 30))
```
cthulhu-rider, line 1001: i realize that these are drops and it is hardly possible, but how about setting `1GB` if everything is less?

carpawell: you mean preventing `0` announcement? well, mb not announce at all then?

@roman-khimov

roman-khimov: The it'd be an implicit zero? I don't think I have any preference here. 1GB when we don't have 1GB is not exactly correct. Zero capacity looks weird (is it a _storage_ node?). No capacity announced is also weird when every node has it. Explicit zero (the way it's now) is not the worst thing to have, at least we can see that the node has tried to calculate it and got zero (or it's configured this way explicitly which can also happen).

cthulhu-rider: at the same time 950M is not much far from 1.05G

in total, lets announce 0 for now (nodes will be chosen anyway), will revise in case of problems if any

roman-khimov:
Conflicts.

PR #2491, opened 2023-08-10
Fix operations with storage groups

cthulhu-rider:

PR #2504, opened 2023-08-15
writecache: Fix initialization in read-only mode

cthulhu-rider:
_originally posted in https://github.com/nspcc-dev/neofs-node/pull/2491#discussion_r1294355353_

### DIFF
--- a/pkg/local_object_storage/writecache/writecache.go
+++ b/pkg/local_object_storage/writecache/writecache.go
@@ -141,10 +141,12 @@ func (c *cache) Open(readOnly bool) error {
 	return c.initCounters()
 }
 
-// Init runs necessary services.
+// Init runs necessary services. No-op in read-only mode.
 func (c *cache) Init() error {
-	c.initFlushMarks()
-	c.runFlushLoop()
+	if !c.db.IsReadOnly() {
+		c.initFlushMarks()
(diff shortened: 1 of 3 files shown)


pkg/local_object_storage/writecache/writecache.go
```diff
@@ -141,10 +141,12 @@ func (c *cache) Open(readOnly bool) error {
 	return c.initCounters()
 }
 
-// Init runs necessary services.
+// Init runs necessary services. No-op in read-only mode.
 func (c *cache) Init() error {
-	c.initFlushMarks()
-	c.runFlushLoop()
+	if !c.db.IsReadOnly() {
```
carpawell, line 146: runtime mode change is possible, have you checked that moving to `R/W` does start workers again?

cthulhu-rider: yep, checked: `SetMode` calls `openStore` too, that's why we rely on the DB state here (it would be possible to have a separate field, but I considered it redundant)

if u mean test, then there is no in-box approach to catch init actions. I may add optional callback which will be used in test only, do u think it's worth it?

roman-khimov (approved):

carpawell (approved):

PR #2515, opened 2023-08-16
neofs-adm: init and update-contracts are the only commands for wallets creation

### LINKED ISSUE #2134: neofs-adm: do not create contract wallet in `set-config`
```
dzeta@wpc ~/r/neofs-dev-env (master)> ../neofs-node/bin/neofs-adm  morph set-config -c neofs-adm.yml --alphabet-wallets ./services/ir MaintenanceModeAllowed=true
Contract group wallet is missing, initialize at services/ir/contract.json
...
```
The only command that can create alphabet wallets is `init`.
The only commands that can create contract wallet are `init` and `update-contracts`.

AliceInHunterland:
added checks for commands before wallet creation

Closes #2134.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ minor release, the component will be purged, so be prepared (see `Updating` sect
 - Storage node no longer ignores unhealthy shards on startup (#2464)
 - Processing of status errors returned by API client from NeoFS SDK RC-9 (#2465)
 - `neofs-lens write-cache list` command duplication (#2505)
+- `neofs-adm` works with contract wallet in `init` and `update-contracts` commands only (#2134)
 
 ### Removed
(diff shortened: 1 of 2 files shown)


cmd/neofs-adm/internal/modules/morph/generate.go
```diff
@@ -35,6 +35,9 @@ func generateAlphabetCreds(cmd *cobra.Command, _ []string) error {
 	if err != nil {
 		return err
 	}
+	if cmd.Name() != "init" {
```
roman-khimov, line 38: I don't think this is needed and it's likely to break `generate-alphabet` command.

roman-khimov (changes requested):
Likely we also need to limit `newInitializeContext` to call `openContractWallet` only when this wallet is really neded, that's not the case for `set-config`, for example (check `ContractWallet` uses).

cthulhu-rider (approved):

roman-khimov (approved):

PR #2521, opened 2023-08-21
cli: unify usage of expired-at flag instead of lifetime flag

### LINKED ISSUE #1574: Unify epoch expiration flag for CLI's commands

AliceInHunterland:
The expired-at flag replaces the deprecated lifetime flag. 

Closes #1574.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ minor release, the component will be purged, so be prepared (see `Updating` sect
 - `peapod` command for `neofs-lens` (#2507)
 -  New CLI exit code for awaiting timeout (#2380)
 - Validation of excessive positional arguments to `neofs-cli` commands (#1941)
+- `--lifetime` flag to `bearer create` and `object put` CLI commands  (#1574) 
+- `--expired-at` flag to `session create` and `storagegroup put` CLI commands (#1574)
 
 ### Fixed
 - `neo-go` RPC connection loss handling (#1337)
--- a/cmd/neofs-cli/modules/bearer/create.go
+++ b/cmd/neofs-cli/modules/bearer/create.go
@@ -3,6 +3,7 @@ package bearer
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"time"
@@ -47,27 +48,32 @@ func init() {
 	createCmd.Flags().String(outFlag, "", "File to write token to")
 	createCmd.Flags().Bool(jsonFlag, false, "Output token in JSON")
 	createCmd.Flags().StringP(commonflags.RPC, commonflags.RPCShorthand, commonflags.RPCDefault, commonflags.RPCUsage)
+	createCmd.Flags().Uint64P(commonflags.Lifetime, "l", 0, "Number of epochs for token to stay valid")
 
 	_ = cobra.MarkFlagFilename(createCmd.Flags(), eaclFlag)
 
 	_ = cobra.MarkFlagRequired(createCmd.Flags(), issuedAtFlag)
 	_ = cobra.MarkFlagRequired(createCmd.Flags(), notValidBeforeFlag)
-	_ = cobra.MarkFlagRequired(createCmd.Flags(), commonflags.ExpireAt)
 	_ = cobra.MarkFlagRequired(createCmd.Flags(), ownerFlag)
 	_ = cobra.MarkFlagRequired(createCmd.Flags(), outFlag)
+	createCmd.MarkFlagsMutuallyExclusive(commonflags.ExpireAt, commonflags.Lifetime)
 }
 
 func createToken(cmd *cobra.Command, _ []string) {
 	iat, iatRelative, err := common.ParseEpoch(cmd, issuedAtFlag)
 	common.ExitOnErr(cmd, "can't parse --"+issuedAtFlag+" flag: %w", err)
 
+	lifetime, _ := cmd.Flags().GetUint64(commonflags.Lifetime)
 	exp, expRelative, err := common.ParseEpoch(cmd, commonflags.ExpireAt)
 	common.ExitOnErr(cmd, "can't parse --"+commonflags.ExpireAt+" flag: %w", err)
+	if exp == 0 && lifetime == 0 {
+		common.ExitOnErr(cmd, "", errors.New("expiration epoch or lifetime period is required"))
+	}
 
 	nvb, nvbRelative, err := common.ParseEpoch(cmd, notValidBeforeFlag)
 	common.ExitOnErr(cmd, "can't parse --"+notValidBeforeFlag+" flag: %w", err)
 
-	if iatRelative || expRelative || nvbRelative {
+	if iatRelative || expRelative || nvbRelative || lifetime != 0 {
 		ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
 		defer cancel()
 
@@ -83,6 +89,9 @@ func createToken(cmd *cobra.Command, _ []string) {
 		}
 		if nvbRelative {
 			nvb += currEpoch
+		}
+		if lifetime != 0 {
+			exp = currEpoch + lifetime
 		}
 	}
 	if exp < nvb {
--- a/cmd/neofs-cli/modules/object/put.go
+++ b/cmd/neofs-cli/modules/object/put.go
@@ -53,10 +53,12 @@ func initObjectPutCmd() {
 	flags.Bool("disable-filename", false, "Do not set well-known filename attribute")
 	flags.Bool("disable-timestamp", false, "Do not set well-known timestamp attribute")
 	flags.Uint64VarP(&putExpiredOn, commonflags.ExpireAt, "e", 0, "The last active epoch in the life of the object")
+	flags.Uint64P(commonflags.Lifetime, "l", 0, "Number of epochs for object to stay valid")
 	flags.Bool(noProgressFlag, false, "Do not show progress bar")
 
 	flags.String(notificationFlag, "", "Object notification in the form of *epoch*:*topic*; '-' topic means using default")
 	flags.Bool(binaryFlag, false, "Deserialize object structure from given file.")
+	objectPutCmd.MarkFlagsMutuallyExclusive(commonflags.ExpireAt, commonflags.Lifetime)
 }
 
 func putObject(cmd *cobra.Command, _ []string) {
@@ -100,6 +102,15 @@ func putObject(cmd *cobra.Command, _ []string) {
 	common.ExitOnErr(cmd, "can't parse object attributes: %w", err)
 
 	expiresOn, _ := cmd.Flags().GetUint64(commonflags.ExpireAt)
+	lifetime, _ := cmd.Flags().GetUint64(commonflags.Lifetime)
+	if lifetime > 0 {
+		endpoint, _ := cmd.Flags().GetString(commonflags.RPC)
+		currEpoch, err := internalclient.GetCurrentEpoch(ctx, endpoint)
+		common.ExitOnErr(cmd, "Request current epoch: %w", err)
+
+		expiresOn = currEpoch + lifetime
+	}
+
 	if expiresOn > 0 {
 		var expAttrFound bool
 		expAttrValue := strconv.FormatUint(expiresOn, 10)
--- a/cmd/neofs-cli/modules/object/util.go
+++ b/cmd/neofs-cli/modules/object/util.go
@@ -278,8 +278,11 @@ func OpenSessionViaClient(ctx context.Context, cmd *cobra.Command, dst SessionPr
 	const sessionLifetime = 10 // in NeoFS epochs
 
 	common.PrintVerbose(cmd, "Opening remote session with the node...")
-
-	err := sessionCli.CreateSession(ctx, &tok, cli, *key, sessionLifetime)
+	endpoint, _ := cmd.Flags().GetString(commonflags.RPC)
+	currEpoch, err := internal.GetCurrentEpoch(ctx, endpoint)
+	common.ExitOnErr(cmd, "can't fetch current epoch: %w", err)
+	exp := currEpoch + sessionLifetime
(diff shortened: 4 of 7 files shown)


cmd/neofs-cli/modules/storagegroup/put.go
```diff
@@ -43,10 +43,19 @@ func initSGPutCmd() {
...
 
 	flags.Uint64(commonflags.Lifetime, 0, "Storage group lifetime in epochs")
-	_ = sgPutCmd.MarkFlagRequired(commonflags.Lifetime)
+	flags.Uint64P(commonflags.ExpireAt, "e", 0, "The last active epoch of the storage group")
+	sgPutCmd.MarkFlagsMutuallyExclusive(commonflags.ExpireAt, commonflags.Lifetime)
+	_ = flags.MarkDeprecated("lifetime", "use --expire-at flag instead for setting last epoch of the storage group")
 }
 
 func putSG(cmd *cobra.Command, _ []string) {
+	//with next release of cobra we can use this instead of below
```
carpawell, line 52: what exact release is required?

carpawell (commented):
@cthulhu-rider, @roman-khimov, have we decied to deprecate `lifetime`? I find it usefull if you do not want to look for the current epoch number. I can imagine the future where every cmd has both `expire-at` and `lifetime` (or a better naming option).

cmd/neofs-cli/modules/storagegroup/put.go
AliceInHunterland: which will consist https://github.com/spf13/cobra/blob/fd865a44e3c48afeb6a6dbddadb8a5519173e029/flag_groups.go#L49 (probably 1.7.1)

AliceInHunterland:
> @carpawell Okay, I thought https://github.com/nspcc-dev/neofs-testcases/pull/601 and "Unify" lead to lifetime depreciation. By "Unify epoch expiration flag" u mean creating both options (lifetime and expired-at) for each applicable command?

cthulhu-rider:
from my pov as user, i don't find it convenient to work with an absolute epoch value

for example, current epoch is `18446740073709551615`. It'd be very incovenient to specify absolute expiration epoch

if we imagine user or script that requests `netmap netinfo` first, then calculates final absolute value - it will `+E` exact value anyway, so `lifetime` is again more appropriate

absolute value may be useful when user wanna specify "infinite" value (extremely big or even max). We can also cover this with `lifetime` - `+inf` is `inf`

lifetime value saves user from past epoch setting, while absolute value allows to test fault conditions (e.g. in neofs-testcases)

in total, `lifetime` covers 99+% usecases while `expires-at` is not very usable in real practice

---

i'd support both flags in all commands and make them mutual exclusive: if both set, command fails (i don't like priorities, pls set one or nothing)

roman-khimov:
> expires-at is not very usable in real practice

I don't think so, you may want to add things over time that should expire all at once. Suppose you have some process running for a day producing content to be stored in NeoFS, but that should be deleted eventually. It'd be a bit inconvenient to have it expire piece by piece in different epochs.

> i'd support both flags in all commands and make them mutual exclusive: if both set, command fails (i don't like priorities, pls set one or nothing)

But I absolutely agree with that, it's not hard to have both.

cthulhu-rider:
>  things over time that should expire all at once

yeah, good example for absolute value. Always forget that CLI usage is much wider than i expect

AliceInHunterland:
>> i'd support both flags in all commands and make them mutual exclusive: if both set, command fails (i don't like priorities, pls set one or nothing)

>But I absolutely agree with that, it's not hard to have both.

however, for example, for `bearer create` command we have "All epoch flags can be specified relative to the current epoch with the +n syntax.", so it means adding the `lifetime` flag is redundant to this command. is it okay to add a lifetime flag or maybe we can do the same approach with a relative epoch to expired-at and deprecate lifetime?

cthulhu-rider:
i'd make it simple - for each command with expiration context:
* if command accepts NeoFS RPC endpoint (`-r`), provide both `lifetime` and `expires-at` flags
* otherwise, provide `expires-at` only

P.S. `lifetime` w/o `-r` is obviously unusable

cmd/neofs-cli/modules/bearer/create.go
```diff
@@ -38,35 +39,39 @@ is set to current epoch + n.
...
 	_ = cobra.MarkFlagRequired(createCmd.Flags(), notValidBeforeFlag)
-	_ = cobra.MarkFlagRequired(createCmd.Flags(), commonflags.ExpireAt)
 	_ = cobra.MarkFlagRequired(createCmd.Flags(), ownerFlag)
 	_ = cobra.MarkFlagRequired(createCmd.Flags(), outFlag)
+	createCmd.MarkFlagsMutuallyExclusive(commonflags.ExpireAt, commonflags.Lifetime)
 }
 
 func createToken(cmd *cobra.Command, _ []string) {
+	lifetime, _ := cmd.Flags().GetUint64(commonflags.Lifetime)
+	exp, expRelative, err := common.ParseEpoch(cmd, commonflags.ExpireAt)
```
carpawell, line 64: `err` is not checked

cmd/neofs-cli/modules/object/put.go
```diff
@@ -99,6 +102,19 @@ func putObject(cmd *cobra.Command, _ []string) {
 	common.ExitOnErr(cmd, "can't parse object attributes: %w", err)
 
 	expiresOn, _ := cmd.Flags().GetUint64(commonflags.ExpireAt)
+	lifetime, _ := cmd.Flags().GetUint64(commonflags.Lifetime)
+	if lifetime > 0 {
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
```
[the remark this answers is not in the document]
carpawell: but in fact, it seems useless at all, we can just use `ctx`

cmd/neofs-cli/modules/bearer/create.go
```diff
@@ -38,35 +39,39 @@ is set to current epoch + n.
...
 func init() {
-	createCmd.Flags().StringP(eaclFlag, "e", "", "Path to the extended ACL table")
-	createCmd.Flags().StringP(issuedAtFlag, "i", "", "Epoch to issue token at")
-	createCmd.Flags().StringP(notValidBeforeFlag, "n", "", "Not valid before epoch")
-	createCmd.Flags().StringP(commonflags.ExpireAt, "x", "", "The last active epoch for the token")
-	createCmd.Flags().StringP(ownerFlag, "o", "", "Token owner")
-	createCmd.Flags().String(outFlag, "", "File to write token to")
-	createCmd.Flags().Bool(jsonFlag, false, "Output token in JSON")
-	createCmd.Flags().StringP(commonflags.RPC, commonflags.RPCShorthand, commonflags.RPCDefault, commonflags.RPCUsage)
+	flags := createCmd.Flags()
```
carpawell, line 42: imo, unexpected change for that PR/commit

cmd/neofs-cli/modules/session/create.go
```diff
@@ -65,7 +68,15 @@ func createSession(cmd *cobra.Command, _ []string) {
...
 	}
-
+	exp, _ := cmd.Flags().GetUint64(commonflags.ExpireAt)
+	if exp != 0 {
+		var netInfoPrm internalclient.NetworkInfoPrm
+		netInfoPrm.SetClient(c)
+		ni, err := internalclient.NetworkInfo(ctx, netInfoPrm)
+		common.ExitOnErr(cmd, "can't get current epoch: %w", err)
+		currEpoch := ni.NetworkInfo().CurrentEpoch()
+		lifetime = exp - currEpoch
```
carpawell, line 78: i think i would be better to change `CreateSession` definition cause you do `NetworkInfo` here to do `NetworkInfo` one more time inside `CreateSession` later

or at least check that `exp > currEpoch`

CHANGELOG.md
```diff
@@ -68,6 +68,7 @@ Deprecated `neofs_node_object_epoch` metric for IR and SN (the same for both)
 has been removed. Use `neofs_node_state_epoch` for SN and `neofs_ir_state_epoch`
 for IR instead.
 Deprecated `--no-precheck` flag in `neofs-cli container set-eacl` use `--force` flag instead for skipping validation checks.
+Unified usage of `--expired-at` and `--lifetime` flags for `neofs-cli`.
```
carpawell, line 71: usually this section contans info that "speaks" with a reader, like "we have done (in non-compatible way) **that** so update **that** cause it can be broken now"

so i would either describe which cmds have a new flags now, or move it to `Changed` or `Added` section

cmd/neofs-cli/modules/session/create.go
```diff
@@ -37,16 +37,19 @@ var createCmd = &cobra.Command{
...
-	_ = cobra.MarkFlagRequired(createCmd.Flags(), outFlag)
-	_ = cobra.MarkFlagRequired(createCmd.Flags(), commonflags.RPC)
+	flags := createCmd.Flags()
+	flags.Uint64P(commonflags.Lifetime, "l", defaultLifetime, "Number of epochs for token to stay valid")
+	flags.Uint64P(commonflags.ExpireAt, "e", 0, "The last active epoch for token to stay valid")
+	flags.StringP(commonflags.WalletPath, commonflags.WalletPathShorthand, commonflags.WalletPathDefault, commonflags.WalletPathUsage)
+	flags.StringP(commonflags.Account, commonflags.AccountShorthand, commonflags.AccountDefault, commonflags.AccountUsage)
+	flags.String(outFlag, "", "File to write session token to")
+	flags.Bool(jsonFlag, false, "Output token in JSON")
+	flags.StringP(commonflags.RPC, commonflags.RPCShorthand, commonflags.RPCDefault, commonflags.RPCUsage)
```
roman-khimov, line 47: Unrelated as well.

carpawell:
BTW, @roman-khimov, @vvarg229, @evgeniiz321 some issues for testing it?

cmd/neofs-cli/modules/session/create.go
```diff
@@ -65,7 +67,17 @@ func createSession(cmd *cobra.Command, _ []string) {
...
 	}
-
+	exp, _ := cmd.Flags().GetUint64(commonflags.ExpireAt)
+	if exp != 0 {
+		var netInfoPrm internalclient.NetworkInfoPrm
+		netInfoPrm.SetClient(c)
+		ni, err := internalclient.NetworkInfo(ctx, netInfoPrm)
+		common.ExitOnErr(cmd, "can't get current epoch: %w", err)
+		currEpoch := ni.NetworkInfo().CurrentEpoch()
+		if exp < currEpoch {
```
carpawell, line 77: so we ensure that `lifetime` gonna be negative (underflowed in fact)?

have you considered my primary idea about changing `CreateSession`?  it can look like `CreateSession(ctx context.Context, dst *session.Object, c *client.Client, key ecdsa.PrivateKey, expireAt uint64)` (look at the last arg). it will not have `NetworkInfo` inside then and will look the same as the other code in the CLI: we are reading both flags but `expire-at` is our primary goal. no subtraction code, only additional (for `lifetime` cases)

cmd/neofs-cli/modules/session/create.go
```diff
@@ -129,9 +131,7 @@ func CreateSession(ctx context.Context, dst *session.Object, c *client.Client, k
 	}
 
 	dst.SetID(idSession)
-	dst.SetNbf(cur)
-	dst.SetIat(cur)
```
roman-khimov, line 133: No nbf/iat?

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ minor release, the component will be purged, so be prepared (see `Updating` sect
 - SN's version and capacity is announced via the attributes automatically but can be overwritten explicitly (#2455, #602)
 - `peapod` command for `neofs-lens` (#2507)
 -  New CLI exit code for awaiting timeout (#2380)
+- To `neofs-cli` commands: `bearer create`, `object put` added `--lifetime` flag (#1574) 
+- To `neofs-cli` commands `session create`, `storagegroup put` added `--expired-at` flag (#1574)
```
roman-khimov, line 22: This is an `added` section, so it can be rephrased like
 * `--lifetime` flag to `bearer create` and `object put` CLI commands
 * `--expire-at` flag to `session create` and `storagegroup put` CLI commands

roman-khimov (approved):

roman-khimov:
Requires a rebase now to solve CHANGELOG conflict.

cmd/neofs-cli/modules/object/util.go
```diff
@@ -278,8 +278,13 @@ func OpenSessionViaClient(ctx context.Context, cmd *cobra.Command, dst SessionPr
 	const sessionLifetime = 10 // in NeoFS epochs
 
 	common.PrintVerbose(cmd, "Opening remote session with the node...")
-
-	err := sessionCli.CreateSession(ctx, &tok, cli, *key, sessionLifetime)
+	endpoint, _ := cmd.Flags().GetString(commonflags.RPC)
+	currEpoch, err := internal.GetCurrentEpoch(ctx, endpoint)
+	common.ExitOnErr(cmd, "can't fetch current epoch: %w", err)
+	exp := currEpoch + sessionLifetime
+	tok.SetNbf(currEpoch)
```
carpawell, line 285: well, i see, some my suggestion conflicts with the reallity. i would not set _some parts_ of a token here and _some_ on the `CreateSession` side cause it bug-prone IMO

can we pass `currEpoch` as a param instead? or any another alternative that does not make a caller set `nbf` and `iat` outside?

AliceInHunterland: Well, we can (made it in the last commit), however, actually it seems that at the beginning (https://github.com/nspcc-dev/neofs-node/pull/2521#discussion_r1301985286) when we were passing `lifetime` to the `CreateSession` it was more optimal in numbers of calls of `internal.GetCurrentEpoch`

carpawell (approved):

PR #2523, opened 2023-08-21
feat/Use container iterators

carpawell:


carpawell:
@roman-khimov, @AnnaShaleva, I was not involved in iterators support in contract repo so that PR is how I interpret it. It _does_ work.

Some things that surprised me:
1. [Get](https://github.com/nspcc-dev/neofs-contract/blob/7625cf7e82f420eccf41eb79976c34cb7ed66080/container/container_contract.go#L560-L571) is deprecated and will be removed, right? So not you can `Put` but can not `Get, only `Iterate` cause returned key is trimmed and that is not that easy to get the original key.
2. Now you have to add some `map` logic cause aggregating already existing extimations in a summarizing struct (like was [done before](https://github.com/nspcc-dev/neofs-contract/blob/7625cf7e82f420eccf41eb79976c34cb7ed66080/container/container_contract.go#L585)) is obviously missing in a new methods (but current (our) implementation does not provide more that one estimation per container).

If both points are ok, PR is ready.

roman-khimov:
1. Why do you want to have some "original key"? `Get` is deprecated, that's true.
2. Well, you're retrieving estimations for a container, so what you get is exactly that, estimations for a single container. Is it a problem?

pkg/innerring/settlement.go
```diff
@@ -246,27 +246,12 @@ func (b basicIncomeSettlementDeps) BasicRate() (uint64, error) {
 }
 
 func (b basicIncomeSettlementDeps) Estimations(epoch uint64) ([]*containerClient.Estimations, error) {
-	estimationIDs, err := b.cnrClient.ListLoadEstimationsByEpoch(epoch)
+	estimations, err := b.cnrClient.ListLoadEstimationsByEpoch(epoch)
```
roman-khimov, line 249: `return b.cnrClient.ListLoadEstimationsByEpoch(epoch)`?

carpawell:
@roman-khimov

> Why do you want to have some "original key"?

I do not want but it was possible before and is not possible after deprecation. Just asking, not needed for that PR.

> Well, you're retrieving estimations for a container

As an IR, I do not want it for a contianer, I want them all (like it is done in the PR). So iterating may lead to a duplication if more than one node has announced estimations for the contianer (not prohibited now). That is why map is required (but was not required when `List` + `Get` were used).

pkg/morph/client/client.go
```diff
@@ -204,6 +205,49 @@ func (c *Client) TestInvoke(contract util.Uint160, method string, args ...any) (
...
+	if val.State != HaltState {
+		return nil, wrapNeoFSError(&notHaltStateError{state: val.State, exception: val.FaultException})
+	}
+
+	sid, iter, err := unwrap.SessionIterator(val, nil)
+	if err != nil {
+		return nil, fmt.Errorf("iterator expansion: %w", err)
+	}
+
+	const cacheItems = 100
```
AnnaShaleva, line 214: Use `config.DefaultMaxIteratorResultItems` from NeoGo, server-side constraint is likey to be the same.

pkg/morph/client/container/load.go
```diff
@@ -53,34 +54,105 @@ type EstimationID []byte
...
-	} else if ln := len(prms); ln != 1 {
-		return nil, fmt.Errorf("unexpected stack item count (%s): %d", listSizesMethod, ln)
 	}
 
-	prms, err = client.ArrayFromStackItem(prms[0])
-	if err != nil {
-		return nil, fmt.Errorf("could not get stack item array from stack item (%s): %w", listSizesMethod, err)
-	}
+	resMap := make(map[cid.ID]*Estimations)
+	for i := range kvs {
```
AnnaShaleva, line 64: Can't we use an `unwrap.Map` for it? Is it a map returned from the contract?

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ minor release, the component will be purged, so be prepared (see `Updating` sect
 - SN's version and capacity is announced via the attributes automatically but can be overwritten explicitly (#2455, #602)
 - `peapod` command for `neofs-lens` (#2507)
 -  New CLI exit code for awaiting timeout (#2380)
+- Contract-side iterators support; receiving estimations via iterators (#2173)
```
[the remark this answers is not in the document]
carpawell: that was about the first commit: just iterators support. now sure what wording should be written here about, updated, check one more time, please

pkg/morph/client/client.go
```diff
@@ -204,6 +205,49 @@ func (c *Client) TestInvoke(contract util.Uint160, method string, args ...any) (
...
+	if c.inactive {
+		return nil, ErrConnectionLost
+	}
+
+	val, err := c.rpcActor.Call(contract, method, args...)
+	if err != nil {
+		return nil, err
+	}
+
+	if val.State != HaltState {
```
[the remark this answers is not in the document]
carpawell: at the beginning of this PR i thought that i needed to just call one method instead of another so i wanted to copy behavior but now the rules have changed and now it seems useless so dropped

> I don't see a lot of value in notHaltStateError.

hm, btw, there also was `wrapNeoFSError` that used to be usefull but not for now so also dropped in a separate commit

pkg/morph/client/container/load.go
```diff
@@ -53,34 +54,105 @@ type EstimationID []byte
 
 // ListLoadEstimationsByEpoch returns a list of container load estimations for to the specified epoch.
 // The list is composed through Container contract call.
-func (c *Client) ListLoadEstimationsByEpoch(epoch uint64) ([]EstimationID, error) {
+func (c *Client) ListLoadEstimationsByEpoch(epoch uint64) ([]*Estimations, error) {
```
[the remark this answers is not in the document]
carpawell: > I'm not sure why a slice is returned here

well, the same: wanted to mimic the previous algorithm but with the current changes it is useless, dropped

pkg/morph/client/container/load.go
```diff
@@ -53,34 +54,105 @@ type EstimationID []byte
...
 		if err != nil {
-			return nil, fmt.Errorf("could not get ID byte array from stack item (%s): %w", listSizesMethod, err)
+			return nil, fmt.Errorf("storage value handling: %w", err)
 		}
 
-		res = append(res, id)
+		if ee, found := resMap[cnr]; found {
+			ee.Values = append(ee.Values, e)
+		} else {
+			resMap[cnr] = &Estimations{
```
AnnaShaleva, line 87: And BTW, it would be nice to implement `stackitem.Convertible` interface for `Estimations` (and leave the `ToStackItem` panicing). Then you can place all this serialization-related code to `FromStackItem` method and make the main method cleaner. But it's just a suggestion.

I think we can also add `Convertible` support to the `unwrap` package.

AnnaShaleva: Ref. https://github.com/nspcc-dev/neo-go/issues/3100

pkg/morph/client/client.go
carpawell: oh, thanks, i forgot to ask your oppinions, so @roman-khimov, @AnnaShaleva, what value do you think is a good one for fetching containers? 100k containers does not look strange and it would be 1k RPCs for interators as i understand. i took `100` exactly because of its defaultness in `neo-go` repo but wanted to change it after review

roman-khimov: You'll be limited by the server most of the time, setting 1K won't help a lot. This can be a setting though, if some network has a lot of containers and has some higher settings server-side then you could use a higher setting client-side as well.

CHANGELOG.md
```diff
@@ -17,7 +17,8 @@ minor release, the component will be purged, so be prepared (see `Updating` sect
 - New `blobovnicza-to-peapod` tool providing blobovnicza-to-peapod data migration (#2453)
 - SN's version and capacity is announced via the attributes automatically but can be overwritten explicitly (#2455, #602)
 - `peapod` command for `neofs-lens` (#2507)
--  New CLI exit code for awaiting timeout (#2380)
+- New CLI exit code for awaiting timeout (#2380)
```
roman-khimov, line 20: Meh.

pkg/morph/client/client.go
```diff
@@ -220,10 +220,6 @@ func (c *Client) TestInvokeIterator(contract util.Uint160, method string, args .
 		return nil, err
 	}
 
-	if val.State != HaltState {
-		return nil, wrapNeoFSError(&notHaltStateError{state: val.State, exception: val.FaultException})
-	}
-
 	sid, iter, err := unwrap.SessionIterator(val, nil)
```
roman-khimov, line 209: And then this should just be `unwrap.SessionIterator(c.rpcActorCall(...))`, that's why it takes an error for input.

pkg/morph/client/container/load.go
roman-khimov: Nope, it's not.

roman-khimov (changes requested):

CHANGELOG.md
carpawell: ok, i have one more PR to take care of it

pkg/morph/client/container/load.go
carpawell: hope, i got you right, check, please. but it not gonna look perfecr because of the iteration values and `Estimation`+`Estimations` pair (there can be more that one estimation per an epoch theoretically)

pkg/morph/client/client.go
carpawell: dont mind using `neo-go`'s default. adding another magic setting isn not worth it yet, IMO

roman-khimov (approved):

pkg/morph/client/client.go
```diff
@@ -204,6 +206,38 @@ func (c *Client) TestInvoke(contract util.Uint160, method string, args ...any) (
...
+	items := make([]stackitem.Item, 0)
+	for {
+		ii, err := c.rpcActor.TraverseIterator(sid, &iter, config.DefaultMaxIteratorResultItems)
+		if err != nil {
+			return nil, fmt.Errorf("iterator traversal; session: %s, error: %w", sid, err)
+		}
+
+		if len(ii) == 0 {
+			break
+		}
```
AnnaShaleva, line 219: How about terminating session?

AnnaShaleva (commented):
LGTM otherwise.

pkg/morph/client/client.go
carpawell: it feels like i have asked about it when sessions were about their release and somebody answers me that it is not critical. but added, no problem

AnnaShaleva: > it is not critical

That's true, but if we can release a bit of server resources, then why not perform it.

pkg/morph/client/client.go
```diff
@@ -235,6 +235,11 @@ func (c *Client) TestInvokeIterator(contract util.Uint160, method string, args .
 		items = append(items, ii...)
 	}
 
+	err = c.rpcActor.TerminateSession(sid)
+	if err != nil {
+		return nil, fmt.Errorf("session (%s) termination: %w", sid, err)
+	}
+
```
AnnaShaleva, line 228: And one more adjustment: let's skip error check here. It's not important for the client, and current version of NeoGo will return an error if session was already terminated.

carpawell: sure. via `defer` then

AnnaShaleva (approved):

PR #2525, opened 2023-08-24
neofs-cli: add check commands positional arguments

### LINKED ISSUE #1941: Check CLI args for commands that do not need them
### Context

Technically correct usage:

```
neofs-cli object put test test test test -r <...> -w <...> ...
```

### Example

A real example of a logically incorrect usage that did not return any error:

```
neofs-cli object hash range 0:11 <...>
```

instead of

```
neofs-cli object hash --range 0:11 <...>
```

### Reason

We [do not check](https://github.com/nspcc-dev/neofs-node/blob/7d456cb4d0c5381020b4bd76bfc9c5ab5683f90a/cmd/neofs-cli/modules/object/put.go#L60) `cobra`'s `args` if we do not use them.

### Suggestion

Check their presence and return an error if they are found but not expected.

AliceInHunterland:
Added Cobra's check of positional arguments to prevent logically incorrect argument usage and passing redundant arguments to the commands.

Closes #1941.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ minor release, the component will be purged, so be prepared (see `Updating` sect
 - SN's version and capacity is announced via the attributes automatically but can be overwritten explicitly (#2455, #602)
 - `peapod` command for `neofs-lens` (#2507)
 -  New CLI exit code for awaiting timeout (#2380)
+- Validation of excessive positional arguments to `neofs-cli` commands (#1941)
 
 ### Fixed
 - `neo-go` RPC connection loss handling (#1337)
--- a/cmd/neofs-cli/modules/control/drop_objects.go
+++ b/cmd/neofs-cli/modules/control/drop_objects.go
@@ -15,7 +15,8 @@ var dropObjectsCmd = &cobra.Command{
 	Use:   "drop-objects",
 	Short: "Drop objects from the node's local storage",
 	Long:  "Drop objects from the node's local storage",
-	Run: func(cmd *cobra.Command, args []string) {
+	Args:  cobra.NoArgs,
+	Run: func(cmd *cobra.Command, _ []string) {
 		ctx, cancel := commonflags.GetCommandContext(cmd)
 		defer cancel()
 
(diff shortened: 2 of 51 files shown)


cthulhu-rider (approved):

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ minor release, the component will be purged, so be prepared (see `Updating` sect
 - SN's version and capacity is announced via the attributes automatically but can be overwritten explicitly (#2455, #602)
 - `peapod` command for `neofs-lens` (#2507)
 -  New CLI exit code for awaiting timeout (#2380)
+- Validation positional arguments to `neofs-cli` commands (#1941)
```
roman-khimov, line 21: `validation of excessive...`

cmd/neofs-cli/modules/control/drop_objects.go
```diff
@@ -15,6 +15,7 @@ var dropObjectsCmd = &cobra.Command{
 	Use:   "drop-objects",
 	Short: "Drop objects from the node's local storage",
 	Long:  "Drop objects from the node's local storage",
+	Args:  cobra.NoArgs,
 	Run: func(cmd *cobra.Command, args []string) {
```
roman-khimov, line 19: Keeping `args` here?

cmd/neofs-cli/modules/netmap/nodeinfo.go
```diff
@@ -16,6 +16,7 @@ var nodeInfoCmd = &cobra.Command{
 	Use:   "nodeinfo",
 	Short: "Get target node info",
 	Long:  `Get target node info`,
+	Args:  cobra.NoArgs,
 	Run: func(cmd *cobra.Command, args []string) {
```
roman-khimov, line 20: And here? I'm not sure these `Run` changes belong to the same commit though (maybe they are, since we're about arguments here).

roman-khimov (commented):
It looks much cleaner with Cobra than what we've done with NeoGo.

cmd/neofs-cli/modules/netmap/nodeinfo.go
carpawell: seems like their removal should be a part of https://github.com/nspcc-dev/neofs-node/commit/b134dbbe02bfb804d3a2c80e07298f0f308dff4b but the linter wasn't that good if we are speaking about func literals...

carpawell (approved):

PR #2530, opened 2023-08-28
feat/Add `object nodes` command

carpawell:
It applies placement policy rules to the current network map for the provided object. Works the same way `container nodes` does.

### DIFF
--- a/cmd/neofs-cli/modules/object/nodes.go
+++ b/cmd/neofs-cli/modules/object/nodes.go
@@ -1,0 +1,78 @@
+package object
+
+import (
+	internalclient "github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/client"
+	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/common"
+	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/commonflags"
+	"github.com/nspcc-dev/neofs-node/pkg/morph/client/netmap"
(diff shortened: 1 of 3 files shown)


cmd/neofs-cli/modules/object/nodes.go
```diff
@@ -0,0 +1,78 @@
...
+
+		placementNodes, err := resmap.NetMap().PlacementVectors(cnrNodes, oID)
+		common.ExitOnErr(cmd, "could not build placement nodes for the given container: %w", err)
+
+		short, _ := cmd.Flags().GetBool(shortFlag)
+
+		for i := range placementNodes {
+			cmd.Printf("Descriptor #%d, REP %d:\n", i+1, policy.ReplicaNumberByIndex(i))
+			for j := range placementNodes[i] {
+				common.PrettyPrintNodeInfo(cmd, placementNodes[i][j], j, "\t", short)
```
roman-khimov, line 64: Do you have output example for it with regular/short flag? I'd expect some mode to just return a list of nodes with zero metadata (like descriptor/rep above).

carpawell: > a list of nodes with zero metadata (like descriptor/rep above)

what do you mean? that is not proved list, it just says which nodes _may_ store the object and what number of copies are _expected_

roman-khimov (approved):
Please resolve conflicts.

PR #2526, opened 2023-08-24
fix/Removed object listing

carpawell:

PR #2535, opened 2023-08-29
fix/Locked but expired objects

carpawell:


carpawell:
It waits https://github.com/nspcc-dev/neofs-node/pull/2526 but some changes will be also added.

roman-khimov (approved):

pkg/local_object_storage/engine/inhume.go
```diff
@@ -179,7 +179,8 @@ func (e *StorageEngine) inhumeAddr(addr oid.Address, prm shard.InhumePrm, checkE
 	return ok, retErr
 }
 
-func (e *StorageEngine) isLocked(addr oid.Address) (bool, error) {
+// IsLocked checks whether an object is locked according to StorageEngine's state.
+func (e *StorageEngine) IsLocked(addr oid.Address) (bool, error) {
```
cthulhu-rider, line 183: previously was called under `execIfNotBlocked` condition only, pls double-check do we need to keep this or not

pkg/local_object_storage/shard/gc.go
```diff
@@ -431,8 +431,17 @@ func (s *Shard) HandleExpiredLocks(lockers []oid.Address) {
...
+		s.log.Warn("expired object filtering",
+			zap.Error(err),
+		)
+
+		return
+	}
+
 	var pInhume meta.InhumePrm
-	pInhume.SetAddresses(lockers...)
+	pInhume.SetAddresses(append(lockers, expired...)...)
```
cthulhu-rider, line 444: lets reflect this in method docs? same for `HandleDeletedLocks`

pkg/local_object_storage/shard/gc.go
```diff
@@ -442,7 +442,7 @@ func (s *Shard) HandleExpiredLocks(lockers []oid.Address) {
 
 	var pInhume meta.InhumePrm
 	pInhume.SetAddresses(append(lockers, expired...)...)
-	pInhume.SetGCMark()
+	pInhume.SetForceGCMark()
```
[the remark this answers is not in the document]
carpawell: > lockers  are unlocked cuz they cannot be locked according to the system protocol

yes, but they can not be removed regularly. that was the last [decision we made](https://github.com/nspcc-dev/neofs-node/issues/1227#issuecomment-1083186577) (and that is true for the current implementation)

> finally, could u pls clarify this to me again?

yeap, lockers are the only reason for that force GC mark

> and btw u didn't add similar change to HandleDeletedLocks

... and missing lockers above in the removing list is the reason not to use force GC

pkg/local_object_storage/shard/gc.go
carpawell: done

pkg/local_object_storage/engine/inhume.go
carpawell: fixed

cthulhu-rider (approved):

PR #2536, opened 2023-08-30
Makefile: Exclude `cmd/internal` from `make [all]` target

cthulhu-rider:
The directory is a Go lib package, not an application.


vvarg229 (approved):

carpawell (commented):
My vote is for moving `internal` to `pkg`. It is a helper pkg for `cobra` CLIs to me.

roman-khimov (approved):

PR #2542, opened 2023-09-01
feat/Check that just bootstrapped SN is available

carpawell:


pkg/innerring/processors/netmap/process_peers.go
```diff
@@ -47,6 +51,28 @@ func (np *Processor) processAddPeer(ev netmapEvent.AddPeer) {
...
+
+	c, err := np.clientSource.Get(info)
+	if err != nil {
+		np.log.Warn("could not create client to the node: %w",
+			zap.Error(err),
+			zap.String("node", hex.EncodeToString(info.PublicKey())))
+		return
+	}
+
+	_, err = c.EndpointInfo(context.Background(), client.PrmEndpointInfo{})
```
carpawell, line 69: @roman-khimov, @cthulhu-rider, @AliceInHunterland, ping me if you think we need to check returned values (IMO, that is useless)

roman-khimov: I'd check them against the data for the map.

pkg/innerring/processors/netmap/process_peers.go
```diff
@@ -47,6 +51,28 @@ func (np *Processor) processAddPeer(ev netmapEvent.AddPeer) {
...
+	if err != nil {
+		np.log.Warn("could not ping node with `EndpointInfo`",
+			zap.Error(err),
+			zap.String("node", hex.EncodeToString(info.PublicKey())))
+	}
+
```
roman-khimov, line 75: Close the connection immediately afterwards?

roman-khimov (commented):
What if we have some temporary networking error/inconsistent behavior across IR nodes? Can this be retried once in a while until the notary request expires?

pkg/innerring/processors/netmap/process_peers.go
carpawell: the client was received from the [client cache](https://github.com/nspcc-dev/neofs-node/blob/9224e9f2038d1932027a5c4b01032ed904a93dcc/pkg/innerring/rpc.go#L46-L55) so it is not easy to close it cause there is some complex logic that may break the IR if a client is closed outside

pkg/innerring/processors/netmap/process_peers.go
carpawell: added some checks

carpawell:
> What if we have some temporary networking error/inconsistent behavior across IR nodes?

Every IR node? More than 5/7?

> Can this be retried once in a while until the notary request expires?

Do you mean some cache of "loser" SN nodes that will be retried _some_ times after _some_ period? Need to think about it. Maybe it would be better to do it on the SN side? It may not find itself as a candidate for the next epoch and retry request (but to whom if the IR (local morph consensus) is down?).

pkg/innerring/processors/netmap/process_peers.go
roman-khimov: We can bypass client cache since it's just a probe, open a connection manually and close it.

roman-khimov:
>> What if we have some temporary networking error/inconsistent behavior across IR nodes?
>
> Every IR node? More than 5/7?

Can be anything.

> Maybe it would be better to do it on the SN side?

That depends on notary request expiration parameters mostly, although nodes can retry even without waiting for request to expire.

pkg/innerring/processors/netmap/process_peers.go
```diff
@@ -37,6 +44,37 @@ func (np *Processor) processAddPeer(ev netmapEvent.AddPeer) {
 		return
 	}
 
+	var info clientcore.NodeInfo
+	err = clientcore.NodeInfoFromRawNetmapElement(&info, netmapCore.Node(nodeInfo))
+	if err != nil {
+		np.log.Warn("could not convert node info: %w", zap.Error(err))
```
cthulhu-rider, line 50:
```suggestion
		np.log.Warn("could not convert node info", zap.Error(err))
```
same below

pkg/innerring/processors/netmap/process_peers.go
```diff
@@ -37,6 +44,37 @@ func (np *Processor) processAddPeer(ev netmapEvent.AddPeer) {
 		return
 	}
 
+	var info clientcore.NodeInfo
```
cthulhu-rider, line 47: why don't we extend `np.nodeValidator.VerifyAndUpdate` procedure below with new check?

pkg/innerring/processors/netmap/process_peers.go
```diff
@@ -37,6 +44,37 @@ func (np *Processor) processAddPeer(ev netmapEvent.AddPeer) {
...
+
+	c, err := np.clientSource.Get(info)
+	if err != nil {
+		np.log.Warn("could not create client to the node: %w",
+			zap.Error(err),
+			zap.String("node", hex.EncodeToString(info.PublicKey())))
+		return
+	}
+
+	res, err := c.EndpointInfo(context.Background(), client.PrmEndpointInfo{})
```
cthulhu-rider, line 62: i'd add some sane deadline

pkg/innerring/processors/netmap/process_peers.go
```diff
@@ -116,3 +154,60 @@ func (np *Processor) processUpdatePeer(ev netmapEvent.UpdatePeer) {
 		np.log.Error("can't invoke netmap.UpdatePeer", zap.Error(err))
 	}
 }
+
+var errUnknownDifference = errors.New("announced and received information differ but not critical")
```
cthulhu-rider, line 158: in current code makes no sense

pkg/innerring/processors/netmap/process_peers.go
```diff
@@ -116,3 +154,60 @@ func (np *Processor) processUpdatePeer(ev netmapEvent.UpdatePeer) {
...
+
+func compareNodeInfos(niExp, niGot netmap.NodeInfo) error {
+	// if a node bootstraps the first time it is OK
+	// to be OFFLINE but send ONLINE bootstrap requests
+	niGot.SetOnline()
```
cthulhu-rider, line 163: `if niGot.IsOffline()`?

pkg/innerring/processors/netmap/process_peers.go
carpawell: have not thought about that. it does not seem to me like a node info structure validator

pkg/innerring/processors/netmap/process_peers.go
carpawell: added 15 sec

pkg/innerring/processors/netmap/process_peers.go
carpawell: dont mind dropping it

pkg/innerring/processors/netmap/process_peers.go
carpawell: > states will be almost definitely different

do not agree, health check pings should be the majority

> maybe SetOnline() for both instances?

agree that it should allow maintenance state transfer, fixed

pkg/innerring/processors/netmap/process_peers.go
```diff
@@ -116,3 +154,60 @@ func (np *Processor) processUpdatePeer(ev netmapEvent.UpdatePeer) {
...
+		return nil
+	}
+
+	var err error
+
+	if exp, got := niExp.Hash(), niGot.Hash(); exp != got {
+		return fmt.Errorf("hash: got %d, expect %d", got, exp)
+	}
+
+	niExp.SortAttributes()
```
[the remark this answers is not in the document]
carpawell: hm, strange, not needed at all, dropped

pkg/innerring/processors/netmap/process_peers.go
```diff
@@ -116,3 +154,60 @@ func (np *Processor) processUpdatePeer(ev netmapEvent.UpdatePeer) {
...
+	niGot.SortAttributes()
+	if exp, got := niExp.NumberOfAttributes(), niGot.NumberOfAttributes(); exp != got {
+		return fmt.Errorf("attr number: got %d, expect %d", got, exp)
+	}
+
+	niExp.IterateAttributes(func(key, value string) {
```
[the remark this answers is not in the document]
carpawell: > btw do we check the uniqueness of attributes and endpoints somewhere?

not found that check on the IR side. but _our_ SN will not accept such config

carpawell:
> That depends on notary request expiration parameters mostly, although nodes can retry even without waiting for request to expire.

So we may do smth, I am not sure I have a strict opinion on that. @cthulhu-rider, @AliceInHunterland

pkg/innerring/processors/netmap/process_peers.go
carpawell: dont mind dropping the cache

pkg/innerring/processors/netmap/process_peers.go
carpawell: ok, moved to a separate validator

pkg/innerring/processors/netmap/process_peers.go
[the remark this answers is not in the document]
carpawell: added a validator for it (extended the existing one) but found out that it is [impossible](https://github.com/nspcc-dev/neofs-sdk-go/blame/3be68fb19e8680b7634fa47153cf4fc7209e605b/netmap/node_info.go#L50-L52) to parse such node info so i can't even add a test for it

cthulhu-rider (approved):

PR #2567, opened 2023-09-11
Upgrade NeoGo to 0.102.0

vvarg229:


vvarg229:
> `testcases:`, please. We have way too many other relations to NeoGo to say that it's all 0.102.0 here.

Added workflows prefix

PR #2568, opened 2023-09-11
fix/`EndpointInfo` serving

carpawell:
Be a flexible storage node: take the updated by the Alphabet `NodeInfo` as the
source of truth and bootstrap with it (if this info is available and SN does not
bootstrap for the first time).


roman-khimov (approved):
This just proves that IR shouldn't mess with endpoint data.

PR #2569, opened 2023-09-12
neofs-cli: fix pass of several object put --attributes flags

### LINKED ISSUE #2427: cli: Improve UX concerning attributes' setting in `object put` command
## Context
_Originally posted by @vvarg229_

```
$  neofs-cli object put <required_flags> --attributes RunNumber=184 --attributes FilePath=184.zip ContentType=application/zip
```
command executed successfully. Resulting object was created **with** `FilePath` attribute and  **without** `ContentType` and `Number` ones.

## Overview
currently `object put` command accepts single flag `attributes` of string type to specify list of user-defined object attributes:
```
$ neofs-cli object put -h
      --attributes string     User attributes in form of Key1=Value1,Key2=Value2
```
with next behavior:
* if multiple flags specified (see example), only the latest one will enter the object
* if value contains spaces, all data starting from the 1st space is ignored

### 1. Change type to repeated strings
it may be incovenient to specify multiple attributes in single comma-separated flag without spaces (imagine 5+ elements).

we can support the ability to repeat the flag, while still maintaining the comma-separated format. It'll be easy todo via Cobra's (NeoFS CLI engine) https://pkg.go.dev/github.com/spf13/pflag#StringSlice flag type. In example above, user could specify
```
--attributes RunNumber=184,FilePath=184.zip --attributes ContentType=application/zip
```
and all 3 attributes would be accepted. It may be useful when, for example, some attributes are related like `RunNumber` and `FilePath` while `ContentType` is still independent.

as an alternative, we can support multiple flags with single attributes (w/o comma-splitting within single value). To do this, Cobra provides https://pkg.go.dev/github.com/spf13/pflag#StringArray flag type. 
```
--attribute RunNumber=184 --attribute FilePath=184.zip --attribute ContentType=application/zip
```
With this approach, the name `--attributes` becomes redundant and can be replaced by `--attribute`. To me, this alternative a bit weaker then the slice one, but also concise.

### 2. Strict positional arguments
as we can see in the original example, space-separated `ContentType=application/zip` input is ignored. It happens cuz Cobra interprets this input part as one of the command's positional arguments, and `object put` command just  ignores them. From the user pov, it happens unexpectedly and the input quietly disappears inside.

as a solution, we can start to strictly verify positional input of the command (worth to do for all commands globally), so user could see `invalid input` message and write his command more precisely.

imo strictness is always much better than freedom bordering on uncertainty

AFAIK, similar improvement were added in NeoGo project, correct me pls @AnnaShaleva @roman-khimov

AliceInHunterland:
The `object put` command previously could only retrieve the last `--attributes` parameter if multiple such parameters were provided. Now, the `object put` command correctly parses all `--attributes` values, whether they are provided as comma-separated or single values.

Closes #2427.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ minor release, the component will be purged, so be prepared (see `Updating` sect
 - Missing removed but locked objects in `SEARCH`'s results (#2526)
 - LOCK objects and regular objects expiration conflicts (#2392)
 - SN responds with a different node information compared to a bootstrapping contract call's argument (#2568)
+- `neofs-cli object put` command processes multiple `--attributes` flags (#2427)
 
 ### Removed
 - Deprecated `morph.rpc_endpoint` SN and `morph.endpoint.client` IR config sections (#2400)
--- a/cmd/neofs-cli/modules/object/put.go
+++ b/cmd/neofs-cli/modules/object/put.go
@@ -48,7 +48,7 @@ func initObjectPutCmd() {
 
 	flags.String(commonflags.CIDFlag, "", commonflags.CIDFlagUsage)
 
-	flags.String("attributes", "", "User attributes in form of Key1=Value1,Key2=Value2")
+	flags.StringSlice("attributes", []string{}, "User attributes in form of Key1=Value1,Key2=Value2")
 	flags.Bool("disable-filename", false, "Do not set well-known filename attribute")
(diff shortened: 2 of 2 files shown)


roman-khimov (approved):
So both `--attributes A=X,B=Y` and `--attributes A=X --attributes B=Y` would work the same? Nice.

cthulhu-rider (approved):

PR #2576, opened 2023-09-13
node: don't mkdir random paths from configuration

roman-khimov:
Unbreak peapods:
  2023/09/13 18:50:36 open shard S4wpuCnzpWW7SbwhER3U1v: could not open *blobstor.BlobStor: open substorage peapod: open BoltDB instance: open /storage/peapod0.db: is a directory

Call `stat()` gently instead walking up. FS mount point has to exist there in any event and we should have some access to it.

util.MkdirAllX will be removed from code in most of the cases.


cmd/neofs-node/config.go
```diff
@@ -931,13 +933,21 @@ func writeSystemAttributes(c *cfg) error {
 	var paths []string
 	for _, sh := range c.applicationConfiguration.EngineCfg.shards {
 		for _, subStorage := range sh.SubStorages {
-			path := subStorage.Path
+			var path = subStorage.Path
```
carpawell, line 936: why this change?

carpawell (approved):

cmd/neofs-node/config.go
roman-khimov: Because it's better this way! I've minimized changes here now.

carpawell (approved):

cthulhu-rider (approved):

PR #2577, opened 2023-09-14
neofs-adm: fix displaying values and improve documentation

### LINKED ISSUE #2555: Improve neofs-adm morph set-config --force documentation
related to https://github.com/nspcc-dev/neofs-testcases/pull/628. It is not clear when --force flag should be used and the behavior is not clear as well. E.g. if I set some unknown_key to 120 with the force flag. I will get this back:

```
COMMAND: neofs-cli --config ./TemporaryDir/empty-password.yml netmap netinfo --rpc-endpoint 's01.neofs.devenv:8080' --wallet '../neofs-dev-env/services/storage/wallet01.json'
RETCODE: 0

STDOUT:
Epoch: 5
Network magic: [net 0x3c2d] 15405
Time per block: 1s
NeoFS network configuration (system)
  Audit fee: 10000
  Storage price: 100000000
  Container fee: 1000
  EigenTrust alpha: 0.1
  Number of EigenTrust iterations: 4
  Epoch duration: 240
  Inner Ring candidate fee: 10000000000
  Maximum object size: 524288
  Withdrawal fee: 100000000
  Homomorphic hashing disabled: false
  Maintenance mode allowed: false
NeoFS network configuration (other)
  SystemDNS: 636f6e7461696e6572
  unknown_key: 313230
```
Unknown key is now equal to 313230. It is not obvious why it is so. 

Please add an example of the 'right' force usage and what to expect in the end.

## Don't forget to add labels!
- component label (`neofs-adm`, `neofs-storage`, ...)
- issue type (`enhancement`, `refactor`, ...)
- `goodfirstissue`, `helpwanted` if needed
- does this issue belong to an epic?
- priority (`P0`-`P4`) if already triaged
- quarter label (`202XQY`) if possible

AliceInHunterland:
Comments about the obligatory usage of the `--force` flag for setting unknown parameters have been added. `neofs-cli netmap netinfo` now displays it correctly.

Closes #2555.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Changelog for NeoFS Node
 ## [Unreleased]
 
 ### Fixed
+- `neofs-cli netmap netinfo` documentation (#2555)
 
 ### Changed
 - FSTree storage now uses more efficient and safe temporary files under Linux (#2566)
--- a/cmd/neofs-adm/internal/modules/morph/root.go
+++ b/cmd/neofs-adm/internal/modules/morph/root.go
@@ -131,6 +131,9 @@ var (
 		Use:                   "set-config key1=val1 [key2=val2 ...]",
 		DisableFlagsInUseLine: true,
 		Short:                 "Add/update global config value in the NeoFS network",
+		Long: `Add/update global config value in the NeoFS network. 
+				If key is unknown, it will be added to the config only with --force flag.
+Values for unknown keys are added exactly the way they're provided, no conversion is made ("123" will be stored as "313233" hexadecimal).`,
 		PreRun: func(cmd *cobra.Command, _ []string) {
 			_ = viper.BindPFlag(alphabetWalletsFlag, cmd.Flags().Lookup(alphabetWalletsFlag))
 			_ = viper.BindPFlag(endpointFlag, cmd.Flags().Lookup(endpointFlag))
--- a/cmd/neofs-cli/modules/netmap/netinfo.go
+++ b/cmd/neofs-cli/modules/netmap/netinfo.go
@@ -14,8 +14,9 @@ import (
 var netInfoCmd = &cobra.Command{
 	Use:   "netinfo",
 	Short: "Get information about NeoFS network",
-	Long:  "Get information about NeoFS network",
-	Args:  cobra.NoArgs,
+	Long: `Get information about NeoFS network.
+		Unknown configuration settings are displayed in hexadecimal format.`,
+	Args: cobra.NoArgs,
 	Run: func(cmd *cobra.Command, _ []string) {
 		ctx, cancel := commonflags.GetCommandContext(cmd)
 		defer cancel()


cthulhu-rider (commented):
changes are okay, but lets split them into 2 commits

cthulhu-rider (approved):

cmd/neofs-cli/modules/netmap/netinfo.go
```diff
@@ -55,7 +54,7 @@ var netInfoCmd = &cobra.Command{
 
 		cmd.Println("NeoFS network configuration (other)")
 		netInfo.IterateRawNetworkParameters(func(name string, value []byte) {
-			cmd.Printf(format, name, hex.EncodeToString(value))
+			cmd.Printf(format, name, string(value))
```
roman-khimov, line 57: The old code was correct here. We don't know if it's a string at all, it can be numeric. This just needs to be explained.

Additionally we can check for UTF-8 and print raw string as well if so, just to make things more human-friendly.

carpawell: do not agree that was a bug too. any system settings that are going to be read by the NeoFS system (and neofs-cli is a NeoFS tool so it should work according to its rules) are printed above so this line is just the best effort: we do not know exactly what it is but we can print _these bytes_ (they are for some NeoFS extensions more likely)

OR

we may add a restriction that the [values](https://github.com/nspcc-dev/neofs-api/blob/7ae1a7438b19fc4ca32c451db67de170edb539a7/netmap/types.proto#L306) are utf8 strings too

carpawell: but +1 that we can check if it is a valid utf-8 string and print it

roman-khimov: They're explicitly not utf8 in case of numeric settings (like fees). Eventually we'll solve nspcc-dev/neofs-contract#314 and all of this will fade away, but currently in this particular case we can expect arbitrary bytes.

carpawell: > They're explicitly not utf8 in case of numeric settings (like fees).

oh, well, i see. also, changing API isnt a good think in general. so i would document `hex` encoding in CLI

cmd/neofs-cli/modules/netmap/netinfo.go
```diff
@@ -14,8 +14,9 @@ import (
 var netInfoCmd = &cobra.Command{
 	Use:   "netinfo",
 	Short: "Get information about NeoFS network",
-	Long:  "Get information about NeoFS network",
-	Args:  cobra.NoArgs,
+	Long: `Get information about NeoFS network. 
+Other parameters of NeoFS network configuration are stored as bytes and printed in HEX.`,
```
carpawell, line 18: > are stored as bytes

well, everything is stored as bytes mb something like "Unknown (non-system) network parameters are displayed in HEX encoding."

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Changelog for NeoFS Node
 - Inability to start node with peapods configured (#2576)
 
 - `neofs-adm morph set-config` command error messages (#2556)
+- `neofs-cli netmap netinfo` documentation (#2555)
```
roman-khimov, line 10: Needs to be moved, 0.38.1 is out.

cmd/neofs-adm/internal/modules/morph/root.go
```diff
@@ -131,6 +131,8 @@ var (
 		Use:                   "set-config key1=val1 [key2=val2 ...]",
 		DisableFlagsInUseLine: true,
 		Short:                 "Add/update global config value in the NeoFS network",
+		Long: `Add/update global config value in the NeoFS network. 
+				If key is unknown, it will be added to the config only with --force flag.`,
```
roman-khimov, line 135: Add something like

Values for unknown keys are added exactly the way they're provided, no conversion is made ("123" will be stored as "313233" hexadecimal)

?

cmd/neofs-cli/modules/netmap/netinfo.go
roman-khimov: "Unknown configuration settings are displayed in hexadecimal format"?

carpawell: ok to me too

CHANGELOG.md
```diff
@@ -1770,8 +1771,7 @@ NeoFS-API v2.0 support and updated brand-new storage node application.
 
 First public review release.
 
-[Unreleased]: https://github.com/nspcc-dev/neofs-node/compare/v0.38.1...master
-[0.38.1]: https://github.com/nspcc-dev/neofs-node/compare/v0.38.0...v0.38.1
```
carpawell, line 1774: why drop this harmless release?

PR #2580, opened 2023-09-14
Introduce private attributes of the storage nodes

### LINKED ISSUE #2280: Verifiable attributes
Any node operator can set any attribute for their node on the network. We can't have some controlled group that uses an attribute no one else can use, while this might be useful.

One simple approach is just to have a public group key and some signature derived from the corresponding private key, similar to the way NEP-15 groups are designed (https://github.com/neo-project/proposals/blob/master/nep-15.mediawiki#user-content-Group). It may be not the most convenient one at the same time.

Another approach could be reusing NNS with some special key/address entries for nodes. Maybe there are some other ones.

cthulhu-rider:
* based on and blocked by https://github.com/nspcc-dev/neofs-contract/milestone/31
* closes #2280


docs/private-node-attributes.md
```diff
@@ -0,0 +1,26 @@
...
+
+Note that if domain for the attribute exists but has no records, attribute is
+non-private.
+
+### Domain record format
+
+For each public key, a record is created - a structure with at least 3 fields:
+1. `ByteString` with name of the corresponding domain
+2. `Integer` that should be `16` (TXT records)
+3. `ByteString` with HEX-encoded public key
```
roman-khimov, line 26: For any address a key can be derived. Any address can be stored using https://github.com/neo-project/proposals/pull/133. Maybe this would be a bit better?

docs/private-node-attributes.md
```diff
@@ -0,0 +1,26 @@
...
+network map. In general, any attributes can be declared, however, some of them
+may be subject to restrictions. In particular, to specify so-called private
+attributes, the node must be in the corresponding access list.
+
+## Access lists
+
+These lists are stored in the NeoFS NNS. For each private attribute value
+`KEY=VALUE`, there is a domain with name `MD5(VALUE).MD5(KEY).private-node-attributes.neofs`
+and records of `TXT` type with HEX-encoded public keys. Here `MD5` is a
+HEX-encoded MD5 checksum. If the domain exists and has non-empty list of public
```
roman-khimov, line 13: How about `base58(SHA256(Key + \x0 + Value))`? Zero byte can be replaced by any other separator that can't be used in keys or values.

docs/private-node-attributes.md
```diff
@@ -0,0 +1,26 @@
...
+this attribute value. The Inner Ring will deny everyone else access to the
+network map.
+
+Note that if domain for the attribute exists but has no records, attribute is
+non-private.
+
+### Domain record format
+
+For each public key, a record is created - a structure with at least 3 fields:
+1. `ByteString` with name of the corresponding domain
```
roman-khimov, line 24: But why a new domain for each key if you can have multiple TXT entries and just `resolve()`?

docs/private-node-attributes.md
cthulhu-rider: about concatenation: i thought about that. Right now it may be unstable since attribute value may be [any](https://github.com/nspcc-dev/neofs-api/blob/c1a97ad6ea03f8bd7bf71e47cba946a9cb1ca9e5/netmap/types.proto#L140-L141), so we cannot select a reserved byte sequence (like `\0` in ur example). But if it existed, we could save on the number of domains. In total, i see either two domains or protocol changes

about encoding: i don't see much differencies b/w md5/sha256 and hex/base58 right now. What potential problem do they solve or why else u suggest to use them?

docs/private-node-attributes.md
cthulhu-rider: no domain per-pubkey, only per-attribute. Node keys are set as TXT records. Or docs confuse this moment?

docs/private-node-attributes.md
cthulhu-rider: sounds like a working option

> Maybe this would be a bit better?

not ready to answer right now. Can we outline particular problem/ability such approach may solve/provide? so that the transition is justified not only by technical feasibility

docs/private-node-attributes.md
```diff
@@ -0,0 +1,26 @@
...
+
+Storage nodes declare key-value string attributes when applying to enter the
+network map. In general, any attributes can be declared, however, some of them
+may be subject to restrictions. In particular, to specify so-called private
+attributes, the node must be in the corresponding access list.
+
+## Access lists
+
+These lists are stored in the NeoFS NNS. For each private attribute value
+`KEY=VALUE`, there is a domain with name `MD5(VALUE).MD5(KEY).private-node-attributes.neofs`
```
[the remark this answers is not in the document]
cthulhu-rider: afaik only `private-node-attributes.neofs` domain's owner and admins may register sub-domains, am i wrong?

roman-khimov (changes requested):
Please adjust the scheme to have a fixed key that contains arbitrary NeoFS NNS domain (that can be registered by anyone and that's the key). This domain then should have a list of TXT records with addresses derived from node keys. IR will check for presence in this list.

cthulhu-rider:
changes are done

https://github.com/nspcc-dev/neofs-contract/releases/tag/v0.18.0 is now available, but it pulls https://github.com/nspcc-dev/neo-go/releases/tag/v0.102.0 along. So, lets wait for #2587 first

pkg/innerring/nns.go
```diff
@@ -0,0 +1,207 @@
...
+					invalidStructFieldError(0, wrongStackItemTypeError(stackitem.ByteArrayT, fields[0].Type())))
+			}
+
+			typ, err := fields[1].TryInteger()
+			if err != nil {
+				return invalidNNSDomainRecordError(domain, "",
+					invalidStructFieldError(1, wrongStackItemTypeError(stackitem.IntegerT, fields[1].Type())))
+			}
+
+			if typ.Cmp(nnsrpc.TXT) != 0 {
```
roman-khimov, line 185: Domain can have some other records as well.

cmd/neofs-adm/internal/modules/morph/verified_domains.go
```diff
@@ -0,0 +1,219 @@
...
+				return fmt.Errorf("address #%d is invalid: %w", i, err)
+			}
+		}
+
+		additionalRecords = strNeoAddresses
+	} else {
+		additionalRecords = make([]string, len(strPublicKeys))
+
+		for i := range strPublicKeys {
+			bKey, err := hex.DecodeString(strPublicKeys[i])
```
roman-khimov, line 94: Just decode a proper public key with https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/crypto/keys#NewPublicKeyFromString , it's not a hot path of some kind. And it'll allow to use https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/crypto/keys#PublicKey.GetScriptHash

cmd/neofs-adm/internal/modules/morph/verified_domains.go
```diff
@@ -0,0 +1,219 @@
...
+			buf := io.NewBufBinWriter()
+			emit.CheckSig(buf.BinWriter, bKey)
+
+			additionalRecords[i] = address.Uint160ToString(hash.Hash160(buf.Bytes()))
+		}
+	}
+
+	walletDir := config.ResolveHomePath(vpr.GetString(alphabetWalletsFlag))
+
+	wallets, err := openAlphabetWallets(vpr, walletDir)
```
roman-khimov, line 114: This particular command doesn't require alphabet wallets.

cmd/neofs-adm/internal/modules/morph/verified_domains.go
```diff
@@ -0,0 +1,219 @@
...
+	if err != nil {
+		return fmt.Errorf("open connection: %w", err)
+	}
+
+	nnsContractAddr, err := nnsrpc.InferHash(n3Client)
+	if err != nil {
+		return fmt.Errorf("get NeoFS NNS contract address: %w", err)
+	}
+
+	actr, err := actor.NewSimple(n3Client, committeeAcc)
```
roman-khimov, line 134: Or committee.

cmd/neofs-adm/internal/modules/morph/verified_domains.go
```diff
@@ -0,0 +1,219 @@
...
+
+		hasOtherRecord = true
+
+		return errBreakIterator
+	})
+	if err != nil {
+		switch {
+		default:
+			return fmt.Errorf("handle domain %q records: %w", domain, err)
+		case errors.Is(err, errMissingDomainRecords):
```
roman-khimov, line 160: Better have some comment for this case.

cmd/neofs-adm/internal/modules/morph/verified_domains.go
```diff
@@ -0,0 +1,219 @@
...
+		scriptBuilder.InvokeMethod(nnsContractAddr, "addRecord",
+			domain, nnsrpc.TXT.Int64(), additionalRecords[i])
+	}
+
+	txScript, err := scriptBuilder.Script()
+	if err != nil {
+		return fmt.Errorf("build transaction script: %w", err)
+	}
+
+	txID, vub, err := actr.SendRun(txScript)
```
roman-khimov, line 201: Why not `actr.Wait(actr.SendRun(txScript))`?

docs/verified-node-domains.md
```diff
@@ -0,0 +1,40 @@
...
+NNS domain and set records of `TXT` type with Neo addresses of the storage
+nodes. After the domain is registered, it becomes an alias to the subnet composed
+only from specified storage nodes. Any storage node trying to associate itself
+with this subnet while trying to enter the network must have public key
+presented in the access list. The Inner Ring will deny everyone else access to
+the network map.
+
+### Domain record format
+
+For each public key, a record is created - a structure with at least 3 fields:
```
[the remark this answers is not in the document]
cthulhu-rider: `AddRecord` only creates the record, then we read the structures. This may be useful for alternative integrations

cmd/neofs-adm/internal/modules/morph/verified_domains.go
cthulhu-rider: to stay closer to other commands

pkg/innerring/nns.go
```diff
@@ -0,0 +1,207 @@
...
+// CheckDomainRecord calls iterating 'getAllRecords' method of the parameterized
+// Neo smart contract passing the given domain name. If contract throws 'token
+// not found' exception, CheckDomainRecord returns errDomainNotFound. Each value
+// in the resulting iterator is expected to be structure with at least 3 fields.
+// If any value has the 2nd field is a number equal to 16 (TXT record type in
+// the NNS) and the 3rd one is a string equal to the specified record,
+// CheckDomainRecord returns nil. Otherwise,
+// [privatedomains.ErrMissingDomainRecord] is returned.
+func (x *neoFSNNS) CheckDomainRecord(domain string, record string) error {
+	sessionID, iter, err := x.contract.GetAllRecords(domain)
```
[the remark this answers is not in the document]
roman-khimov: This doesn't justify all of this code. Resolve can have iterators as well, but it might not have them too,  https://github.com/nspcc-dev/neofs-contract/pull/266/commits/b1f5864942138d8667badd0f7c583417a5bc593c

cmd/neofs-adm/internal/modules/morph/verified_domains.go
cthulhu-rider: done

cmd/neofs-adm/internal/modules/morph/verified_domains.go
cthulhu-rider: done

pkg/innerring/nns.go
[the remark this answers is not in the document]
cthulhu-rider: done

docs/verified-node-domains.md
```diff
@@ -0,0 +1,40 @@
...
+1. `ByteString` with name of the corresponding domain
+2. `Integer` that should be `16` (TXT records)
+3. `ByteString` with Neo address of the storage node's public key
```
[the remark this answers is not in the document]
cthulhu-rider: done

cmd/neofs-adm/internal/modules/morph/verified_domains.go
cthulhu-rider: changed to work with any account

roman-khimov (approved):

PR #2587, opened 2023-09-20
upd/`neo-go` to `v0.102.0`

carpawell:


pkg/morph/deploy/util.go
```diff
@@ -41,7 +41,9 @@ func isErrNotEnoughGAS(err error) bool {
 }
 
 func isErrInvalidTransaction(err error) bool {
-	return errors.Is(err, neorpc.ErrValidationFailed)
+	// neo-go changed -504 code to be a network fee error,
+	// but it was a verification error before v0.102.0
+	return errors.Is(err, neorpc.ErrInsufficientNetworkFee)
```
roman-khimov: That's a straight `isErrNotEnoughGAS`, while `isErrInvalidTransaction` is any of -5xx (you can get a lot of different codes now).

pkg/morph/client/notifications.go
```diff
@@ -107,3 +164,167 @@ func (c *Client) UnsubscribeAll() error {
...
+	curBlockChan  chan *block.Block
+	curNotaryChan chan *result.NotaryRequestEvent
+
+	// cached subscription information
+	subscribedEvents       map[util.Uint160]struct{}
+	subscribedNotaryEvents map[util.Uint160]struct{}
+	subscribedToNewBlocks  bool
+}
+
+func (c *Client) routeNotifications() {
```
roman-khimov, line 199: Why do we still reroute?

pkg/morph/client/constructor.go
```diff
@@ -145,8 +145,10 @@ func New(key *keys.PrivateKey, opts ...Option) (*Client, error) {
 }
 
 func (c *Client) newCli(endpoint string) (*rpcclient.WSClient, *actor.Actor, error) {
-	cli, err := rpcclient.NewWS(c.cfg.ctx, endpoint, rpcclient.Options{
-		DialTimeout: c.cfg.dialTimeout,
+	cli, err := rpcclient.NewWS(c.cfg.ctx, endpoint, rpcclient.WSOptions{
+		Options: rpcclient.Options{
+			DialTimeout: c.cfg.dialTimeout,
```
carpawell, line 178: now sure what do we do with `CloseNotificationChannelIfFull`

roman-khimov: It's a deadlock safeguard basically, likely should be set. @AnnaShaleva?

carpawell: > It's a deadlock safeguard basically

yes, but do we need it? it leads to unexpected connection switches, not sure if it is ok for the current logic. but if we want if, i will recheck it

pkg/morph/deploy/util.go
carpawell: yes, usually i split "big" libraries updates into two (or more) commits: one for the `go.mod` file and the other for the features adoptions. the problem was that `ErrValidationFailed` just does not exist anymore so it could not be complied. that is why i put it here (another -504 code, even if it is named weird for that part of the code). it is dropped right in the next commit

pkg/morph/client/constructor.go
roman-khimov: > it leads to unexpected connection switches...

...or a deadlock. But you can choose which one you like more!

pkg/morph/deploy/util.go
roman-khimov: Yeah, I've seen that as well.

pkg/morph/client/constructor.go
carpawell: well, not sure what is worse fot debugginh: unexpected connection switch or unexpected deadlock so added a commit that turns that new option on

pkg/morph/client/constructor.go
```diff
@@ -71,10 +71,10 @@ func defaultConfig() *cfg {
 	}
 }
 
-// ErrStaleNode is returned from [New] when minimal required height
+// ErrStaleNodes is returned from [New] when minimal required height
 // requirement specified in [WithMinRequiredBlockHeight] is not
 // satisfied by the given nodes.
-var ErrStaleNode = errors.New("RPC node is not yet up to date")
+var ErrStaleNodes = errors.New("RPC nodes are not yet up to date")
```
roman-khimov, line 77: Why this change, really?

carpawell: cause now it is retuned only if _every_ node is _not_ up to date. calling `SwitchRPC` outside of the morph pkg is strange to me

roman-khimov (approved):

pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -352,12 +352,12 @@ func New(cfg Config) (res *Blockchain, err error) {
 	cfgBaseProto.MaxTraceableBlocks = cfg.TraceableChainLength
 	cfgBaseProto.Hardforks = cfg.HardForks
 	if cfg.ValidatorsHistory != nil {
-		cfgBaseProto.ValidatorsHistory = make(map[uint32]int, len(cfg.ValidatorsHistory))
+		cfgBaseProto.ValidatorsHistory = make(map[uint32]uint32, len(cfg.ValidatorsHistory))
```
cthulhu-rider, line 355: no need to copy elements anymore

cthulhu-rider (approved):

Issue #1663, opened 2022-08-05
GC should have some reaction on container removal event


carpawell:
Node is [able](https://github.com/nspcc-dev/neofs-contract/blob/02c2ce3966e68b23c0c08818e88949ad9bbc27c3/container/container_contract.go#L361) to track container removal but currently, it does nothing.

It could: 
1. Mark object of just removed container for removal;
2. Change its behavior (do not replicate objs, do not send estimations for that cnr, etc)
3. Mark container as removed in the Metabase to prevent operation with the objects that are still stored but technically *do not exist anymore* 
4. ...

/cc @realloc

roman-khimov:
Looks like we have some cases like this on mainnet (slightly modified tool from #2491):
```
2023/08/15 09:00:37 read container 5bq36EJ27VWTyzWtyVaeXMBr339TSex5v3uJLwnjMM3w: status: code = 3072 message = container not found
2023/08/15 09:00:37 read container 2pWrCbvejFqLy1Z3jEYLyACVX6ZdhLudnFjt2Ct5HT1B: status: code = 3072 message = container not found
2023/08/15 09:00:37 read container 2pWrCbvejFqLy1Z3jEYLyACVX6ZdhLudnFjt2Ct5HT1B: status: code = 3072 message = container not found
2023/08/15 09:00:37 read container 2pkfDULFobUXxVw4NVvqNjbmUwZHKv7dwU32W8SatrLr: status: code = 3072 message = container not found
2023/08/15 09:00:37 read container 5bq36EJ27VWTyzWtyVaeXMBr339TSex5v3uJLwnjMM3w: status: code = 3072 message = container not found
2023/08/15 09:00:37 read container 5bq36EJ27VWTyzWtyVaeXMBr339TSex5v3uJLwnjMM3w: status: code = 3072 message = container not found
2023/08/15 09:00:37 read container 5qj8tUtmARpJNtfzX1PfF4u8vugqAdM29x3Sv31Qdr5X: status: code = 3072 message = container not found
2023/08/15 09:00:38 read container EShohP7bvDmVoTXo8Xg2MXcXgi5yw1zjSqrSrhEz17rv: status: code = 3072 message = container not found
2023/08/15 09:00:38 read container EShohP7bvDmVoTXo8Xg2MXcXgi5yw1zjSqrSrhEz17rv: status: code = 3072 message = container not found
[...]

PR #2597, opened 2023-09-27
fix/Estimations announcement

carpawell:


carpawell:
Seems like we also need a routine that is going to iterate over containers and be ready to clean up every info on every container removal (also, the storage engine should support this operation too, of course). But it is a too big change so not added it yet to this PR. We have some [removal subscription](https://github.com/nspcc-dev/neofs-node/blob/4bbe8f99e30e4fde727e9c3ea36e695ed58e74cf/cmd/neofs-node/container.go#L109) but it won't work for the removals at the SN's offline time.

Thoughts? @roman-khimov, @cthulhu-rider, @AliceInHunterland

roman-khimov:
> clean up every info on every container removal

Are you talking about #1663 or some kind of locally stored metabase thing related to estimations specifically?

roman-khimov (approved):

carpawell:
> some kind of locally stored metabase thing related to estimations specifically?

Like this, yes. Iff a node is sleeping it won't be able to drop its indexes with the notifications approach (that we already have). But yes, now it does _absolutely nothing_ if we are talking about the estimations but I do not want to fix it fast cause notifications it the only part of potentially useless info in the metabase if a container is dropped, so want to hear your thoughts and maybe create another issue for more accurate solution.

But the current missing container issue is fixed with this PR, although I would not treat it as the best solution for _the exact that_ problem.

roman-khimov:
I'd forget this problem for now, estimation mechanism itself won't be like this forever.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Changelog for NeoFS Node
 ### Fixed
 - `neofs-cli netmap netinfo` documentation (#2555)
 - `GETRANGEHASH` to a node without an object produced `GETRANGE` or `GET` requests (#2541)
+- Iteration over locally collected container estimation values (#)
```
roman-khimov, line 9: Number?

carpawell: fixed

cthulhu-rider (approved):

roman-khimov:
Needs a rebase, CHANGELOG has changed.

Issue #2590, opened 2023-09-25
Configurable inter-object policer delay


roman-khimov:
Policer loops through objects as fast as it can. Let's try making it a bit slower in that with an experimental option that adds a delay between HEADs to other nodes.

PR #2600, opened 2023-09-28
feat/Policer enhancements

carpawell:


carpawell:
Treat almost every commit as a suggestion and feel free to ask, suggest, and discuss.

pkg/services/policer/policer.go
```diff
@@ -108,7 +108,7 @@ func defaultCfg() *cfg {
 	return &cfg{
 		log:           zap.L(),
 		batchSize:     10,
-		cacheSize:     1024, // 1024 * address size = 1024 * 64 = 64 MiB
+		cacheSize:     1 << 20, // 1024*1024 * address size = 1MiB * 64 = 64 MiB
```
roman-khimov, line 114: MiB? Please...

roman-khimov: But the biggest delay here in fact is communication. You're not likely to do more than 1000 RPS per thread, which roughly means 600000 entries for 30s

pkg/services/policer/process.go
```diff
@@ -91,7 +91,7 @@ func (p *Policer) poolCapacityWorker(ctx context.Context) {
 
 			if p.taskPool.Cap() != newCapacity {
 				p.taskPool.Tune(newCapacity)
-				p.log.Debug("tune replication capacity",
+				p.log.Info("tune replication capacity",
```
roman-khimov, line 120: Once per second is too frequent for `info`.

pkg/services/policer/process.go
```diff
@@ -72,6 +79,13 @@ func (p *Policer) shardPolicyWorker(ctx context.Context) {
 				}
 			}
 		}
+
+		select {
+		case <-ctx.Done():
+			return
+		case <-t.C:
```
roman-khimov, line 92: But it's not per-object, isn't it?

carpawell: yes, it is per object replication (per 10 objects for now but can be per any number of objects (e.g. 1)). do you want to fetch objects at once and then replicate them with sleeps? e.g. read 100 objects and replicate the last one after 99+ seconds? what if the batch size is 1000?

pkg/services/policer/process.go
carpawell: i may suggest a metric then. want to check this tunning's values and not sure it can help in any scenarios _consistently_

pkg/services/policer/policer.go
carpawell: > MiB? Please...

it was `MiB` i went `MiB`, fixed

> 600000 entries for 30s

but 30s maybe not 30s. and 1000 RPS may be 100 RPS. not sure we need this cache in general but think that 1024 is senseless for sure. this PR allows us to check it in tests. what is your concern/suggestion?

pkg/services/policer/process.go
roman-khimov: OK, let's try this way.

pkg/services/policer/policer.go
roman-khimov: Let's try, OK.

carpawell: my main question, in fact, is about memory usage. what do you think about 64mb for replication cache storage?

pkg/services/policer/process.go
carpawell: agree that 1s logs are annoying, how do we go here?

roman-khimov: Leave it as debug? Is this behavior configurable?

pkg/services/policer/policer.go
roman-khimov: I'm not sure we need it, but it doesn't seem to be a big problem either, if you have 1M objects it's not a small installation.

pkg/services/policer/process.go
carpawell: > Leave it as debug?

ok, reverted.

> Is this behavior configurable?

max value with this PR -- yes, but the way it tunes itself in [1, `maxCapacity`] is random to me

carpawell:
Made new configs public.

cthulhu-rider:
lets swap 3418af9c24fbf8c775c6cd0c1fbcc59b8cad197d and 2ddf3126f1421fad480e13cd660b6d44d2e36769 in order? first one changes code in order to change nothing. new feature must affect smth insta imo

config/example/node.yaml
```diff
@@ -94,6 +94,11 @@ apiclient:
 
 policer:
   head_timeout: 15s  # timeout for the Policer HEAD remote operation
+  cache_size: 1000001 # recently-handled objects cache size
+  cache_time: 31s # recently-handled objects cache expiration time
```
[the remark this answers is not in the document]
carpawell: > size and time to cache section?

well, could be done but i am almost sure the resulting config will change so lets wait testing results and see what we need

config/example/node.yaml
```diff
@@ -94,6 +94,11 @@ apiclient:
...
+  replication_cooldown: 101ms # cooldown time b/w replication tasks submitting
```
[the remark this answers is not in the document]
carpawell: > it's implemented like interval, not cooldown

what do you mean here? implementation is what i would call a cooldown (but i could be wrong)

carpawell:
> lets swap https://github.com/nspcc-dev/neofs-node/commit/3418af9c24fbf8c775c6cd0c1fbcc59b8cad197d and https://github.com/nspcc-dev/neofs-node/commit/2ddf3126f1421fad480e13cd660b6d44d2e36769 in order?

That was not a mistake, that is how i usually do such things.

> changes code in order to change nothing

The second commit is a summary: now the policer is configurable. The first one is about a policer's internal change (although the defaults work ok and no behavior changes were presented (btw, the commit says that nothing changed)). Dont mind swapping if you think that it is not correct.

cc @roman-khimov

config/example/node.yaml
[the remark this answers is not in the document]
carpawell: > why do we expose all parameters while https://github.com/nspcc-dev/neofs-node/issues/2590 suggests replication_cooldown only?

the whole PR is a suggestion. we have magic consts and _I_ am not sure how they work. so now we have a PR that can be tested

roman-khimov:
Commits are OK as is. And OK the other way around. It doesn't change much here.

config/example/node.yaml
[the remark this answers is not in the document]
cthulhu-rider: if `replication_cooldown` is `10s` and last replication took `5s`, then next replication will start after `5s` or `10s`? If `10s`, then yes, this is cooldown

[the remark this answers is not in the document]
carpawell: lets use some ascii paintings here

Rs -- replication started
Rf -- replication finished

```
Rs-----Rf-----Rs--------------------Rf/Rs
|______||_____||______________________|
   5s      5s            20s
```

a cooldown to me is "i used some func, next time i will be able to use it no earlier than after \*cooldown_time\*"
an interval to me is "i used some func, lets sleep for \*interval_time\*"

roman-khimov (approved):

Issue #2496, opened 2023-08-11
neofs-cli container set-eacl removal of the no-precheck flag


AliceInHunterland:
Remove the "no-precheck" flag, which is equivalent to "force" (skips verification) and has been deprecated since #2490.

roman-khimov:
It's deprecated in #2490, now we need to drop it here.

PR #2490, opened 2023-08-10
cli: set-eacl update

AliceInHunterland:
force flag added, no-precheck deprecated, ownership check added
Refs #2436.

PR #2605, opened 2023-10-02
neofs-cli: container set-eacl removal of the no-precheck flag

AliceInHunterland:
Remove the "no-precheck" flag, which is equivalent to "force" (skips verification) and has been deprecated since #2490.

Closes: #2496.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Changelog for NeoFS Node
 - BoltDB open timeout increased from 100ms to 1s (#2499)
 
 ### Removed
+- deprecated `no-precheck` flag of `neofs-cli container set-eacl` (#2496)
 
 ### Updated
 - Update minimal supported Go version up to v1.19 (#2485)
--- a/cmd/neofs-cli/modules/container/set_eacl.go
+++ b/cmd/neofs-cli/modules/container/set_eacl.go
@@ -16,8 +16,6 @@ import (
 )
 
 var flagVarsSetEACL struct {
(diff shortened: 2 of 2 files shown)


cmd/neofs-cli/modules/container/set_eacl.go
```diff
@@ -41,7 +39,7 @@ Container ID in EACL table will be substituted with ID from the CLI.`,
 		pk := key.GetOrGenerate(cmd)
 		cli := internalclient.GetSDKClientByFlag(ctx, cmd, commonflags.RPC)
 		force, _ := cmd.Flags().GetBool(commonflags.ForceFlag)
-		if !(force || flagVarsSetEACL.noPreCheck) {
+		if !(force) {
```
cthulhu-rider, line 42:
```suggestion
		if !force {
```
surprised that all linters don't care btw

cthulhu-rider (approved):

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ Changelog for NeoFS Node
 - FSTree storage now uses more efficient and safe temporary files under Linux (#2566)
 
 ### Removed
+- `neofs-cli container set-eacl` removal of the `no-precheck` flag (#2496)
```
roman-khimov, line 15: The section is named "removed", therefore this line should be more like `deprecated no-precheck flag of neofs-cli container set-eacl`. It's important that this flag was deprecated previously as well.

roman-khimov (approved):

PR #2607, opened 2023-10-02
fix/PUT service enhancements

carpawell:


cmd/neofs-node/config.go
```diff
@@ -790,6 +791,9 @@ func initObjectPool(cfg *config.Config) (pool cfgObjectRoutines) {
 	pool.putRemote, err = ants.NewPool(pool.putRemoteCapacity, optNonBlocking)
 	fatalOnErr(err)
 
+	pool.putLocal, err = ants.NewPool(-1) // -1 stands for an infinite capacity worker
```
cthulhu-rider, line 794: did u test this change? does it affect anything? afaik https://pkg.go.dev/github.com/nspcc-dev/neofs-node/pkg/util#NewPseudoWorkerPool used by default also provides an infinite pool

pkg/services/object/put/streamer.go
```diff
@@ -175,6 +175,8 @@ func (p *Streamer) preparePrm(prm *PutInitPrm) error {
 		)
 	}
 
+	prm.traverseOpts = append(prm.traverseOpts, placement.WithCopiesNumber(prm.copiesNumber))
```
cthulhu-rider, line 178: `placement.SuccessAfter` does not fit?

carpawell: `SuccessAfter` does some strange things and is used for read operations. it makes a placement a vector of nodes (not a vector of vectors, so just a concatenation) and says that N number is enough (the first N). so if we have `REP 3` (and some more `REP`s) for the first vector with 10 length and call `SuccessAfter(10)` it will place 7 useless replicas that should be dropped later

cmd/neofs-node/config.go
carpawell: let's sync the terms. consider the [example](https://gist.github.com/carpawell/fd443d752aee1961d11efd9c3ef05894) and its output:
```
ants took: 10ms
sync pool took: 1038ms
```

that is what i (and `ants`'s creator) call an infinite pool: if you have a cycle that `Submit`s some task an infinite loop won't lock any iteration ever (in other words `numOfJobs` in the example can be any number and every job will always have a separate routine). on the other hand, `PseudoWorkerPool` just calls a func and always blocks

> if you have a cycle that `Submit`s some task

and that is what [we have](https://github.com/nspcc-dev/neofs-node/blob/c817194269943e4ed56d85eedf33b529bb1b282e/pkg/services/object/put/distributed.go#L180-L224) in our distribution logic: every local put task is done syncly (even if we have more remote node to handle)

carpawell: > used by default also provides an infinite pool

so answering this part: `PseudoWorkerPool` does not provide a pool by design (finite or infinite), it always blocks. not smth we expect when trying to parallelize some i/o operations

cthulhu-rider: got it now, thx

pkg/services/object/put/streamer.go
cthulhu-rider: this is an interesting topic at the intersection of different dimensions

let's try to base the behavior on writing and reading on the same basis. I expect both ops to iterate over same sequence of nodes (unless reading actually ends after the 1st success)

roman-khimov (approved):

roman-khimov:
Conflicts, needs to be rebased.

pkg/services/object/put/streamer.go
carpawell: https://github.com/nspcc-dev/neofs-node/issues/2616

roman-khimov (approved):

PR #2617, opened 2023-10-09
policer: use 1s as the default replicaton cooldown

roman-khimov:
10 objects per second is OK and seems to be similar to what we have on current public networks.

### DIFF
--- a/pkg/services/policer/policer.go
+++ b/pkg/services/policer/policer.go
@@ -107,7 +107,7 @@ func defaultCfg() *cfg {
 		log:           zap.L(),
 		batchSize:     10,
 		rebalanceFreq: 1 * time.Second,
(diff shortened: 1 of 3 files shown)


pkg/services/policer/policer.go
```diff
@@ -107,7 +107,7 @@ func defaultCfg() *cfg {
 		log:           zap.L(),
 		batchSize:     10,
 		rebalanceFreq: 1 * time.Second,
-		repCooldown:   0,
+		repCooldown:   1 * time.Second,
```
carpawell, line 110: this change should also be applied to the `cmd/config/policer` pkg

roman-khimov: Fixed.

PR #2620, opened 2023-10-17
Move bbcz into migrator

roman-khimov:


pkg/local_object_storage/shard/dump.go
```diff
@@ -31,7 +31,7 @@ func (p *DumpPrm) WithStream(r io.Writer) {
 }
 
 // WithIgnoreErrors is an Dump option to allow ignore all errors during iteration.
-// This includes invalid blobovniczas as well as corrupted objects.
+// This includes invalid peapods as well as corrupted objects.
```
carpawell, line 34: strange comment but at least there is always one peapod per shard

PR #2624, opened 2023-10-19
fstree: handle unsupported O_TMPFILE, fix #2609

### LINKED ISSUE #2609: fstree: Unsupported `O_TMPFILE` flag
_Originally posted by @smallhive_

```
warn	engine/engine.go:159	could not put object to shard	{"shard_id": "Amjk6muSeEhGPqoKe8typ8", "error count": 1, "error": "could not put object to BLOB storage: put object to sub-storage fstree: unix open: operation not supported"}
```

```
$ docker info
Client: Docker Engine - Community
 Version:    24.0.2
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.10.5
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.18.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 33
  Running: 1
  Paused: 0
  Stopped: 32
 Images: 516
 Server Version: 24.0.2
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 3dce8eb055cbb6872793272b4f20ed16117344f8
 runc version: v1.1.7-0-g860f061
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
 Kernel Version: 5.4.0-163-generic
 Operating System: Linux Mint 20.3
 OSType: linux
 Architecture: x86_64
 CPUs: 16
 Total Memory: 15.44GiB
 Name: ice
 ID: 6K3A:4WAL:YKHV:LSW3:M2E5:EUVQ:CV7D:QE4D:WFFQ:SUZL:NHBG:6JEU
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 Username: belazar13
 Experimental: false
 Insecure Registries:
  89.208.208.118:5000
  127.0.0.0/8
 Live Restore Enabled: false
```

Host machine filesystem is `ext4`

---

`O_TMPFILE` may be not supported by running filesystem and/or kernel. As 59250aae0b88dd6ab29520ae2140ee52dff459e9 says it's expected that `modern FSs` are gonna be used, but this is not obvious for admins. Me personally dont expect runtime node error like above

## Possible solution
require flag support, document this and pre-check on initialization. If flag is not supported, abort application start. This doesn't cover 100% of possible cases, but will work in practice

as an alternative, we may also fallback to operation without the flag if it's not supported. But i find requirements to use the most modern platforms better

roman-khimov:
Probe it during Init.

### DIFF
--- a/pkg/local_object_storage/blobstor/fstree/fstree_write_linux.go
+++ b/pkg/local_object_storage/blobstor/fstree/fstree_write_linux.go
@@ -4,26 +4,47 @@ package fstree
 
 import (
 	"errors"
+	"io/fs"
 	"strconv"
 
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor/common"
 	"golang.org/x/sys/unix"
 )
 
-func (t *FSTree) writeData(p string, data []byte) error {
-	err := t.writeFile(p, data)
+type linuxWriter struct {
+	root  string
+	perm  uint32
+	flags int
+}
+
+func newSpecificWriteData(root string, perm fs.FileMode, noSync bool) func(string, []byte) error {
+	flags := unix.O_WRONLY | unix.O_TMPFILE | unix.O_CLOEXEC
+	if !noSync {
+		flags |= unix.O_DSYNC
+	}
+	fd, err := unix.Open(root, flags, uint32(perm))
+	if err != nil {
+		return nil // Which means that OS-specific writeData can't be created and FSTree should use the generic one.
+	}
+	_ = unix.Close(fd) // Don't care about error.
+	w := &linuxWriter{
+		root:  root,
+		perm:  uint32(perm),
+		flags: flags,
+	}
+	return w.writeData
+}
+
+func (w *linuxWriter) writeData(p string, data []byte) error {
+	err := w.writeFile(p, data)
 	if errors.Is(err, unix.ENOSPC) {
 		return common.ErrNoSpace
 	}
 	return err
 }
 
-func (t *FSTree) writeFile(p string, data []byte) error {
-	flags := unix.O_WRONLY | unix.O_TMPFILE | unix.O_CLOEXEC
-	if !t.noSync {
-		flags |= unix.O_DSYNC
-	}
-	fd, err := unix.Open(t.RootPath, flags, uint32(t.Permissions))
+func (w *linuxWriter) writeFile(p string, data []byte) error {
+	fd, err := unix.Open(w.root, w.flags, w.perm)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


pkg/local_object_storage/blobstor/fstree/fstree_write_linux.go
```diff
@@ -4,26 +4,47 @@ package fstree
...
+func newSpecificWriteData(root string, perm fs.FileMode, noSync bool) func(string, []byte) error {
+	flags := unix.O_WRONLY | unix.O_TMPFILE | unix.O_CLOEXEC
+	if !noSync {
+		flags |= unix.O_DSYNC
+	}
+	fd, err := unix.Open(root, flags, uint32(perm))
+	if err != nil {
+		return nil
+	}
+	unix.Close(fd)
```
carpawell, line 29:
```suggestion
	_ = unix.Close(fd)
```
?

pkg/local_object_storage/blobstor/fstree/fstree_write_linux.go
```diff
@@ -4,26 +4,47 @@ package fstree
...
+}
+
+func newSpecificWriteData(root string, perm fs.FileMode, noSync bool) func(string, []byte) error {
+	flags := unix.O_WRONLY | unix.O_TMPFILE | unix.O_CLOEXEC
+	if !noSync {
+		flags |= unix.O_DSYNC
+	}
+	fd, err := unix.Open(root, flags, uint32(perm))
+	if err != nil {
+		return nil
```
carpawell, line 27: can we comment this key difference b/w the old and the new behaviors? it took me some time to understand how exactly this PR fixes #2609

roman-khimov: Four cited lines look wonderful on their own, I agree.

pkg/local_object_storage/blobstor/fstree/fstree_write_linux.go
roman-khimov: Fixed.

carpawell (approved):

PR #2625, opened 2023-10-19
Drop contract group scope

roman-khimov:


pkg/morph/client/nns.go
```diff
@@ -48,7 +43,7 @@ var (
 // NNSAlphabetContractName returns contract name of the alphabet contract in NNS
 // based on alphabet index.
 func NNSAlphabetContractName(index int) string {
-	return "alphabet" + strconv.Itoa(index) + ".neofs"
```
carpawell, line 51: why `.neofs` has left us?

pkg/morph/client/nns.go
```diff
@@ -107,76 +102,29 @@ func nnsResolveItem(c *rpcclient.WSClient, nnsHash util.Uint160, domain string)
...
-		{
-			Type:  smartcontract.ByteArrayType,
-			Value: []byte(domain),
-		},
-	}, nil)
+func exists(nnsReader *nns.ContractReader, domain string) (bool, error) {
+	_, err := nnsReader.OwnerOf([]byte(domain))
+
 	if err != nil {
+		if strings.Contains(err.Error(), "invocation failed") {
```
carpawell: is it the best we can suggest here?

pkg/innerring/innerring.go
```diff
@@ -985,15 +983,13 @@ func (s *Server) createClient(ctx context.Context, p chainParams, errChan chan<-
...
 		return nil, fmt.Errorf("%s chain client endpoints not provided", p.name)
 	}
-
-	return client.New(
-		p.key,
+	var options = []client.Option{
 		client.WithContext(ctx),
 		client.WithLogger(p.log),
-		client.WithDialTimeout(p.cfg.GetDuration(p.name+".dial_timeout")),
+		client.WithDialTimeout(p.cfg.GetDuration(p.name + ".dial_timeout")),
```
carpawell, line 989: go fmt did this? constructor (func args) vs slice creation? interesting

pkg/morph/client/nns.go
```diff
@@ -90,94 +86,64 @@ func (c *Client) NNSHash() (util.Uint160, error) {
...
+		return fmt.Errorf("resolving neofsid: %w", err)
+	}
+
+	conf.signer = &transaction.Signer{
+		Scopes: transaction.CalledByEntry | transaction.Rules,
+		Rules: []transaction.WitnessRule{{
+			Action: transaction.WitnessAllow,
+			Condition: &transaction.ConditionAnd{
+				(*transaction.ConditionCalledByContract)(&cntHash),
+				(*transaction.ConditionScriptHash)(&balanceHash),
```
carpawell, line 126: what error/log/side effect will be seen when a new dev comes and changes a contract and the morph client breaks?

cmd/neofs-adm/internal/modules/morph/initialize.go
```diff
@@ -265,37 +255,44 @@ func (c *initializeContext) nnsContractState() (*state.Contract, error) {
...
 
-	groupKey, err := nnsResolveKey(c.ReadOnlyInvoker, nnsCs.Hash, morphClient.NNSGroupKeyName)
-	if err == nil {
-		c.groupKey = groupKey
+	var nnsReader = nns.NewReader(invoker.New(c.Client, nil), nnsHash)
 
-		signer.Scopes = transaction.CustomGroups
-		signer.AllowedGroups = keys.PublicKeys{groupKey}
+	balanceHash, err := nnsReader.ResolveFSContract(nns.NameBalance)
+	if err != nil && c.Contracts[balanceContract] != nil {
```
carpawell, line 276: but what if `err != nil` and no hash has been cached?

cmd/neofs-adm/internal/modules/morph/initialize.go
```diff
@@ -369,34 +366,36 @@ loop:
 }
 
 // sendCommitteeTx creates transaction from script, signs it by committee nodes and sends it to RPC.
-// If tryGroup is false, global scope is used for the signer (useful when
-// working with native contracts).
-func (c *initializeContext) sendCommitteeTx(script []byte, tryGroup bool) error {
-	return c.sendMultiTx(script, tryGroup, false)
+// If fancyScope is true, universal NeoFS scope is used for the signer (useful when
```
carpawell, line 369: is it still "useful when working with native contracts"?

carpawell:
Conflicts.

pkg/morph/client/nns.go
roman-khimov: Because https://pkg.go.dev/github.com/nspcc-dev/neofs-contract/rpc/nns#ContractReader.ResolveFSContract includes it.

pkg/morph/client/nns.go
roman-khimov: Until https://github.com/nspcc-dev/neo-go/issues/3130 yes.

pkg/innerring/innerring.go
roman-khimov: Seems to be like fmt, yes.

pkg/morph/client/nns.go
roman-khimov: Some invocation will fail, likely even before it's created (test invocation), therefore it'll be seen pretty quickly.

cmd/neofs-adm/internal/modules/morph/initialize.go
roman-khimov: It means that we don't know the hash and it stays zero. In this case it's irrelevant, we don't have data from contract, we don't have data from the network, let's hope it works, but if it doesn't we can't help much.

cmd/neofs-adm/internal/modules/morph/initialize.go
roman-khimov: Not really, fixed.

pkg/morph/client/nns.go
carpawell: so this is not `contract name of the alphabet contract in NNS based on alphabet index.` right now?

cmd/neofs-adm/internal/modules/morph/initialize.go
carpawell: passing zero hash and continuing working scares me. but ok

carpawell (approved):
I would wait for @AnnaShaleva's review.

pkg/morph/client/nns.go
roman-khimov: It's still a contract name. Not an FQDN, but a name.

carpawell:
> I would wait for @AnnaShaleva's review.

Ok, not this time.

PR #2629, opened 2023-10-25
upd/HRW library

carpawell:


pkg/local_object_storage/engine/evacuate.go
```diff
@@ -96,11 +96,6 @@ func (e *StorageEngine) Evacuate(prm EvacuateShardPrm) (EvacuateShardRes, error)
 	}
 	e.mtx.RUnlock()
 
-	weights := make([]float64, 0, len(shards))
-	for i := range shards {
-		weights = append(weights, e.shardWeight(shards[i].Shard))
```
roman-khimov, line 101: What if we have different capacity for every shard?

roman-khimov: We better keep the functionality as is then and create an issue to solve it in the future. We will need this, shard can be put into/retrieved from the metabase.

carpawell: > shard can be put into/retrieved from the metabase.

if meta is _not per shard_ (not the current case)

roman-khimov: Sure, sure.

carpawell: > But the weighting itself should be kept.

you mean this PR? this code is currently broken for sure: it always uses `0` and confuses. i suggest returning weighting (reworking in fact, the current version can't be used at all) when we need it and when we are ready for it

roman-khimov: Why do you think it needs a complete rework?

carpawell: 1. it is not used and not usable. i mean totally. it is commented as `Amount of free disk space. Measured in kilobytes.` and this field is never changed (always zero). and never was changed from the beginnings
2. shard currently does not know how much it stores. just no code uses it, nobody ever asked about it. some estimation metrics per container, some calculations inside bbolt's transactions, some internal changes by background workers, and so on. all of it does not use this field (see `1.`)
3. if we want it to be used as a weight, it will be the first storage use case, so _meta per shard_ should be reworked

i do not mean weighting is not hard to implement, but some things should be done first, and currently it just does wrong things

roman-khimov (approved):

pkg/local_object_storage/engine/shards.go
```diff
@@ -276,7 +268,7 @@ func (e *StorageEngine) HandleNewEpoch(epoch uint64) {
 }
 
 func (s hashedShard) Hash() uint64 {
-	return hrw.Hash(
+	return murmur3.Sum64(
```
carpawell, line 271: @roman-khimov, so what do you think about it? is it OK to have `hrw` lib without `Hash` func?

roman-khimov: I've seen that. You can wrap murmur3 into `Hash`, but you can use just about any other hash as well. Providing some default can be useful.

roman-khimov: Will you update it to the new `hrw` with `Hash()`?

carpawell: sure. done

PR #2635, opened 2023-10-31
upd/Neo-go to v0.103.0

carpawell:


cmd/neofs-adm/internal/modules/morph/initialize_transfer.go
```diff
@@ -144,10 +141,12 @@ func (c *initializeContext) transferGASToProxy() error {
 		return err
 	}
 
-	tx, err := createNEP17MultiTransferTx(c.Client, c.CommitteeAcc, []rpcclient.TransferTarget{{
-		Token:   gas.Hash,
-		Address: proxyCs.Hash,
-		Amount:  initialProxyGASAmount,
+	from := c.CommitteeAcc.Contract.ScriptHash()
+
+	tx, err := createNEP17MultiTransferTx(c.Client, c.CommitteeAcc, gas.Hash, []nep17.TransferParameters{{
```
AnnaShaleva, line 146: Maybe then `createNEP17MultiTransferTx` can be removed completely? And you transfer it from the committee account anyway, so you can use committee actor directly from `c.CommitteeAct` without creating simple actor. Is there any pitfalls?

AnnaShaleva (commented):
Is there any other places where `emit` NeoGo package is uses for custom scripts creation instead of `smartcontract`'s Builder? If so, then we should also adjust them (if possible).

cmd/neofs-adm/internal/modules/morph/initialize_transfer.go
carpawell: > Maybe then createNEP17MultiTransferTx can be removed completely?

can be done, ok

> Is there any pitfalls?

do not think there are any

carpawell:
> Is there any other places where emit NeoGo package is uses for custom scripts creation instead of smartcontract's Builder? If so, then we should also adjust them (if possible).

There are some for sure. @roman-khimov, do we need it? How long `neofs-adm` gonna live (or be unchanged) after we merge auto deploy?

roman-khimov:
`neofs-adm` gives much more abilities than just a deploy and we need some low-level instruments as well, so it won't disappear tomorrow.

cmd/neofs-adm/internal/modules/morph/initialize_transfer.go
AnnaShaleva: If we've agreed to implement Builder-related changes in a separate PR, then let's fix this issue, and otherwise LGTM.

carpawell: > Maybe then createNEP17MultiTransferTx can be removed completely?

fixed

carpawell:
> neofs-adm gives much more abilities than just a deploy and we need some low-level instruments as well

`neofs-amd` without deploying commands is `neofs-cli` to me.

> If so, then we should also adjust them (if possible).

Done.

roman-khimov (approved):

cmd/neofs-adm/internal/modules/morph/balance.go
```diff
@@ -126,17 +124,17 @@ func dumpBalances(cmd *cobra.Command, _ []string) error {
 	if dumpAlphabet {
 		alphaList := make([]accBalancePair, len(irList))
 
-		w := io.NewBufBinWriter()
+		b := smartcontract.NewBuilder()
 		for i := range alphaList {
-			emit.AppCall(w.BinWriter, nnsCs.Hash, "resolve", callflag.ReadOnly,
-				getAlphabetNNSDomain(i),
-				int64(nns.TXT))
+			b.InvokeMethod(nnsCs.Hash, "resolve", getAlphabetNNSDomain(i), int64(nns.TXT))
```
AnnaShaleva, line 129: BTW, builder invokes this with `callflag.All`, and it was ReadOnly before. But anyway, it's our contract and there's no harm in it.

cmd/neofs-adm/internal/modules/morph/balance.go
```diff
@@ -126,17 +124,17 @@ func dumpBalances(cmd *cobra.Command, _ []string) error {
...
 		}
-		if w.Err != nil {
-			panic(w.Err)
+
+		script, err := b.Script()
+		if err != nil {
+			return fmt.Errorf("resolbing alphabet hashes script: %w", err)
```
AnnaShaleva, line 134: s/resolbing/resolving

AnnaShaleva (commented):
LGTM otherwise.

cmd/neofs-adm/internal/modules/morph/balance.go
carpawell: oh, yeah, forgot to ping you about that. @roman-khimov, yes, in some cases there were readonly calls for readonly operations but nothing should break because of it. tbh, i really do not see much difference b/w `InvokeMethod` and `AppCall` except the call flags (and sometimes we need them)

cmd/neofs-adm/internal/modules/morph/balance.go
carpawell: fixed

cmd/neofs-adm/internal/modules/morph/balance.go
AnnaShaleva: I think it's OK to have this RO -> ALL replacement in Adm.

AnnaShaleva (approved):

cmd/neofs-adm/internal/modules/morph/balance.go
roman-khimov: I've seen this, it doesn't matter much for these calls, most of them are test invocations only.

PR #2638, opened 2023-11-09
feat/Clean up engine after container removal

carpawell:


pkg/local_object_storage/metabase/containers.go
```diff
@@ -104,3 +107,113 @@ func changeContainerSize(tx *bbolt.Tx, id cid.ID, delta uint64, increase bool) e
...
+			return fmt.Errorf("parent index cleanup: %w", err)
+		}
+
+		err = tx.DeleteBucket(splitBucketName(cID, buff))
+		if err != nil && !errors.Is(err, bbolt.ErrBucketNotFound) {
+			return fmt.Errorf("split id index cleanup: %w", err)
+		}
+
+		// Attributes index
+		var keysToDelete [][]byte // see https://github.com/etcd-io/bbolt/issues/146
```
roman-khimov, line 203: How many of those can we expect?

pkg/local_object_storage/metabase/get.go
```diff
@@ -170,3 +171,98 @@ func getSplitInfoError(tx *bbolt.Tx, cnr cid.ID, key []byte) error {
...
 	return logicerr.Wrap(apistatus.ObjectNotFound{})
 }
+
+// GetContainer returns objects stored in the metabase that
+// belong to the provided container.
+// Note: metabase can store information about a locked object,
+// but it will not be included to the result if the object is
+// not stored in the metabase (in other words, no information
+// in the regular objects index).
+func (db *DB) GetContainer(cID cid.ID) ([]oid.Address, error) {
```
roman-khimov, line 181: All 1M objects? Doesn't seem to be a good idea to me.

pkg/local_object_storage/shard/container.go
```diff
@@ -39,3 +45,86 @@ func (s *Shard) ContainerSize(prm ContainerSizePrm) (ContainerSizeRes, error) {
...
+				}
+			}
+
+			return nil
+		})
+	}
+
+	wg.Go(func() error {
+		for _, obj := range objs {
+			_, err := s.blobStor.Delete(common.DeletePrm{Address: obj})
```
roman-khimov, line 81: Why can't we just mark them for deletion by GC?

pkg/local_object_storage/metabase/containers.go
carpawell: any number. any container can be dropped and it may have this or that number

cmd/neofs-node/config.go
```diff
@@ -750,6 +751,15 @@ func initLocalStorage(c *cfg) {
 		ls.HandleNewEpoch(ev.(netmap2.NewEpoch).EpochNumber())
 	})
 
+	subscribeToContainerRemoval(c, func(e event.Event) {
+		ev := e.(containerEvent.DeleteSuccess)
+
+		err := ls.DeleteContainer(c.ctx, ev.ID)
+		if err != nil {
+			c.log.Warn("could not delete container", zap.Stringer("cID", ev.ID), zap.Error(err))
```
cthulhu-rider, line 763: message reflects the actual action too vaguely: we dont delete the container, we update local object storage after the container removal event. Otherwise, we react to `delete container` event with `delete container`

this only confirms the dubiousness of the method name

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Changelog for NeoFS Node
 - Policer's setting to the SN's application configuration (#2600)
 - Support of verified domains for the storage nodes (#2280)
 - `neofs-lens storage status` CLI command (#2550)
+- Missing container cleanup (#1663)
```
cthulhu-rider, line 10: `containers`

pkg/local_object_storage/engine/engine.go
```diff
@@ -33,6 +34,11 @@ type StorageEngine struct {
 	}
 }
 
+// ContainersSource must return actual container list in the map form.
```
cthulhu-rider, line 37: interface is defined in the form of a requirement - looks very strange. I'd say
```
// ContainerSource is a source of information about NeoFS containers.
```
and
```
// GetContainers returns a list of containers for which data is subject to further storage.
```

pkg/local_object_storage/engine/control.go
```diff
@@ -39,6 +39,11 @@ func (e *StorageEngine) Init() error {
 		}
 	}
 
+	err := e.deleteNotFoundContainers()
```
cthulhu-rider, line 42: lets show it in engine docs, e.g.
```
// If [WithContainersSource] is used, StorageEngine deletes all stored containers outside [ContainersSource.GetContainers] set.
```

pkg/local_object_storage/shard/container.go
```diff
@@ -39,3 +45,86 @@ func (s *Shard) ContainerSize(prm ContainerSizePrm) (ContainerSizeRes, error) {
...
+// - Write-cache (if configured).
+func (s *Shard) DeleteContainer(_ context.Context, cID cid.ID) error {
+	s.m.RLock()
+	defer s.m.RUnlock()
+
+	var wg errgroup.Group
+
+	objs, err := s.metaBase.GetContainer(cID)
+	if err != nil {
+		return fmt.Errorf("fetching '%s' container objects: %w", cID, err)
```
cthulhu-rider, line 63: no need to add func input to err context (same below)
```suggestion
		return fmt.Errorf("fetching container objects: %w", err)
```

pkg/local_object_storage/shard/container.go
```diff
@@ -39,3 +45,86 @@ func (s *Shard) ContainerSize(prm ContainerSizePrm) (ContainerSizeRes, error) {
...
+	}
+
+	if s.cfg.useWriteCache {
+		wg.Go(func() error {
+			for _, obj := range objs {
+				err := s.writeCache.Delete(obj)
+				if err != nil && !errors.As(err, new(apistatus.ObjectNotFound)) {
+					return fmt.Errorf("removing %s obj from write-cache: %w", cID, err)
```
cthulhu-rider, line 71:
```suggestion
					return fmt.Errorf("removing %s obj from write-cache: %w", obj, err)
```

pkg/local_object_storage/shard/container.go
```diff
@@ -39,3 +45,86 @@ func (s *Shard) ContainerSize(prm ContainerSizePrm) (ContainerSizeRes, error) {
...
+
+			return nil
+		})
+	}
+
+	wg.Go(func() error {
+		for _, obj := range objs {
+			_, err := s.blobStor.Delete(common.DeletePrm{Address: obj})
+			if err != nil && !errors.As(err, new(apistatus.ObjectNotFound)) {
+				return fmt.Errorf("removing %s obj from blobstor: %w", cID, err)
```
cthulhu-rider, line 83:
```suggestion
				return fmt.Errorf("removing %s obj from blobstor: %w", obj, err)
```

pkg/util/convert.go
```diff
@@ -0,0 +1,22 @@
...
+func MapToSlice[V comparable](m map[V]struct{}) []V {
+	res := make([]V, 0, len(m))
+	for v := range m {
+		res = append(res, v)
+	}
+
+	return res
+}
+
+// SliceToMap converts any slice with comparable elements to a unique filter map.
```
cthulhu-rider, line 14: rly need `unique filter` words?

pkg/util/convert.go
```diff
@@ -0,0 +1,22 @@
+package util
+
+// MapToSlice converts the unique map's key to a slice. Order is not
```
cthulhu-rider, line 3: map keys are always unique

cthulhu-rider (commented):
3rd commit: `Contract container` -> `Container contract`

pkg/local_object_storage/metabase/containers.go
```diff
@@ -104,3 +107,113 @@ func changeContainerSize(tx *bbolt.Tx, id cid.ID, delta uint64, increase bool) e
...
+	}
+
+	cIDRaw := make([]byte, cidSize)
+	cID.Encode(cIDRaw)
+	buff := make([]byte, addressKeySize)
+
+	return db.boltDB.Update(func(tx *bbolt.Tx) error {
+		// Estimations
+		bktEstimations := tx.Bucket(containerVolumeBucketName)
+		err := bktEstimations.Delete(cIDRaw)
```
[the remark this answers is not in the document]
carpawell: i think yes. we are expect this bucket to exist anyway: https://github.com/nspcc-dev/neofs-node/blob/bbfbd9bce59a09c99e61d6b937ae18fcab847e4b/pkg/local_object_storage/metabase/control.go#L103

pkg/local_object_storage/metabase/containers.go
```diff
@@ -104,3 +107,113 @@ func changeContainerSize(tx *bbolt.Tx, id cid.ID, delta uint64, increase bool) e
...
+		// SG objects
+		err = tx.DeleteBucket(storageGroupBucketName(cID, buff))
+		if err != nil && !errors.Is(err, bbolt.ErrBucketNotFound) {
+			return fmt.Errorf("SG bucket cleanup: %w", err)
+		}
+
+		// TS objects
+		err = tx.DeleteBucket(tombstoneBucketName(cID, buff))
+		if err != nil && !errors.Is(err, bbolt.ErrBucketNotFound) {
+			return fmt.Errorf("TS bucket cleanup: %w", err)
```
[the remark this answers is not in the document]
carpawell: do not see real differences but ok

cmd/neofs-node/config.go
carpawell: added some details to the message

pkg/util/convert.go
carpawell: why not? removed

pkg/local_object_storage/metabase/containers.go
```diff
@@ -104,3 +107,113 @@ func changeContainerSize(tx *bbolt.Tx, id cid.ID, delta uint64, increase bool) e
 
 	return containerVolume.Put(key, buf)
 }
+
+// DeleteContainer removes any information that the metabase has
+// associated with the provided container (its objects).
+//
+// Note: does not clean graveyard.
+func (db *DB) DeleteContainer(cID cid.ID) error {
```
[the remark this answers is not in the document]
cthulhu-rider: > it is designed to delete all the container info

```suggestion
func (db *DB) DeleteContainerInfo(cID cid.ID) error {
```
?

pkg/util/convert.go
cthulhu-rider: i mean map is always unique

pkg/local_object_storage/metabase/graveyard.go
```diff
@@ -259,3 +260,104 @@ func (db *DB) DropGraves(tss []TombstonedObject) error {
...
+				return fmt.Errorf("parsing deleted address: %w", err)
+			}
+
+			if _, handled := alreadyHandledContainers[addrBuff.Container()]; handled {
+				continue
+			}
+
+			resObjects = append(resObjects, addrBuff)
+
+			if len(resObjects) >= limit {
```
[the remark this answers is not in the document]
carpawell: agree. but i am too afraid of not having any limits (in some bug scenarios if more than one object was appended). i feel more confident this way. done it on purpose

pkg/local_object_storage/metabase/graveyard.go
```diff
@@ -259,3 +260,104 @@ func (db *DB) DropGraves(tss []TombstonedObject) error {
...
+
+			addrBuff.SetContainer(cidBuff)
+			for obj := range uniqueObjectsMap {
+				addrBuff.SetObject(obj)
+				resObjects = append(resObjects, addrBuff)
+			}
+
+			alreadyHandledContainers[cidBuff] = struct{}{}
+
+			if len(resObjects) < limit {
```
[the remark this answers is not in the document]
carpawell: you mean it would save us if `listContainerObjects` is buggy? it is an unexported func with its own `limit` arg, i think we should not duplicate checks because of that

pkg/local_object_storage/metabase/graveyard.go
```diff
@@ -259,3 +260,104 @@ func (db *DB) DropGraves(tss []TombstonedObject) error {
...
+
+	db.modeMtx.RLock()
+	defer db.modeMtx.RUnlock()
+
+	if db.mode.NoMetabase() {
+		return nil, nil, ErrDegradedMode
+	}
+
+	const reasonableLimit = 1000
+	initLimit := reasonableLimit
```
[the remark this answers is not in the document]
carpawell: fixed

pkg/local_object_storage/metabase/get.go
```diff
@@ -173,12 +174,13 @@ func getSplitInfoError(tx *bbolt.Tx, cnr cid.ID, key []byte) error {
 }
 
 // ListContainerObjects returns objects stored in the metabase that
-// belong to the provided container.
+// belong to the provided container. No more than limit objects per
```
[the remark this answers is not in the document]
carpawell: > expected?

no, i would say. make them useless and added this info to the comment

pkg/local_object_storage/metabase/get.go
```diff
@@ -187,60 +189,92 @@ func (db *DB) ListContainerObjects(cID cid.ID) ([]oid.ID, error) {
...
-		return nil
-	})
+		return fmt.Errorf("small objects iteration: %w", err)
+	}
+
+	bktRoot := tx.Bucket(rootBucketName(cID, buff))
+	err = expandObjectsFromBucket(bktRoot, unique, limit)
 	if err != nil {
-		return nil, err
+		if errors.Is(err, errBreakIter) {
```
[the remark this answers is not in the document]
carpawell: you mean to return some bool flag? or check len again instead of err? why do you think it would be better?

pkg/local_object_storage/metabase/graveyard.go
```diff
@@ -259,3 +260,104 @@ func (db *DB) DropGraves(tss []TombstonedObject) error {
...
+		c := bkt.Cursor()
+
+		for k, _ := c.First(); k != nil; k, _ = c.Next() {
+			err := cidBuff.Decode(k)
+			if err != nil {
+				return fmt.Errorf("parsing raw CID: %w", err)
+			}
+
+			// another container, clean the map
+			uniqueObjectsMap = make(map[oid.ID]struct{}, initLimit)
```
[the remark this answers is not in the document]
cthulhu-rider:
```go
if uniqueObjectsMap == nil {
	uniqueObjectsMap = make(map[oid.ID]struct{}, initLimit)
} else {
	for k := range uniqueObjectsMap {
		delete(uniqueObjectsMap, k)
	}
}
```

> why not just GC it?

then why we try to preallocate slices and reuse them if they have enough capacity? Ok, `s = s[:0]` is much shorter, but still

and, as u mentioned, we may later use `clear` here. If we leave `uniqueObjectsMap = make(map[oid.ID]struct{}, initLimit)` we'll forget to do this

pkg/local_object_storage/metabase/get.go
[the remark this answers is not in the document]
cthulhu-rider: i'd check len cuz it's strange to me to see that function returned an error when nicely done its purpose. But current approach also works (but i'd name service error smth like `errComplete` cuz `errBreakIter` is absolutely unexpected to see according to the function signature). Leave up to u

pkg/local_object_storage/metabase/graveyard.go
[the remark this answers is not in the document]
cthulhu-rider: no, i missed that `listContainerObjects` limit is updated with `resObjects ` progress, so my comment is irrelevant

besides this: can't we add `cidBuff` to the result if `len(resObjects) == limit`?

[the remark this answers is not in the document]
carpawell: in the current design -- no. we cant say for sure if it is a limit restriction (and we have more objects but we wont get them for this iteration) or we just dont have any more objects. can be fixed a little with some bool flags and more complex code but i do not think it is worth it: that should be a really rare case with no bad affects (container will be deleted just in the next iteration)

pkg/local_object_storage/metabase/graveyard.go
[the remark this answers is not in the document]
carpawell: fixed

cthulhu-rider (approved):

roman-khimov:
Wait for 0.39.0 release.

roman-khimov:
Conflicts, rebase pls.

roman-khimov (approved):

Issue #2612, opened 2023-10-04
Do not change node attributes in IR


roman-khimov:
After #2611 it'll finally be possible to not change SN attributes by IR, so check them (if locode is specified, then all the other attributes should match it), but do not change them. Make sure neofs-cli can produce a set of attributes for a node based on locode (for admins to fix their configurations).

roman-khimov:
This also allows to not override SN bootstrap notary request, rather just completing it with IR signatures if it's OK. Can be done here as well.

Issue #2279, opened 2023-03-15
Configurable and extendable IR node validation checks


roman-khimov:
## Is your feature request related to a problem? Please describe.
Current node validation logic mixes logical correctness and policy checks. For example, address validation is just a correctness checks, same thing for maintenance mode. However locode check is a policy, networks may want to do it or may not want to. They at the same time may want to check some network-specific attributes according to any network-specific logic.

## Describe the solution you'd like
Make embedded policy checks configurable and provide a generic extension hook to check node data via an external program.

## Describe alternatives you've considered
Embedded ones can be moved out of the node as well, but maybe that'd make them a bit less convenient to use.

PR #2648, opened 2023-11-15
feat/Move locode attrs expansion on SN side

carpawell:


cmd/neofs-node/attributes.go
```diff
@@ -11,4 +15,52 @@ func parseAttributes(c *cfg) {
...
+	}
+	if subDivName := record.SubDivName; subDivName != "" {
+		setLocodeAttr(n, "SubDiv", subDivName)
+	}
+}
+
+func getRecord(lc string) (*locodedb.Key, locodedb.Record, error) {
+	country, location := lc[:2], lc[2:]
+	if lc[2] == ' ' {
+		location = lc[3:]
```
roman-khimov, line 49: Doing these dances just for `locodedb.NewKey()` is no fun.

cmd/neofs-node/attributes.go
```diff
@@ -11,4 +15,52 @@ func parseAttributes(c *cfg) {
...
+	}
+
+	key, record, err := getRecord(locAttr)
+	if err != nil {
+		fatalOnErr(fmt.Errorf("could not get locode record from DB: %w", err))
+	}
+
+	n := &c.cfgNodeInfo.localInfo
+
+	setLocodeAttr(n, "CountryCode", key.CountryCode().String())
```
roman-khimov, line 34: The way I saw it in #2612 is:
 * `./neofs-cli util locode get-attributes "BE RUS"`
 * copy-paste into config and run

But this code auto-expands one attribute into multiple attributes (and requires some memory on the SN side!).

Now I'm not sure which way is better. Having all parameters listed explicitly in the config with no auto-magic seems to be a bit more robust and predictable, but have some magic already with version and capacity, so maybe it's still OK if documented properly. @532910?

pkg/innerring/processors/netmap/nodevalidation/locode/calls_test.go
```diff
@@ -70,18 +70,18 @@ func TestValidator_VerifyAndUpdate(t *testing.T) { // test record with valid but
...
+			n.SetAttribute("SubDiv", r.SubDivName)
+			n.SetAttribute("Continent", r.Cont.String())
 
-		require.Equal(t, k.CountryCode().String(), n.Attribute("CountryCode"))
-		require.Equal(t, r.Country, n.Attribute("Country"))
-		require.Equal(t, r.Location, n.Attribute("Location"))
-		require.Equal(t, r.SubDivCode, n.Attribute("SubDivCode"))
-		require.Equal(t, r.SubDivName, n.Attribute("SubDiv"))
-		require.Equal(t, r.Cont.String(), n.Attribute("Continent"))
+			require.NoError(t, validator.VerifyAndUpdate(n))
```
roman-khimov: So tests that fail on attributes also?

cmd/neofs-node/attributes.go
carpawell: > requires some memory on the SN side

agree. that is why i also thought about cleaning memory in the locode lib (a separate call that makes it unusable)

> Now I'm not sure which way is better.

me neither. but i am sure that it is more convenient for a user (taking into account that some attrs are already doing such a magic expansion)

> Having all parameters listed explicitly in the config with no auto-magic seems to be a bit more robust and predictable

and an administrator can do it himself, that is why user's value are prioritized in this PR

cmd/neofs-node/attributes.go
carpawell: copy-pasted from the approved PR

pkg/innerring/processors/netmap/nodevalidation/locode/calls_test.go
carpawell: done

cmd/neofs-node/attributes.go
roman-khimov: > cleaning memory in the locode lib (a separate call that makes it unusable)

I had this thought as well. And we can also use an mmaped tmp file for unpacked data (that can be swapped out).

pkg/innerring/processors/netmap/processor.go
```diff
@@ -38,15 +38,10 @@ type (
...
+		// Verify must verify NodeInfo structure.
 		//
 		// Must return an error if NodeInfo input is invalid.
-		// Must return an error if it is not possible to correctly
-		// change the structure for sending to the network map.
-		//
-		// If no error occurs, the parameter must point to the
-		// ready-made NodeInfo structure.
-		VerifyAndUpdate(*netmap.NodeInfo) error
+		Verify(*netmap.NodeInfo) error
```
cthulhu-rider, line 44: then we should no longer use pointer

cthulhu-rider, line 44: then we should no longer use pointer

cmd/neofs-node/attributes.go
```diff
@@ -11,4 +15,52 @@ func parseAttributes(c *cfg) {
 	}
 
 	fatalOnErr(attributes.ReadNodeAttributes(&c.cfgNodeInfo.localInfo, nodeconfig.Attributes(c.appCfg)))
+
+	// expand UN/LOCODE attribute if any found; keep user's attributes
+	// if any conflicts appear
+
+	locAttr := c.cfgNodeInfo.localInfo.LOCODE()
+	if locAttr == "" {
+		return
```
cthulhu-rider, line 24: lets changelog that storage nodes no longer obliged to set LOCODE attribute from now

cthulhu-rider (commented):
3rd commit body contains backslash, last one - imperfect grammar

cmd/neofs-node/attributes.go
roman-khimov: After additional discussions, let's leave it as is:
 * with nspcc-dev/locode-db#28 it's just 4M overhead and we manage at least of number of 64M objects
 * it makes configuration 100% compatible
 * it's a bit easier to manage
 * it's friendly to configuration reloads (`Close()` is not)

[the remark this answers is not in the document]
carpawell: > https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go@v1.0.0-rc.11/netmap#NodeInfo.SetCountryCode fits well for this

cmd/neofs-node/config.go
```diff
@@ -819,21 +819,15 @@ func initObjectPool(cfg *config.Config) (pool cfgObjectRoutines) {
...
-	} else {
-		c.cfgNodeInfo.localInfo.WriteToV2(&res)
-	}
+	c.cfgNodeInfo.localInfo.WriteToV2(&res)
 
 	return &res, nil
 }
 
-// handleLocalNodeInfo rewrites local node info from the NeoFS network map.
+// handleNetworkNodeInfo rewrites chain node info from the NeoFS network map.
```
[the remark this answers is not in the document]
carpawell: > chain confuses me, how about cached?

fixed

> and why ur removing local?

we have both local and network state vars. local for read config (the var is named `local*`) and this func (named `handle*`) is called when we get info from the network

pkg/innerring/processors/netmap/nodevalidation/locode/calls.go
```diff
@@ -25,16 +25,38 @@ func (v *Validator) VerifyAndUpdate(n *netmap.NodeInfo) error {
 		return fmt.Errorf("could not get locode record from DB: %w", err)
 	}
 
-	n.SetCountryCode(key.CountryCode().String())
-	n.SetCountryName(record.Country)
-	n.SetLocationName(record.Location)
-	n.SetContinentName(record.Cont.String())
-	if subDivCode := record.SubDivCode; subDivCode != "" {
-		n.SetSubdivisionCode(subDivCode)
+	err = checkAttribute(n, "CountryCode", key.CountryCode().String())
```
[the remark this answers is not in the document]
carpawell: what getters?

pkg/innerring/processors/netmap/nodevalidation/locode/calls.go
```diff
@@ -25,16 +25,38 @@ func (v *Validator) VerifyAndUpdate(n *netmap.NodeInfo) error {
...
+
+	return nil
+}
 
-	if subDivName := record.SubDivName; subDivName != "" {
-		n.SetSubdivisionName(subDivName)
+func checkAttribute(n *netmap.NodeInfo, key, expectedVal string) error {
+	val := n.Attribute(key)
+	if val != expectedVal {
+		return fmt.Errorf("unexpected '%s' attribute value: want '%s', got '%s'", key, expectedVal, val)
```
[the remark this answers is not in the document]
carpawell: fixed

cmd/neofs-node/attributes.go
carpawell: why do you think this is a _change_?

cmd/neofs-node/attributes.go
```diff
@@ -11,4 +15,52 @@ func parseAttributes(c *cfg) {
...
+	if subDivCode := record.SubDivCode; subDivCode != "" {
+		setLocodeAttr(n, "SubDivCode", subDivCode)
+	}
+	if subDivName := record.SubDivName; subDivName != "" {
+		setLocodeAttr(n, "SubDiv", subDivName)
+	}
+}
+
+func getRecord(lc string) (*locodedb.Key, locodedb.Record, error) {
+	country, location := lc[:2], lc[2:]
```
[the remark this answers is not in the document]
carpawell: that is not my code but added

cmd/neofs-node/attributes.go
```diff
@@ -11,4 +15,52 @@ func parseAttributes(c *cfg) {
...
+	if lc[2] == ' ' {
```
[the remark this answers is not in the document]
carpawell: that is not my code but fixed

cmd/neofs-node/attributes.go
roman-khimov: You can add some `locodedb.CodeToKey(string)`. It's always to possible to do something better than the nearest PR. And we should be doing that.

roman-khimov (approved):

cmd/neofs-node/attributes.go
carpawell: > You can add some locodedb.CodeToKey(string)

you mean locode repo?

> It's always to possible to do something better than the nearest PR.

i agree but do not know what we can do here. removed that check and used country code constants instead

roman-khimov: > you mean locode repo?

Sure.

carpawell: 5 char len string to some type to 2 char len string (same but shorter) then?

roman-khimov: OK this way too, you get the data first which ensures that locode is correct and its first two bytes can be used this way safely.

cmd/neofs-node/attributes.go
cthulhu-rider: cuz they were obliged before

cmd/neofs-node/attributes.go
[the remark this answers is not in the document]
cthulhu-rider: > does not allow "set if empty"

yep, and that's perfect. Lets check emptiness of these fields. All required `record` fields must not be empty (`subDivCode` and `subDivName` are checked explicitly anyway)

cmd/neofs-node/config.go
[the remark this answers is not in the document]
cthulhu-rider: reasonable, so what u think about `handleLocalNodeInfoFromNetwork/handleLocalNodeInfoFromConfig` naming pattern? So we dont collide b/w local/remote node and local/network info

pkg/innerring/processors/netmap/nodevalidation/locode/calls.go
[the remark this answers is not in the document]
cthulhu-rider: oh, there were no getters until https://github.com/nspcc-dev/neofs-sdk-go/pull/534/. Now we have them, so lets use

cmd/neofs-node/attributes.go
cthulhu-rider: nope, they werent bcfd33d9fb1476b2555b807c476d8e951c54ace9
my bad sry

cthulhu-rider: ...but now we have https://github.com/nspcc-dev/neofs-sdk-go/blob/d72d76ccb0baa0ed61b32a214ee67997661303de/netmap/node_info.go#L317-L318

finally: LOCODE attrs' absence is valid or invalid?

cmd/neofs-node/attributes.go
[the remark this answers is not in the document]
carpawell: > Lets check emptiness of these fields.

not possible cause there are no getters in the sdk

> must not be empty

but they can be empty according to https://github.com/nspcc-dev/neofs-node/pull/2648#discussion_r1398962507
but if they are not, we decided to leave them as is to allow overwriting (if a user wants so)

> subDivCode and subDivName are checked explicitly anyway

did not think much about it, just moved checks on the SN side (IR did this before, SN does this now, just copy-pasted)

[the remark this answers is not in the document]
cthulhu-rider: > not possible cause there are no getters in the sdk

im about `record`

> but they can be empty according to https://github.com/nspcc-dev/neofs-node/pull/2648#discussion_r1398962507

ok, anyway
```
countryCode := key.CountryCode().String()
if countryCode != "" {
  n.SetCountryCode(countryCode)
}
```
we see that code may be empty, use SDK and follow its requirements. And dont care that country code is some attribute

cmd/neofs-node/attributes.go
carpawell: > finally: LOCODE attrs' absence is valid or invalid?

need to be decided. you specified the commit, you mentioned the SDK getter and i can add some words about "well-known" attrs from the API that i am not sure i can understand clearly: https://github.com/nspcc-dev/neofs-api/blob/1424834f743842b88e02ec6ac79c64ef2d2d3e75/netmap/types.proto#L156-L158

@roman-khimov, lets fight

P.S.

> but now we have

not now, for a year (before the refactor, and maybe even more in fact)

cmd/neofs-node/attributes.go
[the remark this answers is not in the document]
carpawell: > im about record

you mean setting only if we have non-empty values in our db?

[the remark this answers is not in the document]
cthulhu-rider: > > im about record
> 
> you mean setting only if we have non-empty values in our db?

sry typo'ed

yeah, just like in code from prev comment. Or with config prevalence (like in current PR):
```go
if n.CoutryCode() == "" {
  countryCode := key.CountryCode().String()
  if countryCode != "" {
    n.SetCountryCode(countryCode)
  } 
}
```

cmd/neofs-node/attributes.go
cthulhu-rider: #2612 requires all attributes match the DB record when LOCODE is specified. So lets adjust SDK docs and storage node code to this requirement at the moment. Btw in current PR storage node doesn't check config settings against the DB

cmd/neofs-node/attributes.go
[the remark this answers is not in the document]
carpawell: done but TBW not sure that is a correct way, mb setting empty string if we do not have it in our base is a better way. or not

pkg/innerring/processors/netmap/nodevalidation/locode/calls.go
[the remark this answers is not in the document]
carpawell: done

cmd/neofs-node/attributes.go
carpawell: > Btw in current PR storage node doesn't check config settings against the DB

that is the idea. if a user wants to place some attrs, he can do it

> https://github.com/nspcc-dev/neofs-node/issues/2612 requires all attributes match the DB record when LOCODE is specified

on the IR side. but what if a user wants Chelyabinsk to be in USA and does everything he can as an admin (config)?

> So lets adjust SDK docs and storage node code to this requirement at the moment.

so what behavior do you want? cannot fix what i cannot understand

roman-khimov: It's a policy thing that can vary from network to network (hi, #2279), for public networks it's more like mandatory, in private ones no one cares. But until we have #2279 we better have it optional.

carpawell: ok, do we need a warning at least?

roman-khimov: You can add a warning.

carpawell: done

cthulhu-rider (approved):

cmd/neofs-node/attributes.go
```diff
@@ -31,27 +32,69 @@ func parseAttributes(c *cfg) {
...
+
+	fromUser = n.SubdivisionName()
+	if fromUser != "" && fromUser != record.SubDivName {
+		c.log.Warn("locode attribute mismatch, configuration value kept: subdivision name",
+			zap.String("configuration_value", fromUser),
+			zap.String("database_value", record.SubDivName))
+	} else {
+		setIfNotEmpty(n.SetSubdivisionName, record.SubDivName)
 	}
+
```
roman-khimov, line 88: Meh.

carpawell: took into account, analyzed, researched and finally fixed

roman-khimov: It made linter red!

carpawell: yeah, i missed that, thanks

Issue #2585, opened 2023-09-18
control port must be opened and handled on start


532910:
The same as  #2428 but for control port.

532910:
moreover, program must not start in case of control port opening failure

carpawell:
Do we want to provide a separate control status for every operation (e.g. "starting service X") instead of [generic](https://github.com/nspcc-dev/neofs-node/blob/bb9256fa164597ce31833aba1b9c328d92ad4d91/pkg/innerring/innerring.go#L162-L165) "starting", "ready"?

@cthulhu-rider, @roman-khimov

AliceInHunterland:
> Yes. We have very distinct states especially wrt integrated CN.

what is wrt?

532910:
not fixed
```
[root@tenes6:~]# grep -A2 'control:' /etc/neofs/ir/config.yml 
control:
  grpc:
    endpoint: localhost:8090

[root@tenes6:~]# s lsof -i -s tcp:listen -P | grep neofs-ir
neofs-ir  25696        neofs-ir    7u  IPv4 1968265      0t0  TCP localhost:9902 (LISTEN)
neofs-ir  25696        neofs-ir    8u  IPv4 1968267      0t0  TCP localhost:9901 (LISTEN)
[root@tenes6:~]# s lsof -i -s tcp:listen -P | grep 8090    
```

log:
```
Dec 15 07:23:21 tenes6 systemd[1]: Started NeoFS InnerRing node.
Dec 15 07:23:21 tenes6 neofs-ir[25696]: 2023-12-15T07:23:21.220Z        info        neofs-ir/main.go:128        prometheus is enabled
Dec 15 07:23:21 tenes6 neofs-ir[25696]: 2023-12-15T07:23:21.220Z        info        neofs-ir/main.go:128        pprof is enabled
Dec 15 07:23:23 tenes6 neofs-ir[25696]: 2023-12-15T07:23:23.384Z        info        innerring/innerring.go:494        auto-deployment configured, initializing Sidechain...
Dec 15 07:23:23 tenes6 neofs-ir[25696]: 2023-12-15T07:23:23.407Z        info        deploy/deploy.go:243        initializing NNS contract on the chain...
Dec 15 07:23:23 tenes6 neofs-ir[25696]: 2023-12-15T07:23:23.408Z        info        deploy/util.go:69        listening to new blocks...
[...]

PR #2655, opened 2023-11-21
fix/Nil panic in control service

carpawell:
Split service creation on two steps:
1. service creation with key initialization and basic health checks;
2. service real initialization that includes internals reading (such as storage engine, tree service, etc).

The first one handles incoming requests when the app starts, the second one is responsible for the full work mode (storage engine initializations may take time in the real runs with the real objects stored).

### DIFF
--- a/cmd/neofs-node/control.go
+++ b/cmd/neofs-node/control.go
@@ -36,19 +36,7 @@ func initControlService(c *cfg) {
 		rawPubs = append(rawPubs, pubs[i].Bytes())
 	}
 
-	ctlSvc := controlSvc.New(
-		controlSvc.WithKey(&c.key.PrivateKey),
-		controlSvc.WithAuthorizedKeys(rawPubs),
-		controlSvc.WithHealthChecker(c),
-		controlSvc.WithNetMapSource(c.netMapSource),
-		controlSvc.WithContainerSource(c.cfgObject.cnrSource),
-		controlSvc.WithReplicator(c.replicator),
-		controlSvc.WithNodeState(c),
-		controlSvc.WithLocalStorage(c.cfgObject.cfgLocalStorage.localStorage),
-		controlSvc.WithTreeService(treeSynchronizer{
-			c.treeService,
-		}),
-	)
+	c.shared.control = controlSvc.New(&c.key.PrivateKey, rawPubs, c)
 
 	lis, err := net.Listen("tcp", endpoint)
 	if err != nil {
@@ -62,7 +50,7 @@ func initControlService(c *cfg) {
 		stopGRPC("NeoFS Control API", c.cfgControlService.server, c.log)
 	})
 
-	control.RegisterControlServiceServer(c.cfgControlService.server, ctlSvc)
+	control.RegisterControlServiceServer(c.cfgControlService.server, c.shared.control)
 	c.wg.Add(1)
 	go func() {
 		runAndLog(c, "control", false, func(c *cfg) {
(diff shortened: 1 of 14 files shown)


pkg/services/control/server/helpers.go
```diff
@@ -11,10 +16,18 @@ func (s *Server) getShardIDList(raw [][]byte) []*shard.ID {
...
 	res := make([]*shard.ID, 0, len(info.Shards))
 	for i := range info.Shards {
 		res = append(res, info.Shards[i].ID)
 	}
 	return res
 }
+
+func (s *Server) ready() error {
+	if !s.available.Load() {
+		return status.Error(codes.Unavailable, "node has not started fully yet")
```
cthulhu-rider, line 29: to me `node` is smth higher level. I'd report `service has not been completely initialized yet`

pkg/services/control/server/server.go
```diff
@@ -13,6 +14,12 @@ import (
 // Server is an entity that serves
 // Control service on storage node.
 type Server struct {
+	// initialization sync; control service
+	// is the first service to run so no other
```
cthulhu-rider, line 18: `first service to run` is strange to see on the individual service: this field doesnt "know" what the app does, the app knows how `Service` behaves

cmd/neofs-node/control.go
```diff
@@ -70,6 +58,17 @@ func initControlService(c *cfg) {
 			c.wg.Done()
 		})
 	}()
+
+	c.workers = append(c.workers, newWorkerFromFunc(func(ctx context.Context) {
```
cthulhu-rider, line 62: why async worker? U know exactly when all needed components are ready

pkg/services/control/server/server.go
```diff
@@ -65,89 +72,34 @@ type cfg struct {
...
-}
-
-// WithTreeService returns an option to set tree service.
-func WithTreeService(s TreeService) Option {
-	return func(c *cfg) {
-		c.treeService = s
-	}
+// MarkReady marks server available. Before this call none of the other calls
+// are available except for the health checks.
+func (s *Server) MarkReady(e *engine.StorageEngine, nm netmap.Source, c container.Source, r *replicator.Replicator, st NodeState, tr TreeService) {
```
cthulhu-rider, line 96: suggest to require all prms to be non-nil and insta panic otherwise

cthulhu-rider (approved):
overall LGTM, left few suggestions

pkg/services/control/server/server.go
carpawell: fixed

cmd/neofs-node/control.go
carpawell: think that is our common pattern but it was about IR, fixed

pkg/services/control/server/server.go
carpawell: i can understand you but think that is too much of an abstraction then, do not think the control service can be reused anywhere else, only for the node and only as a main controlling service

> this field doesnt "know" what the app does

yes, but we need this service to be split now. not because it is good for it but because we need it for its caller and this comment would allow a future reader to understand it faster

anyway, fixed the comment

pkg/services/control/server/helpers.go
carpawell: fixed

roman-khimov (commented):
So, what useful can we do with the control service if it's not ready yet?

carpawell:
> So, what useful can we do with the control service if it's not ready yet?

Health checks (the can be improved, see https://github.com/nspcc-dev/neofs-node/issues/2585#issuecomment-1725809368) and closing https://github.com/nspcc-dev/neofs-node/issues/2585.

roman-khimov (approved):

roman-khimov:
OK, we can have some more in the future.

PR #2657, opened 2023-11-23
replicator: Copy object from the local storage without decoding

cthulhu-rider:
* based on and blocked by https://github.com/nspcc-dev/neofs-sdk-go/pull/533/


cthulhu-rider:
* not needed anymore due to https://github.com/nspcc-dev/neofs-node/pull/2685, https://github.com/nspcc-dev/neofs-api/pull/280

PR #2664, opened 2023-12-01
ir: Do not configure network settings

cthulhu-rider:


cthulhu-rider:
deployment works with https://github.com/nspcc-dev/neofs-dev-env/pull/282 but IR does not become healthy, will take a look

pkg/morph/deploy/deploy.go
```diff
@@ -462,7 +440,53 @@ func Deploy(ctx context.Context, prm Prm) error {
 
 	prm.Logger.Info("Audit contract successfully synchronized", zap.Stringer("address", auditContractAddress))
```
roman-khimov, line 441: You still deploy audit the first (from regular non-NNS contracts), but it's named `02-audit` now. Probably a rename or reorder is needed.

roman-khimov (commented):
OK otherwise

pkg/morph/deploy/deploy.go
cthulhu-rider: Audit contract is deployed after the Proxy one

roman-khimov: OK, then the numbering here and the numbering in contracts don't match exactly. Not a big deal though.

roman-khimov (approved):

PR #2665, opened 2023-12-04
ir: Support auto-deployment in non-local consensus launch mode

cthulhu-rider:


pkg/morph/client/notifications.go
```diff
@@ -1,328 +1,358 @@
...
+
+		delete(demux, f)
+	}
+
+	for c := range mToClose {
+		close(c)
 	}
+}
+
+func addSub[Event any, Filter comparable](
```
roman-khimov, line 291: Looks like you're trying to create "stable" subscriptions that survive disconnects again. Please, do not do this, it can't be done reliably. Subscribe to the current client, throw an error and force users to restart upon connection failure. It's _much_ simpler and it's more reliable this way.

pkg/morph/client/client.go
```diff
@@ -107,29 +117,167 @@ func (c *Client) CallAndExpandIterator(contract util.Uint160, method string, max
...
+// GetCommittee returns current public keys of the committee of the Neo
+// blockchain the Client connected to.
+func (c *Client) GetCommittee() (keys.PublicKeys, error) {
+	return c.Committee()
+}
+
+// GetContractStateByID returns current state of the identified Neo smart
+// contract deployed in the blockchain the Client connected to. Returns
+// [neorpc.ErrUnknownContract] if requested contract is missing.
+func (c *Client) GetContractStateByID(id int32) (*state.Contract, error) {
```
roman-khimov, line 258: These proxies are OK.

pkg/morph/client/constructor.go
```diff
@@ -119,16 +120,9 @@ func New(key *keys.PrivateKey, opts ...Option) (*Client, error) {
...
-			subscribedEvents:       make(map[util.Uint160]struct{}),
-			subscribedNotaryEvents: make(map[util.Uint160]struct{}),
-			subscribedToNewBlocks:  false,
-
-			curNotifyChan: make(chan *state.ContainedNotificationEvent),
-			curBlockChan:  make(chan *block.Block),
-			curNotaryChan: make(chan *result.NotaryRequestEvent),
+			blocks:         make(eventDemux[*block.Block, *neorpc.BlockFilter]),
+			notifications:  make(eventDemux[*state.ContainedNotificationEvent, *neorpc.NotificationFilter]),
+			notaryRequests: make(eventDemux[*result.NotaryRequestEvent, *neorpc.TxFilter]),
```
roman-khimov, line 125: But subscriptions better be left as is.

pkg/morph/client/notifications.go
cthulhu-rider: ok I'll adjust deploy procedure to this flow then

cthulhu-rider: done

cthulhu-rider:
> `ERROR: for ir-healthcheck Container "3fe89d10c8f1" is unhealthy.`

see `docker logs ir01`

guess it wants NNS e-mail config https://github.com/nspcc-dev/neofs-node/blob/752d8a79d2e8ce2dde7ba7f44810a78d509ee90e/config/example/ir.yaml#L83-L84 that does not default now

pkg/innerring/deploy.go
```diff
@@ -6,20 +6,35 @@ import (
...
 type neoFSSidechain struct {
+	neoFSSidechainCommonRPC
+
 	client *client.Client
 
 	netmapContractMtx sync.RWMutex
 	netmapContract    *netmap.Client
 
-	*rpcclient.WSClient
+	wsClient *rpcclient.WSClient
```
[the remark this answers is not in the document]
cthulhu-rider: as we discussed `client.Client` and `rpcclient.WSClient` cannot share the whole needed interface due to event listening methods. Another inconvenience is that `client.Client` is one-time in terms of subscriptions. So, for non-local consensus we use `client.Client` dedicated for deployment only (opened, used, closed, then use another instance), while in local one we use single (local) `rpcclient.WSClient` with which we can cancel deploy subscriptions and continue as if from scratch

i could have 2 different types for these 2 modes, but decided to not do so for now

roman-khimov:
Still isn't healthy.

cthulhu-rider:
fixed, should work now

roman-khimov (approved):

PR #2671, opened 2023-12-06
refactor/Object `transformer` package

carpawell:


pkg/core/object/target.go
```diff
@@ -1,4 +1,4 @@
-package transformer
+package object
```
cthulhu-rider, line 1: correct, but how about locate package in the `pkg/services/object/internal` if it's not intended to be used outside Object service?

pkg/services/object/put/slice.go
```diff
@@ -21,23 +21,23 @@ type slicingTarget struct {
...
 	homoHashDisabled bool
 
-	nextTarget transformer.ObjectTarget
+	nextTarget internal.Target
 
 	payloadWriter *slicer.PayloadWriter
 }
 
-// returns transformer.ObjectTarget for raw root object streamed by the client
+// returns [objectcore.Target] for raw root object streamed by the client
```
roman-khimov, line 29: Is it `internal` or `objectcore`?

pkg/services/object/put/slice.go
```diff
@@ -4,7 +4,7 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/nspcc-dev/neofs-node/pkg/services/object_manager/transformer"
+	"github.com/nspcc-dev/neofs-node/pkg/services/object/internal"
```
roman-khimov, line 7: Usually, `internal/smth` is used, because `internal` itself doesn't mean a lot.

roman-khimov: It's strange enough we have a whole package just for this interface. To me it looks a lot like a part of `pkg/services/object/put`. But there it'd be internal. But then who needs an internal interface?

carpawell: https://github.com/nspcc-dev/neofs-node/pull/2671#discussion_r1420316138 describes where i saw this interface and why

part of it is used by the get service too. i wanted to see more usages but somehow we have so many `WriteChunk` and `Write(object.Object) error` (not the `oi.Writer`) so it does not look so useful sadly. but we can continue this refactor

roman-khimov: `get` doesn't need it. It has the thing it needs already.

carpawell: dont you like how `HeaderWriter` is included in the `Target` and how every object interface can be placed in a single common file? every header should be written with the same interface does not matter what service does it

pkg/services/object/put/slice.go
carpawell: has not pushed the last commit rebase, fixed

cthulhu-rider (approved):

PR #2683, opened 2023-12-18
Several NeoFS chain autodeploy/update fixes

cthulhu-rider:


pkg/morph/deploy/funds.go
```diff
@@ -448,3 +445,44 @@ func distributeNEOToAlphabetContracts(ctx context.Context, prm distributeNEOToAl
...
+			rem.Sub(rem, big.NewInt(1))
+		} else if amount.Sign() == 0 {
+			return
+		}
+
+		f(i, amount)
+	}
+}
+
+func divideFundsEvenlyU64(fullAmount uint64, n int, f func(ind int, amount uint64)) {
```
roman-khimov, line 473: One version is enough. We can't have more than 100M NEO.

roman-khimov (approved):

pkg/morph/deploy/funds.go
cthulhu-rider: done

PR #2685, opened 2023-12-18
Optimize verification of the request signatures

cthulhu-rider:
will implement PoC for `ObjectService.Put` as the most signature verification-laden RPCs

### DIFF
--- a/pkg/network/transport/object/grpc/service_test.go
+++ b/pkg/network/transport/object/grpc/service_test.go
@@ -1,0 +1,298 @@
+package object_test
+
+import (
+	"context"
+	"encoding/base64"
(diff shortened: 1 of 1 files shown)


roman-khimov:
Still relevant?

cthulhu-rider:
now we have https://github.com/nspcc-dev/neofs-node/tree/master/internal/protobuf/protoscan. It may be used to handle request signatures much faster than currently used https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go/crypto#VerifyRequestWithBufferN3

moreover, we can now access request fields only when needed w/o full deserialization in advance

PR #2690, opened 2023-12-19
adm: Allow generation more than one SN wallet per call

carpawell:
Add `wallets-number` flag that "duplicates" requests to the `neofs-adm`. Passwords and labels are the same for all the wallets being created. Names differ with the suffix-numbers.


carpawell:
I decided not to create an additional command.

roman-khimov (approved):

PR #2694, opened 2023-12-20
node/client-cache: Log errors that led to a connection drops

carpawell:
It simplifies connection drop inspection cause after the first connection drop, there are a lot of `client has recently failed, skipping` logs, and it is not clear what exact reason did not allow the node to communicate with a "problem" one. It could be dropped in the future, but a connection loss for 30s (default reconnection interval for now) is not a normal situation in general, and it causes replication processes, data duplication, etc., so logs about it can be kept.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Changelog for NeoFS Node
 ## [Unreleased]
 
 ### Added
+- Logs on a connection drop in the cache of NeoFS API clients (#2694)
 
 ### Fixed
 
--- a/pkg/network/cache/client.go
+++ b/pkg/network/cache/client.go
@@ -6,6 +6,7 @@ import (
 
 	clientcore "github.com/nspcc-dev/neofs-node/pkg/core/client"
 	"github.com/nspcc-dev/neofs-sdk-go/client"
+	"go.uber.org/zap"
 )
 
 // DefaultBufferSize describes default max GRPC message size. Unfortunately GRPC lib contains this const in private.
@@ -27,6 +28,7 @@ type (
 		ResponseCallback func(client.ResponseMetaInfo) error
 		AllowExternal    bool
 		Buffers          *sync.Pool
+		Logger           *zap.Logger
 	}
 )
 
--- a/pkg/network/cache/multi.go
+++ b/pkg/network/cache/multi.go
@@ -17,6 +17,7 @@ import (
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	reputationSDK "github.com/nspcc-dev/neofs-sdk-go/reputation"
 	"github.com/nspcc-dev/neofs-sdk-go/user"
+	"go.uber.org/zap"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
@@ -166,6 +167,7 @@ func (x *multiClient) iterateClients(ctx context.Context, f func(clientcore.Clie
 		}
 
 		if err != nil {
+			err = fmt.Errorf("%s node client: %w", addr, err)
 			x.ReportError(err)
 		}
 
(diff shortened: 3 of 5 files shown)


pkg/network/cache/multi.go
```diff
@@ -192,6 +193,13 @@ func (x *multiClient) ReportError(err error) {
 		return
 	}
 
+	x.addrMtx.RLock()
+	group := x.addr
+	x.addrMtx.RUnlock()
+
+	x.opts.Logger.Warn("multi-cache: dropping clients to the node",
```
roman-khimov, line 200: Warn? @532910 won't like it. It's a regular operation thing.

roman-khimov (commented):
No changelog? It affects the behavior.

The other thing is, can we just suppress additional errors from the same connection?

pkg/network/cache/client.go
```diff
@@ -27,6 +28,7 @@ type (
 		ResponseCallback func(client.ResponseMetaInfo) error
 		AllowExternal    bool
 		Buffers          *sync.Pool
+		Logger           *zap.Logger
```
cthulhu-rider, line 31: `NewSDKClientCache` must fallback to no-op log if opt is unset

pkg/network/cache/multi.go
```diff
@@ -192,6 +193,13 @@ func (x *multiClient) ReportError(err error) {
 		return
 	}
 
+	x.addrMtx.RLock()
+	group := x.addr
```
cthulhu-rider, line 197: `x.addr` can be already unrelated to the `err` at this moment and, if so, we'll see a lie in the log

pkg/network/cache/multi.go
cthulhu-rider: nobody except developers will understand this message: what is `multi-cache`?. So, if u wanna make it visible for admins, lets make text clear and then `info`. Otherwise - `debug`

pkg/network/cache/multi.go
carpawell: why is it a problem? we are invalidating clients currently associated with _this_ address group

added the exact problem address to the err on the higher level

pkg/network/cache/client.go
carpawell: not sure i agree cause it is kinda required thing in the current context and this PR means we _want to log it_ but ok

pkg/network/cache/multi.go
carpawell: I provide my ideas in the commit message why it is an important thing for us but since 2>1 made it INFO

carpawell:
> No changelog? It affects the behavior.

Dont mind, done.

> The other thing is, can we just suppress additional errors from the same connection?

Not sure I got you. You mean do not duplicate errors in the logs?

pkg/network/cache/multi.go
cthulhu-rider: it's not about democracy. I didn't got the maintenance flow: does admin do anything when he see this message?

pkg/network/cache/multi.go
cthulhu-rider: u fixed message and now it conveys a fact

pkg/network/cache/client.go
cthulhu-rider: but i told about `NewSDKClientCache`... ok, nvm

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ Changelog for NeoFS Node
 ## [Unreleased]
 
 ### Added
+- Logs on a connection drop in the SDK client multi-cache (#2694)
```
cthulhu-rider, line 7:
```suggestion
- Logs on a connection drop in the cache of NeoFS API clients (#2694)
```

pkg/network/cache/multi.go
carpawell: > does admin do anything when he see this message?

tries to figure out why some node is dead and objects are going to another node for at least the next 30 seconds

pkg/network/cache/multi.go
carpawell: > is it guaranteed that the x.addr here exactly related to the x.clients below?

reorganized mutexes a little, now it can be

pkg/network/cache/multi.go
cthulhu-rider: imo admin does not care about cache invalidating, only about observable system degradation requiring certain actions for correction. And this is on alerting system plane

cthulhu-rider (approved):

pkg/network/cache/multi.go
carpawell: > only about observable system degradation requiring certain actions for correction

system is degrading, cause one node thinks that another is dead. not for an object, not for a single connection, not for some operation but in general. i do not like this behavior but it was done for failover tests. should be discussed cause it solves something but make it hard in another cases

cthulhu-rider: >system is degrading

ok, admin sees this and the message u've added, then do u know

> certain actions for correction

by admin right now? Restart current node/restart problem node/anything else? 

> i do not like this behavior 

i believe, and tryin to realize the problem. Ref me to an existing report about this potential problem or create the new issue pls

PR #2695, opened 2023-12-20
deploy: Fix witnesses for Balance and Container contract deployments

cthulhu-rider:
tested with 1/4/7 node envs

### DIFF
--- a/pkg/morph/deploy/contracts.go
+++ b/pkg/morph/deploy/contracts.go
@@ -13,6 +13,7 @@ import (
 	"github.com/nspcc-dev/neo-go/pkg/rpcclient/actor"
 	"github.com/nspcc-dev/neo-go/pkg/rpcclient/management"
 	"github.com/nspcc-dev/neo-go/pkg/rpcclient/notary"
+	"github.com/nspcc-dev/neo-go/pkg/smartcontract"
 	"github.com/nspcc-dev/neo-go/pkg/smartcontract/manifest"
 	"github.com/nspcc-dev/neo-go/pkg/smartcontract/nef"
 	"github.com/nspcc-dev/neo-go/pkg/util"
@@ -23,6 +24,12 @@ import (
 // various common methods of the NeoFS contracts.
 const (
 	methodUpdate = "update"
+)
+
+const (
+	_ uint8 = iota
+	witnessValidators
+	witnessValidatorsAndCommittee
 )
 
 // syncNeoFSContractPrm groups parameters of syncNeoFSContract.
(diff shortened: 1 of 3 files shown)


pkg/morph/deploy/contracts.go
```diff
@@ -126,11 +135,54 @@ func syncNeoFSContract(ctx context.Context, prm syncNeoFSContractPrm) (util.Uint
...
-		deployCommitteeActor, err := newCommitteeNotaryActorWithCustomCommitteeSigner(prm.blockchain, prm.localAcc, prm.committee, func(s *transaction.Signer) {
-			s.Scopes = transaction.CustomContracts
-			s.AllowedContracts = append(prm.extraCommitteeDeployAllowedContracts, prm.nnsContract)
-		})
+	if prm.deployWitness > 0 {
+		if prm.deployWitness != witnessAlphabet && prm.deployWitness != witnessAlphabetAndCommittee {
+			panic(fmt.Sprintf("unexpected deploy witness mode value %v", prm.deployWitness))
+		}
+
+		committeeDefaultMultiSigAcc := wallet.NewAccountFromPrivateKey(prm.localAcc.PrivateKey())
```
roman-khimov, line 143: Technically, it's not a committee address. It's a validator multisig (when you take validators only, of course, so `ValidatorsCount` can be applied, although NeoFS has it set to the size of the committee). Your initial NEO and GAS reside on this account, for example.

pkg/morph/deploy/contracts.go
```diff
@@ -126,11 +135,54 @@ func syncNeoFSContract(ctx context.Context, prm syncNeoFSContractPrm) (util.Uint
...
+		signers[0].Signer.Account = prm.localAcc.ScriptHash()
+		signers[0].Signer.Scopes = transaction.None
+		// Alphabet
+		signers[1].Account = committeeDefaultMultiSigAcc
+		signers[1].Signer.Account = committeeDefaultMultiSigAcc.ScriptHash()
+		signers[1].Signer.Scopes = transaction.CustomContracts
+		signers[1].Signer.AllowedContracts = prm.alphabetDeployAllowedContracts
+
+		if prm.deployWitness == witnessAlphabetAndCommittee {
+			committeeMajorityMultiSigAcc := wallet.NewAccountFromPrivateKey(prm.localAcc.PrivateKey())
```
roman-khimov, line 161: And this is a proper committee, as is, not some special "majority" one.

roman-khimov (commented):
The fix is OK, but we better improve the terminology at the same time, otherwise we'll leave some misconceptions in the code.

pkg/morph/deploy/contracts.go
cthulhu-rider: ok i agree, competely do not like inconsistency of terms

btw what u think of Alphabet? it's mentioned in many places like https://pkg.go.dev/github.com/nspcc-dev/neofs-contract@v0.19.1/netmap#SubscribeForNewEpoch

roman-khimov: I doubt NeoFS will survive in `len(committee) != ValidatorsCount` scenario now, btw. But it's a bit different story.

cthulhu-rider: left "validators" for now as more Neo-known term

roman-khimov (approved):

PR #2704, opened 2023-12-22
ir: Make exit code be `1` if app fails

carpawell:
Also, wait for http servers to stop before exiting the app.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Changelog for NeoFS Node
 ### Added
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


cthulhu-rider (commented):
code is ok (dont forget linter), but to me these are two independent changes in single commit

carpawell:
> dont forget linter

Fixed. 

> but to me these are two independent changes in single commit

Dont mind, done.

cthulhu-rider (approved):
i'd appreciate imperative form in the 2nd commit

roman-khimov (approved):

PR #2705, opened 2023-12-22
morph: Try all endpoints on client creation

carpawell:
Do not fail if the first one is unavailable but more endpoints are provided.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Changelog for NeoFS Node
 ### Fixed
 - IR does not wait for HTTP servers to stop gracefully before exiting (#2704)
 - Zero exit code if IR fails (#2704)
+- Neo RPC client failure when the first endpoint is unavailable even if there are more endpoints to try (#2703)
 
 ### Changed
 - Created files are not group writable (#2589)
--- a/pkg/morph/client/constructor.go
+++ b/pkg/morph/client/constructor.go
@@ -160,8 +160,14 @@ func New(key *keys.PrivateKey, opts ...Option) (*Client, error) {
 		}
 
 		cli.endpoints = cfg.endpoints
+		for _, e := range cli.endpoints {
+			cli.client, act, err = cli.newCli(e)
+			if err != nil {
+				cli.logger.Warn("Neo RPC connection failure", zap.String("endpoint", e), zap.Error(err))
+				continue
+			}
+		}
 
-		cli.client, act, err = cli.newCli(cli.endpoints[0])
 		if err != nil {
 			return nil, fmt.Errorf("could not create RPC client: %w", err)
 		}


pkg/morph/client/constructor.go
```diff
@@ -160,8 +160,14 @@ func New(key *keys.PrivateKey, opts ...Option) (*Client, error) {
...
 		cli.endpoints = cfg.endpoints
+		for _, e := range cli.endpoints {
+			cli.client, act, err = cli.newCli(e)
+			if err != nil {
+				cli.logger.Warn("morph client connection", zap.String("endpoint", e), zap.Error(err))
+				continue
+			}
+		}
 
-		cli.client, act, err = cli.newCli(cli.endpoints[0])
```
cthulhu-rider, line 164: previously we returned `cli.endpoints[0]` connection error, now - the last one. Not sure this is an expected behavior change

pkg/morph/client/constructor.go
```diff
@@ -160,8 +160,14 @@ func New(key *keys.PrivateKey, opts ...Option) (*Client, error) {
 		}
 
 		cli.endpoints = cfg.endpoints
+		for _, e := range cli.endpoints {
+			cli.client, act, err = cli.newCli(e)
+			if err != nil {
+				cli.logger.Warn("morph client connection", zap.String("endpoint", e), zap.Error(err))
```
cthulhu-rider, line 166:
```suggestion
				cli.logger.Warn("NeoFS chain connection failure", zap.String("endpoint", e), zap.Error(err))
```
level is also doubtful, `Info` is enough imo

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Changelog for NeoFS Node
 ### Added
 
 ### Fixed
+- Morph client dies when the first endpoint is unavailable even if there are more endpoints to try (#2703)
```
cthulhu-rider, line 9: can be confusing: is it about what was fixed or after changes? `from now` / `no more` help

and `Morph client` -> `NeoFS chain connection`

carpawell: > is it about what was fixed or after changes?

it always bothers me when i fill `Fixed`. i came to a mentioning a problem itself (so every point here describes _not_ expected things)

> `from now` / `no more` help

i can understand you but not sure we want to see hundreds of `from now` it this file. mb we need to find some common approach or try to paraphrase it. @roman-khimov

> and Morph client -> NeoFS chain connection

this is a common chain client

pkg/morph/client/constructor.go
carpawell: the level is the same as it is done in [reconnection](https://github.com/nspcc-dev/neofs-node/blob/eba32a060dd0b1d919976d420b4ea740c268e9b0/pkg/morph/client/multi.go#L55)

> NeoFS chain

we use this client for mainchain too

pkg/morph/client/constructor.go
carpawell: can it be important?

roman-khimov: Can, but not likely.

CHANGELOG.md
roman-khimov: fixed _smth_, like "RPC client failure when..."

roman-khimov (approved):

pkg/morph/client/constructor.go
cthulhu-rider: lets leave until real problems then

pkg/morph/client/constructor.go
cthulhu-rider: level - :ok_hand: 

> we use this client for mainchain too

then its extra not morph client, `Neo RPC connection failure`

CHANGELOG.md
carpawell: agree, fixed. but still think that we need to have some common behavior pattern here. an headache every time I need to expand this section even if I really want it to be clear

pkg/morph/client/constructor.go
carpawell: fixed

cthulhu-rider (approved):

PR #2710, opened 2023-12-26
fix/Storage engine panics

carpawell:


carpawell:
Local big suit run was ok (but I did some refactoring just right before the last push, as I love).

cmd/neofs-node/config.go
```diff
@@ -17,6 +17,7 @@ import (
 
 	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
 	neogoutil "github.com/nspcc-dev/neo-go/pkg/util"
+	utilneogo "github.com/nspcc-dev/neo-go/pkg/util"
```
cthulhu-rider, line 20: we already have this import aliased with `neogoutil`

cmd/neofs-node/config.go
```diff
@@ -88,27 +80,45 @@ const notificationHandlerPoolSize = 10
...
+	Policer struct {
 		maxCapacity         uint32
 		headTimeout         time.Duration
 		replicationCooldown time.Duration
 		objectBatchSize     uint32
 	}
+
+	Morph struct {
+		endpoints                 []string
+		dialTimout                time.Duration
```
cthulhu-rider, line 109:
```suggestion
		dialTimeout                time.Duration
```

cmd/neofs-node/config.go
```diff
@@ -88,27 +80,45 @@ const notificationHandlerPoolSize = 10
 // values. It should not store any application helpers structs (pointers to shared
 // structs).
 // It must not be used concurrently.
+//
+// REFACTOR IS NOT FINISHED.
```
cthulhu-rider, line 84: unless it refers some issue or explain how the code should be maintained until the refactor is completed, this comment doesn't help anyone

roman-khimov, line 84: Useless without some issue reference.

cthulhu-rider (commented):
commend the refactor, but pls dont mess it with bugfix in the single commit next time

it can be done like this:
1. i implement fix itself with minimum diff
2. did u saw how ugly the fix looked like? this due to undone refactor, so lets do it insta

for this PR lets leave as is

cmd/neofs-node/config.go
```diff
@@ -88,27 +80,45 @@ const notificationHandlerPoolSize = 10
...
 
-	EngineCfg struct {
+	Engine struct {
 		errorThreshold uint32
 		shardPoolSize  uint32
 		shards         []storage.ShardCfg
 	}
 
-	PolicerCfg struct {
+	Policer struct {
```
roman-khimov, line 100: Questionable to me. It's still some config in most cases. Processed, ready to go, but not some component.

cmd/neofs-node/config.go
```diff
@@ -88,27 +80,45 @@ const notificationHandlerPoolSize = 10
...
 		maxCapacity         uint32
 		headTimeout         time.Duration
 		replicationCooldown time.Duration
 		objectBatchSize     uint32
 	}
+
+	Morph struct {
```
roman-khimov, line 107: We dont't comment them?

cmd/neofs-node/config.go
carpawell: fixed

cmd/neofs-node/config.go
carpawell: fixed

cmd/neofs-node/config.go
carpawell: > this comment doesn't help anyone

IMO, it conflicts with the description: `stores component-specific configuration`, but the most of the config sections are omitted here

I added links to the original issue about SIGHUP and the original PR that does it, is it ok?

carpawell: lets continue here: https://github.com/nspcc-dev/neofs-node/pull/2710#discussion_r1437162957

cmd/neofs-node/config.go
carpawell: we _literally_ have both `MorphCfg` and `cfgMorph` in the code. basic usage of _this_ struct is: `c.applicationConfiguration.Morph` so that is why I think _this_ struct can miss `Cfg` suffix (`applicationConfiguration` is the struct that has exact values from a `yml` file and the other sources by my design)

cmd/neofs-node/config.go
carpawell: they were not commented initially and i can't say why they are exported made them unexported

carpawell:
> commend the refactor, but pls dont mess it with bugfix in the single commit next time

Totally agree, but as I remember (I really tried it a few times), it just could not be fixed easily after the first two attempts without panicking somewhere else.

cmd/neofs-node/config.go
carpawell: struct renamings are welcome

cmd/neofs-node/config.go
carpawell: btw, added a linter for it, we like double importing a little

cmd/neofs-node/config.go
cthulhu-rider: understandable to me, but i'd appreciate issue in form `have this -> do this to got this`, pls fill the issue it async

cthulhu-rider (approved):
cant wait untill all tests run

roman-khimov (approved):

PR #2711, opened 2024-01-09
ir: Do not hash public keys to check orig/announced SN diff

carpawell:
Logs like "*number* differs *number*" may confuse, public keys are more common.

### DIFF
--- a/pkg/innerring/processors/netmap/nodevalidation/availability/validator.go
+++ b/pkg/innerring/processors/netmap/nodevalidation/availability/validator.go
@@ -85,8 +85,8 @@ func compareNodeInfos(niExp, niGot netmap.NodeInfo) error {
 
 	var err error
 
-	if exp, got := niExp.Hash(), niGot.Hash(); exp != got {
-		return fmt.Errorf("hash: got %d, expect %d", got, exp)
+	if exp, got := niExp.PublicKey(), niGot.PublicKey(); !bytes.Equal(exp, got) {
+		return fmt.Errorf("public key: got %x, expect %x", got, exp)
 	}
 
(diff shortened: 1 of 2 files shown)


roman-khimov (approved):

pkg/innerring/processors/netmap/nodevalidation/availability/validator.go
```diff
@@ -85,8 +86,8 @@ func compareNodeInfos(niExp, niGot netmap.NodeInfo) error {
 
 	var err error
 
-	if exp, got := niExp.Hash(), niGot.Hash(); exp != got {
-		return fmt.Errorf("hash: got %d, expect %d", got, exp)
+	if exp, got := niExp.PublicKey(), niGot.PublicKey(); !bytes.Equal(exp, got) {
+		return fmt.Errorf("public key: got %s, expect %s", hex.EncodeToString(got), hex.EncodeToString(exp))
```
cthulhu-rider, line 90:
```suggestion
		return fmt.Errorf("public key: got %x, expect %x", got, exp)
```

cthulhu-rider (approved):

pkg/innerring/processors/netmap/nodevalidation/availability/validator.go
carpawell: dont mind but `hex.EncodeToString` is more readable (especially for a different language reader) imo

cthulhu-rider: this is not innovation:
* https://docs.python.org/3/library/string.html
* https://en.cppreference.com/w/c/io/fprintf
* https://docs.oracle.com/javase/8/docs/api/java/util/Formatter.html

yeah, not all langs, but pretty understandable overall

PR #2716, opened 2024-01-12
fix/Child objects ACL

carpawell:


carpawell:
Well, I need Go 1.20 now. Otherwise, I am ready.

carpawell:
Not sure what is wrong with the tests: it creates `REP 2 IN X CBF 1 SELECT 4 FROM * AS X` container PUTs a complex object and HEADs its parts with TTL 1 to every container node (2 of 4 answers correct, the other 2 say 404). Need to be investigated, not sure why all 4 nodes are expected to answer. Generally can be reviewed, draft status is caused by that strange test.

cmd/neofs-node/object.go
```diff
@@ -535,3 +546,52 @@ func (e storageEngine) Lock(locker oid.Address, toLock []oid.ID) error {
...
+	h *objectSDK.Object
+}
+
+func (h *headerWriter) WriteHeader(o *objectSDK.Object) error {
+	h.h = o
+	return nil
+}
+
+func (h headerSource) Head(address oid.Address) (*objectSDK.Object, error) {
+	if h.cache != nil {
```
roman-khimov, line 575: Why would it be `nil`?

pkg/local_object_storage/metabase/VERSION.md
```diff
@@ -57,6 +57,10 @@ Numbers stand for a single byte value.
   - Name: container ID + `9`
   - Key: object ID
   - Value: marshaled object
+- Buckets containing object or LINK type
+  - Name: container ID + `18`
```
roman-khimov, line 61: I thought they're sorted here.

pkg/local_object_storage/metabase/put.go
```diff
@@ -520,14 +529,38 @@ func splitInfoFromObject(obj *objectSDK.Object) (*objectSDK.SplitInfo, error) {
 	return info, nil
 }
 
-// isLinkObject returns true if object contains parent header and list
-// of children.
+// isLinkObject returns true if
+// V1: object contains parent header and list
+//
+//	of children
```
roman-khimov, line 535: Wow.

pkg/services/object/split/verify.go
```diff
@@ -0,0 +1,125 @@
...
+//  2. NeoFS should answer to the HEAD requests with children whose object size corresponds to the link's info
+//  3. Every child should have a provided first object in its split header (the first object itself is an exception)
+func (v *Verifier) VerifySplit(ctx context.Context, cnr cid.ID, firstID oid.ID, childrenFromLink []object.MeasuredObject) error {
+	// can be limited, depends on the expected big objects payload length
+	var wg errgroup.Group
+	receivedObjects := make([]*object.Object, len(childrenFromLink))
+
+	for i := range childrenFromLink {
+		iCopy := i
+		wg.Go(func() error {
```
roman-khimov, line 39: 1M goroutines?

pkg/services/object/split/verify.go
```diff
@@ -0,0 +1,125 @@
...
+			var shouldHaveFirstObject *oid.ID
+			if iCopy != 0 {
+				shouldHaveFirstObject = &firstID
+			}
+
+			var err error
+			receivedObjects[iCopy], err = v.verifySinglePart(ctx, cnr, shouldHaveFirstObject, childrenFromLink[iCopy])
```
roman-khimov, line 46: 1M * 16K for headers?

roman-khimov (commented):
@cthulhu-rider, join the party. We need to merge this ASAP to unblock SDK updates.

cthulhu-rider:
@roman-khimov affirmative

@carpawell wt readiness? PR is still draft

carpawell:
> @carpawell wt readiness? PR is still draft

Basically can be reviewed since the push that allows building. Some fixes are done (and will be done but are based on my work with int test cases) a little and after the review (when the base is fixed), I would add some missing unit tests.

pkg/local_object_storage/metabase/put.go
carpawell: wow, my IDE did that, fixed

pkg/local_object_storage/metabase/VERSION.md
carpawell: they have a logical title (`Unique index buckets`) so i decided them to be together

cmd/neofs-node/object.go
carpawell: `cacheSize` _may be_ nil so no caching is done then. hardcoded for now but may be useful in the future

pkg/services/object/split/verify.go
carpawell: that is why the comment above says that it can be limited. what number do you suggest?

pkg/services/object/split/verify.go
carpawell: how child objects may be 16K? optimizations may be presented here but not sure they changes smth for now (64mb are required for the most optimal part of the object storing if _NO_ get-check-drop is implemented and all the objects are fetched before any validation is being done)

pkg/services/object/split/verify.go
roman-khimov: It must be limited. Start small, 16 isn't bad.

pkg/services/object/split/verify.go
roman-khimov: Fetch and verify group by group, headers are headers and can be up to 16KB.

cmd/neofs-cli/modules/object/util.go
```diff
@@ -483,3 +480,72 @@ func collectObjectRelatives(ctx context.Context, cmd *cobra.Command, cli *client
...
+	addrObj.SetObject(linkID)
+	prmGet.SetClient(cli)
+	prmGet.SetPrivateKey(*key)
+	prmGet.SetAddress(addrObj)
+	prmGet.SetRawFlag(false)
+	prmGet.SetPayloadWriter(pw)
+	Prepare(cmd, &prmGet)
+
+	_, err := internal.GetObject(ctx, prmGet)
+	if err == nil {
```
cthulhu-rider, line 507: easier to check `err != nil` instead

cmd/neofs-cli/modules/object/util.go
```diff
@@ -483,3 +480,72 @@ func collectObjectRelatives(ctx context.Context, cmd *cobra.Command, cli *client
...
+		var link object.Link
+		err = link.Unmarshal(pw.payload)
+		if err != nil {
+			common.PrintVerbose(cmd, "failed to get linking object's header: %w", err)
```
cthulhu-rider, line 511: header is gotten, decode failure

but this message is also useful

cmd/neofs-cli/modules/object/util.go
```diff
@@ -483,3 +480,72 @@ func collectObjectRelatives(ctx context.Context, cmd *cobra.Command, cli *client
...
+	prmHead.SetRawFlag(false)
+	Prepare(cmd, &prmHead)
+
+	res, err := internal.HeadObject(ctx, prmHead)
+	if err == nil {
+		children := res.Header().Children()
+
+		common.PrintVerbose(cmd, "Received split members from the linking object: %v", children)
+
+		// include linking object
```
cthulhu-rider, line 546: done in both cases, caller can deduplicate

cmd/neofs-cli/modules/object/util.go
```diff
@@ -483,3 +480,72 @@ func collectObjectRelatives(ctx context.Context, cmd *cobra.Command, cli *client
...
+func getChildrenFromV2Link(ctx context.Context, cmd *cobra.Command, cli *client.Client, key *ecdsa.PrivateKey, linkID oid.ID) ([]oid.ID, bool) {
+	var addrObj oid.Address
+	var prmGet internal.GetObjectPrm
+	pw := &payloadWriter{}
+
+	addrObj.SetObject(linkID)
+	prmGet.SetClient(cli)
+	prmGet.SetPrivateKey(*key)
+	prmGet.SetAddress(addrObj)
+	prmGet.SetRawFlag(false)
```
cthulhu-rider, line 502: no need to set default explicitly

same below

cthulhu-rider (changes requested):
finished small commits, will return to the main one

cmd/neofs-cli/modules/object/util.go
carpawell: copy-pasted, fixed

cmd/neofs-cli/modules/object/util.go
carpawell: fixed

cmd/neofs-cli/modules/object/util.go
carpawell: fixed

cmd/neofs-cli/modules/object/util.go
carpawell: this was copy-pasted. dropped

pkg/services/object/util/chain.go
```diff
@@ -34,45 +38,178 @@ func IterateAllSplitLeaves(r HeadReceiver, addr oid.Address, h func(*object.Obje
...
+
+	lastID, ok := info.LastPart()
+	if ok {
+		addr.SetObject(lastID)
+		return iterateFromLastObject(r, addr, handler)
 	}
 
-	for i := len(leaves) - 1; i >= 0; i-- {
-		if h(leaves[i]) {
+	return errors.New("nor link, nor last object ID is found")
```
roman-khimov, line 100: neither nor

pkg/services/object/split/verify.go
roman-khimov: This is still a problem, but can be left for another issue.

roman-khimov (approved):

pkg/core/object/fmt.go
```diff
@@ -191,6 +241,32 @@ func (v *FormatValidator) ValidateContent(o *object.Object) (ContentMeta, error)
...
+		}
+
+		var testLink object.Link
+
+		err := o.ReadLink(&testLink)
+		if err != nil {
+			return ContentMeta{}, fmt.Errorf("reading link object's payload: %w", err)
+		}
+
+		err = v.sv.VerifySplit(context.Background(), cnr, firstObjID, testLink.Objects())
```
cthulhu-rider, line 266: why ctx is parameterized if we pass `context.Background` only?

pkg/local_object_storage/metabase/put.go
```diff
@@ -112,15 +112,18 @@ func (db *DB) put(
 		return nil
 	}
 
-	if par := obj.Parent(); par != nil && !isParent { // limit depth by two
-		parentSI, err := splitInfoFromObject(obj)
-		if err != nil {
-			return err
-		}
+	par := obj.Parent()
```
cthulhu-rider, line 115: just to shorten `if` statement or u forgot to use `par` below?

pkg/local_object_storage/metabase/util.go
```diff
@@ -157,6 +162,11 @@ func smallBucketName(cnr cid.ID, key []byte) []byte {
 	return bucketName(cnr, smallPrefix, key)
 }
 
+// linkObjectsBucketName returns link objects bucket key (`17<CID>`).
```
cthulhu-rider, line 165: so finally `17` or `18` (VERSION.md)?

pkg/services/object/get/assembly_v2.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	linkID, set := si.Link()
+	if set && exec.processV2Link(linkID) {
+		return
+	}
+
+	// fallback to the full chain assembly from the last part
+	prev, set := si.LastPart()
+	if set {
```
cthulhu-rider, line 27: else?

pkg/services/object/get/assembly_v2.go
```diff
@@ -0,0 +1,178 @@
...
+				rngPerChild.SetLength(uint64(child.ObjectSize()) - uint64(lastBound+firstOffset))
+			}
+		}
+
+		part, ok := exec.getChild(child.ObjectID(), rngPerChild, false)
+		if !ok {
+			return false
+		}
+
+		if exec.writeObjectPayload(part) {
```
cthulhu-rider, line 123: shouldn't this be inversion?

pkg/services/object/get/assembly_v2.go
```diff
@@ -0,0 +1,178 @@
...
+func (exec *execCtx) processV2Split(si *objectSDK.SplitInfo) {
+	if _, set := si.FirstPart(); !set {
+		exec.log.Debug("no first ID found in V2 split")
+		exec.err = errors.New("v2 split without first object ID")
+
+		return
+	}
+
+	linkID, set := si.Link()
+	if set && exec.processV2Link(linkID) {
```
cthulhu-rider, line 21: `processV2Link` could return `false` on partial success: `writeObjectPayload` and `writeCollectedHeader` are called inside. I doubt we should continue processing when they fail

pkg/services/object/get/assemble.go
```diff
@@ -44,6 +44,14 @@ func (exec *execCtx) assemble() {
 		}
 	}
 
+	if splitInfo.SplitID() == nil {
+		exec.log.Debug("handling V2 split")
+		exec.processV2Split(splitInfo)
```
cthulhu-rider, line 49: i thought we wont need completely separate branch of code for this (schemes differ by LINK obj format and quick access to piece sizes). But ok, done as done

cthulhu-rider (commented):
huge amount of work has been done, such that I am not able to closely examine some of its parts. These are my final comments, lets merge after them

cmd/neofs-node/object.go
carpawell: we have a config param that explicitly says "no caching, please" (but for morph values mostly)

> cacheSize may be zero

i meant you can change one const and it is not caching anymore, you may make it configurable and just init it with a value you read and zero value is already supported. i agree, it may look strange without a real use-case and config param but since it is already done, why not?

pkg/services/object/util/chain.go
```diff
@@ -4,17 +4,21 @@ import (
...
+	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 )
 
-// HeadReceiver is an interface of entity that can receive
+// SplitInformer is an interface of entity that can receive
 // object header or the information about the object relations.
-type HeadReceiver interface {
+type SplitInformer interface {
```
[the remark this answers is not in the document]
carpawell: went `ObjectSource` way

pkg/services/object/util/chain.go
```diff
@@ -34,45 +38,178 @@ func IterateAllSplitLeaves(r HeadReceiver, addr oid.Address, h func(*object.Obje
...
 			break
 		}
+
+		addr.SetObject(prevOID)
+	}
+
+	for i := len(idBuff) - 1; i >= 0; i-- {
+		addr.SetObject(idBuff[i])
+
+		childObj, err := headFromInformer(r, addr)
```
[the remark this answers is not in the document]
carpawell: agree, looks ugly. but what if we have 1M 16K children as @roman-khimov loves? saving 1M 32-bytes IDs would be better than storing so many headers. not sure what is worse: duplicating RPCs or having 16GB memory for simple SG object put

also, this case is not expected since link object are always preferable

also, we are not expecting to have ID in the right order in SG as i remember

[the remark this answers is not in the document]
carpawell: > also, we are not expecting to have ID in the right order in SG as i remember

woops, no: https://github.com/nspcc-dev/neofs-api/blob/5c3535423c564fe63991812cb84d8334db1c553a/storagegroup/types.proto#L32

cmd/neofs-node/object.go
```diff
@@ -535,3 +546,52 @@ func (e storageEngine) Lock(locker oid.Address, toLock []oid.ID) error {
...
+		head, ok := h.cache.Get(address)
```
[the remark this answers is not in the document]
carpawell: fixed

pkg/core/object/fmt.go
carpawell: i am dreaming of having context in every multithread/network/slice-parameterised operations. `SplitVerifier` wants it but `FormatValidator` does not have it yet

pkg/local_object_storage/metabase/put.go
carpawell: returned back, not sure why i needed it

pkg/local_object_storage/metabase/util.go
carpawell: `18`, yes. copy-pasted comment problem

pkg/services/object/get/assembly_v2.go
carpawell: just `return`. that is how `execCtx` works. if no status and result is set inside some of the func it returns 404. this line is the last. but agree, added Debug logs here

pkg/services/object/get/assembly_v2.go
carpawell: hm, seems so, strange, fixed

pkg/services/object/get/assemble.go
carpawell: i wanted them to be separate to drop v1 some time in the future

also, i find it more readable. storing some internal state (v1/v2 split) is harder is read IMO. assembly code it complex enough

pkg/services/object/get/assembly_v2.go
carpawell: oh, that is the reason, yes: https://github.com/nspcc-dev/neofs-node/pull/2716#discussion_r1524678153

returned back

pkg/services/object/get/assembly_v2.go
carpawell: changed retuned bool in one place and added comments everywhere

pkg/core/object/fmt.go
cthulhu-rider: this is a good dream, so lets accept context prm then

cthulhu-rider (approved):

pkg/core/object/fmt.go
```diff
@@ -49,6 +51,18 @@ type Locker interface {
...
+// SplitVerifier represent split validation unit. It verifies V2 split
+// chains based on the link object's payload (list of ID+ObjectSize pairs).
+type SplitVerifier interface {
+	// VerifySplit must verify split hierarchy and return any error that did
+	// not allow processing the chain. Must break (if possible) any internal
+	// computations if context is done. The second and the third args are the
+	// first part's address used as a chain unique identifier that also must
+	// be checked. The fourth arg is guaranteed to be the full list from the
+	// link's payload without item order change.
+	VerifySplit(context.Context, cid.ID, oid.ID, []object.MeasuredObject) error
```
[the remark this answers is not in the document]
carpawell: i think i have lost my answer here somehow... it was smth about "can we make this distinguishing any useful?"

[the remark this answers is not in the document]
cthulhu-rider: these cases differ:
1. we got all we needed and chain is invalid. This is a persistent fail
2. we timed out to request smth, this does not matter. Retry could succeed

i realize that right now nobody cares, just fail

Issue #830, opened 2021-09-16
Reconsider storage node healthcheck mechanism


alexvanin:
Storage nodes send bootstrap request once per couple of epochs. Due to https://github.com/nspcc-dev/neofs-node/issues/750 these bootstrap requests will produce fallback transactions, which are totally unnecessary.

cthulhu-rider:
Maybe related to netmap local view?

IR modifies source TX, so user TX doesn't exec. Solution?

Postpone to notary  only?

roman-khimov:
@carpawell, can you please remind me if we're still rewriting the original transaction in the handler?

carpawell:
@roman-khimov, hmmm, it seems like we are still dropping the original TX and [creating](https://github.com/nspcc-dev/neofs-node/blob/df2cc08a08191ce0f6ff84f36c393ed9b26f20da/pkg/innerring/processors/netmap/process_peers.go#L74-L80) a new one with the alphabet witness only. Moreover, we still have method [duplication](https://github.com/nspcc-dev/neofs-contract/blob/9dcf82fc4045b959181e1bb509a8dff31725e5a0/contracts/netmap/contract.go#L255-L290) that adapted contracts to the previous bootstraps.
So we have done some optimizations with overwriting original TXs but have not fixed other kludges.

PR #2717, opened 2024-01-12
ir: Do not create new notary requests for SN netmap entrances

carpawell:
See #2612, https://github.com/nspcc-dev/neofs-contract/pull/324. Refs #830.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Changelog for NeoFS Node
 
 ### Changed
 - Created files are not group writable (#2589)
+- IR does not create new notary requests for the SN's bootstraps but signs the received ones instead (#2717)
 
 ### Removed
 - Deprecated `neofs-adm [...] inspect` commands (#2603)
--- a/pkg/innerring/processors/netmap/process_peers.go
+++ b/pkg/innerring/processors/netmap/process_peers.go
@@ -3,7 +3,6 @@ package netmap
 import (
 	"encoding/hex"
 
-	netmapclient "github.com/nspcc-dev/neofs-node/pkg/morph/client/netmap"
 	netmapEvent "github.com/nspcc-dev/neofs-node/pkg/morph/event/netmap"
 	"github.com/nspcc-dev/neofs-sdk-go/netmap"
 	"go.uber.org/zap"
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Changelog for NeoFS Node
 
 ### Changed
 - Created files are not group writable (#2589)
+- IR does not create new notary requests for the SN's bootstraps but resigns the received ones instead (#2717)
```
roman-khimov, line 18: signs

roman-khimov (approved):
It certainly fixes #830. Contracts can be improved, but gradually and in a separate issue, we need to ensure only one method is called for at least a release and then we can remove the other one.

pkg/innerring/processors/netmap/process_peers.go
```diff
@@ -62,26 +61,9 @@ func (np *Processor) processAddPeer(ev netmapEvent.AddPeer) {
...
-			originalRequest.MainTransaction.Nonce,
-			nil,
-			methodAddPeerNotary,
-			nodeInfoBinary,
-		)
-
+		err = np.netmapClient.Morph().NotarySignAndInvokeTX(ev.NotaryRequest().MainTransaction)
 		if err != nil {
-			np.log.Error("can't invoke netmap.AddPeer", zap.Error(err))
+			np.log.Error("can't sign notary netmap.AddPeer", zap.Error(err))
```
cthulhu-rider, line 66: `send` matters too
```suggestion
			np.log.Error("can't sign and send notary request calling netmap.AddPeer", zap.Error(err))
```

cthulhu-rider (approved):
i'd prefer to not embody commit by issues only. Text and other commit refs are more reliable, issue refs complement

pkg/innerring/processors/netmap/process_peers.go
```diff
@@ -62,26 +61,9 @@ func (np *Processor) processAddPeer(ev netmapEvent.AddPeer) {
...
-		err = np.netmapClient.Morph().NotaryInvoke(
-			np.netmapClient.ContractAddress(),
-			0,
-			originalRequest.MainTransaction.Nonce,
-			nil,
-			methodAddPeerNotary,
-			nodeInfoBinary,
-		)
-
+		err = np.netmapClient.Morph().NotarySignAndInvokeTX(ev.NotaryRequest().MainTransaction)
```
cthulhu-rider, line 64: btw can we use `originalRequest` here? code will reflect "original request" processing more clear

carpawell: i do not see any api about signing notary requests in `neo-go`. it is always TXs

cthulhu-rider: i mean we have https://github.com/nspcc-dev/neofs-node/blob/de84b12e02dc6fe5bc564928d74398b2a32e26d6/pkg/innerring/processors/netmap/process_peers.go#L21-L22

they are the same as `ev.NotaryRequest()` right? seems like that. If so, it's worth to pass `tx` or `originalRequest.MainTransaction` them as method arg

Issue #2559, opened 2023-09-08
epoch stale after mainchain rpc massive restart


532910:
to reproduce: stop all mainchain rpc (for upgrade for exmaple), then start

workaround: restart all ir nodes

roman-khimov:
Container can't be created as well in this case until all nodes are restarted.

carpawell:
> stop all mainchain rpc

@532910, how long RPCs were down? Seconds? Minutes?

> until all nodes are restarted

@roman-khimov, what exact nodes? IR? SN?

532910:
> how long

let start with minutes

> what exact nodes? IR? SN?

IR, I believe

carpawell:
No progress for now: one node local consensus does not allow reproducing. The best I can do here is wait until the next update and see the logs/profiles. If that is unacceptable I may try the local 4/7 nodes consensus.

roman-khimov:
Can't be reproduced at this stage, waiting for another case in some network.

roman-khimov:
Seems to be reproducible if 2/3 RPC nodes to go offline for some time.

evgeniiz321:
Got a stable reproduction during new payment tests `tests.payment.test_container_payments.TestContainerPayments#test_container_payments`:  https://rest.fs.neo.org/HXSaMJXk2g8C14ht8HSi7BBaiYZ1HeWh2xnWPGQCg4H6/477-1722817756/index.html#suites/44e11ced39071e8d0cfc11c5b94622ba/d3f9a85ef9ecb46a/.

carpawell:
@evgeniiz321, as I understand, in _your case_ you want to make the test be faster so you change epoch duration to 20 blocks, right? I do not see any epoch ticks after changing so it is impossible to change epoch duration immediately, we do not have notifications about config changes and cannot recalculate the next block for epoch handling. As I understand you wait for a new epoch no longer than 60 seconds, probably, a new epoch will not happen if you used 240-second epoch before (1-second block for the default 240 epoch duration). Can you either try to tick epoch after network setting tuning or increase max awaiting to 240?

carpawell:
We have not seen this exact issue for so long so I close it, it can be reopened once it happens at least once.

carpawell:
Still was not the case. See https://github.com/nspcc-dev/neofs-node/issues/3007, it is more related to the situation.

PR #2720, opened 2024-01-17
morph: Fix subscription deadlock

carpawell:
Scenario:

0. at least one subscription has been performed
1. another subscription is being done
2. a notification from one of the `0.` point's subs is received

If `2.` happens b/w `0.` and `1.` a deadlock appears since the notification routing process is locked on the subscription lock while the subscription lock cannot be unlocked since the subscription RPC cannot be done before the just-arrived notification is handled (read from the neo-go subscription channel).
Relates #2559.

### DIFF
--- a/pkg/morph/client/notifications.go
+++ b/pkg/morph/client/notifications.go
@@ -206,8 +206,6 @@ func (c *Client) Notifications() (<-chan *state.ContainedNotificationEvent, <-ch
 }
 
 type subscriptions struct {
-	sync.RWMutex
-
 	// notification consumers (Client sends
 	// notifications to these channels)
 	notifyChan chan *state.ContainedNotificationEvent
@@ -219,6 +217,8 @@ type subscriptions struct {
 	curNotifyChan chan *state.ContainedNotificationEvent
 	curBlockChan  chan *block.Block
 	curNotaryChan chan *result.NotaryRequestEvent
+
+	sync.RWMutex // for subscription fields only
 
 	// cached subscription information
 	subscribedEvents            map[util.Uint160]struct{}
@@ -237,11 +237,11 @@ func (c *Client) routeNotifications() {
 routeloop:
 	for {
 		var connLost bool
-		c.subs.RLock()
+		c.switchLock.RLock()
 		notifCh := c.subs.curNotifyChan
 		blCh := c.subs.curBlockChan
 		notaryCh := c.subs.curNotaryChan
-		c.subs.RUnlock()
+		c.switchLock.RUnlock()
 		select {
 		case <-c.closeChan:
 			break routeloop
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Changelog for NeoFS Node
 - Incorrect address mapping of the Alphabet contracts in NNS produced by deployment procedure (#2713)
 - IR compares and logs public keys difference, not hash of keys difference at SN verification check (#2711)
 - Incorrect handling of notary request leading to inability to collect signatures in some cases (#2715)
+- Deadlock in autodeploy routine (#2720)
 
 ### Changed
 - Created files are not group writable (#2589)


pkg/morph/client/notifications.go
```diff
@@ -120,22 +120,23 @@ func (c *Client) ReceiveNotaryRequests(txSigner util.Uint160) error {
 		return ErrConnectionLost
 	}
 
-	c.subs.Lock()
-	defer c.subs.Unlock()
```
roman-khimov, line 124: It's very recent code, likely not the cause of #2559.

pkg/morph/client/notifications.go
```diff
@@ -120,22 +120,23 @@ func (c *Client) ReceiveNotaryRequests(txSigner util.Uint160) error {
...
+		c.subs.RUnlock()
 		return nil
 	}
 
-	if _, ok := c.subs.subscribedNotaryEvents[txSigner]; ok {
-		return nil
-	}
+	c.subs.RUnlock()
 
 	_, err := c.client.ReceiveNotaryRequests(&neorpc.TxFilter{Signer: &txSigner}, c.subs.curNotaryChan)
```
roman-khimov, line 132: Can we have two threads here?

pkg/morph/client/notifications.go
```diff
@@ -120,22 +120,23 @@ func (c *Client) ReceiveNotaryRequests(txSigner util.Uint160) error {
 		return ErrConnectionLost
 	}
 
-	c.subs.Lock()
-	defer c.subs.Unlock()
+	c.subs.RLock()
 
-	if c.subs.subscribedToAllNotaryEvents {
+	if _, ok := c.subs.subscribedNotaryEvents[txSigner]; ok || c.subs.subscribedToAllNotaryEvents {
```
cthulhu-rider, line 125: no need to touch `subscribedNotaryEvents` if `subscribedToAllNotaryEvents`

cthulhu-rider (commented):
i havent dived deeply to the problem yet, but...

to me its ok that subscribing ops acquire subs lock. And they must not cause any deadlocks. If a deadlock happens - it's almost definitely the routing/handling fault and should be resolve there

pkg/morph/client/notifications.go
carpawell: yeap, that is why i do not say it closes smth. but that is the reason (and the only one) for the recent deadlock that I have found (the info was not full since I was a little late)

pkg/morph/client/notifications.go
carpawell: do you mean subscribing? two routines that subscribe for smth?

pkg/morph/client/notifications.go
carpawell: is it critical? do not want more lines for not really important things

carpawell:
> to me its ok that subscribing ops acquire subs lock. And they must not cause any deadlocks. If a deadlock happens - it's almost definitely the routing/handling fault and should be resolve there

I would say https://github.com/nspcc-dev/neofs-node/commit/5655887bca9c1a5f275461ff5804b50a17ee7c1a was not necessary. We can add the third lock but that is extremely hard and not necessary for the client, it would do too many things then.

I do not know what to say here, is it the 5th or the 10th time I am fixing a deadlock caused by the impossibility of doing any RPC calls if nothing reads subs channel? neo-go is changing, neofs is changing but still, we face an intuitive regular code that deadlocks.

pkg/morph/client/notifications.go
cthulhu-rider: not critical if this were code from scratch, but it is not. Extra lines are not a problem if they have a reason

cthulhu-rider:
> 5th or the 10th time I am fixing a deadlock caused by the impossibility of doing any RPC calls if nothing reads subs channel

let's get rid of this minefield once and for all. Half actions don't work

temporary unlocked subscription is not a solution to me. Nothing prevents the listener from discarding events that are not yet in demand

carpawell:
> let's get rid of this minefield once and for all

What do you suggest? We need to hold a few connections and be able to switch b/w them so a lock is there. But neo-go requires to read subs. But making a new sub is an RPC that is impossible without reading the subs.

> temporary unlocked subscription is not a solution to me

If you prohibit subscription change, you may experience a connection switch (and a subscription lock) while trying to make a new subscription (another subscription lock).

roman-khimov:
You can make `routeNotifications` not require `s.subs` lock for regular (non-reconnecting) operation. Likely this will solve the problem. Locking here looks correct.

carpawell:
> You can make routeNotifications not require s.subs lock for regular (non-reconnecting) operation. Likely this will solve the problem.

Legit. But I do not like it cause that will work only based on some info _we_ know: no routines change the channels if no RPC failures happen (or they even happen but their handling is locked cause we are reading some internal channels).
Anyway, done. Small diff, fixes the problem.

roman-khimov (approved):
You can do it in another way: acquire channels once, keep them until `connLost`. But this works too.

cthulhu-rider (approved):

roman-khimov (approved):

PR #2727, opened 2024-01-22
fix: Do not allow broken session token by SN

### LINKED ISSUE #2466: Possible incorrect error when a container is deleted by a static session token without the right permissions
I write tests on the issue https://github.com/nspcc-dev/neofs-testcases/issues/579
and I found that neofs-cli container delete handle differently in the negative case when a non-valid wallet is specified.
The difference is which static session token I use - the one created on the container owner's wallet, or the one created on another wallet.

## Expected Behavior
I'm not sure it's exactly a bug, I'd like to discuss what behavior we expect in which case.

1. Create container
2. Create static session token using container owner wallet
3. Create static session token using another wallet
4. Try to delete container using static session token from step 2 - you'll get a "timed out after 90 seconds" error.
5. Try to delete container using static session token from step 3 - you'll get a "session issuer differs with the container owner" error.

See allure reports:

[report.zip](https://github.com/nspcc-dev/neofs-node/files/12232955/report.zip)
[report_2.zip](https://github.com/nspcc-dev/neofs-node/files/12232956/report_2.zip)

Test code:
```
    @allure.title("Not owner and not trusted party can NOT delete container")
    def test_static_session_token_container_delete_only_trusted_party_proved_by_the_container_owner(
            self,
            owner_wallet: WalletFile,
            user_wallet: WalletFile,
            stranger_wallet: WalletFile,
            scammer_wallet: WalletFile,
            static_sessions: dict[ContainerVerb, str],
            temp_directory: str,
            not_owner_wallet,
    ):
        with allure.step("Create container"):
            cid = create_container(
                owner_wallet.path,
                shell=self.shell,
                endpoint=self.cluster.default_rpc_endpoint,
            )

        user_token = self.static_session_token(owner_wallet, user_wallet, self.shell, temp_directory)
        stranger_token = self.static_session_token(user_wallet, stranger_wallet, self.shell, temp_directory)

        with allure.step("Try to delete container using scammer token"):
            with pytest.raises(RuntimeError, match=NOT_SESSION_CONTAINER_OWNER):
                delete_container(
                    wallet=scammer_wallet.path,
                    cid=cid,
                    session_token=stranger_token[ContainerVerb.DELETE],
                    shell=self.shell,
                    endpoint=self.cluster.default_rpc_endpoint,
                    await_mode=True,
                )

        with allure.step("Try to delete container using scammer token"):
            with pytest.raises(RuntimeError, match=NOT_SESSION_CONTAINER_OWNER):
                delete_container(
                    wallet=scammer_wallet.path,
                    cid=cid,
                    session_token=user_token[ContainerVerb.DELETE],
                    shell=self.shell,
                    endpoint=self.cluster.default_rpc_endpoint,
                    await_mode=True,
                )
```

## Your Environment
neo-go
0.101.1
neofs-s3-authmate
0.27.1
neofs-cli
0.36.1-125-g8cda197e
neofs-adm
0.36.1-125-g8cda197e
AWS
aws-cli/2.9.19 Python/3.11.4 Linux/6.2.0-25-generic source/x86_64.ubuntu.23 prompt/off

carpawell:
If container session token is incorrect, there is no need in Alphabet's check, the error can be returned immediately to a client; otherwise he has to wait for the TX to persist infinitely with not feedback. Closes #2466.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Changelog for NeoFS Node
 - Created files are not group writable (#2589)
 - IR does not create new notary requests for the SN's bootstraps but signs the received ones instead (#2717)
 - IR can handle third-party notary requests without reliance on receiving the original one (#2715)
+- SN validates container session token's issuer to be container's owner (#2466)
 
 ### Removed
 - Deprecated `neofs-adm [...] inspect` commands (#2603)
--- a/pkg/services/container/morph/executor.go
+++ b/pkg/services/container/morph/executor.go
@@ -1,13 +1,16 @@
 package container
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
 
+	"github.com/mr-tron/base58"
 	"github.com/nspcc-dev/neofs-api-go/v2/container"
 	"github.com/nspcc-dev/neofs-api-go/v2/refs"
 	sessionV2 "github.com/nspcc-dev/neofs-api-go/v2/session"
+	"github.com/nspcc-dev/neofs-api-go/v2/util/signature"
 	containercore "github.com/nspcc-dev/neofs-node/pkg/core/container"
 	containerSvc "github.com/nspcc-dev/neofs-node/pkg/services/container"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
@@ -80,6 +83,11 @@ func (s *morphExecutor) Put(_ context.Context, tokV2 *sessionV2.Token, body *con
 		if err != nil {
 			return nil, fmt.Errorf("invalid session token: %w", err)
 		}
+
+		err = s.validateToken(tokV2, nil, sessionV2.ContainerVerbPut)
+		if err != nil {
+			return nil, fmt.Errorf("session token validation: %w", err)
+		}
 	}
 
 	idCnr, err := s.wrt.Put(cnr)
@@ -119,6 +127,11 @@ func (s *morphExecutor) Delete(_ context.Context, tokV2 *sessionV2.Token, body *
 		err := tok.ReadFromV2(*tokV2)
 		if err != nil {
 			return nil, fmt.Errorf("invalid session token: %w", err)
+		}
+
+		err = s.validateToken(tokV2, body.GetContainerID(), sessionV2.ContainerVerbDelete)
+		if err != nil {
+			return nil, fmt.Errorf("session token validation: %w", err)
 		}
 	}
 
@@ -228,6 +241,11 @@ func (s *morphExecutor) SetExtendedACL(_ context.Context, tokV2 *sessionV2.Token
 		if err != nil {
 			return nil, fmt.Errorf("invalid session token: %w", err)
 		}
+
+		err = s.validateToken(tokV2, body.GetEACL().GetContainerID(), sessionV2.ContainerVerbSetEACL)
+		if err != nil {
+			return nil, fmt.Errorf("session token validation: %w", err)
+		}
 	}
 
 	err = s.wrt.PutEACL(eaclInfo)
@@ -273,4 +291,73 @@ func (s *morphExecutor) GetExtendedACL(_ context.Context, body *container.GetExt
 	res.SetSessionToken(tokV2)
 
 	return res, nil
+}
+
+type sessionDataSource struct {
+	t    *sessionV2.Token
+	size int
+}
+
+func (d sessionDataSource) ReadSignedData(buff []byte) ([]byte, error) {
+	if len(buff) < d.size {
+		buff = make([]byte, d.size)
+	}
+
+	res := d.t.GetBody().StableMarshal(buff)
+
+	return res[:d.size], nil
+}
+
+func (d sessionDataSource) SignedDataSize() int {
+	return d.size
+}
+
+func newDataSource(t *sessionV2.Token) sessionDataSource {
+	return sessionDataSource{
+		t:    t,
+		size: t.GetBody().StableSize(),
+	}
+}
+
+func (s *morphExecutor) validateToken(t *sessionV2.Token, cIDV2 *refs.ContainerID, op sessionV2.ContainerSessionVerb) error {
+	c := t.GetBody().GetContext()
+	cc, ok := c.(*sessionV2.ContainerSessionContext)
+	if !ok {
+		return errors.New("session is not container-related")
+	}
+
+	if verb := cc.Verb(); verb != op {
+		return fmt.Errorf("wrong container session operation: %s", verb)
+	}
+
+	err := signature.VerifyDataWithSource(newDataSource(t), t.GetSignature)
+	if err != nil {
+		return fmt.Errorf("incorrect token signature: %w", err)
+	}
+
+	if cIDV2 == nil { // can be nil for PUT or wildcard may be true
+		return nil
+	}
+
+	if sessionCID := cc.ContainerID().GetValue(); !bytes.Equal(sessionCID, cIDV2.GetValue()) {
+		return fmt.Errorf("wrong container: %s", base58.Encode(sessionCID))
+	}
+
+	var cID cid.ID
+
+	err = cID.ReadFromV2(*cIDV2)
+	if err != nil {
+		return fmt.Errorf("invalid container ID: %w", err)
+	}
+
+	cnr, err := s.rdr.Get(cID)
+	if err != nil {
+		return fmt.Errorf("reading container from the network: %w", err)
+	}
+
+	if issuer := t.GetBody().GetOwnerID().GetValue(); !bytes.Equal(cnr.Value.Owner().WalletBytes(), issuer) {
+		return fmt.Errorf("session was not issued by the container owner, issuer: %q", issuer)
+	}
+
+	return nil
 }
--- a/pkg/services/container/morph/executor_test.go
+++ b/pkg/services/container/morph/executor_test.go
@@ -2,6 +2,7 @@ package container_test
 
 import (
 	"context"
+	"crypto/sha256"
 	"testing"
 
 	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ Changelog for NeoFS Node
 - Created files are not group writable (#2589)
 - IR does not create new notary requests for the SN's bootstraps but signs the received ones instead (#2717)
 - IR can handle third-party notary requests without reliance on receiving the original one (#2715)
+- SN validates container session token's issuer to be container's owner (#2466)
```
cthulhu-rider, line 22: not only ownership is validated

pkg/services/container/morph/executor.go
```diff
@@ -80,6 +82,11 @@ func (s *morphExecutor) Put(_ context.Context, tokV2 *sessionV2.Token, body *con
 		if err != nil {
 			return nil, fmt.Errorf("invalid session token: %w", err)
 		}
+
+		err = s.validateToken(tokV2, nil, sessionV2.ContainerVerbPut)
+		if err != nil {
+			return nil, fmt.Errorf("token validation: %w", err)
```
cthulhu-rider, line 88: if client-facing, then
```suggestion
			return nil, fmt.Errorf("session token validation: %w", err)
```

pkg/services/container/morph/executor.go
```diff
@@ -274,3 +291,41 @@ func (s *morphExecutor) GetExtendedACL(_ context.Context, body *container.GetExt
...
+
+func (s *morphExecutor) validateToken(t *sessionV2.Token, cID *refs.ContainerID, op sessionV2.ContainerSessionVerb) error {
+	c := t.GetBody().GetContext()
+	cc, ok := c.(*sessionV2.ContainerSessionContext)
+	if !ok {
+		return errors.New("session is not container related")
+	}
+
+	if verb := cc.Verb(); verb != op {
+		return fmt.Errorf("wrong container session operation: %s", verb)
```
cthulhu-rider, line 303: error should not contain input imo cuz client knows what he sent (same for other cases)

pkg/services/container/morph/executor.go
```diff
@@ -274,3 +291,41 @@ func (s *morphExecutor) GetExtendedACL(_ context.Context, body *container.GetExt
 
 	return res, nil
 }
+
+func (s *morphExecutor) validateToken(t *sessionV2.Token, cID *refs.ContainerID, op sessionV2.ContainerSessionVerb) error {
+	c := t.GetBody().GetContext()
+	cc, ok := c.(*sessionV2.ContainerSessionContext)
+	if !ok {
+		return errors.New("session is not container related")
```
cthulhu-rider, line 299:
```suggestion
		return errors.New("session is not container-related")
```

pkg/services/container/morph/executor.go
```diff
@@ -274,3 +291,41 @@ func (s *morphExecutor) GetExtendedACL(_ context.Context, body *container.GetExt
...
+		return fmt.Errorf("cID converting: %w", err)
+	}
+
+	cnr, err := s.rdr.Get(cIDSDK)
+	if err != nil {
+		return fmt.Errorf("reading container from the network: %w", err)
+	}
+
+	if issuer := t.GetBody().GetOwnerID().GetValue(); !bytes.Equal(cnr.Value.Owner().WalletBytes(), issuer) {
+		return fmt.Errorf("session session was issued not by the owner, issuer: %x", issuer)
```
cthulhu-rider, line 327:
```suggestion
		return fmt.Errorf("session was not issued by the container owner %s", cnr.Value.Owner())
```

pkg/services/container/morph/executor.go
```diff
@@ -274,3 +291,41 @@ func (s *morphExecutor) GetExtendedACL(_ context.Context, body *container.GetExt
...
+		return nil
+	}
+
+	if sessionCID := cc.ContainerID().GetValue(); !bytes.Equal(sessionCID, cID.GetValue()) {
+		return fmt.Errorf("wrong container: %s", base58.Encode(sessionCID))
+	}
+
+	var cIDSDK cid.ID
+	err := cIDSDK.ReadFromV2(*cID)
+
```
cthulhu-rider, line 316: not needed linebreak

pkg/services/container/morph/executor.go
```diff
@@ -274,3 +291,41 @@ func (s *morphExecutor) GetExtendedACL(_ context.Context, body *container.GetExt
...
+	if cID == nil { // can be nil for PUT
+		return nil
+	}
+
+	if sessionCID := cc.ContainerID().GetValue(); !bytes.Equal(sessionCID, cID.GetValue()) {
+		return fmt.Errorf("wrong container: %s", base58.Encode(sessionCID))
+	}
+
+	var cIDSDK cid.ID
```
cthulhu-rider, line 314: seeing `SDK` here confuses me. There is only one type - container ID

pkg/services/container/morph/executor.go
```diff
@@ -274,3 +291,41 @@ func (s *morphExecutor) GetExtendedACL(_ context.Context, body *container.GetExt
...
+	err := cIDSDK.ReadFromV2(*cID)
+
+	if err != nil {
+		return fmt.Errorf("cID converting: %w", err)
```
cthulhu-rider, line 318:
```suggestion
		return fmt.Errorf("invalid container ID: %w", err)
```

pkg/services/container/morph/executor.go
```diff
@@ -274,3 +291,41 @@ func (s *morphExecutor) GetExtendedACL(_ context.Context, body *container.GetExt
 
 	return res, nil
 }
+
+func (s *morphExecutor) validateToken(t *sessionV2.Token, cID *refs.ContainerID, op sessionV2.ContainerSessionVerb) error {
```
cthulhu-rider, line 295: worth unit testing

pkg/services/container/morph/executor.go
```diff
@@ -274,3 +291,41 @@ func (s *morphExecutor) GetExtendedACL(_ context.Context, body *container.GetExt
...
+	cnr, err := s.rdr.Get(cIDSDK)
+	if err != nil {
+		return fmt.Errorf("reading container from the network: %w", err)
+	}
+
+	if issuer := t.GetBody().GetOwnerID().GetValue(); !bytes.Equal(cnr.Value.Owner().WalletBytes(), issuer) {
+		return fmt.Errorf("session session was issued not by the owner, issuer: %x", issuer)
+	}
+
+	return nil
```
cthulhu-rider, line 330: why are not all fields checked?

cthulhu-rider (commented):
talking about system testing: with these changes, broken sessions' verification by the IR wont be tested anymore since SNs wont bypass such sessions, right? IR tests are critical, so we must not lose them

CHANGELOG.md
carpawell: do you mean providing a list there? the issue has the exact problem described

pkg/services/container/morph/executor.go
carpawell: changed but the user will not see it because of how we unwrap the errors (status gets the lowest one only)

pkg/services/container/morph/executor.go
carpawell: > client knows what he sent

it sounds to me "error doer knows that he does an error". it can be a bug inside our node, or it can be a bug on the client side, i do not think it is critical to mention that session has a verb that does not relate to the operation it is attached to

pkg/services/container/morph/executor.go
carpawell: this func already has `cID` and it is from api-go. but `morphExecutor` can also get container by cID from the SDK (and I need it here) but SDK struct is required for it. what bothers you?

pkg/services/container/morph/executor.go
carpawell: not valuable to me cause i had to fix tests a little (the only one in that pkg named "InvalidToken") to make them feel good. also, that is V2 level so most of the tests look like "create byte array one, create byte array two, make sure byte array one does not equal the second one". but ok, added

pkg/services/container/morph/executor.go
carpawell: what else fields to check? that is not IR validation before we try to change smth, just a logical validation that you are not trying to make totally incorrect things (container owner do not know about your operation but you still trying it). did not want to copy IR behavior

epoch can not be ensured: the final check will be done on the IR side, no one knows at what state and when

carpawell:
> with these changes, broken sessions' verification by the IR wont be tested anymore since SNs wont bypass such sessions, right?

Seems so. But that is how the issues like "do not want to wait too long before IR reacts, pls, make it on the SN side" are fixed. Do not see any problems here.

> IR tests are critical, so we must not lose them

Can be checked with `neo-go` and notary requests anyway.

pkg/services/container/morph/executor.go
cthulhu-rider: there are other stateless conditions like `lifetime` field presence, `signature` correctness, etc. And some stateful conditions are consistent, e.g. expiraiton claim

pkg/services/container/morph/executor.go
carpawell: i can understand you but still think issuer is smth that can be lost (and a token itself) while container owner is a constant. also, i would say that errors should highlight mistakes (container owner can not be an error, but incorrect issuer can)

pkg/services/container/morph/executor.go
carpawell: > there are other stateless conditions like lifetime field presence

i would say it is more like a protocol validation (if we are talking about the presence, and epoch values i wouldn't check ever by the intermediate SN ever, as I've already said)

> signature correctness

well... ok, done but not valuable to me

pkg/services/container/morph/executor.go
cthulhu-rider: > issuer is smth that can be lost

suggest leaving this to the transport security layer

ok, i dont insist, for now this is negligible data

CHANGELOG.md
cthulhu-rider: imo issue is just a ref, changelog contains meaningful changes

```suggestion
- SN now validates container session tokens (#2466)
```

pkg/services/container/morph/executor.go
cthulhu-rider: > well... ok, done but not valuable to me

words of some undefined behavior...

cthulhu-rider:
> Can be checked with neo-go and notary requests anyway.

not everybody will be ready for this. I'd add a remark to `Updating` section of the changelog

carpawell:
> I'd add a remark to Updating section of the changelog

Not sure what an updater should see in this line. Not sure how @532910 will react. But I tried a little.

cthulhu-rider:
> But I tried a little.

anything i can look at?

cthulhu-rider (approved):

carpawell:
> anything i can look at?

Oh, has not pushed, my bad.

PR #2737, opened 2024-02-09
CHANGELOG: release 0.40.0

roman-khimov:


carpawell (approved):
Do we still support debian pkg build? Release docs say about its version changelog file but it is 0.33 currently.
Also, your branch is not `release/*`. Neither of the ways is cool or not cool to me, but we have docs for some reason in this repo.

The whole comment is more about our docs, not this PR.

PR #2740, opened 2024-02-13
neofs-adm: use registerTLD to register TLD

roman-khimov:
It's the only way since neofs-contract 0.18.0, "register" just fails with

    Error: could not perform test invocation: script failed (FAULT state) due to an error: at instruction 1794 (THROW): unhandled exception: "TLD denied"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Changelog for NeoFS Node
 ### Added
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


cthulhu-rider (approved):

carpawell (approved):
Commit body format looks strange

roman-khimov:
It's a quote, that's why.

Issue #2744, opened 2024-02-21
Reconsider status responses


carpawell:
Some status errors in NeoFS are known and self-describing. But there are also "general errors" with 1024 status codes and some "good" human-readable messages.

## Expected Behavior
If I have a general error, I expect NOT a general message that can help me understand what is happening. E.g. session token does not relate to the container a request is trying to change: 
```
status: code = 1024 message = could not execute SetEACL request: session token validation: wrong container: 
```

## Current Behavior
Only the last error in the errors chain is attached:
```
status: code = 1024 message = wrong container: 
```

## Possible Solution
1. [errors.Join](https://go.dev/pkg/errors/#Join) for meaningful errros inside a package? (like session tokens errors can be wrapped with a common session token error with detailed errors so the `Unwrap` does not unwrap more than it is needed as the package thinks)
2. Make all the low-level errors well described and have more context than they have now? (can be a huge work)
3. Attach the full error chain if it is a 1024 error? (chain could be too long and too golang specific)
4. Attach 1-2 more errors from the lowest one? (a strange rule can fix smth but not all the cases)

## Steps to Reproduce (for bugs)
Any 1024 "general" error case (like incorrect session token; even though session token error may become another status case, there always be some "general" errors that should be described correctly).

## Context
https://github.com/nspcc-dev/neofs-node/blob/996b1eae7bd1f1ad5693758b38c9c025523842ac/pkg/services/util/sign.go#L225-L229

## Regression
No.

## Your Environment
0.40.0

roman-khimov:
Server should return as much data as it can (except when it somehow affects security), simple as that. Joining/chaining errors is fine for that and the protocol allows for it, `message` can be anything. Then this message should be a part of client error.

carpawell:
@roman-khimov, so you are for the full error message?

cthulhu-rider:
big :+1: to this proposal

pretty simple approach comes to my mind:
1. server always returns full error context in text message, i.e. drop mentioned pre-unwrapping
2. CLI checks the code first, and prints clear output according to it. If there are some details, CLI adds them just like we add context to Go errors. For example, for [`2048`](https://github.com/nspcc-dev/neofs-api/blob/e360d265f85182fc19aa890b5b26b65430ff9018/status/types.proto#L108-L111) w/o details, CLI just types:
```
2048 Fobidden
```
or, if status has detail:
```
2048 Fobidden
access to operation OBJECT_PUT is denied by basic ACL check
```
[...]

PR #2747, opened 2024-02-21
Fix/session token errors details

carpawell:


pkg/services/container/morph/executor.go
```diff
@@ -361,7 +361,7 @@ func (s *morphExecutor) validateToken(t *sessionV2.Token, cIDV2 *refs.ContainerI
 
 	err = keyFromToken.Decode(t.GetSignature().GetKey())
 	if err != nil {
-		return errors.New("error while decoding public key from the token's signer")
+		return fmt.Errorf("error while decoding public key from the token's signer: %w", err)
```
roman-khimov, line 364: `decoding key from signer:`? It's rather obvious that it's an error and likely we know that it happens in token, messages better be concise.

carpawell: > we know that it happens in token

hehehehe (https://github.com/nspcc-dev/neofs-node/issues/2744)

> decoding key from signer:?

don't mind, fixed

roman-khimov (approved):

PR #2748, opened 2024-02-21
acl: Perceive tombstone saving as delete operation

cthulhu-rider:


roman-khimov (approved):
Needs to be rebased onto 0.40.1, but looks good otherwise (needs a hell lot of testing though).

carpawell (approved):

PR #2742, opened 2024-02-14
eacl: Support new operators

cthulhu-rider:
- [x]  IR verification
- [x] SN processing (covered by https://github.com/nspcc-dev/neofs-sdk-go/pull/554)
- [x] CLI support

all ready, waiting for SDK merge

PR #2749, opened 2024-02-22
object: Limit header length

cthulhu-rider:
* based on and blocked by https://github.com/nspcc-dev/neofs-api/issues/262


pkg/core/object/fmt.go
```diff
@@ -12,6 +12,13 @@ import (
 	"github.com/nspcc-dev/neofs-sdk-go/storagegroup"
 )
 
+// MaxHeaderLen is a maximum allowed length of binary object header to be
+// created via NeoFS API protocol.
+const MaxHeaderLen = 16 << 10
```
roman-khimov, line 17: Belongs to SDK to me.

cthulhu-rider: now there

cmd/neofs-cli/modules/util/acl.go
```diff
@@ -140,7 +140,6 @@ func eaclFiltersToString(fs []eacl.Filter) string {
 			_, _ = tw.Write([]byte("\t==\t"))
 		case eacl.MatchStringNotEqual:
 			_, _ = tw.Write([]byte("\t!=\t"))
-		case eacl.MatchUnknown:
```
roman-khimov, line 143: Can be left as is not requiring subsequent linter changes.

cthulhu-rider: i bet it was previously added to satisfy the linter. Dont see any reason to have empty case

roman-khimov: Well, you either have this line to satisfy the linter, or another line to do the same thing. If there is no real difference, whatever we already have wins to me. We just don't gain anything with the change.

cthulhu-rider: > have this line to satisfy the linter

having it dont fix the linter, otherwise i wouldnt touch this code place at all

roman-khimov (approved):

pkg/core/object/fmt.go
carpawell: will it be added to SDK?

carpawell: will it be added to SDK?

cthulhu-rider: > now there

already

carpawell: > now there

oh, got it wrong

cmd/neofs-cli/modules/util/acl.go
carpawell: i strongly disagree with dropping cases and using `nolint` for them. in general, it shows that you skip smth intentionally and does not require to think about it if you are looking for a bug. not critical though

carpawell (approved):

carpawell:
But conflicts.

cmd/neofs-cli/modules/util/acl.go
roman-khimov: Likely can be solved with `default`, btw, one line, a bit more meaning than `//nolint`.

carpawell: so think i: https://github.com/nspcc-dev/neofs-node/pull/2742#discussion_r1507113032

roman-khimov: https://github.com/nspcc-dev/go-ordered-json/pull/12/commits/c406b37478706212bddd8de3434a44e76a08ef59

The other option of course is to not use `exhaustive` like in https://github.com/nspcc-dev/neo-go/blob/24041451335d2f6ff64b8e334401d2dff7c3b211/.golangci.yml#L44

cthulhu-rider: i don't realize the benefit of `exhaustive`. It helped me 0 times, but swears all the time

carpawell: it helped me with updating for sure. if you add more enums, the linter does not allow you to skip places where enums are used, you _have_ to look at them and be sure they are ok or extend switches

PR #2753, opened 2024-02-26
engine: Support reading objects without decoding

cthulhu-rider:


pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -297,6 +299,62 @@ func (t *FSTree) Get(prm common.GetPrm) (common.GetRes, error) {
...
+		return b, fmt.Errorf("read all %d bytes from object file %q: %w", sz, p, err)
+	}
+
+	if !t.IsCompressed(b) {
+		return b, nil
+	}
+
+	dec, err := t.DecompressForce(b)
+	if err != nil {
+		if cap(dec) > cap(b) {
```
carpawell, line 349: what is it for?

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -297,6 +299,62 @@ func (t *FSTree) Get(prm common.GetPrm) (common.GetRes, error) {
...
+	}
+
+	var b []byte
+	if opts.GetBuffer != nil {
+		b = opts.GetBuffer(int(sz))
+	} else {
+		b = make([]byte, sz)
+	}
+
+	_, err = io.ReadFull(f, b)
```
carpawell, line 335: why not streaming?

carpawell (commented):
BTW, not valuable to me as a separate PR. Not clear why we need it, operation params are questionable, the issue does not describe why this should be done (it appeared earlier than the one we are trying to resolve now).

pkg/local_object_storage/blobstor/fstree/fstree.go
cthulhu-rider: biggest allocated buffer is returned

pkg/local_object_storage/blobstor/fstree/fstree.go
cthulhu-rider: u mean return `io.Reader`? no need for now, may be added later

cthulhu-rider:
> Not clear why we need it

tried to explain in the commit. This is 100%-needed functionality for any storage IMO - just get what u store as `[]byte`

pkg/local_object_storage/blobstor/common/get.go
```diff
@@ -15,3 +15,11 @@ type GetRes struct {
 	Object  *objectSDK.Object
 	RawData []byte
 }
+
+// GetBytesOptions groups options for [Storage.GetBytes].
+type GetBytesOptions struct {
```
roman-khimov, line 20: Please, let's not add more `...Options` if we're not talking about tens of them. Functions have arguments, that's what they're for.

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -297,6 +299,62 @@ func (t *FSTree) Get(prm common.GetPrm) (common.GetRes, error) {
...
+		}
+		return nil, fmt.Errorf("open object file %q: %w", p, err)
+	}
+
+	fi, err := f.Stat()
+	if err != nil {
+		return nil, fmt.Errorf("stat object file %q: %w", p, err)
+	}
+	sz := fi.Size()
+	if sz > math.MaxInt {
```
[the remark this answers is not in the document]
roman-khimov: Only from garbage in the storage maybe.

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -297,6 +299,62 @@ func (t *FSTree) Get(prm common.GetPrm) (common.GetRes, error) {
...
+	if !t.IsCompressed(b) {
+		return b, nil
+	}
+
+	dec, err := t.DecompressForce(b)
+	if err != nil {
+		if cap(dec) > cap(b) {
+			b = dec
+		}
+		return b, fmt.Errorf("decompress object file data %q: %w", p, err)
```
roman-khimov, line 352: Why returning `b` here?

pkg/local_object_storage/blobstor/common/get.go
```diff
@@ -15,3 +15,11 @@ type GetRes struct {
...
+	// GetBuffer specifies function replacing built-in make to allocate memory
+	// buffer of ln length. Buffer allocated via GetBuffer is returned from GetBytes
+	// regardless of error so caller could handle it properly.
+	GetBuffer func(ln int) []byte
```
roman-khimov, line 24: I'd rather have `buf []byte` passed. In most cases we know the size already.

pkg/local_object_storage/engine/get.go
```diff
@@ -181,3 +182,33 @@ func Get(storage *StorageEngine, addr oid.Address) (*objectSDK.Object, error) {
...
 }
+
+// GetBytes reads object from the StorageEngine by address into memory buffer in
+// a canonical NeoFS binary format. Returns [apistatus.ObjectNotFound] if object
+// is missing.
+func (e *StorageEngine) GetBytes(addr oid.Address, opts common.GetBytesOptions) ([]byte, error) {
+	var b []byte
+	var err error
+	getBufferBase := opts.GetBuffer
+	opts.GetBuffer = func(ln int) []byte {
```
roman-khimov, line 193: Why this override?

pkg/local_object_storage/shard/get.go
```diff
@@ -157,7 +163,48 @@ func (s *Shard) fetchObjectData(addr oid.Address, skipMeta bool, cb storFetcher,
...
+	return s.getBytesWithMetadataLookup(addr, opts, false)
+}
+
+func (s *Shard) getBytesWithMetadataLookup(addr oid.Address, opts common.GetBytesOptions, skipMeta bool) ([]byte, bool, error) {
+	s.m.RLock()
+	defer s.m.RUnlock()
+
+	var b []byte
+	getBufferBase := opts.GetBuffer
+	opts.GetBuffer = func(ln int) []byte {
```
roman-khimov, line 190: And again and again?

roman-khimov (changes requested):
We need better coverage as well.

pkg/local_object_storage/blobstor/fstree/fstree.go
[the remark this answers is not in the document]
cthulhu-rider: otherwise data cannot be fully read

pkg/local_object_storage/blobstor/common/get.go
cthulhu-rider: structured code is always better to me, seeing calls like `call(addr, nil)` makes you just imagine what it means. Required prms are explicit (1 in this case) 

so, i keep proposing this

pkg/local_object_storage/blobstor/fstree/fstree.go
cthulhu-rider: to return buffer to the caller, as written in `GetBytesOptions.GetBuffer`

pkg/local_object_storage/blobstor/common/get.go
cthulhu-rider: we never know it, only inside end storage (fstree, peapod)

pkg/local_object_storage/engine/get.go
cthulhu-rider: to allocate buffer once

pkg/local_object_storage/shard/get.go
cthulhu-rider: same case

pkg/local_object_storage/blobstor/common/get.go
roman-khimov: Is `call(addr, GetBytesOptions{})` better than `call(addr, nil)`?

pkg/local_object_storage/blobstor/common/get.go
roman-khimov: Meta has payload size, we request things after request to metabase IINM.

pkg/local_object_storage/blobstor/fstree/fstree.go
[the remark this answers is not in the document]
carpawell: but how int64 can be bigger than MaxInt? do we target such systems? can we run on such systems?

pkg/local_object_storage/blobstor/fstree/fstree.go
carpawell: we have big problems with memory, i thought that extending the storage interface with row data should be done for memory optimizations mainly

pkg/local_object_storage/blobstor/fstree/fstree.go
carpawell: with a non-nil error? what usecase can be this for?

pkg/local_object_storage/blobstor/fstree/fstree.go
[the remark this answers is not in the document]
cthulhu-rider: on 32-bit systems

> do we target such systems? can we run on such systems?

yes, yes

pkg/local_object_storage/blobstor/fstree/fstree.go
cthulhu-rider: we allocated 64M buffer, got some logical error, but still wanna reuse such a big buffer

pkg/local_object_storage/blobstor/fstree/fstree.go
cthulhu-rider: > for memory optimizations

it is: as i mentioned in the commit, in some cases we wanna skip decode-encode round

pkg/local_object_storage/blobstor/common/get.go
cthulhu-rider: yes, but here `ln` stand for complete binary object, not only payload. Dynamic size resolution is easier and more reliable (blob storage could diverge with metabase in theory)

at the same time, we _can_ add API to request full binary size from metabase, but in my opinion it's not worth it for this particular goal

cthulhu-rider:
increased test coverage. Fixed old bug on the way

roman-khimov:
We need to have some broader context for this. It looks like a beginning of a custom memory allocator which just shouldn't be done. We can mmap things, we can stream things, we can pool some chunks (but it doesn't work well for objects of different sizes unfortunately). If nothing else works in-place allocation is an option too (which we had for years anyway, just in a bit different form), especially given that it will be more effective already (one flat buffer, no decoding).

cthulhu-rider:
ok lets use make only for now

cthulhu-rider:
dropped options

pkg/local_object_storage/blobstor/fstree/fstree.go
cthulhu-rider: gone

pkg/local_object_storage/blobstor/common/get.go
cthulhu-rider: gone

pkg/local_object_storage/blobstor/fstree/fstree.go
cthulhu-rider: gone

pkg/local_object_storage/blobstor/common/get.go
cthulhu-rider: gone

pkg/local_object_storage/engine/get.go
cthulhu-rider: gone

pkg/local_object_storage/shard/get.go
cthulhu-rider: gone

pkg/local_object_storage/blobstor/fstree/fstree.go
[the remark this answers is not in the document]
carpawell: for me it is no, no. i had to google examples of such machines (meaning years of their production), also it seems that you read a file bigger than such machines can have ram

pkg/local_object_storage/blobstor/fstree/fstree.go
carpawell: i mean reading a full object in memory at once is not good to me in a PR that tries to beat memory problems

pkg/local_object_storage/blobstor/fstree/fstree.go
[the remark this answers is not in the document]
cthulhu-rider: > machines (meaning years of their production)

this may be at the OS level

cthulhu-rider:
> we need object streaming RPC for a correct optimization

two different topics. Here we cover redundant decoding, while ur talking about smaller bufferization (which is also good to be done, but unrelated to the current PR/issue)

carpawell:
> two different topics

Not different to me, the main idea is the same.

> ur talking about smaller bufferization

Im talking about less memory consumption and/or allocations number. RPC that streams an object with partial reading may improve important things by orders of magnitude, while reading the whole object into memory and transferring it as a single message should only make it a few times better.

PR #2763, opened 2024-03-05
cli: Use `MarkFlagsOneRequired` for mutex but required flags

cthulhu-rider:


cmd/neofs-cli/modules/control/shards_set_mode.go
```diff
@@ -155,9 +155,7 @@ func getShardIDList(cmd *cobra.Command) [][]byte {
 	}
 
 	sidList, _ := cmd.Flags().GetStringSlice(shardIDFlag)
-	if len(sidList) == 0 {
-		common.ExitOnErr(cmd, "", fmt.Errorf("either --%s or --%s flag must be provided", shardIDFlag, shardAllFlag))
-	}
+	// guaranteed by MarkFlagsMutuallyExclusive
```
roman-khimov, line 158: Not sure it's obvious what's guaranteed here without this diff.

roman-khimov (approved):

cmd/neofs-cli/modules/control/shards_set_mode.go
cthulhu-rider: agree, explained

carpawell (approved):

Issue #1240, opened 2022-03-15
Limit amount of incoming connections


alexvanin:
It seems that bottleneck caused by https://github.com/nspcc-dev/neofs-node/pull/1236 leads to node goroutine overload and OOM: if request intensity is high enough to hit application limits (goroutine limit), then storage node is going to be killed by OS. 

The better option is to provide some kind of limits in config and do not accept new incoming connections in case of high load.

fyrchik:
This setting can influence default metabase boltdb batch size: if there are no more than X parallel operations on shard, there is no reason to have batch size >X, we better use `X / shard_count` in this case.

fyrchik:
GRPC doesn't allow to limit this amount directly, though we can try to restrict the waiting queue.
I have 2 solutions in mind:
1. Maintain a counter for the amount of operations being executed, fail request immediately if the counter exceeds a threshold.
2. Introduce a pool instead of launching new goroutine, submit task, fail if it hasn't started to execute in 2 seconds (any other time).

Also we better solve this on a case-by-case basis.

PR #2776, opened 2024-03-13
Feat/grpc conn limits

carpawell:


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ Changelog for NeoFS Node
 - Support of `GT`, `GE`, `LT` and `LE` numeric comparison operators in CLI (#2733)
 - SN eACL processing of NULL and numeric operators (#2742)
 - CLI now allows to create and print eACL with numeric filters (#2742)
+- gRPC connection limits per endpoitn (#1240)
```
cthulhu-rider, line 11:
```suggestion
- gRPC connection limits per endpoint (#1240) 
```

cmd/neofs-node/config/grpc/config.go
```diff
@@ -31,6 +31,20 @@ func (x *Config) Endpoint() string {
 	return v
 }
 
+// ConnectionLimit returns the value of "conn_limit" config parameter.
+//
+// Returns 0 if the value is negative.
```
cthulhu-rider, line 36:
```suggestion
// Returns 0 if the value is not a positive integer.
```
e.g. `Hello, world!` -> zero too

config/example/node.yaml
```diff
@@ -36,12 +36,14 @@ node:
 
 grpc:
   - endpoint: s01.neofs.devenv:8080  # endpoint for gRPC server
+    conn_limit: 1  # connection limits; exceeding connection will not be declined, just blocked before active number decreases or client timeouts
```
cthulhu-rider, line 39: i dont think static value can be usable. How user will choose the exact value?

as a basis - quite possible, but then we still need to rely on the real load. For example, there is a difference between 20 clients constantly writing data and 10K ones sending a simple request (like healthcheck)

cmd/neofs-node/grpc.go
```diff
@@ -42,6 +43,10 @@ func initGRPC(c *cfg) {
 			return
 		}
 
+		if connLimit := sc.ConnectionLimit(); connLimit > 0 {
+			lis = netutil.LimitListener(lis, connLimit)
```
cthulhu-rider, line 47: this opens up vulnerability to idle clients: when `connLimit` of them become idle and dont close the connection, all new conns are refused. Overall, i'd not delegate this job to some hard-limiting tool like `netutil.LimitListener`. There are some suggestions in the issue https://github.com/nspcc-dev/neofs-node/issues/1240#issuecomment-1084216928

carpawell: we can control it with server params: https://pkg.go.dev/google.golang.org/grpc#KeepaliveParams

> all new conns are refused

not refused but blocked until some connection is closed in fact

config/example/node.yaml
carpawell: can be discussed, i solved exactly the problem described in the issue: go routines number that lead to the OOM

> i dont think static value can be usable

i can say smth like that about almost every config value but we need to stick to smth

cmd/neofs-node/grpc.go
cthulhu-rider: > we can

i know we can, but we dont

> not refused but blocked

to be more precise, yeah, but the fact remains - no connection

carpawell: > i know we can, but we don't

we can start with this PR, not sure what bound values to choose. depends on whether you still think this approach is usable or not
@roman-khimov

config/example/node.yaml
carpawell: > the trap is that we dont know what exact number cuz it's not const and depends on too many factors

but we need limits according to the issue. so you suggest a dynamic limit?

> static conn limit is not an option for this IMO

what is an option then?

> But lets focus on the introduced one

sure, i just meant that we should test it somehow (if we are ok with a static limit or course) and be able to change depending on the machine and possible load

cthulhu-rider: there are some proposed suggestions https://github.com/nspcc-dev/neofs-node/issues/1240#issuecomment-1084216928. That's all we have for now, and hard net conn limit isn't there

carpawell: > there are some proposed suggestions https://github.com/nspcc-dev/neofs-node/issues/1240#issuecomment-1084216928

i do not think returning some error is a good idea, i would prefer blocking instead and waiting for some timeout (on the client side) or succeeding with some delay

> hard net conn limit isn't there

that is here and in the issue's title

> hard net conn

once again, how do you suggest to make it non-hard?

cthulhu-rider: comment suggests at least 2 approaches

config/example/node.env
```diff
@@ -42,6 +42,11 @@ NEOFS_GRPC_1_ENDPOINT=s02.neofs.devenv:8080
 ### TLS config
 NEOFS_GRPC_1_TLS_ENABLED=false
 
+## 2 server
+NEOFS_GRPC_2_ENDPOINT=s03.neofs.devenv:8080
+### TLS config
+NEOFS_GRPC_2_TLS_ENABLED=true
```
roman-khimov, line 50: What key/cert is used for it?

config/example/node.yaml
roman-khimov: Yeah, it's not easy to choose a proper value and the default might be unlimited, but this handle can be handy to have in case of some case-specific problems. Just like `MaxWebSocketClients` in https://github.com/nspcc-dev/neo-go/blob/master/docs/node-configuration.md#RPC-Configuration, just like `limit_conn` in nginx, just like https://pkg.go.dev/golang.org/x/net/netutil#LimitListener.

cmd/neofs-node/grpc.go
roman-khimov: > to be more precise, yeah, but the fact remains - no connection

Just like with any other limit. Unlimited connections is a vulnerability of one kind, limited is a vulnerability of the other.

The proposal is to add it this way for now and then if/when needed try tuning it. If there are better options found, we'll implement them. This handle itself doesn't cost a lot.

roman-khimov (approved):

config/example/node.env
carpawell: well, none, they are for config reading tests only. [json](https://github.com/nspcc-dev/neofs-node/blob/98f0918b074aa9e53b7c67354cadefd0257f71f5/config/example/node.json#L54-L59) and [yml](https://github.com/nspcc-dev/neofs-node/blob/98f0918b074aa9e53b7c67354cadefd0257f71f5/config/example/node.yaml#L47-L49) are the same. env file was just not tested. do you want to make it false then?

roman-khimov: It's an example, it should have something more or less realistic, that's the only thing I care about wrt this question.

config/example/node.yaml
cthulhu-rider: ok will see how it'll gonna be used

cmd/neofs-node/grpc.go
cthulhu-rider: ok lets see how it'll go

cthulhu-rider (approved):

config/example/node.env
carpawell: fixed

PR #2782, opened 2024-03-21
Fix/v2 split range

carpawell:


pkg/services/object/get/assembly_v2.go
```diff
@@ -116,7 +116,7 @@ func (exec *execCtx) rangeFromLink(link objectSDK.Link) bool {
 				rngPerChild.SetOffset(uint64(firstOffset))
 			}
 			if i == last {
-				rngPerChild.SetLength(uint64(child.ObjectSize()) - uint64(lastBound+firstOffset))
+				rngPerChild.SetLength(uint64(lastBound - firstOffset))
```
roman-khimov, line 119: Isn't it just `uint64(lastBound)`?

carpawell: we may range only part of the first object, then we do not need its offset

roman-khimov: But what if we have multiple objects? You get `X:$END` of `first`, OK, then you go over some set of complete objects and then you need `0:Y`, while you do `0:Y-X`. Or not?

carpawell: hm, seems like you are right but why do tests work then? we never range over the bounds in the tests?

roman-khimov (approved):

cthulhu-rider (commented):
> Seems like a hotfix after some review or smth.

i'd rather formulate essence of the problem instead of obvious abstract things

carpawell:
> i'd rather formulate essence of the problem

I have nothing to add. This line should have the exact number, it did not. The number it was had nothing about the required range, I do remember that i fixed something there but have no idea what that meant. OK, dropped the body.

cthulhu-rider (approved):

PR #2784, opened 2024-03-26
Fix/range over v2 split

carpawell:


pkg/services/object/get/assembly_v2.go
```diff
@@ -114,6 +114,7 @@ func (exec *execCtx) rangeFromLink(link objectSDK.Link) bool {
 
 			if i == first {
 				rngPerChild.SetOffset(uint64(firstOffset))
+				rngPerChild.SetLength(uint64(child.ObjectSize()) - uint64(firstOffset))
```
roman-khimov, line 117: I thought that unset length implies `[x:]` semantics.

carpawell: > I thought that unset length implies [x:] semantics.

me too. also, i could not imagine that it may be a bug but only a small session test will fail, not the range tests itself

PR #2799, opened 2024-04-01
neofs-lens: Add a simple link objects inspector

carpawell:
V2 link scheme does not allow analyzing link objects without additional tools due to the missing children info in the split header: children are in the link object's payload. Add a simple command for inspecting new link objects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@ Changelog for NeoFS Node
 - Support of `GT`, `GE`, `LT` and `LE` numeric comparison operators in CLI (#2733)
 - SN eACL processing of NULL and numeric operators (#2742)
 - CLI now allows to create and print eACL with numeric filters (#2742)
-- gRPC connection limits per endpoint (#1240) 
+- gRPC connection limits per endpoint (#1240)
+- `neofs-lens object link` command for the new link object inspection (#2799)
 
 ### Fixed
 - Access to `PUT` objects no longer grants `DELETE` rights (#2261)
--- a/cmd/neofs-lens/internal/object/root.go
+++ b/cmd/neofs-lens/internal/object/root.go
@@ -1,0 +1,33 @@
+package object
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+var vPath string
+
+// Root defines root command for operations with NeoFS objects.
+var Root = &cobra.Command{
+	Use:   "object",
+	Short: "NeoFS object inspection",
+}
+
+const filePathFlagKey = "file"
+
+func init() {
+	pFlags := Root.PersistentFlags()
+	pFlags.StringVar(&vPath, filePathFlagKey, "", "Path to file with NeoFS object")
+
+	err := cobra.MarkFlagFilename(pFlags, filePathFlagKey)
+	if err != nil {
+		panic(fmt.Errorf("MarkFlagFilename for %s flag: %w", filePathFlagKey, err))
+	}
+	err = cobra.MarkFlagRequired(pFlags, filePathFlagKey)
+	if err != nil {
+		panic(fmt.Errorf("MarkFlagRequired for %s flag: %w", filePathFlagKey, err))
+	}
+
+	Root.AddCommand(linkCMD)
+}
--- a/cmd/neofs-lens/internal/object/link.go
+++ b/cmd/neofs-lens/internal/object/link.go
@@ -1,0 +1,53 @@
+package object
+
+import (
+	"errors"
+	"fmt"
+	"os"
+
+	common "github.com/nspcc-dev/neofs-node/cmd/neofs-lens/internal"
+	"github.com/nspcc-dev/neofs-sdk-go/object"
+	"github.com/spf13/cobra"
+)
+
+var linkCMD = &cobra.Command{
+	Use:   "link",
+	Short: "Inspect link object",
+	Run:   linkFunc,
+}
+
+func linkFunc(cmd *cobra.Command, _ []string) {
+	if vPath == "" {
+		common.ExitOnErr(cmd, errors.New("empty path to file"))
+	}
+
+	raw, err := os.ReadFile(vPath)
+	common.ExitOnErr(cmd, common.Errf("reading file: %w", err))
+
+	var link object.Link
+	err = link.Unmarshal(raw)
+	if err != nil {
+		cmd.Printf("%q file does not contain raw link payload, trying to decode it as a full NeoFS object", vPath)
+
+		var obj object.Object
+		err = obj.Unmarshal(raw)
+		common.ExitOnErr(cmd, common.Errf("decoding NeoFS object: %w", err))
+
+		if typeGot := obj.Type(); typeGot != object.TypeLink {
+			common.ExitOnErr(cmd, fmt.Errorf("unexpected object type (not %s): %s", object.TypeLink, typeGot))
+		}
+
+		err = obj.ReadLink(&link)
+	}
+	common.ExitOnErr(cmd, common.Errf("decoding link object: %w", err))
+
+	if len(link.Objects()) == 0 {
+		common.ExitOnErr(cmd, errors.New("empty children list"))
+	}
+
+	cmd.Println("Children (sorted according to the read payload):")
+
+	for _, measuredObject := range link.Objects() {
+		cmd.Printf("Size: %d, object ID: %s\n", measuredObject.ObjectSize(), measuredObject.ObjectID())
+	}
+}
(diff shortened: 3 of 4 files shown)


carpawell:
```
▶ neofs-lens objects link --file /tmp/test.file
Children (sorted according to the read payload):
Child size: 524288, child ID: 2vZARtk8kzFYCpmxPByuN6jPcjk4faJMGtqrZHM5CbSZ
Child size: 524288, child ID: 2TGoZuA5a8XMooMCTWoZHJ7uix2FcsrZ7G2TRBWSwNTi
Child size: 524288, child ID: CkdrkMiezVKhbyG3ALkGM36UrpzbTUHbBLYwHjpsab4t
Child size: 1000, child ID: 2hb9CctME7xRT5wrvvy8rGvFPYgadv5dB8a4MhknzyLo
```

cmd/neofs-lens/internal/objects/root.go
```diff
@@ -0,0 +1,21 @@
+package objects
+
+import "github.com/spf13/cobra"
+
+var vPath string
+
+// Root defines root command for operations with NeoFS objects.
+var Root = &cobra.Command{
+	Use:   "objects",
```
roman-khimov, line 9: Why plural?

CHANGELOG.md
```diff
@@ -8,7 +8,8 @@ Changelog for NeoFS Node
 - Support of `GT`, `GE`, `LT` and `LE` numeric comparison operators in CLI (#2733)
 - SN eACL processing of NULL and numeric operators (#2742)
 - CLI now allows to create and print eACL with numeric filters (#2742)
-- gRPC connection limits per endpoint (#1240) 
+- gRPC connection limits per endpoint (#1240)
```
cthulhu-rider, line 11: any change here? GH doesnt mark any on my side

CHANGELOG.md
```diff
@@ -8,7 +8,8 @@ Changelog for NeoFS Node
...
+- `neofs-lens objects link` command for the new link object inspection (#)
```
cthulhu-rider, line 12: dont forget to add number

cmd/neofs-lens/internal/objects/link.go
```diff
@@ -0,0 +1,39 @@
...
+	if vPath == "" {
+		common.ExitOnErr(cmd, errors.New("empty path to file"))
+	}
+
+	raw, err := os.ReadFile(vPath)
+	common.ExitOnErr(cmd, common.Errf("reading file: %w", err))
+
+	var link object.Link
+	err = link.Unmarshal(raw)
+	common.ExitOnErr(cmd, common.Errf("decoding link object: %w", err))
```
cthulhu-rider, line 28: i'd also support (as fallback) binary linking objects that can be downloaded via `object get --binary`. What u think?

cmd/neofs-lens/internal/objects/link.go
```diff
@@ -0,0 +1,39 @@
...
+
+	if len(link.Objects()) == 0 {
+		common.ExitOnErr(cmd, errors.New("empty children list"))
+	}
+
+	cmd.Println("Children (sorted according to the read payload):")
+
+	for _, measuredObject := range link.Objects() {
+		cmd.Printf("Child size: %d, child ID: %s\n", measuredObject.ObjectSize(), measuredObject.ObjectID())
```
cthulhu-rider, line 37: `Children` is already printed, may be not write `child` each time?

cmd/neofs-lens/internal/objects/root.go
```diff
@@ -0,0 +1,21 @@
...
+	Short: "NeoFS objects inspection",
+}
+
+const filePathFlagKey = "file"
+
+func init() {
+	Root.PersistentFlags().StringVar(&vPath, filePathFlagKey, "", "Path to file with raw NeoFS object payload")
+	_ = Root.MarkFlagFilename(filePathFlagKey)
```
cthulhu-rider, line 17:
```suggestion
	_ = Root.MarkPersistentFlagFilename(filePathFlagKey)
```
now u'll get 'flag not found' error (ignored). I see u check this in `linkFunc`, this case is still better to be caught by Cobra engine

same below

if arrived with copypaste, lets leave an issue for other cases (or make instafix). Seems like ignoring such errors is not so good

cthulhu-rider (commented):
tried: covered case works well

LGTM overall, left several code suggestions

cmd/neofs-lens/internal/objects/root.go
carpawell: fixed

> if arrived with copypaste, lets leave an issue for other cases

no, done myself but remembered that we ignore this err usually. do you mean we need not to ignore such error ever or just fixing `*Persistent*` here?

CHANGELOG.md
carpawell: redundant space at the end

cmd/neofs-lens/internal/objects/root.go
carpawell: planned to have `neofs-lens objects link`, `neofs-lens objects storage-group`, `neofs-lens objects lock`, etc. it should be either `objects` or `inspect-object` to me. your suggestion?

CHANGELOG.md
roman-khimov: Usually they can be seen in GH diff.

cmd/neofs-lens/internal/objects/root.go
roman-khimov: `object`. `ip route`/`ip address`/etc. Ideally commands are verbs, but it's not the case already and you're always working with a single object here.

CHANGELOG.md
cthulhu-rider: > redundant space

oh rly

> Usually they can be seen in GH diff.

seems like additional line makes it invisible in diff. Being a separate commit, it'd be visible at least in `Commits` (not in `Files changes` i suppose)

cmd/neofs-lens/internal/objects/link.go
```diff
@@ -0,0 +1,49 @@
...
+	err = link.Unmarshal(raw)
+	if err != nil {
+		cmd.Printf("%q file does not contain raw link payload, trying to decode it as a full NeoFS object", vPath)
+
+		var obj object.Object
+		err = obj.Unmarshal(raw)
+		common.ExitOnErr(cmd, common.Errf("decoding NeoFS object: %w", err))
+
+		raw = obj.Payload()
+		err = link.Unmarshal(raw)
```
cthulhu-rider, line 36: cool! this can also be used for contents of storages like FS tree

btw
```suggestion
		err = obj.ReadLink(&link)
```
'd be even more laconic, huh?

and since the method `Makes sense only if object has [TypeLock] type`, it's also worth to either pre-check object type or include type to error context on decode failure

cthulhu-rider (approved):

cmd/neofs-lens/internal/objects/root.go
carpawell: > first u get fs := Root.PersistentFlags() and then working with fs only

ok, lets try this

> leave an issue for other cases

https://github.com/nspcc-dev/neofs-node/issues/2800

CHANGELOG.md
carpawell: > Usually they can be seen in GH diff.

i guess that is caused by the moving `\n` to the next line

> Being a separate commit, it'd be visible at least in Commits

IMO, a separate commit is too much. i know we love commits but that is moving nonvisible symbols to another place because of a new line

cmd/neofs-lens/internal/objects/root.go
carpawell: ok, let it be `object`. i would go `objects` as a root command that works with different kinds of objects

cmd/neofs-lens/internal/object/root.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+var vPath string
+
+// Root defines root command for operations with NeoFS objects.
+var Root = &cobra.Command{
+	Use:   "objects",
```
roman-khimov, line 13: Still `objects`

CHANGELOG.md
cthulhu-rider: im not callin for a separate commit (although independent fix in fact), just tried to imagine and understand why it was visible in diff of past PRs

cmd/neofs-lens/internal/object/root.go
cthulhu-rider: dont forget changelog

roman-khimov (approved):

PR #2801, opened 2024-04-05
engine: Support writing already encoded objects

cthulhu-rider:
* reverse side of https://github.com/nspcc-dev/neofs-node/pull/2753

### DIFF
--- a/pkg/local_object_storage/engine/put.go
+++ b/pkg/local_object_storage/engine/put.go
@@ -16,6 +16,10 @@ import (
 // PutPrm groups the parameters of Put operation.
 type PutPrm struct {
 	obj *objectSDK.Object
+
+	binSet bool
+	objBin []byte
+	hdrLen int
 }
 
 // PutRes groups the resulting values of Put operation.
@@ -28,6 +32,16 @@ var errPutShard = errors.New("could not put object to any shard")
 // Option is required.
 func (p *PutPrm) WithObject(obj *objectSDK.Object) {
 	p.obj = obj
+}
+
+// SetObjectBinary allows to provide the already encoded object in
+// [StorageEngine] format. Object header must be a prefix with specified length.
+// If provided, the encoding step is skipped. It's the caller's responsibility
+// to ensure that the data matches the object structure being processed.
+func (p *PutPrm) SetObjectBinary(objBin []byte, hdrLen int) {
+	p.binSet = true
+	p.objBin = objBin
+	p.hdrLen = hdrLen
 }
 
 // Put saves the object to local storage.
@@ -72,7 +86,7 @@ func (e *StorageEngine) put(prm PutPrm) (PutRes, error) {
 			return false
 		}
 
-		putDone, exists := e.putToShard(sh, ind, pool, addr, prm.obj)
+		putDone, exists := e.putToShard(sh, ind, pool, addr, prm)
 		finished = putDone || exists
 		return finished
 	})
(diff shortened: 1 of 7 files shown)


roman-khimov (approved):
Seems to be legit and can go into 0.41.0.

pkg/local_object_storage/engine/put.go
```diff
@@ -16,6 +16,10 @@ import (
 // PutPrm groups the parameters of Put operation.
 type PutPrm struct {
 	obj *objectSDK.Object
+
+	binSet bool
```
carpawell, line 20: `binSet` == `objBin != nil`?

pkg/local_object_storage/shard/put.go
```diff
@@ -38,9 +52,18 @@ func (s *Shard) Put(prm PutPrm) (PutRes, error) {
...
+	var data []byte
+	var err error
+	if prm.binSet {
+		data = prm.objBin
+	} else {
+		data, err = prm.obj.Marshal()
+		if err != nil {
+			return PutRes{}, fmt.Errorf("cannot marshal object: %w", err)
+		}
+		// TODO: currently, we don't need to calculate prm.hdrLen in this case.
```
carpawell, line 64: todo with no issue?

pkg/local_object_storage/engine/put.go
cthulhu-rider: `binSet` prevents duality b/w uncalled and called with nil cases

carpawell: > called with nil cases

is it valid?

pkg/local_object_storage/shard/put.go
carpawell: how can you be sure? we had a work once where the dev searched for such lost cases

pkg/local_object_storage/engine/put.go
cthulhu-rider: depends on the caller. This is his responsibility, docs say that

pkg/local_object_storage/shard/put.go
cthulhu-rider: > how can you be sure?

this is what git is made for

pkg/local_object_storage/engine/put.go
```diff
@@ -30,6 +34,16 @@ func (p *PutPrm) WithObject(obj *objectSDK.Object) {
 	p.obj = obj
 }
 
+// SetObjectBinary allows to provide the already encoded object in
+// [StorageEngine] format. Object header must be a prefix with specified length.
```
carpawell, line 38: btw not that clear what such a format is

pkg/local_object_storage/engine/put.go
carpawell: undesired complexity to me. calling with empty bytes ends with no error (i guess) while that is incorrect for sure

carpawell (approved):

pkg/local_object_storage/shard/put.go
carpawell: > git is made for

searching for `TODO` strings? my point is that `TODO` is about smth that _should_ be done. good to have a corresponding issue then. otherwise that is not `TODO`

roman-khimov: To me it's not even a TODO, just a remark on the specifics of the current code. It's nothing actionable at this point.

PR #2803, opened 2024-04-05
metrics: Purge deprecated counters

cthulhu-rider:


roman-khimov (changes requested):
You need to also delete them from `objectServiceMetrics` (cleaning up still remaining setter afterwards).and then sett

cthulhu-rider:
dropped

roman-khimov (approved):

PR #2805, opened 2024-04-09
Fix panic on metrics init

cthulhu-rider:


roman-khimov (approved):
How come https://rest.fs.neo.org/HXSaMJXk2g8C14ht8HSi7BBaiYZ1HeWh2xnWPGQCg4H6/1487-1712598182/index.html is not all red then? We don't have them enabled? @evgeniiz321?

PR #2810, opened 2024-04-16
Fix/complex objects relations

carpawell:


pkg/services/object/tombstone/verify.go
```diff
@@ -0,0 +1,195 @@
...
+		if err != nil {
+			return fmt.Errorf("heading %s object that was searched: %w", addr, err)
+		}
+
+		if len(h.Children()) != 0 {
+			return nil
+		}
+	}
+
+	return errors.New("no link object found")
```
roman-khimov, line 140: No link here -> can delete.

pkg/services/object/tombstone/verify.go
```diff
@@ -0,0 +1,195 @@
...
+		return nil
+	}
+
+	// it is a child object removal somehow; only allowed for incomplete object put
+	// (garbage collecting) so check if there is a correct link object for the chain
+
+	if sID != nil {
+		err = v.verifyV1Child(ctx, cnr, *sID)
+		if err != nil {
+			return fmt.Errorf("verify V1 split: %w", err)
```
roman-khimov, line 96: But an error is returned here.

roman-khimov (commented):
Some tests would be helpful.

pkg/services/object/tombstone/verify.go
carpawell: fixed

carpawell:
> Some tests would be helpful.

Added some.

roman-khimov (approved):
Let's try it.

Issue #1951, opened 2022-10-21
Shrink bolt databases


fyrchik:
The size of bbolt databases can only grow https://github.com/boltdb/bolt/issues/308

This is a bit misleading from the user POV, who can do `df -h` and also leaves less space for big objects.
We can extend GC with a vacuum, which (rarely) tries to compact blobovnicza trees by trying to move all objects on a single level to a smaller set of leaves.

There is also https://github.com/etcd-io/bbolt/blob/master/compact.go, but it needs 2 files.

roman-khimov:
1. Objects are not stored in DBs, so it's not a problem for objects.
2. It is a problem for metadata, but this DB is always open and the only way to compact it is to run `bbolt compact` (https://github.com/etcd-io/bbolt/blob/c0f7c3c0ff156c0bd51039aa21276e386612bad6/cmd/bbolt/README.md).
3. Practically this shouldn't matter much if meta space is sufficient for the purpose.
4. If it still is --- we can't do anything meaningful node-side.

PR #2814, opened 2024-04-18
Combined writing for FSTree

roman-khimov:
We can think of replacing peapod with it.


roman-khimov:
Notice that this scheme also helps with #1951 and #2107 (!) if we're to store meta separately.

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -346,6 +379,38 @@ func (t *FSTree) GetBytes(addr oid.Address) ([]byte, error) {
...
+		lengthOff = addrOff + addrSize
+		dataOff   = lengthOff + lengthSize
+	)
+
+	var notFound bool
+
+	for len(data) > dataOff && data[0] == combinedPrefix {
+		notFound = true // The file _is_ combined, so the object _must_ be there.
+		var l = binary.BigEndian.Uint32(data[lengthOff:dataOff])
+		if bytes.Compare(data[addrOff:lengthOff], addr[:]) == 0 {
```
carpawell, line 375: why not `bytes.Equal`?

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -346,6 +379,38 @@ func (t *FSTree) GetBytes(addr oid.Address) ([]byte, error) {
 	return dec, nil
 }
 
+func extractCombinedObject(addr oid.ID, data []byte) ([]byte, error) {
```
carpawell, line 359: `addr` is not `oid.ID`, would not name it this way

pkg/local_object_storage/blobstor/fstree/fstree.go
carpawell, line 375: it was not supposed to get bytes this way... we do not follow our own rules... it is either the rules are bad, or we should use `addr.Encode` here

pkg/local_object_storage/blobstor/fstree/fstree_write_linux.go
```diff
@@ -29,19 +58,159 @@ func newSpecificWriteData(root string, perm fs.FileMode, noSync bool) func(strin
...
+
+func (b *syncBatch) wait() error {
+	<-b.ready
+	return b.err
+}
+
+func (b *syncBatch) write(id oid.ID, p string, data []byte) error {
+	var (
+		err  error
+		pref [1 + len(id) + 4]byte
```
carpawell, line 126: we have consts for this numbers but they are defined in a func for some reason

carpawell (commented):
No changelog at all; the last commit was not DCOed.

So what is the peapod's fate?

roman-khimov:
> So what is the peapod's fate?

TBD

pkg/local_object_storage/blobstor/fstree/fstree.go
roman-khimov: This naming is already used in some places and OID is an address in some ways.

pkg/local_object_storage/blobstor/fstree/fstree.go
roman-khimov: Current code is obviously more efficient, even more importantly it's symmetric to `(*syncBatch).write`, so unless the type changes radically we don't care much about the contents.

pkg/local_object_storage/blobstor/fstree/fstree_write_linux.go
roman-khimov: It's harder to handle in the reader, more code with more fancy calculations. I had a thought about moving them into upper-layer definitions, but they'd need some `combined` prefix that would make them _so_ long. But can be done.

pkg/local_object_storage/blobstor/fstree/fstree.go
roman-khimov: Fixed though.

pkg/local_object_storage/blobstor/fstree/fstree.go
roman-khimov: Fixed

carpawell:
>> So what is the peapod's fate?
>
> TBD

I am asking cause we have to rework some things if there is no "small" storage anymore.

Have you considered GET bench? Isn't it slower to read a combined object for only a part of it if there are a few parallel GET requests? Like reading 8MB for 100 bytes only 128 times _can be slower_ than reading 100 bytes 128 times from a peadod.

roman-khimov:
> Isn't it slower to read

For sure. But I doubt it's noticeable. I'll try to bench it.

roman-khimov:
Ironically, it doesn't have any conflicts with current master even without a rebase. So my proposal is to merge this as is and then deal with peapod. @cthulhu-rider, @carpawell?

cthulhu-rider (approved):

carpawell:
> Have you considered GET bench? Isn't it slower to read a combined object for only a part of it if there are a few parallel GET requests? Like reading 8MB for a 4KB object 128 times can be slower than reading 4KB 128 times from a peadod.

@roman-khimov, my questions are the same. I am afraid there (at least this PR) is not enough info about if it is really a _better_ one than we had before, from _any_ perspective (i consider it a worse solution for GETs). If there is a single slower thing in results, we should discuss it and find out (at least write it somewhere) what is better for us.

roman-khimov:
But you've seen the test results. GETs are the same, PUTs are ~15% better for the target cases. GETs do incur some additional overhead, but it's totally acceptable, there is no real difference in practice.

roman-khimov:
Rebased, otherwise lens failed to build after the merge (even though the merge itself is clean).

carpawell (approved):

PR #2830, opened 2024-04-24
pkg/morph/container: Do not create client with notary status

carpawell:
Provide notary signature status with explicit parameter. Some calls are always expected to be Alphabet signed (at least for the current contracts), make it explicit for a code reader, that is more error proof. This commit also fixes container size estimations writing: they were tried to be notary signed, but the Alphabet does not handle such requests. This is also the reason why this kinda refactor does not touch other contracts; general client improvements can be done with another PR.

### DIFF
--- a/pkg/morph/client/container/client.go
+++ b/pkg/morph/client/container/client.go
@@ -60,10 +60,6 @@ func NewFromMorph(cli *client.Client, contract util.Uint160, fee fixedn.Fixed8, 
 		o.staticOpts = append(o.staticOpts, client.WithCustomFee(putNamedMethod, o.feePutNamed))
 	}
 
-	if !o.disableNotarySigning {
-		o.staticOpts = append(o.staticOpts, client.TryNotary())
-	}
-
 	sc, err := client.NewStatic(cli, contract, fee, o.staticOpts...)
 	if err != nil {
 		return nil, fmt.Errorf("can't create container static client: %w", err)
--- a/pkg/morph/client/container/client.go
+++ b/pkg/morph/client/container/client.go
@@ -86,24 +86,11 @@ type opts struct {
 	feePutNamedSet bool
 	feePutNamed    fixedn.Fixed8
 
-	disableNotarySigning bool
-
 	staticOpts []client.StaticClientOption
 }
 
 func defaultOpts() *opts {
 	return new(opts)
-}
-
-// DisableNotarySigning returns option to disable
-// notary request signing. With that option, every
-// call will be created, signed with provided key
-// (a single regular sign) and sent to the side
-// chain.
-func DisableNotarySigning() Option {
-	return func(o *opts) {
-		o.disableNotarySigning = true
-	}
 }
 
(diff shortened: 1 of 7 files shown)


pkg/morph/client/static.go
```diff
@@ -97,6 +99,14 @@ func (i *InvokePrmOptional) SetHash(hash util.Uint256) {
 	i.hash = &hash
 }
 
+// NotaryAlphabetSignature makes client send notary request instead of a
+// regular signed transaction. Such a request should be received and signed by
+// the Alphabet, otherwise the plug (empty) transaction will be added to the
+// chain.
+func (ti *InvokePrmOptional) NotaryAlphabetSignature(signByAlphabet bool) {
```
roman-khimov, line 106: Can be `RequireAlphabetSignature()`, the default is `false`, this will just set it to `true`.

pkg/morph/client/container/client.go
```diff
@@ -90,26 +86,13 @@ type opts struct {
...
 func defaultOpts() *opts {
 	return new(opts)
 }
 
-// DisableNotarySigning returns option to disable
-// notary request signing. With that option, every
-// call will be created, signed with provided key
-// (a single regular sign) and sent to the side
-// chain.
-func DisableNotarySigning() Option {
```
roman-khimov, line 107: This doesn't seem to be used, can be a separate patch.

roman-khimov (commented):
While the intention is good, I fear it can break things.

roman-khimov (commented):
How about `balance`? Likely you can't mint/burn without alphabet support. NeoFSID? Netmap?

roman-khimov:
You only change `container` defaults though, so likely these are fine.

pkg/morph/client/container/client.go
carpawell: ok

pkg/morph/client/static.go
carpawell: dont mind, wanted it to be flexible, but that is just prm struct, not needed. fixed

pkg/morph/client/container/client.go
```diff
@@ -60,10 +60,6 @@ func NewFromMorph(cli *client.Client, contract util.Uint160, fee fixedn.Fixed8,
 		o.staticOpts = append(o.staticOpts, client.WithCustomFee(putNamedMethod, o.feePutNamed))
 	}
 
-	if !o.disableNotarySigning {
-		o.staticOpts = append(o.staticOpts, client.TryNotary())
```
roman-khimov, line 64: But this changes the default behavior, so it's not just `drop ...`

carpawell: ok, chagned the message, check it out

carpawell: changed one more time. let defaults be changed in the first commit while the second can be still "Drop"

roman-khimov:
Yup, convincing.

roman-khimov (approved):

PR #2832, opened 2024-04-24
meta: Do not stop iteration in GC on errors

carpawell:
Errors in GC cycle can be logged, and the cycle can be continued, there is no reason not to handle leftover garbage. Make it async (up to 1000 concurrent calls) cause most of the meta code is performed in bbolt transaction, so blocking will slow general performance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Changelog for NeoFS Node
 - Container size estimation contract writing (#2819)
 - Custom contract deployment with custom zone via neofs-adm (#2827)
 - Errors in neofs-adm morph dump-names output (#2831)
+- GC stops if any non-critical "content" errors happen (#2823)
 
 ### Changed
 
--- a/pkg/local_object_storage/metabase/iterators.go
+++ b/pkg/local_object_storage/metabase/iterators.go
@@ -10,6 +10,7 @@ import (
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"go.etcd.io/bbolt"
+	"go.uber.org/zap"
 )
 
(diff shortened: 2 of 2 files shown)


pkg/local_object_storage/metabase/control.go
```diff
@@ -94,11 +94,18 @@ func (db *DB) Reset() error {
...
+// asyncErrHandlingCapacity defines max async errors number
+// that can be handled in parallel.
+const asyncErrHandlingCapacity = 1000
+
 func (db *DB) init(reset bool) error {
 	if db.mode.NoMetabase() || db.mode.ReadOnly() {
 		return nil
 	}
 
+	db.errChan = make(chan error, asyncErrHandlingCapacity)
```
roman-khimov, line 106: But why are you complicating it this much? In normal case we expect exactly zero errors, so this code is not supposed to be running at all, let alone with concurrency. If it wastes some microseconds for logging it's not critical.

carpawell: i do not want to spend this time in the bolt TX

> so this code is not supposed to be running at all

but if it run, it is supposed to be run in cycle until we fix it or drop problem object

carpawell: so you are for just logging?

roman-khimov: > but if it run, it is supposed to be run in cycle until we fix it or drop problem object

This cycle starts once in a while and ends pretty quickly (depending on the DB).

> so you are for just logging?

I think it's sufficient.

carpawell: cant say i fully agree, IMO this code is not complex at all, and i would never log anything in a db transaction, but let it be your way

roman-khimov (approved):

Issue #2824, opened 2024-04-22
Run neofs-testcases for linux-arm64


roman-khimov:
## Is your feature request related to a problem? Please describe.
I'm always frustrated when I don't know if our node works for linux-arm64.

## Describe the solution you'd like
Run testcases on linux-arm64 and see how it goes.

## Additional context
This will allow us to have proper arm64 releases and cut down some infra costs (arm64 machines are cheaper). This will also allow to have Docker images that work for ARM-based Macs.

roman-khimov:
It may require macos-14 runner and darwin-arm64 in fact, but this combination fits the purpose as well.

evgeniiz321:
@roman-khimov here is the PR with a try - https://github.com/nspcc-dev/neofs-node/pull/2825
2 problems there:
1. Current binaries (neo-go/neofs-s3-gw/neofs-rest-gw) from releases are not suitable to run on macos runners, even though some binaries are marked as darwin/arm64, all of them fail with the invalid CPU type or something similar. To make it work I had to build everything manually inside the workflow.
2. Allure reports publishing to neofs is not supported for arm64. There are problems both in our action and in `simple-elf/allure-report-action@v1.6` that we use inside to generate reports, it only supports linux runners.

roman-khimov:
1. linux-arm64 images should work for MacOS as well. @mike-petrov, you're using them, right?
2. Our action depends on neofs-cli, so in absence of linux-arm64 binary it's hard for it to work. But `allure-report-action` dependency is more worrisome.

At the same time, can you build everything locally and run testcases on Mac? What's the status for them in this case? If they're confirmed to be OK we can at least start enabling appropriate builds which will solve a part of the problem.

mike-petrov:
I don't know which image I'm using, but everything works

roman-khimov:
Fixed in #2835, please be more careful with "fixes/closes".

PR #2835, opened 2024-04-25
workflows: build arm64/darwin binaries

roman-khimov:
Related to #2824.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,8 +37,8 @@ jobs:
     runs-on: ${{matrix.os.name}}
     strategy:
       matrix:
-        os: [{ name: ubuntu-22.04, bin-name: linux }] # { name: windows-2022, bin-name: windows }, { name: macos-12, bin-name: darwin }
-        arch: [amd64] # arm64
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -37,8 +37,8 @@ jobs:
     runs-on: ${{matrix.os.name}}
     strategy:
       matrix:
-        os: [{ name: ubuntu-22.04, bin-name: linux }] # { name: windows-2022, bin-name: windows }, { name: macos-12, bin-name: darwin }
-        arch: [amd64] # arm64
+        os: [{ name: ubuntu-22.04, bin-name: linux }, { name: macos-14, bin-name: darwin }] # { name: windows-2022, bin-name: windows }
```
evgeniiz321, line 40: For neo-go we use macos-12 - https://github.com/nspcc-dev/neo-go/blob/master/.github/workflows/build.yml#L40

evgeniiz321: fun fact, macos-12 runs on x86, macos-14 on arm, so I guess macos-14 is the correct one

roman-khimov: For builds it doesn't really matter, the resulting binary is determined by GOOS/GOARCH and you can build any binary on ubuntu-22.04 in fact (one of Go surprises).

carpawell (approved):

PR #2840, opened 2024-04-27
Workflow fixes

roman-khimov:


.github/workflows/go.yml
```diff
@@ -31,12 +57,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Run go test
-        run: go test -coverprofile=coverage.txt -covermode=atomic ./...
-
-      - name: Codecov
-        env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
-        run: bash <(curl -s https://codecov.io/bash)
+        run: go test -race ./...
```
carpawell, line 71: why `race` here now?

.github/workflows/go.yml
```diff
@@ -15,11 +15,37 @@ on:
...
+        uses: actions/setup-go@v5
+        with:
+          cache: true
+          go-version: '1.22'
+
+      - name: Check out code
+        uses: actions/checkout@v4
+
+      - name: Run go test
+        run: go test -coverprofile=coverage.txt -covermode=atomic ./...
```
carpawell, line 32: and no `race` here?

.github/workflows/go.yml
```diff
@@ -15,11 +15,37 @@ on:
...
+          token: ${{ secrets.CODECOV_TOKEN }}
+          verbose: true
+
   test:
     runs-on: ubuntu-latest
+    name: Unit tests
     strategy:
       matrix:
-        go: [ '1.20.x', '1.21.x', '1.22' ]
+        go: [ '1.20.x', '1.21.x' ]
```
carpawell: does it solve something? i do not like this version split so would do only if it really helps us with something

.github/workflows/go.yml
```diff
@@ -45,7 +45,18 @@ jobs:
     name: Unit tests
     strategy:
       matrix:
-        go: [ '1.20.x', '1.21.x' ]
+        os: [ubuntu-22.04, macos-14]
+        go: [ '1.20.x', '1.21.x', '1.22' ]
+        exclude:
+          # Only latest Go version for Windows and MacOS.
```
carpawell, line 51: but no windows is found here

carpawell (commented):
Tests died.

.github/workflows/go.yml
roman-khimov: It should've always been this way. `-race` is good.

.github/workflows/go.yml
roman-khimov: This one is about coverage.

.github/workflows/go.yml
cthulhu-rider: start of racing is a separate change to me

cthulhu-rider (approved):
good overall

.github/workflows/go.yml
carpawell: and is it a problem? we can run tests for 3 versions, and upload (and run test the second time) with the latest one. i do not like the idea of moving the whole version run to a separate job with `Coverage` name when the only reason is "maybe action is a bit more reliable"

carpawell:
> Yeah, -race works.

.github/workflows/go.yml
carpawell: so 2 runs with `race` and one run with no `race`?

.github/workflows/go.yml
roman-khimov: Likely it can be reworked this way, but an additional job is rather easy to have. It also makes it easier to separate coverage from other tests for various purposes (like branch protection rules).

.github/workflows/go.yml
roman-khimov: Fixed.

PR #2826, opened 2024-04-22
neofs-adm: Add container estimations inspector

carpawell:

PR #2841, opened 2024-04-27
adm: never use GetContractStateByID(1) to retrieve NNS hash

roman-khimov:
And use a bit more of NNS wrapper.

Notice that this changes "alphabet X" to "alphabetX" in dump output, but I consider this to be an improvement (since that's the contract name).

It also deoptimizes some carefully crafted code to extract a number of names in one call. Likely it's not a performance issue you care about and it's actually somewhat broken in that it assumes one result per invocation (but you can get zero or more than one).

Many more things can be improved, but let's start with this.

Refs. #2826.

### DIFF
--- a/cmd/neofs-adm/internal/modules/morph/balance.go
+++ b/cmd/neofs-adm/internal/modules/morph/balance.go
@@ -7,7 +7,6 @@ import (
 	"math/big"
 
 	"github.com/nspcc-dev/neo-go/pkg/core/native/noderoles"
-	"github.com/nspcc-dev/neo-go/pkg/core/state"
 	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
 	"github.com/nspcc-dev/neo-go/pkg/encoding/address"
 	"github.com/nspcc-dev/neo-go/pkg/encoding/fixedn"
@@ -19,7 +18,7 @@ import (
 	"github.com/nspcc-dev/neo-go/pkg/util"
 	"github.com/nspcc-dev/neo-go/pkg/vm/stackitem"
 	"github.com/nspcc-dev/neo-go/pkg/vm/vmstate"
-	"github.com/nspcc-dev/neofs-contract/contracts/nns"
+	"github.com/nspcc-dev/neofs-contract/rpc/nns"
 	"github.com/nspcc-dev/neofs-sdk-go/netmap"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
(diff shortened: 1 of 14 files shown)


cmd/neofs-adm/internal/modules/morph/balance.go
```diff
@@ -54,12 +53,13 @@ func dumpBalances(cmd *cobra.Command, _ []string) error {
 	inv := invoker.New(c, nil)
 
 	if dumpStorage || dumpAlphabet || dumpProxy {
-		nnsCs, err = c.GetContractStateByID(1)
```
carpawell, line 57: is the juice worth the squeeze? it's almost always the same or even more code lines in the current adm when you need to get hash. also, nns package has public `ID` const

roman-khimov: It's the same when we need hash. But there is potential to drop even more code, `nnsResolveHash` can be dropped eventually.

cmd/neofs-adm/internal/modules/morph/config.go
```diff
@@ -102,12 +104,13 @@ func setConfigCmd(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("can't initialize context: %w", err)
 	}
 
-	cs, err := wCtx.Client.GetContractStateByID(1)
+	nnsHash, err := nns.InferHash(wCtx.Client)
 	if err != nil {
-		return fmt.Errorf("can't get NNS contract info: %w", err)
+		return err
```
cthulhu-rider, line 109: lets keep the context as everywhere
```suggestion
		return fmt.Errorf("can't get NNS contract hash: %w", err)
```

cmd/neofs-adm/internal/modules/morph/balance.go
carpawell: > It's the same when we need hash

this exact code part is a 3 > 4 change

> But there is potential to drop even more code

i do not doubt using nns wrappers is the best way to go, i just doubt this PR makes it any easier/better/simpler, i would refactor it some deeper or keep it as is

cthulhu-rider (approved):

cmd/neofs-adm/internal/modules/morph/config.go
roman-khimov: Fixed!

cmd/neofs-adm/internal/modules/morph/balance.go
roman-khimov: I've cut a bit more lines now and we'll do more after https://github.com/nspcc-dev/neofs-contract/pull/397.

PR #2846, opened 2024-05-06
node: Return status responses only

cthulhu-rider:
support for non-status returns makes no sense after such a period

### DIFF
--- a/pkg/services/util/sign.go
+++ b/pkg/services/util/sign.go
@@ -50,8 +50,6 @@ type RequestMessageStreamer struct {
 
 	respCons ResponseConstructor
 
-	statusSupported bool
-
 	sendErr error
 }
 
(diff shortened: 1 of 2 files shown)


pkg/services/util/sign.go
```diff
@@ -195,31 +172,21 @@ func (s *SignService) HandleUnaryRequest(ctx context.Context, req any, handler U
...
 		setStatusV2(resp, err)
 	}
 
 	// sign the response
-	if err = signResponse(s.key, resp, statusSupported); err != nil {
-		return nil, err
+	if err = signature.SignServiceMessage(s.key, resp); err != nil {
+		// We can't pass this error as status code since response will be unsigned.
+		// Isn't expected in practice, so panic is ok here.
+		panic(err)
```
roman-khimov, line 184: Brutal. We can log and close the connection.

cthulhu-rider: #2848

roman-khimov (approved):

PR #2847, opened 2024-05-06
Use more of NNS wrapper in adm

roman-khimov:


go.mod
```diff
@@ -17,7 +17,7 @@ require (
 	github.com/nspcc-dev/locode-db v0.6.0
 	github.com/nspcc-dev/neo-go v0.105.1
 	github.com/nspcc-dev/neofs-api-go/v2 v2.14.1-0.20240305074711-35bc78d84dc4
-	github.com/nspcc-dev/neofs-contract v0.19.2-0.20240427112531-286b14199fd0
+	github.com/nspcc-dev/neofs-contract v0.19.2-0.20240506202632-e78d64ecdfc2
```
carpawell, line 20: can you specify this update somehow in the commit's message? usually, i prefer to mention updating a dep as a commit's reason to exist and some usage bonuses rather than fixing something and absolutely no info about updating a dep

carpawell (approved):

go.mod
roman-khimov: It's _so_ internal and we'll be updating it again.

roman-khimov: Fixed though.

PR #2852, opened 2024-05-20
Storage list for lens

roman-khimov:


cthulhu-rider (approved):
overall good

desc says that cmd lists objects in blobstor. At the same time, it does not access the blobstor, only metabase is read. May be this worth to be clarified

roman-khimov:
> May be this worth to be clarified

Done.

cmd/neofs-lens/internal/storage/list.go
```diff
@@ -0,0 +1,48 @@
...
+
+func listFunc(cmd *cobra.Command, _ []string) {
+	// other targets can be supported
+	w := cmd.OutOrStderr()
+
+	storage := openEngine(cmd)
+	defer storage.Close()
+
+	var p engine.ListWithCursorPrm
+	p.WithCount(1024)
```
carpawell, line 32: magic? can lose 1025-th object and do not understand it?

carpawell (commented):
@roman-khimov, as @cthulhu-rider said that is a meta operation and the lens' `meta` subcommand already has it, so why do we need it? It can be confusing.

cmd/neofs-lens/internal/storage/list.go
roman-khimov: It will list everything. Yeah, magic.

carpawell: oh, almost endless `for`, did not see it. check context b/w iteration then to make it controllable?

carpawell (approved):
The work is done and it does what it claims so OK, but I would not add this command as the blobstor is a too complex thing to inspect IMO, and it is not about storage only but about managing and background processes too. Also, it is changed quite often.

PR #2863, opened 2024-06-10
Fix/GC big objects parts

### LINKED ISSUE #2858: GC has not collected an expired object
## Expected Behavior
Expired objects are deleted. Even if the node was down for some time, it should delete them afterwards.

## Current Behavior
```
May 30 14:54:27 metis2 neofs-node[656]: 2024-05-30T14:54:27.505Z        error        replicator/process.go:76        could not replicate object        {"component": "Object Replicator", "node": "03aeff8a19f0202090afb0916b1c00b432321be7e8623a06c9b9b5db8ee5c053a4", "object": "HXSaMJXk2g8C14ht8HSi7BBaiYZ1HeWh2xnWPGQCg4H6/6Ha2WhpvBsAeT23jsa9Ak8DVrqVB97xLXTSeLrjYnoGD", "error": "copy object using NeoFS API client of the remote node: status: code = 1024 message = failed to verify and store object locally: validate object format: object did not pass expiration check: object has expired"}
```
It tries to replicate the same set of objects again and again which means GC has failed to do its job.

## Possible Solution
Fix GC.

Shut an existing node down for some time and expand it with one shard (3->4).

## Your Environment
* Version used: 0.42.0
* Server setup and configuration: mainnet
* Operating System and version (`uname -a`): Debian stable

carpawell:
Closes #2858.

### DIFF
--- a/pkg/core/object/object.go
+++ b/pkg/core/object/object.go
@@ -2,6 +2,7 @@ package object
 
 import (
 	"errors"
+	"strconv"
 
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
@@ -25,4 +26,21 @@ func AddressOf(obj *object.Object) oid.Address {
 	}
 
 	return addr
+}
+
+// ErrNoExpiration means no expiration was set.
+var ErrNoExpiration = errors.New("missing expiration epoch attribute")
+
+// Expiration searches for expiration attribute in the object. Returns
+// ErrNoExpiration if not found.
+func Expiration(obj object.Object) (uint64, error) {
+	for _, a := range obj.Attributes() {
+		if a.Key() != object.AttributeExpirationEpoch {
+			continue
+		}
+
+		return strconv.ParseUint(a.Value(), 10, 64)
+	}
+
+	return 0, ErrNoExpiration
 }
--- a/pkg/local_object_storage/engine/engine_test.go
+++ b/pkg/local_object_storage/engine/engine_test.go
@@ -14,6 +14,7 @@ import (
 	meta "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/metabase"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/pilorama"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/shard"
+	"github.com/nspcc-dev/neofs-node/pkg/util"
 	"github.com/nspcc-dev/neofs-sdk-go/checksum"
 	checksumtest "github.com/nspcc-dev/neofs-sdk-go/checksum/test"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
@@ -28,10 +29,12 @@ import (
 	"go.uber.org/zap"
 )
 
-type epochState struct{}
+type epochState struct {
+	e uint64
+}
 
 func (s epochState) CurrentEpoch() uint64 {
-	return 0
+	return s.e
 }
 
 func BenchmarkExists(b *testing.B) {
@@ -128,7 +131,15 @@ func testNewShard(t testing.TB, id int) *shard.Shard {
 			meta.WithPath(filepath.Join(t.Name(), fmt.Sprintf("%d.metabase", id))),
 			meta.WithPermissions(0700),
 			meta.WithEpochState(epochState{}),
-		))
+		),
+		shard.WithGCWorkerPoolInitializer(func(sz int) util.WorkerPool {
+			pool, err := ants.NewPool(sz)
+			if err != nil {
+				panic(err)
+			}
+
+			return pool
+		}))
 
 	require.NoError(t, s.Open())
 	require.NoError(t, s.Init())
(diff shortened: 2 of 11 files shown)


carpawell:
Found corner case in shard distribution, will add tests and fix.

carpawell:
OK, a fresh view and an expended test showed that it did not work. Let it be simple and error-proof.

pkg/core/object/object.go
```diff
@@ -26,3 +27,20 @@ func AddressOf(obj *object.Object) oid.Address {
 
 	return addr
 }
+
+// ErrNoExpiration means no expiration was set.
+var ErrNoExpiration = errors.New("missing expiration epoch attribute")
+
+// Expiration searches for expiration attribute in the object. Returns
+// ErrNoExpiration if not found.
+func Expiration(obj *object.Object) (uint64, error) {
```
cthulhu-rider, line 36:
```suggestion
func Expiration(obj object.Object) (uint64, error) {
```

pkg/local_object_storage/engine/engine_test.go
```diff
@@ -28,10 +29,12 @@ import (
...
 )
 
-type epochState struct{}
+type epochState struct {
+	e uint64
+}
 
-func (s epochState) CurrentEpoch() uint64 {
-	return 0
+func (s *epochState) CurrentEpoch() uint64 {
```
cthulhu-rider, line 36: why switched to pointer?

pkg/services/replicator/process.go
```diff
@@ -45,6 +49,11 @@ func (p *Replicator) HandleTask(ctx context.Context, task Task, res TaskResult)
 			stream = client.DemuxReplicatedObject(stream)
 		}
 	} else {
+		if err = p.checkExpiration(task.obj); err != nil {
```
cthulhu-rider, line 52: minor: this is not Replicator's concern imo, i'd make such exclusions on caller side

pkg/services/replicator/process.go
```diff
@@ -45,6 +49,11 @@ func (p *Replicator) HandleTask(ctx context.Context, task Task, res TaskResult)
...
+			p.log.Debug("expiration check", zap.Error(err))
+			return
```
cthulhu-rider, line 54: iiuc `checkExpiration` eternal objects w/o parent are never replicated. I doubt this is a correct behavior

pkg/services/replicator/process.go
```diff
@@ -85,3 +94,24 @@ func (p *Replicator) HandleTask(ctx context.Context, task Task, res TaskResult)
 		}
 	}
 }
+
+func (p *Replicator) checkExpiration(o *objectSDK.Object) error {
```
cthulhu-rider, line 98:
```suggestion
func (p *Replicator) checkExpiration(o objectSDK.Object) error {
```

pkg/services/replicator/process.go
```diff
@@ -45,6 +49,11 @@ func (p *Replicator) HandleTask(ctx context.Context, task Task, res TaskResult)
 			stream = client.DemuxReplicatedObject(stream)
 		}
 	} else {
+		if err = p.checkExpiration(task.obj); err != nil {
+			p.log.Debug("expiration check", zap.Error(err))
```
cthulhu-rider, line 53: i'd `Info` (at least) cases with broken exp attributes cuz we don't expect them in general

carpawell: ok, added details and made it warn

pkg/services/replicator/process.go
carpawell: fixed

pkg/services/replicator/process.go
carpawell: objects are taken in a batch and handled via a pool of _some size_ in the Policer, so Replicator to me is the closest thing to the PUT/REPLICATE operation and is the best thing to abort operation if an object is expired b/w engine's SELECT and network replication. but i understand you, can move it to Policer @roman-khimov

pkg/local_object_storage/engine/engine_test.go
carpawell: rework artifact, rolled back

pkg/services/replicator/process.go
roman-khimov: Caller doesn't have object header to judge on this IIUC. And replicator has to get it anyway.

roman-khimov: Ah, it can have it in this case. Then policer would be a better place to keep this code doing exactly one thing.

carpawell: @roman-khimov, @cthulhu-rider, i rechecked, a full obj can be here only in evacuation case, in a normal replication case it is now binary replication, so it is impossible to get expiration (or make additional unmarshaling after we make replication without unmarshaling 😋). so only evacuation can be improved (good, can be done but is almost useless to me), i would just drop this commit then (evacuation can be improved, but this is a control function, so even no logs can be done if smth is skipped)

carpawell: dropped commit

pkg/core/object/object.go
```diff
@@ -26,3 +27,20 @@ func AddressOf(obj *object.Object) oid.Address {
 
 	return addr
 }
+
+// ErrNoExpiration means no expiration was set.
+var ErrNoExpiration = errors.New("missing expiration epoch attribute")
+
+// Expiration searches for expiration attribute in the object. Returns
+// ErrNoExpiration if not found.
+func Expiration(obj object.Object) (uint64, error) {
```
roman-khimov, line 36: Do we need this if we're not touching replicator?

cthulhu-rider (approved):

pkg/core/object/object.go
carpawell: this can be useful in general as i mentioned in the commit, i just had not searched if more places can reused it

Issue #2800, opened 2024-04-02
Do not ignore errors ever


carpawell:
## Expected Behavior
Never ignore errors in gonang.

## Current Behavior
Code may ignore some errors (even if they are not expected at all, that is not good).

## Possible Solution
Panic if an error is not expected at all.

## Context
At least CLIs are affected: [binding flags](https://github.com/nspcc-dev/neofs-node/blob/97a2dc7414ca80092fcf4b55dd2d347d7d4d00d5/cmd/neofs-cli/modules/root.go#L70), [getting flags](https://github.com/nspcc-dev/neofs-node/blob/97a2dc7414ca80092fcf4b55dd2d347d7d4d00d5/cmd/neofs-lens/root.go#L24), [marking flags](https://github.com/nspcc-dev/neofs-node/blob/97a2dc7414ca80092fcf4b55dd2d347d7d4d00d5/cmd/neofs-lens/internal/flags.go#L27-L28).
Some other places may be affected too.

roman-khimov:
Narrow down the scope, please. This is too broad.

PR #2867, opened 2024-06-11
adm: add mint-balance command to manually top-up balances

roman-khimov:


cmd/neofs-adm/internal/modules/morph/root.go
```diff
@@ -146,6 +150,36 @@ var (
 		},
 	}
 
+	mintBalanceCmd = &cobra.Command{
+		Use:   "mint-balance",
+		Short: "Mint new NEOFS tokens in the morph network",
+		PreRun: func(cmd *cobra.Command, _ []string) {
+			_ = viper.BindPFlag(alphabetWalletsFlag, cmd.Flags().Lookup(alphabetWalletsFlag))
```
carpawell, line 157: https://github.com/nspcc-dev/neofs-node/pull/2799#discussion_r1548117070 and https://github.com/nspcc-dev/neofs-node/issues/2800?

roman-khimov: I'm just doing what every other command does here. And the problem is that we can't return error.

carpawell: this func should not break so we can just exit/panic here

PR #2868, opened 2024-06-11
fix/`TestEvacuateShard` test

carpawell:


carpawell:
Disgusting debugging.

On the other hand, [this](https://github.com/nspcc-dev/neofs-node/blob/b105cbf7609cf2839f7c5ff550498b3669fd336b/pkg/local_object_storage/engine/shards.go#L120) (`WithNonblocking`) looks not good to me and can be changed. It may be the reason of objects-on-wrong-shards effects and also may lead to an error when all shards are busy (i would prefer to wait for it and fail with deadline). @roman-khimov, @cthulhu-rider

pkg/local_object_storage/engine/put.go
```diff
@@ -138,6 +143,10 @@ func (e *StorageEngine) putToShard(sh hashedShard, ind int, pool util.WorkerPool
 				}
 			}
 
+			e.log.Debug("object put: object already exist",
```
roman-khimov, line 146: exists

pkg/local_object_storage/engine/put.go
```diff
@@ -163,6 +172,7 @@ func (e *StorageEngine) putToShard(sh hashedShard, ind int, pool util.WorkerPool
 
 		putSuccess = true
 	}); err != nil {
+		e.log.Warn("object put: pool task submitting", zap.Error(err))
```
cthulhu-rider, line 175: how this affects the system and/or what's the admin's reaction to this message?

pkg/local_object_storage/engine/evacuate.go
```diff
@@ -164,6 +164,8 @@ mainLoop:
 						}
 						continue loop
 					}
+
+					e.log.Debug("could not put to shard, trying another", zap.String("shard", shards[j].ID().String()))
```
cthulhu-rider, line 168: if last shard failed there wont be more `trying`

pkg/local_object_storage/engine/evacuate_test.go
```diff
@@ -28,7 +28,7 @@ func newEngineEvacuate(t *testing.T, shardNum int, objPerShard int) (*StorageEng
 
 	e := New(
 		WithLogger(zaptest.NewLogger(t)),
-		WithShardPoolSize(1))
+		WithShardPoolSize(uint32(objPerShard)))
```
cthulhu-rider, line 31: in general, doubtful fix to me possibly hiding buggy implementation or test

pkg/local_object_storage/engine/put.go
carpawell: load to this shard currently is bigger than it can handle. also, we have a system that says where to put an object but now (this line) it is not gonna work: an object goes to another shard (we had (have?) bugs that relate such cases). in fact, it always bothered me that object put was always kinda random, you never know from logs if this placement is ok or a "bad" shard was taken as a best-effort

pkg/local_object_storage/engine/evacuate.go
carpawell: not a problem to me, classic iteration: "try another shard; shard list is over; i finished cycle"

carpawell: what exactly would you expect here?

pkg/local_object_storage/engine/evacuate_test.go
carpawell: > doubtful fix to me

why? this is _literally_ the problem, i proofed it locally, required 100k+ test runs. `ants.Pool` has background workers and some logic that blocks execution (calculates num of wip workers), if it does not free it before another iteration is being done, current shard logic just tries another shard and duplicates objects

carpawell: in other words, 1-sized pool cannot guarantee that an object will be put if you want to put 2+ objects. on a few-core machine, it is more critical, on my notebook it is like ~ 0.003%. but an evacuate test should not suffer because of the put problem IMO, it should test evacuation logic and it cannot, because objects are duplicated and go to unexpected shards

pkg/local_object_storage/engine/put.go
carpawell: this PR (and the issue) is a real example BTW. if i had this log, i wouldn't have had to run this test so many times trying to understand what was happening, it would have taken 1 min looking at logs

roman-khimov (approved):

PR #2872, opened 2024-06-14
Update neofs-contract to v0.20.0

roman-khimov:
As a part of https://github.com/nspcc-dev/neofs-contract/issues/384 I'd like to unify different code dealing with them.


roman-khimov:
@evgeniiz321, tests fail here and I can't get any logs.

roman-khimov:
This updates contracts as well, so it's gonna be postponed till we have a proper 0.20.0.

carpawell (changes requested):
Tests adoption.

cmd/neofs-adm/internal/modules/morph/initialize_deploy.go
```diff
@@ -429,12 +435,14 @@ func readContractsFromArchive(file io.Reader, names []string) (map[string]*contr
 		}
 
 		switch {
-		case strings.HasSuffix(h.Name, filepath.Join(ctrName, ctrName+"_contract.nef")):
+		case strings.HasSuffix(h.Name, filepath.Join(ctrName, ctrName+"_contract.nef")) ||
+			strings.HasSuffix(h.Name, "contract.nef"):
```
cthulhu-rider, line 439: 2nd operand is enough

cmd/neofs-adm/internal/modules/morph/initialize_deploy.go
```diff
@@ -372,11 +372,17 @@ func (c *initializeContext) readContracts(names []string) error {
 func readContract(ctrPath, ctrName string) (*contractState, error) {
 	rawNef, err := os.ReadFile(filepath.Join(ctrPath, ctrName+"_contract.nef"))
 	if err != nil {
-		return nil, fmt.Errorf("can't read NEF file for %s contract: %w", ctrName, err)
+		rawNef, err = os.ReadFile(filepath.Join(ctrPath, "contract.nef"))
```
carpawell, line 375: go this way only if file not found? also, do you play to drop it in the future?

pkg/morph/client/netmap/netmap.go
```diff
@@ -4,7 +4,7 @@ import (
 	"fmt"
 
 	"github.com/nspcc-dev/neo-go/pkg/vm/stackitem"
-	netmapcontract "github.com/nspcc-dev/neofs-contract/contracts/netmap"
+	netmaprpc "github.com/nspcc-dev/neofs-contract/rpc/netmap"
```
cthulhu-rider, line 7: i liked previous alias more tbh

cthulhu-rider (commented):
good overall

cmd/neofs-adm/internal/modules/morph/initialize_deploy.go
cthulhu-rider: @carpawell is faster, but i'll leave my comment anyway

err may be other than 404, so i'd either:

1. try other file on 404 only
2. try anyway, but keep both errors when it fails too

same for manifest files

cmd/neofs-adm/internal/modules/morph/initialize_deploy.go
roman-khimov: Yep, fixed.

cmd/neofs-adm/internal/modules/morph/initialize_deploy.go
carpawell: > also, do you plan to drop it in the future?

why do we try to support both ways? how long?

roman-khimov: Because it still is a generic tool. We'll see. An issue can be created.

carpawell (approved):

cthulhu-rider (approved):

PR #2873, opened 2024-06-18
Fix/drop control call

### LINKED ISSUE #2822: Dropping object via control service leaves traces in the metabase
## Expected Behavior
`neofs-cli control drop-objects` removing all data about the object from the node.

## Current Behavior
Apparently, some traces are still left in the metabase after it, this:
```
Apr 20 04:09:14 titan4 neofs-node[420363]: 2024-04-20T04:09:14.760Z warn shard/gc.go:339 iterator over expired locks failed {"shard_id": "2N1drpyfAuTXes7ShXUAF4", "error": "could not parse expiration epoch: strconv.ParseUint: parsing "'or'1'='1": invalid syntax"}
```
only ends with metabase resync (broken object is deleted via drop-objects).

## Possible Solution
Clean up _all_ metabase data on `drop-objects`.

Add some broken object (likely not that easy now), delete it, see that GC is still broken.

## Context
Testnet.

## Your Environment
* Version used: 0.41.0

carpawell:
Overall, should close #2822, hard to say, there were many commits merged about it, and "problem" objects are already dropped.

### DIFF
--- a/pkg/local_object_storage/engine/delete.go
+++ b/pkg/local_object_storage/engine/delete.go
@@ -1,11 +1,8 @@
 package engine
 
 import (
-	"errors"
-
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/shard"
 	apistatus "github.com/nspcc-dev/neofs-sdk-go/client/status"
-	objectSDK "github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"go.uber.org/zap"
 )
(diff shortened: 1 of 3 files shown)


pkg/local_object_storage/engine/delete.go
```diff
@@ -58,102 +55,25 @@ func (e *StorageEngine) delete(prm DeletePrm) (DeleteRes, error) {
...
-		if err != nil {
-			e.log.Warn("error during searching for object children",
-				zap.Stringer("addr", addr),
-				zap.String("error", err.Error()))
-			return false
-		}
-
-		for _, addr := range res.AddressList() {
-			inhumePrm.MarkAsGarbage(addr)
+	_, err := e.inhumeAddr(prm.addr, inhumePrm)
```
roman-khimov, line 76: Looks like it was a bit more efficient previously with _sorted_ shards.

pkg/local_object_storage/engine/delete.go
```diff
@@ -58,102 +55,25 @@ func (e *StorageEngine) delete(prm DeletePrm) (DeleteRes, error) {
...
-
-		resExists, err := sh.Exists(existsPrm)
-		if err != nil {
-			if shard.IsErrRemoved(err) || shard.IsErrObjectExpired(err) {
-				return true
-			}
-
-			var splitErr *objectSDK.SplitInfoError
-			if !errors.As(err, &splitErr) {
-				if !shard.IsErrNotFound(err) {
```
roman-khimov, line 82: This one differs from the two implementations.

pkg/local_object_storage/metabase/delete.go
```diff
@@ -141,25 +130,15 @@ func (db *DB) deleteGroup(tx *bbolt.Tx, addrs []oid.Address, sizes []uint64) (ui
...
-			}
-		}
-	}
-
 	return rawDeleted, availableDeleted, nil
 }
 
-// delete removes object indexes from the metabase. Counts the references
-// of the object that is being removed.
+// delete removes object indexes from the metabase.
```
roman-khimov, line 138: So, you're relying on engine to drop every object from the metabase now, right?

pkg/local_object_storage/engine/delete.go
carpawell: why do you think it is unsorted now?

pkg/local_object_storage/engine/delete.go
carpawell: yes, that is the key of my suggestion: why do we differ this logic? we do the same things (and one may always have more or less num of bugs) when delete/inhume objects, why some of them should be collected (if an object is a big one) one way while the other should be handled differently? about this PR: i have improved and fixed a little `Inhume` some time ago and could not understand why i had do it again one more time when looking at `Delete`

pkg/local_object_storage/metabase/delete.go
```diff
@@ -103,7 +104,8 @@ func (db *DB) deleteGroup(tx *bbolt.Tx, addrs []oid.Address, sizes []uint64) (ui
 	for i := range addrs {
 		removed, available, size, err := db.delete(tx, addrs[i], currEpoch)
 		if err != nil {
-			return 0, 0, err // maybe log and continue?
+			db.log.Warn("failed to delete object", zap.Error(err), zap.Stringer("addr", addrs[i]))
+			return 0, 0, nil
```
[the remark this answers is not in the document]
carpawell: sounds reasonable, will change

pkg/local_object_storage/metabase/delete.go
carpawell: yes, that is the point. i started to think that "complex" logic should be done on the engine level and that is why the engine exists. If a metabase is trying to handle parts/parents that it may lose (resync or it just never saw any part with parent header), there are always some problems. that is what i want to discuss and open another issue if any needed

pkg/local_object_storage/engine/delete.go
roman-khimov:
```
        e.iterateOverUnsortedShards(func(sh hashedShard) (stop bool) {
```
in `inhumeAddr`

carpawell: you mean [this](https://github.com/nspcc-dev/neofs-node/blob/0e4f7c919e712ca77e7ce15140599de0e644bb69/pkg/local_object_storage/engine/inhume.go#L150)? it is root object handling, there is no information about link and last objects that can be acquired via root object's ID

pkg/local_object_storage/metabase/delete.go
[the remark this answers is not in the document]
roman-khimov: Anything changed here?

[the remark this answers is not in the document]
carpawell: a regular simple change, the other threads worries me more. but ok, fixed

pkg/local_object_storage/metabase/delete.go
```diff
@@ -116,6 +127,12 @@ func (db *DB) deleteGroup(tx *bbolt.Tx, addrs []oid.Address, sizes []uint64) (ui
 		}
 	}
 
+	if firstErr != nil {
+		all := uint64(len(addrs))
+		success := all - errorCount
+		return 0, 0, fmt.Errorf("deleted %d out of %d objects, first error: %w", success, all, firstErr)
```
cthulhu-rider, line 133: lets mention best effort also in `Delete` docs?

pkg/local_object_storage/engine/delete.go
```diff
@@ -58,102 +55,25 @@ func (e *StorageEngine) delete(prm DeletePrm) (DeleteRes, error) {
...
-
-		var shPrm shard.InhumePrm
-		shPrm.MarkAsGarbage(prm.addr)
-		if prm.forceRemoval {
-			shPrm.ForceRemoval()
-		}
-
-		_, err = sh.Inhume(shPrm)
+	if !prm.forceRemoval {
+		locked, err := e.IsLocked(prm.addr)
```
cthulhu-rider, line 59:
```suggestion
		locked, err := e.isLocked(prm.addr)
```
no need to acquire `blockExec` mutex again

pkg/local_object_storage/metabase/delete.go
```diff
@@ -105,16 +96,25 @@ func (db *DB) Delete(prm DeletePrm) (DeleteRes, error) {
...
+	var errorCount uint64
+	var firstErr error
 
 	for i := range addrs {
-		removed, available, size, err := db.delete(tx, addrs[i], refCounter, currEpoch)
+		removed, available, size, err := db.delete(tx, addrs[i], currEpoch)
 		if err != nil {
-			return 0, 0, err // maybe log and continue?
+			errorCount++
+			err = fmt.Errorf("%s object delete fail: %w", addrs[i], err)
```
cthulhu-rider, line 110: this makes no sense for errors after the 1st: only needed for the log, and it will contain a tautology then. Suggest
```
db.log.Warn("failed to delete object", zap.Stringer("addr", addres[i]), zap.Error(err))
if firstErr == nil {
  err = fmt.Errorf("%s object delete fail: %w", addrs[i], err)
}
```

pkg/local_object_storage/metabase/delete.go
```diff
@@ -127,6 +127,12 @@ func (db *DB) deleteGroup(tx *bbolt.Tx, addrs []oid.Address, sizes []uint64) (ui
 		}
 	}
 
+	if firstErr != nil {
+		all := uint64(len(addrs))
```
cthulhu-rider, line 131: `var errorCount int` and no cast and temp vars  needed

cthulhu-rider (commented):
5b6b5a4ad54abe224a88fdc16c6fae07d665a3e2 explains its changes by 6bf698489acc1313eafffb189ea13e2ef7e5ac55, maybe just squash them then?

carpawell:
> 5b6b5a4 explains its changes by 6bf6984, maybe just squash them then?

I prefer a couple of commits. The first one makes `Delete` work as `Inhume`, and the second drops no-op code since the previous change. IMO, they are read better this way.

cthulhu-rider:
> the second drops no-op code

i thought that `Delete` code became less optimal after the 1st. If just no-op, then yeah - separate commits are much better

pkg/local_object_storage/engine/delete.go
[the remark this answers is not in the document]
cthulhu-rider: >  i could not recall why we have them separated

no reason mentioned #1461

> it scares me

on the contrary, this inspires me to break duality and merge them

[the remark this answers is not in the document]
carpawell: > this inspires me to break duality and merge them

me too. consider the whole PR as a suggestion, the PR can be extended, more issues can be opened, etc. an option (`force`) looks more reliable than having two methods

> no reason mentioned #1461

it was always like `Delete` + `Inhume` and a lot of changes were made, so i think it is a little complex to find a reason (and it is not worth it, imo)

cthulhu-rider (approved):

carpawell:
> @carpawell worth changelog? the issue occurred in the pub network

Not sure I can find correct words about the whole PR. But if you are about the original issue, we may add it, yes.

PR #2875, opened 2024-06-26
Feat/cli delete several objects

carpawell:


cmd/neofs-cli/modules/object/delete.go
```diff
@@ -28,53 +28,67 @@ func initObjectDeleteCmd() {
 	flags := objectDelCmd.Flags()
 
 	flags.String(commonflags.CIDFlag, "", commonflags.CIDFlagUsage)
-	flags.String(commonflags.OIDFlag, "", commonflags.OIDFlagUsage)
+	flags.StringSlice(commonflags.OIDFlag, nil, commonflags.OIDFlagUsage)
 	flags.Bool(binaryFlag, false, "Deserialize object structure from given file.")
 	flags.String(fileFlag, "", "File with object payload")
+
+	_ = objectDelCmd.MarkFlagRequired(commonflags.CIDFlag)
```
roman-khimov, line 35: Separate thing?

roman-khimov (commented):
IIUC old commands (`--oid XYZ`) still work the same way with the same CLI output?

cmd/neofs-cli/modules/object/delete.go
carpawell: it was not possible to implement multiple objects without this change so it is the same change to me, but ok, one more commit has landed

cmd/neofs-cli/modules/object/util.go
```diff
@@ -250,13 +245,7 @@ func OpenSession(ctx context.Context, cmd *cobra.Command, dst SessionPrm, key *e
...
 // If provided SessionPrm is of type internal.DeleteObjectPrm, OpenSessionViaClient
 // spreads the session to all object's relatives.
-func OpenSessionViaClient(ctx context.Context, cmd *cobra.Command, dst SessionPrm, cli *client.Client, key *ecdsa.PrivateKey, cnr cid.ID, obj *oid.ID) {
-	var objs []oid.ID
-
-	if obj != nil {
-		objs = []oid.ID{*obj}
-	}
-
+func OpenSessionViaClient(ctx context.Context, cmd *cobra.Command, dst SessionPrm, cli *client.Client, key *ecdsa.PrivateKey, cnr cid.ID, objs ...oid.ID) {
```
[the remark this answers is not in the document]
carpawell: do not understand your question. do you mean protobuff restrictions or cobra?

carpawell:
> IIUC old commands (--oid XYZ) still work the same way with the same CLI output?

Yes (almost, it now prints an object ID that has been removed).

cmd/neofs-cli/modules/object/util.go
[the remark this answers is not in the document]
roman-khimov: Protocol. CLI is more predictable to me here.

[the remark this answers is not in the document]
carpawell: why are you interested in? it is used only when you create an object session token and is attached to the object (a tombstone in our case), so the current header limitation says _no more_ than 512 objects in one session token. tombstones are temporary objects so it is not that critical but i made it a one-object token per request, check it. was it smth that bothered you?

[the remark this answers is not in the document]
roman-khimov: We need to know our limits, that's it. Multi-object token is perfectly fine.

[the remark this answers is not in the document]
carpawell: > We need to know our limits, that's it.

a header is 16K max, token should be placed here, so 512 is our limit (if ~200 bytes are always reserved for some useful data, it is 505 then)

> Multi-object token is perfectly fine.

still think it is useless in this case, storing a big token in every tombstone is strange

cmd/neofs-cli/modules/object/util.go
```diff
@@ -250,13 +245,7 @@ func OpenSession(ctx context.Context, cmd *cobra.Command, dst SessionPrm, key *e
 //
 // If provided SessionPrm is of type internal.DeleteObjectPrm, OpenSessionViaClient
```
cthulhu-rider, line 246: i read https://github.com/nspcc-dev/neofs-node/pull/2875/files#r1655445343, then noticed this place, and wanted to clarify that input could be much smaller for big objects. But this is irrelevant since e97444aa743322da7876441b1b95c563fc902b10. Let drop, and pls more attention to docs

cmd/neofs-cli/modules/object/delete.go
```diff
@@ -28,53 +28,67 @@ func initObjectDeleteCmd() {
...
 
-	res, err := internalclient.DeleteObject(ctx, prm)
-	common.ExitOnErr(cmd, "rpc error: %w", err)
+	for i, addr := range objAddrs {
+		ReadOrOpenSession(ctx, cmd, &prm, pk, cnr, objs[i])
+		prm.SetAddress(addr)
 
-	tomb := res.Tombstone()
+		res, err := internalclient.DeleteObject(ctx, prm)
+		common.ExitOnErr(cmd, "rpc error: %w", err)
```
cthulhu-rider, line 87: lets add `addr`

cmd/neofs-cli/modules/object/delete.go
```diff
@@ -28,53 +28,67 @@ func initObjectDeleteCmd() {
...
+	var objAddrs []oid.Address
 
 	binary, _ := cmd.Flags().GetBool(binaryFlag)
 	if binary {
 		filename, _ := cmd.Flags().GetString(fileFlag)
 		if filename == "" {
 			common.ExitOnErr(cmd, "", fmt.Errorf("required flag \"%s\" not set", fileFlag))
 		}
-		objAddr = readObjectAddressBin(cmd, &cnr, &obj, filename)
+		var obj oid.ID
```
cthulhu-rider, line 53: create `objs` first and no temp var needed

cmd/neofs-cli/modules/object/delete.go
```diff
@@ -28,53 +28,67 @@ func initObjectDeleteCmd() {
...
-			common.ExitOnErr(cmd, "", fmt.Errorf("required flag \"%s\" not set", commonflags.CIDFlag))
-		}
+		readCID(cmd, &cnr)
+		oIDVals, _ := cmd.Flags().GetStringSlice(commonflags.OIDFlag)
 
-		oidVal, _ := cmd.Flags().GetString(commonflags.OIDFlag)
-		if oidVal == "" {
-			common.ExitOnErr(cmd, "", fmt.Errorf("required flag \"%s\" not set", commonflags.OIDFlag))
-		}
+		var obj oid.ID
```
cthulhu-rider, line 62: same here

cmd/neofs-cli/modules/object/delete.go
```diff
@@ -28,53 +28,67 @@ func initObjectDeleteCmd() {
...
 
 func deleteObject(cmd *cobra.Command, _ []string) {
 	ctx, cancel := commonflags.GetCommandContext(cmd)
 	defer cancel()
 
 	var cnr cid.ID
-	var obj oid.ID
-	var objAddr oid.Address
+	var objs []oid.ID
+	var objAddrs []oid.Address
```
cthulhu-rider, line 45: single `objAddr` is enough

cmd/neofs-cli/modules/object/delete.go
```diff
@@ -28,53 +28,67 @@ func initObjectDeleteCmd() {
...
-		if oidVal == "" {
-			common.ExitOnErr(cmd, "", fmt.Errorf("required flag \"%s\" not set", commonflags.OIDFlag))
-		}
+		var obj oid.ID
+		var objAddr oid.Address
+		objAddr.SetContainer(cnr)
+
+		for _, oIDraw := range oIDVals {
+			err := obj.DecodeString(oIDraw)
+			common.ExitOnErr(cmd, "decode object ID string: %w", err)
```
cthulhu-rider, line 68: lets add `oIDraw`

cmd/neofs-cli/modules/object/delete.go
```diff
@@ -28,53 +28,67 @@ func initObjectDeleteCmd() {
...
 
-	tomb := res.Tombstone()
+		res, err := internalclient.DeleteObject(ctx, prm)
+		common.ExitOnErr(cmd, "rpc error: %w", err)
 
-	cmd.Println("Object removed successfully.")
-	cmd.Printf("  ID: %s\n  CID: %s\n", tomb, cnr)
+		tomb := res.Tombstone()
+
+		cmd.Printf("Object %s removed successfully.\n", addr)
```
cthulhu-rider, line 91: may be
```suggestion
		cmd.Printf("Object %s removed successfully.\n", objs[i])
```
? it's redundant to write same container again and again

cthulhu-rider:
good overall

cmd/neofs-cli/modules/object/delete.go
```diff
@@ -28,53 +28,67 @@ func initObjectDeleteCmd() {
...
+
+		objs = []oid.ID{obj}
+		objAddrs = []oid.Address{objAddr}
 	} else {
-		cidVal, _ := cmd.Flags().GetString(commonflags.CIDFlag)
-		if cidVal == "" {
-			common.ExitOnErr(cmd, "", fmt.Errorf("required flag \"%s\" not set", commonflags.CIDFlag))
-		}
+		readCID(cmd, &cnr)
+		oIDVals, _ := cmd.Flags().GetStringSlice(commonflags.OIDFlag)
```
cthulhu-rider, line 60: lets check `len(oIDVals) > 0`, seems like `objectDelCmd.MarkFlagRequired(commonflags.OIDFlag)` is not enough
```
$ neofs-cli object delete --cid F74twfixT89Wqc4qh32Yj5vzP6p9gQ7dxJCzp3C311an 
Error: required flag(s) "oid" not set
$ neofs-cli object delete --cid F74twfixT89Wqc4qh32Yj5vzP6p9gQ7dxJCzp3C311an --oid '' --oid '' && echo $?
0
```

#2879

cmd/neofs-cli/modules/object/delete.go
carpawell: dropped `objs` var

cmd/neofs-cli/modules/object/delete.go
carpawell: not sure i understand you. it is not always more than one object (if `binary` is provided) and we do not always read `GetStringSlice(commonflags.OIDFlag)` and know its size in advance, only in the first `if` branch. you mean creating a nonempty `objs` slice and just changing it? i do not like this approach usually, it seems error-prone to me, i prefer always to `append`

if you mean to create common vars for all branched, then it will be 
```go
	var cnr cid.ID
	var obj oid.ID
	var objs []oid.ID
	var addr oid.Address
	var objAddrs []oid.Address
```

confusing to me too, creating a var in a place where it is used is a better option for me

cmd/neofs-cli/modules/object/delete.go
carpawell: dont mind

roman-khimov (approved):

cthulhu-rider (approved):

Issue #2844, opened 2024-04-28
Replicator can't fetch local object


roman-khimov:
## Expected Behavior
No errors.

## Current Behavior
```
Apr 28 08:34:14 titan4 neofs-node[709]: 2024-04-28T08:34:14.639Z        error        replicator/process.go:37        could not get object from local storage        {"component": "Object Replicator", "object": "7dppsTsDszYwN5P9yH7G7ubEbrxRAAVC27SZ356QRkvY/4tvV8gaSiCvT5CgVr35vV4EeFz9WDtgqyak1tFRsQBLH", "error": "status: code = 2049 message = object not found"}
Apr 28 08:37:47 titan4 neofs-node[709]: 2024-04-28T08:37:47.609Z        error        replicator/process.go:37        could not get object from local storage        {"component": "Object Replicator", "object": "EgDAUwMadH55dZznrsentnHJFfqnE2gPUAW7zDmDvYuE/2C2oEuAzNLA1zH8u1DKTR5494vQtPgmJ7Wysuih3Eqe8", "error": "status: code = 2049 message = object not found"}
Apr 28 08:40:32 titan4 neofs-node[709]: 2024-04-28T08:40:32.906Z        error        replicator/process.go:37        could not get object from local storage        {"component": "Object Replicator", "object": "D6AJH9UwLiEG2hh8JHaCt798WePzCYDob7TYEWfCE11M/3aFs6YJBXnGcqPXL3AgWuLYJvUGp1FPF8p4iBdHn6uTj", "error": "status: code = 2049 message = object not found"}
Apr 28 08:40:40 titan4 neofs-node[709]: 2024-04-28T08:40:40.373Z        error        replicator/process.go:37        could not get object from local storage        {"component": "Object Replicator", "object": "DivTTmXeugVPyV7hv5u7RN9Y2tMvqnnyD6mrmoSmAT1g/AQtrN3dgCTfNkMAgTsfYg8ueBhnoFXn7o68HQwWxDaY3", "error": "status: code = 2049 message = object not found"}
```

## Possible Solution
Unknown. Metabase was resynchronized some time ago, so it's unexpected.

## Context
Testnet.

## Your Environment
* Version used: 0.41.1

roman-khimov:
## Expected Behavior
No errors.

## Current Behavior
```
[...]

Issue #2506, opened 2023-08-15
Object sanity checks in lens


roman-khimov:
## Is your feature request related to a problem? Please describe.
The tool briefly introduced by #2491 tried to check some aspects of an already existing stored object.

## Describe the solution you'd like
It can be moved into `neofs-lens` and extended with more checks for internal object consistency (add comments here).

## Describe alternatives you've considered
Standalone tool like that doesn't make much sense, it's exactly for `lens` to have.

roman-khimov:
This can help with https://github.com/nspcc-dev/neofs-api/issues/293 as well (extending existing tool is easier than creating another one).

PR #2882, opened 2024-07-02
Feat/add meta indexes inspection to lens

carpawell:


cmd/neofs-lens/internal/meta/status.go
```diff
@@ -0,0 +1,48 @@
...
+	cmd.Printf("Starage's ID: %s\n", res.StorageID)
+	cmd.Println("Indexes:")
+	for _, bucket := range res.Buckets {
+		valStr := "<empty>"
+		if bucket.Value != nil {
+			valStr = fmt.Sprintf("%x", bucket.Value)
+		}
+
+		cmd.Printf("\tBucket: %d\n"+
+			"\tValue (HEX): %s\n", bucket.BucketIndex, valStr)
```
cthulhu-rider, line 46: `(HEX)` is incorrect for `<empty>`

cmd/neofs-lens/internal/meta/status.go
```diff
@@ -0,0 +1,48 @@
...
+	common.ExitOnErr(cmd, common.Errf("invalid address argument: %w", err))
+
+	db := openMeta(cmd, true)
+	defer db.Close()
+
+	res, err := db.ObjectStatus(addr)
+	common.ExitOnErr(cmd, common.Errf("reading object status: %w", err))
+
+	cmd.Printf("Object state: %s\n", res.State)
+	cmd.Printf("Starage's ID: %s\n", res.StorageID)
```
cthulhu-rider, line 37: can `res.StorageID` be empty? if so, this will be printed unclear

cmd/neofs-lens/internal/meta/status.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"fmt"
+
+	common "github.com/nspcc-dev/neofs-node/cmd/neofs-lens/internal"
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	"github.com/spf13/cobra"
+)
+
+var statCMD = &cobra.Command{
+	Use:   "stat",
```
cthulhu-rider, line 12: `status` or `state`, `stat` usually stands for statistics

pkg/local_object_storage/metabase/status.go
```diff
@@ -64,3 +79,59 @@ func (db *DB) ObjectStatus(address oid.Address) (ObjectStatus, error) {
...
+
+	for _, bucketKey := range objectBuckets {
+		b := tx.Bucket(bucketKey)
+		if b == nil {
+			continue
+		}
+
+		res = append(res, BucketValue{
+			BucketIndex: int(bucketKey[0]), // the first byte is always a prefix
+			Value:       b.Get(objKey),
```
cthulhu-rider, line 110:
```suggestion
			Value:       bytes.Clone(b.Get(objKey)),
```
here and everywhere

pkg/local_object_storage/metabase/status.go
```diff
@@ -1,14 +1,23 @@
...
 
+	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"go.etcd.io/bbolt"
 )
 
+// BucketValue pairs a bucket index and a value that relates
+// an object.
+type BucketValue struct {
+	BucketIndex int
```
cthulhu-rider, line 14: this is metabase version-specific info, right? i'd also return the version in `ObjectStatus`

pkg/local_object_storage/metabase/status.go
```diff
@@ -64,3 +79,59 @@ func (db *DB) ObjectStatus(address oid.Address) (ObjectStatus, error) {
...
+
+	containerBuckets := []byte{
+		containerVolumePrefix,
+		garbageContainersPrefix,
+	}
+
+	for _, bucket := range containerBuckets {
+		res = append(res, BucketValue{
+			BucketIndex: int(bucket),
+			Value:       tx.Bucket([]byte{bucket}).Get(cIDRaw),
```
cthulhu-rider, line 122: while according to the code such buckets must exist, even the almost zero possibility of such a panic is incorrect, after all, this is a vendor library

pkg/local_object_storage/metabase/status.go
```diff
@@ -64,3 +79,59 @@ func (db *DB) ObjectStatus(address oid.Address) (ObjectStatus, error) {
...
+	}
+
+	for _, bucketKey := range objectBuckets {
+		b := tx.Bucket(bucketKey)
+		if b == nil {
+			continue
+		}
+
+		res = append(res, BucketValue{
+			BucketIndex: int(bucketKey[0]), // the first byte is always a prefix
```
cthulhu-rider, line 109: 1st i read ur comment, then looked into `VERSION.md` file: seems like it contains incorrect info. For example:
```
- Bucket mapping first object ID to a list of objects IDs
  - Name: container ID + `19`
```
to me it's unclear that `19` is a prefix (although it's it), more like a suffix. I suggest to fix docs

cthulhu-rider:
essentially looks good

cthulhu-rider:
@carpawell can we say this relates to #2506?

pkg/local_object_storage/metabase/status.go
```diff
@@ -27,6 +28,11 @@ func (db *DB) ObjectStatus(address oid.Address) (ObjectStatus, error) {
 	storageID := StorageIDPrm{}
 	storageID.SetAddress(address)
 	resStorageID, err := db.StorageID(storageID)
+	if err != nil {
+		res.Error = fmt.Errorf("reading storage ID: %w", err)
```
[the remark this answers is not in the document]
carpawell: do you want it to be done in this PR? the whole `ObjectStatus` is strange for me, but i needed smth like this in my new command so i decided to extend it, not create a new one

cmd/neofs-lens/internal/meta/status.go
carpawell: agree. do you have an alternative? first, it was just an empty string but i did not like it as a user when i saw it so it was added. also, it is quite clear that it (`<empty>`) is not a HEX string and it cannot be misread, imo

cmd/neofs-lens/internal/meta/status.go
carpawell: yes, it is "peapod" for peapod and empty for FSTree. do you mean the same `<empty>` or smth like that?

cmd/neofs-lens/internal/meta/status.go
carpawell: i was inspired by linux: https://linux.die.net/man/2/stat, it says "get file status"

carpawell: but mb that is not clear enough, so let it be `status`

cmd/neofs-lens/internal/meta/status.go
cthulhu-rider: > `<empty>` or smth like that?

yep, similar to the code below

btw
```suggestion
	cmd.Printf("Storage's ID: %s\n", res.StorageID)
```

pkg/local_object_storage/metabase/status.go
carpawell: > this is metabase version-specific info, right?

yes, but we have never changed buckets, only extend them. fixed

carpawell:
> @carpawell can we say this relates to https://github.com/nspcc-dev/neofs-node/issues/2506?

No, it is more like a #2844's helper. But i would not mentions it.

cthulhu-rider (approved):

PR #2883, opened 2024-07-04
Check the account alongside the public key

smallhive:
Refs https://github.com/nspcc-dev/neofs-api/issues/278


pkg/services/object/acl/v2/classifier.go
```diff
@@ -36,8 +37,9 @@ func (c senderClassifier) classify(
 	// if request owner is the same as container owner, return RoleUser
 	if ownerID.Equals(cnr.Owner()) {
 		return &classifyResult{
-			role: acl.RoleOwner,
-			key:  ownerKey,
+			role:    acl.RoleOwner,
+			key:     ownerKey,
+			account: ownerID.WalletBytes(),
```
cthulhu-rider, line 42: this returns **not** script hash but 25B Neo address. We need to finally decide which format to use and sync implementations/docs

see also https://github.com/nspcc-dev/neofs-sdk-go/pull/592#discussion_r1665733136

cthulhu-rider (commented):
looks good, but lets fix the format accurately

cthulhu-rider:
* _based on and blocked by https://github.com/nspcc-dev/neofs-sdk-go/pull/592_

pkg/services/object/acl/v2/classifier.go
smallhive: Updated

cthulhu-rider: correct now

roman-khimov (approved):
Looks OK and green.

pkg/services/object/acl/acl.go
```diff
@@ -194,14 +194,19 @@ func (c *Checker) CheckEACL(msg any, reqInfo v2.RequestInfo) error {
...
 		WithRole(eaclRole).
 		WithOperation(eaclSDK.Operation(reqInfo.Operation())).
 		WithContainerID(&cnr).
 		WithSenderKey(reqInfo.SenderKey()).
 		WithHeaderSource(hdrSrc).
-		WithEACLTable(&table),
-	)
+		WithEACLTable(&table)
+
+	if sa := reqInfo.SenderAccount(); sa != nil && !sa.IsZero() {
```
cthulhu-rider, line 205: can it ever be nil? seems like we can get rid of a pointer

pkg/services/container/morph/executor.go
```diff
@@ -231,8 +231,16 @@ func (s *morphExecutor) SetExtendedACL(_ context.Context, tokV2 *sessionV2.Token
 		return nil, errors.New("missing signature")
 	}
 
+	var table eaclSDK.Table
+	eacl := body.GetEACL()
+	if eacl != nil {
+		if err := table.ReadFromV2(*body.GetEACL()); err != nil {
+			return nil, fmt.Errorf("can parse eacl body: %w", err)
```
cthulhu-rider, line 238:
```suggestion
			return nil, fmt.Errorf("invalid eACL: %w", err)
```
we `can` not, and we decode whole eACL, not just `body`

go.mod
```diff
@@ -18,7 +18,7 @@ require (
 	github.com/nspcc-dev/neo-go v0.106.3
 	github.com/nspcc-dev/neofs-api-go/v2 v2.14.1-0.20240305074711-35bc78d84dc4
 	github.com/nspcc-dev/neofs-contract v0.20.0
-	github.com/nspcc-dev/neofs-sdk-go v1.0.0-rc.12
+	github.com/nspcc-dev/neofs-sdk-go v1.0.0-rc.12.0.20240807160341-3528eb5bb1cc
```
carpawell, line 21: oh, i really dont like we are updating SDK at this moment (it is being changed a lot now) in one of the pre-release PR, but if we _need_ it to be done, ok

pkg/services/object/acl/acl.go
carpawell: missing sender is a strange thing to me, an error should appear earlier, no?

carpawell (commented):
Can you wrap some too long lines in the commit body? We have some rules about it: https://github.com/nspcc-dev/.github/blob/master/git.md
Otherwise, ok.

cthulhu-rider (approved):
nice overall

@smallhive i left several suggestions up to u

pkg/services/object/acl/acl.go
smallhive: Before this change, the account wasn't represented at all. It was only `key []byte`. According to the code, at this point, it can't be nil, but in the future - nobody knows.

Do you suggest removing the pointer?

pkg/services/container/morph/executor.go
smallhive: Replaced

carpawell (approved):

PR #2885, opened 2024-07-05
*: Convert slices to arrays instead of `copy` where possible

cthulhu-rider:


cmd/neofs-adm/internal/modules/morph/container.go
```diff
@@ -373,8 +374,7 @@ func getCIDFilterFunc(cmd *cobra.Command) (func([]byte) bool, error) {
 	}
 	sort.Strings(rawIDs)
 	return func(rawID []byte) bool {
-		var v [32]byte
-		copy(v[:], rawID)
+		v := [sha256.Size]byte(rawID)
```
carpawell, line 377: overall the same review as in SDK: should be OK in most cases but once something bad happens, users see panic and we open another PR that checks len. can we just do it here now?

cmd/neofs-node/reputation/intermediate/contract.go
```diff
@@ -80,11 +81,8 @@ func (fw FinalWriter) WriteIntermediateTrust(t eigentrust.IterationTrust) error
 	apiTrust.SetValue(t.Value().Float64())
 	apiTrust.SetPeer(t.Peer())
 
-	var managerPublicKey [33]byte
-	copy(managerPublicKey[:], fw.pubKey)
-
 	var apiMangerPeerID apireputation.PeerID
-	apiMangerPeerID.SetPublicKey(managerPublicKey[:])
+	apiMangerPeerID.SetPublicKey(bytes.Clone(fw.pubKey))
```
carpawell, line 85: does not relate commit message too?

carpawell (approved):

PR #2892, opened 2024-07-11
Cache storage policy application results (`Search`, `Replicate`)

cthulhu-rider:
code and unit tests are ready, intergration tests pass. I also want to take a memory profile in devenv

---

i created 5 containers and ran 30 routines sending 1000 search queries (each time to random container)

### `v0.42.1`
Screenshot from 2024-07-11 19-25-14

### `current branch`
Screenshot from 2024-07-11 19-35-30

---

<details>
<summary>Go script</summary>

```go
func TestDevenv(t *testing.T) {
	strCnr := []string{
		"3SDKVLon6bk6ALm8PCEjaGHUeg8cP5d6cvSJdpuMJ3qy",
		"5gGHUYnAXfB7y5XSFJb4SBr2g3NvEnkAELDAV9rxFdWL",
		"65NX4euCiJDDkduNeSc8puo7oKa3ddZC2dHxNMSS3UBA",
		"9EPJGmDDDv9iNqP1rAy7pyYuaJHRy2XSyMZt81DL35cD",
		"ArUd3NXPUhaDLNZuEKYHjbc4GS7u4nViDTKbyLiDtENW",
	}
	const endpoint = "s01.neofs.devenv:8080"
	const workers = 30
	ctx := context.Background()
	signer := test.RandomSignerRFC6979(t)

	cnrs := make([]cid.ID, len(strCnr))
	for i := range strCnr {
		require.NoError(t, cnrs[i].DecodeString(strCnr[i]))
	}

	c, err := client.New(client.PrmInit{})
	require.NoError(t, err)
	var prm client.PrmDial
	prm.SetServerURI(endpoint)
	require.NoError(t, c.Dial(prm))
	t.Cleanup(func() { c.Close() })

	var wg sync.WaitGroup
	st := time.Now()
	for i := 0; i < workers; i++ {
		wg.Add(1)
		go func() {
			defer wg.Done()

			for i := 0; i < 1000; i++ {
				if i%100 == 0 {
					fmt.Println(i)
				}
				rdr, err := c.ObjectSearchInit(ctx, cnrs[rand.Int()%len(cnrs)], signer, client.PrmObjectSearch{})
				if err != nil {
					log.Println("init search:", err)
					rdr.Close()
					continue
				}
				err = rdr.Iterate(func(id oid.ID) bool { return false })
				if err != nil {
					log.Println("read search response:", err)
					rdr.Close()
					continue
				}
				rdr.Close()
			}
		}()
	}
	wg.Wait()
	fmt.Println(time.Since(st))
}
```
</details>

## Potential
besides obvisuly required implementing the same approach to other components/RPC which is gonna TBD in next PRs, i got following thoughts:
1. the past data is less in demand than the current one, so less space can be allocated for it
2. alternatives to LRU strategy
3. dynamic volume estimation taking into account the number of requested containers and storage nodes in the network
4. storing indices instead of node info structures (https://github.com/nspcc-dev/neofs-sdk-go/issues/541)
5. invalidate cached results on container removal. Normally, they will be automatically evicted according to the LRU principle. But it’s still better to force the displacement
6. in proposal, the cache is invalidated by changing the epoch. However, the network map may not change, or the change does not affect the policies of some containers

### DIFF
--- a/cmd/neofs-node/policy.go
+++ b/cmd/neofs-node/policy.go
@@ -3,22 +3,50 @@ package main
 import (
 	"fmt"
 
+	lru "github.com/hashicorp/golang-lru/v2"
 	"github.com/nspcc-dev/neofs-node/pkg/core/container"
 	"github.com/nspcc-dev/neofs-node/pkg/core/netmap"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 	netmapsdk "github.com/nspcc-dev/neofs-sdk-go/netmap"
 )
 
+// storagePolicyRes structures persistent storage policy application result for
+// particular container and network map incl. error.
+type storagePolicyRes struct {
+	nodeSets [][]netmapsdk.NodeInfo
+	err      error
+}
+
+type containerNodesCacheKey struct {
+	epoch uint64
+	cnr   cid.ID
+}
+
+// max number of container storage policy applications results cached by
+// containerNodes.
+const cachedContainerNodesNum = 1000
+
 // containerNodes wraps NeoFS network state to apply container storage policies.
+//
+// Since policy application results are consistent for fixed container and
+// network map, they could be cached. The containerNodes caches up to
+// cachedContainerNodesNum LRU results.
 type containerNodes struct {
 	containers container.Source
 	network    netmap.Source
+
+	cache *lru.Cache[containerNodesCacheKey, storagePolicyRes]
 }
 
 func newContainerNodes(containers container.Source, network netmap.Source) (*containerNodes, error) {
+	l, err := lru.New[containerNodesCacheKey, storagePolicyRes](cachedContainerNodesNum)
+	if err != nil {
+		return nil, fmt.Errorf("create LRU container node cache for one epoch: %w", err)
+	}
 	return &containerNodes{
 		containers: containers,
 		network:    network,
+		cache:      l,
 	}, nil
 }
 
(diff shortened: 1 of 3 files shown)


cmd/neofs-node/policy.go
```diff
@@ -32,58 +119,106 @@ func (x *containerNodes) forEachContainerNodePublicKeyInLastTwoEpochs(cnrID cid.
...
+	id         cid.ID
+	containers container.Source
+	network    netmap.Source
+	// dynamic
+	cnr *container.Container
+}
+
+// applyAtEpoch applies storage policy of container referenced by parameterized
+// ID to the network map at the specified epoch. If cache is non-nil,
+// applyAtEpoch read existing results from it and stores new results in it.
```
carpawell, line 195:
```suggestion
// applyAtEpoch reads existing results from it and stores new results in it.
```

cmd/neofs-node/policy.go
```diff
@@ -32,58 +119,106 @@ func (x *containerNodes) forEachContainerNodePublicKeyInLastTwoEpochs(cnrID cid.
...
+	return nil
+}
 
-	for i := range ns {
-		for j := range ns[i] {
-			if !f(ns[i][j]) {
-				return nil
-			}
+// preserves context of storage policy processing for the particular container.
+type containerPolicyContext struct {
```
carpawell, line 184: it has only one method, do we need it? make more complex, not helps imo

cmd/neofs-node/policy.go
```diff
@@ -2,26 +2,113 @@ package main
...
+type storagePolicyRes struct {
+	nodeSets [][]netmapsdk.NodeInfo
+	err      error
+}
+
+// containerNodesAtEpoch is a thread-safe LRU cache mapping containers into
+// storage policy application results for particular network map.
+type containerNodesAtEpoch struct {
+	mtx sync.RWMutex
+	lru simplelru.LRUCache[cid.ID, storagePolicyRes]
```
carpawell, line 25: why have you decided to use the simple version of it?

carpawell:
Also, the last commit says it was tried to be done unsuccessfully before, can you, please, provide more info about it?

cmd/neofs-node/policy.go
roman-khimov: Looks strange given that we have an `mtx` here anyway dealing with concurrency.

Why not a `[cid, epoch] -> [nodes, error]`, btw?

cmd/neofs-node/policy.go
```diff
@@ -2,26 +2,113 @@ package main
...
+}
+
+// containerNodesCaches groups caches for all tracked epochs.
+type containerNodesCaches struct{ cur, prev *containerNodesAtEpoch }
```
roman-khimov, line 29: We can avoid having cur/prev here with `[cid, epoch]` key.

cmd/neofs-node/policy.go
```diff
@@ -2,26 +2,113 @@ package main
...
+// cachedContainerNodesPerEpochNum results for the latest and the previous
+// epochs. The previous one is required to support data migration after the
+// epoch tick. Older epochs are not cached.
 type containerNodes struct {
 	containers container.Source
 	network    netmap.Source
+
+	lastSeenCurrentEpochMtx sync.Mutex
+	lastSeenCurrentEpoch    uint64
+	epochCaches             containerNodesCaches
```
roman-khimov, line 48: And maybe even simplify these dances.

cmd/neofs-node/policy.go
cthulhu-rider: can drop, worth it?

cmd/neofs-node/policy.go
cthulhu-rider: does not worth it for exactly 2 epochs imo. Moreover, we are likely to decide having less capacity for previous epoch as less in demand. This approach allows such flexibility, while the total number of items will be the same

cmd/neofs-node/policy.go
cthulhu-rider: i dont find them complex. U may leave an issue for the future

cmd/neofs-node/policy.go
cthulhu-rider: > why have you decided to use the simple version of it?

to not make several calculations simultaneously. Sync cache does not prevent this

cthulhu-rider: > Why not a [cid, epoch] -> [nodes, error], btw?

to not block current epoch cache due to the acquired previous one

cmd/neofs-node/policy.go
roman-khimov: But with `[cid, epoch]` key you may not worry about magic 2-3-5-10 epochs at all, it'll all be naturally added/evicted as time goes.

cthulhu-rider: >  naturally added/evicted as time goes

i dont think this is good. We know exactly when some results become stale, and we should free them for GC at this point

roman-khimov: It's not worth the additional management code. It's a very limited amount of data and combined key simplifies this cache a lot.

cthulhu-rider: > It's not worth the additional management code

on the contrary, the code does not change much while respecting the expected behavior. App proceeds only two latest epochs, caching older results and hope they will be evicted by N new requests is incorrect to me. Previously, app didnt cache stale items. And now let it be? Nonsense

cmd/neofs-node/policy.go
cthulhu-rider: fixed

cthulhu-rider:
> Also, the last commit says it was tried to be done unsuccessfully before, can you, please, provide more info about it (to the commit)?

done

cmd/neofs-node/policy.go
cthulhu-rider: redone

cmd/neofs-node/policy.go
cthulhu-rider: redone

cmd/neofs-node/policy.go
```diff
@@ -3,22 +3,56 @@ package main
...
+}
+
+type containerNodesCacheKey struct {
+	epoch uint64
+	cnr   cid.ID
+}
+
+// containerNodesCache is a thread-safe LRU cache mapping (container, epoch)
+// pairs into corresponding storage policy application results.
+type containerNodesCache struct {
```
roman-khimov, line 27: Can be dropped.

roman-khimov (approved):

cmd/neofs-node/policy.go
cthulhu-rider: done

carpawell (approved):

PR #2896, opened 2024-07-17
Cache storage policy application results (`Get`/`Head`/`GetRange`)

cthulhu-rider:
* continues #2892

---

i created 2 containers with 5 object in each and ran 30 routines sending 1000 GET queries (each time random object)

### `v0.42.1`
Screenshot from 2024-07-18 16-27-59

### `current branch`
Screenshot from 2024-07-18 16-35-55

---

<details>
<summary>Go script</summary>

```go
strObjs := []string{
		"47pRu4GKKHL3yJ5wkXRDZj5QsyfH3wUUy27x9eEp7wgK/7GRiZmkxHruwrR5YbwfiTCMzo5rnEHr8HmyNAPwAEzex",
		"47pRu4GKKHL3yJ5wkXRDZj5QsyfH3wUUy27x9eEp7wgK/3ZSUCYqacwGY3teqJm3f9gDYSsEtB4uccXUp5in6jXBc",
		"47pRu4GKKHL3yJ5wkXRDZj5QsyfH3wUUy27x9eEp7wgK/EMdsnvaUK2XQtxLf1zFjVeFFtbszEwKu8mvYt9W15oMw",
		"47pRu4GKKHL3yJ5wkXRDZj5QsyfH3wUUy27x9eEp7wgK/En1K6HVfzoQtX9Vn2nW85szWduXqepdbmeAt9t2NMQJL",
		"47pRu4GKKHL3yJ5wkXRDZj5QsyfH3wUUy27x9eEp7wgK/B7UjiPD8UcCQsWHpDLZWswATnA7q4bWvbHAvJJXCtbPq",
		"2GN486btkTpxsYHsVjz5tZzug65YiSFUAYqAQiEkg4sg/FtRXxUBiTteyCM5QyeKKwFsJiJsNQGDMG31dTchagep3",
		"2GN486btkTpxsYHsVjz5tZzug65YiSFUAYqAQiEkg4sg/FJijkw5XauTY9N2TKUSTf8QxJoDjBFGVS9Qc9V6ejW1i",
		"2GN486btkTpxsYHsVjz5tZzug65YiSFUAYqAQiEkg4sg/HhbtK3cTcwu8uzLaMjMFg9Cnj1ZkmPAvPGPAXBNoYM6E",
		"2GN486btkTpxsYHsVjz5tZzug65YiSFUAYqAQiEkg4sg/DcLuSr69JEEgeqTE3AaMAQqcVzPDHxtK88ZXdDVycRfM",
		"2GN486btkTpxsYHsVjz5tZzug65YiSFUAYqAQiEkg4sg/Afykr7S6kgwbkVA5fsmw6bjkv8LrUQ7hxzLWreneoERh",
	}
	const endpoint = "s01.neofs.devenv:8080"
	const workers = 30
	ctx := context.Background()
	signer := test.RandomSignerRFC6979(t)

	objs := make([]oid.Address, len(strObjs))
	for i := range strObjs {
		require.NoError(t, objs[i].DecodeString(strObjs[i]))
	}

	c, err := client.New(client.PrmInit{})
	require.NoError(t, err)
	var prm client.PrmDial
	prm.SetServerURI(endpoint)
	require.NoError(t, c.Dial(prm))
	t.Cleanup(func() { c.Close() })

	var wg sync.WaitGroup
	st := time.Now()
	for i := 0; i < workers; i++ {
		wg.Add(1)
		go func() {
			defer wg.Done()

			for i := 0; i < 1000; i++ {
				if i%100 == 0 {
					fmt.Println(i)
				}
				a := objs[rand.Int()%len(objs)]
				_, rdr, err := c.ObjectGetInit(ctx, a.Container(), a.Object(), signer, client.PrmObjectGet{})
				if err != nil {
					log.Println("init search:", err)
					continue
				}
				rdr.Close()
			}
		}()
	}
	wg.Wait()
	fmt.Println(time.Since(st))
```
</details>

### DIFF
--- a/cmd/neofs-node/policy.go
+++ b/cmd/neofs-node/policy.go
@@ -8,6 +8,7 @@ import (
 	"github.com/nspcc-dev/neofs-node/pkg/core/netmap"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 	netmapsdk "github.com/nspcc-dev/neofs-sdk-go/netmap"
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 )
 
 // storagePolicyRes structures persistent storage policy application result for
@@ -151,4 +152,48 @@ func (x *containerPolicyContext) applyAtEpoch(epoch uint64, cache *lru.Cache[con
 	result.nodeSets, result.err = networkMap.ContainerNodes(x.cnr.Value.PlacementPolicy(), x.id)
 	cache.Add(cacheKey, result)
 	return result, nil
+}
+
+// getNodesForObject reads storage policy of the referenced container from the
+// underlying container storage, reads network map at the specified epoch from
+// the underlying storage, applies the storage policy to it and returns sorted
+// lists of selected storage nodes along with the per-list numbers of primary
+// object holders. Resulting slices must not be changed.
+func (x *containerNodes) getNodesForObject(addr oid.Address) ([][]netmapsdk.NodeInfo, []uint, error) {
+	epoch, err := x.network.Epoch()
+	if err != nil {
+		return nil, nil, fmt.Errorf("read current NeoFS epoch: %w", err)
+	}
+	cnrID := addr.Container()
+	cnr, err := x.containers.Get(cnrID)
+	if err != nil {
+		return nil, nil, fmt.Errorf("read container by ID: %w", err)
+	}
+	networkMap, err := x.network.GetNetMapByEpoch(epoch)
+	if err != nil {
+		return nil, nil, fmt.Errorf("read network map at epoch #%d: %w", epoch, err)
+	}
+
+	policy := cnr.Value.PlacementPolicy()
+	nodeLists, err := networkMap.ContainerNodes(policy, cnrID)
+	if err != nil {
+		return nil, nil, fmt.Errorf("apply container's storage policy to the network map at epoch #%d: %w", epoch, err)
+	}
+	if nodeLists, err = networkMap.PlacementVectors(nodeLists, addr.Object()); err != nil {
+		return nil, nil, fmt.Errorf("sort container nodes from the network map at epoch #%d: %w", epoch, err)
+	}
+	if len(nodeLists) != policy.NumberOfReplicas() {
+		return nil, nil, fmt.Errorf("invalid result of container's storage policy application to the network map at epoch #%d: "+
+			"diff number of storage node lists (%d) and required replica descriptors (%d)", epoch, len(nodeLists), policy.NumberOfReplicas())
+	}
+
+	primaryCounts := make([]uint, len(nodeLists))
+	for i := range nodeLists {
+		if primaryCounts[i] = uint(policy.ReplicaNumberByIndex(i)); primaryCounts[i] > uint(len(nodeLists[i])) {
+			return nil, nil, fmt.Errorf("invalid result of container's storage policy application to the network map at epoch #%d: "+
+				"invalid storage node list #%d: number of nodes (%d) is less than minimum required by the container policy (%d)",
+				epoch, i, len(nodeLists), policy.NumberOfReplicas())
+		}
+	}
+	return nodeLists, primaryCounts, nil
 }
--- a/cmd/neofs-node/policy.go
+++ b/cmd/neofs-node/policy.go
@@ -39,6 +39,11 @@ const (
 	cachedObjectNodesNum = 10000
 )
 
+type (
+	getContainerNodesFunc  = func(netmapsdk.NetMap, netmapsdk.PlacementPolicy, cid.ID) ([][]netmapsdk.NodeInfo, error)
+	sortContainerNodesFunc = func(netmapsdk.NetMap, [][]netmapsdk.NodeInfo, oid.ID) ([][]netmapsdk.NodeInfo, error)
+)
+
 // containerNodes wraps NeoFS network state to apply container storage policies.
 //
 // Since policy application results are consistent for fixed container and
@@ -50,6 +55,10 @@ type containerNodes struct {
 
 	cache    *lru.Cache[containerNodesCacheKey, storagePolicyRes]
 	objCache *lru.Cache[objectNodesCacheKey, storagePolicyRes]
+
+	// for testing
+	getContainerNodesFunc  getContainerNodesFunc
+	sortContainerNodesFunc sortContainerNodesFunc
 }
 
 func newContainerNodes(containers container.Source, network netmap.Source) (*containerNodes, error) {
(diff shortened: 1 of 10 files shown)


cmd/neofs-node/policy.go
```diff
@@ -152,3 +153,45 @@ func (x *containerPolicyContext) applyAtEpoch(epoch uint64, cache *lru.Cache[con
...
+	networkMap, err := x.network.GetNetMapByEpoch(epoch)
+	if err != nil {
+		return nil, nil, fmt.Errorf("read network map at epoch #%d: %w", epoch, err)
+	}
+
+	policy := cnr.Value.PlacementPolicy()
+	nodeLists, err := networkMap.ContainerNodes(policy, cnrID)
+	if err != nil {
+		return nil, nil, fmt.Errorf("apply container's storage policy to the network map at epoch #%d: %w", epoch, err)
+	} else if nodeLists, err = networkMap.PlacementVectors(nodeLists, addr.Object()); err != nil {
```
carpawell: why `else if`? just a regular algorithm, step by step, stop on any error. additional nesting it not required

pkg/services/object/get/container.go
```diff
@@ -15,71 +17,84 @@ func (exec *execCtx) executeOnContainer() {
...
-		addrs := traverser.Next()
-		if len(addrs) == 0 {
-			exec.log.Debug("no more nodes, abort placement iteration")
-
-			return false
+	var endpoints, externalEndpoints network.AddressGroup
+	var j, jLim uint
+	primary := true
+
+	for i := 0; i < len(nodeLists); i++ { // do not use for-range!
```
carpawell, line 37: why? the comment does not explain, so it is useless (or even harmful)

pkg/services/object/get/service.go
```diff
@@ -2,19 +2,39 @@ package getsvc
...
-	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
+	netmapsdk "github.com/nspcc-dev/neofs-sdk-go/netmap"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"go.uber.org/zap"
 )
 
+// NeoFS provides access to the NeoFS network to get information necessary for
+// the [Service] to work.
+type NeoFS interface {
```
carpawell, line 15: i was not agree last time about such naming so i have to write it one more time: i do not understand this interface, everything in this repo is about NeoFS. i would have understood if it was about some NeoFS constants or the things that are fixed and cannot be explained in NeoFS, but the methods here are about Network Map and that is it.
it really confuses me when i see this:
```go
type Service struct {
	*cfg
	neoFS NeoFS
}
```

cmd/neofs-node/policy.go
```diff
@@ -36,18 +48,24 @@ type containerNodes struct {
 	containers container.Source
 	network    netmap.Source
 
-	cache *lru.Cache[containerNodesCacheKey, storagePolicyRes]
+	cache    *lru.Cache[containerNodesCacheKey, storagePolicyRes]
```
carpawell, line 56: rename too?

cmd/neofs-node/policy.go
```diff
@@ -152,3 +153,45 @@ func (x *containerPolicyContext) applyAtEpoch(epoch uint64, cache *lru.Cache[con
...
+// object holders. Resulting slices must not be changed.
+func (x *containerNodes) getNodesForObject(addr oid.Address) ([][]netmapsdk.NodeInfo, []uint, error) {
+	epoch, err := x.network.Epoch()
+	if err != nil {
+		return nil, nil, fmt.Errorf("read current NeoFS epoch: %w", err)
+	}
+	cnrID := addr.Container()
+	cnr, err := x.containers.Get(cnrID)
+	if err != nil {
+		return nil, nil, fmt.Errorf("read container by ID: %w", err)
```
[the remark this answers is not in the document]
cthulhu-rider: cid is the input part, no need to duplicate it in the err

cmd/neofs-node/policy.go
cthulhu-rider: idk what to say, just a code style

pkg/services/object/get/container.go
cthulhu-rider: usually range is used for this, but here it wouldn't work

pkg/services/object/get/service.go
cthulhu-rider: not a problem, can be renamed to any. `NeoFSNetwork` / `NeoFSChain`?

cmd/neofs-node/policy.go
cthulhu-rider: no need, it contains container nodes

cmd/neofs-node/policy.go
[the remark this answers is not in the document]
carpawell: addr is in an input. but ok

cmd/neofs-node/policy.go
carpawell: we have already discussed it, leaving the same link is the only thing i can do here: https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html

pkg/services/object/get/container.go
carpawell: only after reading it twice (to the end), i saw `i = -1`, why not mention it? it is a comment, it should help. any code can be changed, so comments should explain something, and if they are outdated, it should be easy to understand. not saying it is required here but it is a general rule i would say

pkg/services/object/get/service.go
carpawell: `NeoFSNetwork` is ok to me

carpawell:
Some conflicts.

pkg/services/object/get/container.go
roman-khimov: Tricky one. Can be simplified by inverting i/j loops: walk over the top ones from each list first and then go deeper till there are no more nodes to check. It can be more optimal in some cases (top nodes from different lists are still more likely to have the object), but maybe not for all of them (policies may have local nodes first).

The other way of course is just `for _, primary := range []bool{true, false}` and a regular range loop over `nodeLists` then. Less tricks, same meaning.

pkg/services/object/get/service.go
roman-khimov: `Netmapper`.

pkg/services/object/get/container.go
cthulhu-rider: > top nodes from different lists are still more likely to have the object

they are homogeneous in terms of policies

> `for _, primary := range []bool{true, false}`

i can make any style you think is readable. I'm fine with current. It can be simply divided into two loops. No tricks at all

carpawell:
What's the status here? Some threads are not fully answered and no review request yet.

cthulhu-rider:
> What's the status here?

i've already processed all of them, pls ref me to the exact ones

cmd/neofs-node/policy.go
carpawell: this one is still arguable

cthulhu-rider: elif no more

carpawell: oh, i see. but no GH's `compare` says me about it sadly

carpawell (approved):

PR #2900, opened 2024-07-26
Upd/contracts dep

carpawell:


roman-khimov:
Have you seen #2872?

carpawell:
@roman-khimov, no, I usually skip drafts without explicit review requests. Why is it a draft? See no explanation why no contract release is not done yet but at the same time we still have to wait so long before the deploy is finished in our testings so we need it asap in master (and further release).

carpawell:
Ok after #2900 update it is not needed.

PR #2901, opened 2024-07-29
object/put: Refactor storage policy processing

cthulhu-rider:
* based on and blocked by #2896


cmd/neofs-node/object.go
```diff
@@ -256,14 +256,13 @@ func initObjectService(c *cfg) {
 		searchsvcV2.WithKeyStorage(keyStorage),
 	)
 
-	sPut := putsvc.NewService(&transport{clients: putConstructor},
+	sPut := putsvc.NewService(&transport{clients: putConstructor}, c,
```
carpawell, line 259: btw, why not an option? looks strange when there are position args and options but _both_ are required in fact

cmd/neofs-node/object.go
```diff
@@ -739,3 +738,52 @@ func (c *cfg) IsLocalNodePublicKey(b []byte) bool { return c.IsLocalKey(b) }
...
 	return c.cfgObject.containerNodes.getNodesForObject(addr)
 }
+
+// GetContainerNodes reads storage policy of the referenced container from the
+// underlying container storage, reads current network map from the underlying
+// storage, applies the storage policy to it, gathers storage nodes matching the
+// policy and returns sort interface.
+//
+// GetContainerNodes implements [putsvc.NeoFSNetwork].
+func (c *cfg) GetContainerNodes(cnrID cid.ID) (putsvc.ContainerNodes, error) {
```
carpawell, line 748: why isn't it a part of `containerNodes`? our `cfg` becomes too complex. this method only uses `cfg`'s `c.cfgObject.containerNodes`

pkg/services/object/put/distributed.go
```diff
@@ -22,15 +24,11 @@ type preparedObjectTarget interface {
 }
 
 type distributedTarget struct {
-	traversalState traversal
-	traverser      *placement.Traverser
-
-	remotePool, localPool util.WorkerPool
+	placementIterator
```
carpawell, line 27: why embedded?

pkg/services/object/put/distributed.go
```diff
@@ -211,95 +145,247 @@ func (t *distributedTarget) sendObject(node nodeDesc) error {
 	return nil
 }
 
-func (t *distributedTarget) iteratePlacement(f func(nodeDesc) error) (oid.ID, error) {
-	id, _ := t.obj.ID()
-	var resErr atomic.Value
+type errNotEnoughNodes struct {
```
carpawell, line 148: why it exists?

pkg/services/object/put/distributed.go
```diff
@@ -211,95 +145,247 @@ func (t *distributedTarget) sendObject(node nodeDesc) error {
...
+	linearReplNum uint
+	// whether to perform additional best-effort of sending the object replica to
+	// all reserve nodes of the container
+	broadcast bool
+}
 
-		for i := range addrs {
-			if t.traversalState.processed(addrs[i]) {
-				// it can happen only during additional container broadcast
+func (x placementIterator) iterateNodesForObject(obj oid.ID, f func(nodeDesc) error) error {
```
carpawell, line 176: can you, please, explain why this refactor was required (in commit message too)? does it solve something? the previous version was split into subfunctions and nesting was not that deep and the whole `iteratePlacement` was shorter. overall, i would say the previous approaches were closer to the best practices

cmd/neofs-node/object.go
cthulhu-rider: to not forget to pass it. I agree that in the end this should be brought to a general form

pkg/services/object/put/distributed.go
cthulhu-rider: not required, just allows to directly access its fields from `distributedTarget`

pkg/services/object/put/distributed.go
cthulhu-rider: done

carpawell:
https://rest.fs.neo.org/HXSaMJXk2g8C14ht8HSi7BBaiYZ1HeWh2xnWPGQCg4H6/1842-1724261309/index.html

Something is not good with this PR.

roman-khimov (approved):

carpawell (approved):

PR #2903, opened 2024-07-29
node/container: check session tokens on SN side

### LINKED ISSUE #2898: Validate session token on the node
Using session tokens with bucket operations like create/delete bucket is an async story. In the SDK we use Waiter to get the successful result of such operations.

Despite this, if the session token is expired, the waiter will wait until timeout, because the request was correct, and correctly signed.

It would be perfect if the node were requested immediately if the session token expired, was wrong by any parameter, etc.
Or even in any other requests, when node can check something

## Additional context
Receiving any error on any request to the node can make the developer's life easier

carpawell:
Closes #2898.


roman-khimov (approved):

pkg/services/container/morph/executor.go
```diff
@@ -337,6 +341,22 @@ func (s *morphExecutor) validateToken(t *sessionV2.Token, cIDV2 *refs.ContainerI
 		return fmt.Errorf("incorrect token signature: %w", err)
 	}
 
+	if lt := t.GetBody().GetLifetime(); lt != nil {
```
cthulhu-rider, line 344: why u decided to make lifetime optional? IR denies such tokens

cthulhu-rider (commented):
good overall

pkg/services/container/morph/executor.go
```diff
@@ -337,6 +341,22 @@ func (s *morphExecutor) validateToken(t *sessionV2.Token, cIDV2 *refs.ContainerI
...
+		currEpoch := s.epocher.CurrentEpoch()
+		exp := t.GetBody().GetLifetime().GetExp()
```
cthulhu-rider, line 346:
```suggestion
		exp := lt.GetExp()
```
for all

pkg/services/container/morph/executor.go
carpawell: tbh, it was not easy to check:
1. API does not claim it as a required field: https://github.com/nspcc-dev/neofs-api/blob/4bd8b4da96c810e69a9a1336adb5e4e84bac62eb/session/types.proto#L98-L110
2. IR unmarshals it (and it can be done without any errors for a token without lifetime) and only when i looked at [`InvalidAt`](https://github.com/nspcc-dev/neofs-sdk-go/blob/fced97d5dfae54de0b5b2cfa60edecfa217509ac/session/common.go#L274-L285) i found that it should be set (so it is not required now in fact, it is: unset lifetime == expired lifetime)

fixed, but not sure why it was so difficult to prove (and not sure what error should be)

cthulhu-rider: > API does not claim it as a required field

u goddamn right! This brings us to the next topic

lets consider corner case not really expected in practice:
* current epoch is 0
* 1st request has no embedded lifetime field
* 2nd one has it with all zeros

according to lifetime claim docs and protobuf, both tokens are valid. Then why we require the field? I think we should start treating the field as optional (fix SDK code) and simply work with values as they are (defaulting to 0). What do u think?

carpawell: > What do u think?

i think in any case we should add some words to the API (optional/required). about optionality: i think it is ok to have a token without a lifetime, though it can be considered a security issue

@roman-khimov

roman-khimov: We better have lifetime mandatory.

cthulhu-rider (approved):

PR #2911, opened 2024-08-08
Fix replication service denial after `MaxObjectSize` increase

cthulhu-rider:
overall, solution is very simple: listen to config changes and update the server limit. But...

## gRPC 
lib does not provide option to update max recv msg size on the runnig server. https://pkg.go.dev/google.golang.org/grpc#MaxRecvMsgSize is static. It can be very easily supported, so i [patched](https://github.com/cthulhu-rider/grpc-go/commit/00d000d30657f9fbff8b73fe5904e445b5d4ab81) my fork. If we decide this proposal is good, i suggest to add an organization fork (and later propose it to the upstream)

there is an alternative approach i dont like completely: 
1. gracefully stop the running server
2. create new one with updated limit and run it
although `MaxObjectSize` is expected to be changed very rarely, such method looks very ugly taking into account the native option's simplicity

## Notifications
we dont have them now (https://github.com/nspcc-dev/neofs-contract/issues/427). Polling is used as a temp soluition

---

this brings us closer to fix the [test](https://rest.fs.neo.org/HXSaMJXk2g8C14ht8HSi7BBaiYZ1HeWh2xnWPGQCg4H6/1795-1723053724/index.html#suites/44e11ced39071e8d0cfc11c5b94622ba/610abeaabb96d590/) within which the bug was originally detected. It should be noted that there is not 100% stable solution for it: there will always be a gap b/w contract update and node reaction. The only option i see for now is to ignore overflow errors for about 1-2 minutes and retry
cc @evgeniiz321

### DIFF
--- a/cmd/neofs-node/cache.go
+++ b/cmd/neofs-node/cache.go
@@ -446,12 +446,28 @@ type ttlMaxObjectSizeCache struct {
 	lastUpdated time.Time
 	lastSize    uint64
 	src         putsvc.MaxSizeSource
+	onChange    func(uint64)
 }
 
-func newCachedMaxObjectSizeSource(src putsvc.MaxSizeSource) putsvc.MaxSizeSource {
+func newCachedMaxObjectSizeSource(src putsvc.MaxSizeSource, onChange func(uint64)) *ttlMaxObjectSizeCache {
 	return &ttlMaxObjectSizeCache{
-		src: src,
+		src:      src,
+		onChange: onChange,
 	}
+}
+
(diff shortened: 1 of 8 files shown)


roman-khimov:
> If we decide this proposal is good

It's OK to me (a new feature there, solves a problem, not very invasive). But I wouldn't like to use any forks of the upstream library, so server restart is perfectly fine to me as well as a temporary solution (until your patch is accepted). This is a very rare event.

We can consider some handling on the test side as well. Node restart will solve this for sure.

roman-khimov:
Please upstream the library improvement.

PR #2912, opened 2024-08-09
Feat/control service object status

carpawell:


pkg/services/control/server/status.go
```diff
@@ -0,0 +1,95 @@
...
+	for _, sh := range st.Shards {
+		respSh := new(control.ObjectStatusResponse_Body_Shard)
+		respSh.ShardId = sh.ID
+
+		if len(sh.Shard.Metabase.State) == 0 {
+			// can be reconsidered since it is possible to get
+			// resynchronized state when meta knows nothing about
+			// stored objects in blob; however, it is a control
+			// service, not a debug util
+			continue
```
cthulhu-rider, line 50: IIUC ur comment: if it is in write cache and/or blob storage, some info should be returned anyway. We can fallback this case to `MISSING` status for metabase

pkg/services/control/server/status.go
```diff
@@ -0,0 +1,95 @@
...
+			)
+		}
+
+		var wcStatus string
+		switch {
+		case sh.Shard.Writecache.PathDB != "":
+			wcStatus = fmt.Sprintf("database path: %q", sh.Shard.Writecache.PathDB)
+		case sh.Shard.Writecache.PathFSTree != "":
+			wcStatus = fmt.Sprintf("fsTree path: %q", sh.Shard.Writecache.PathFSTree)
+		default:
```
cthulhu-rider, line 73: not needed

pkg/services/control/server/status.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	// check availability
+	err = s.ready()
+	if err != nil {
+		return nil, err
+	}
+
+	var addr oid.Address
+	err = addr.DecodeString(string(request.Body.ObjectAddress))
```
cthulhu-rider, line 27: can `request.Body` be nil?

pkg/services/control/service.proto
```diff
@@ -345,3 +348,49 @@ message FlushCacheResponse {
     Body body = 1;
     Signature signature = 2;
 }
+
+// ObjectStatus request.
+message ObjectStatusRequest {
+    // Request body structure.
+    message Body {
+        // Requested object.
+        bytes object_address = 1;
```
cthulhu-rider, line 357:
```suggestion
        string object_address = 1;
```

pkg/services/control/server/status.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return nil, status.Error(codes.Internal, fmt.Sprintf("parsing object address: %s", err.Error()))
```
cthulhu-rider, line 29: `codes.InvalidArgument` fits better here

cthulhu-rider, line 29:
```suggestion
		return nil, status.Errorf(codes.Internal, "parsing object address: %v", err)
```
same below

cmd/neofs-cli/modules/control/objects.go
```diff
@@ -0,0 +1,80 @@
...
+		resp, err = control.ObjectStatus(client, req)
+		return err
+	})
+	if err != nil {
+		return fmt.Errorf("rpc error: %w", err)
+	}
+
+	verifyResponse(cmd, resp.GetSignature(), resp.GetBody())
+
+	for _, shard := range resp.Body.Shards {
```
cthulhu-rider, line 65: we'll see nothing if `resp.Body.Shads` is empty. I'd write smth, e.g. `No shards`

same for `shard.Storages` below

cthulhu-rider (commented):
nice overall

pkg/services/control/server/status.go
carpawell: ok, but still %s if you dont mind

pkg/services/control/service.proto
carpawell: hm, we have [this](https://github.com/nspcc-dev/neofs-node/blob/73e84148f8564c5ae71e8b537b6cf0014a96dd47/pkg/services/control/server/gc.go#L32-L42) and [this](https://github.com/nspcc-dev/neofs-node/blob/73e84148f8564c5ae71e8b537b6cf0014a96dd47/pkg/services/control/service.proto#L114-L116), and it confuses me (but i copied it somehow). but agree that in a new command we should use `string`, fixed

pkg/services/control/server/status.go
carpawell: hm, it can, changed to getters

pkg/services/control/server/status.go
carpawell: i prefer explicit "we do nothing" usually, don't mind dropping

pkg/services/control/server/status.go
carpawell: > if it is in write cache and/or blob storage, some info should be returned anyway

that is why i left comment: i do not think we should search for a object if a metabase in this shard does not know it. this is not `neofs-lens`, it is a control service

pkg/services/control/server/status.go
cthulhu-rider: `%s` for `error` is `err.Error()`, no need to call the method

pkg/services/control/server/status.go
cthulhu-rider: > // can be reconsidered

what do u mean then?

pkg/services/control/server/status.go
carpawell: yes, that is why i changed it to `status.Errorf`, dropped `.Error` call but kept `%s` (not `%v`), that is why my answer

pkg/services/control/server/object_status.go
```diff
@@ -0,0 +1,94 @@
...
+
+	var addr oid.Address
+	err = addr.DecodeString(request.GetBody().GetObjectAddress())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "parsing object address: %s", err)
+	}
+
+	st, err := s.storage.ObjectStatus(addr)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "storage engine error: %s", err.Error())
```
cthulhu-rider, line 34:
```suggestion
		return nil, status.Errorf(codes.Internal, "storage engine error: %s")
```

pkg/services/control/server/status.go
cthulhu-rider: oh i see now, u just did it diff in two cases

pkg/services/control/server/object_status.go
carpawell: missed that place, fixed

pkg/services/control/server/status.go
carpawell: i meant somebody can disagree and we may change it in the future. if too many devs disagree now, may change it now

cthulhu-rider: let bs/wc aint in sync with mb - what output will be currently shown?

cthulhu-rider (approved):

cmd/neofs-cli/modules/control/objects.go
```diff
@@ -0,0 +1,92 @@
...
+	})
+	if err != nil {
+		return fmt.Errorf("rpc error: %w", err)
+	}
+
+	verifyResponse(cmd, resp.GetSignature(), resp.GetBody())
+
+	shards := resp.GetBody().GetShards()
+	if len(shards) == 0 {
+		cmd.Println("<emtpy response>")
```
roman-khimov, line 67: empty

cmd/neofs-cli/modules/control/objects.go
```diff
@@ -0,0 +1,92 @@
...
+		return nil
+	}
+
+	for _, shard := range shards {
+		cmd.Printf("Shard ID: %s\n", shard.ShardId)
+		storages := shard.GetStorages()
+		if len(storages) == 0 {
+			cmd.Println("\t<emtpy response>")
```
roman-khimov, line 75: Ditto.

pkg/services/control/server/status.go
carpawell: > let bs/wc aint in sync with mb

if bs/wc _has_ info, but metabase does not know about an object, it will be `<empty response>` currently

> And what GET would respond

it really depends on the degree of resynchronization, it meta know absolutely nothing about an object, than it is 404: https://github.com/nspcc-dev/neofs-node/blob/73e84148f8564c5ae71e8b537b6cf0014a96dd47/pkg/local_object_storage/shard/get.go#L146-L148

roman-khimov (approved):

PR #2913, opened 2024-08-09
node/acl: fix acl checks for non-container nodes

### LINKED ISSUE #2909: Impossible to put big object via non-container node without HEAD allowed
After split V2 scheme adoption, every big object PUT requires to HEAD the first part. If it is denied by container policy, the whole PUT fails. However, this operation does not look like smth strange and the system should allow this behaviour.

## Expected Behavior
PUT big object through a non-container node to a container with HEAD denied to others should work OK.

## Current Behavior
Returns 2048 error (access denied).

## Possible Solution
Do not check ACL objects if a node is not expected to store an object (does not belong to a container). Even for session cases it splits and signs an object and tries to PUT it to a container node. If it does not pass ACL checks, then it is what it is, return error it gets.
Also, re-checks, that if an object is signed (completed, PUT-without-session case), it is just forwarded to a container node; this code does not work usually (or maybe not used at all since it was written).

Run https://github.com/nspcc-dev/neofs-s3-gw/issues/969 test case, see errors.

## Context
https://github.com/nspcc-dev/neofs-s3-gw/issues/969 PUTs objects via SDK's Pool. It uses every node it has, however only 3/4 nodes belong to the test container, so every time part it PUT to a non-container node, the test fails.

## Regression
Yes, split V2.

carpawell:
Split objects are validated and it required to do some additional requests for it. Such operation can be restricted for a non-container node, so checking for it is an unsolvable problem, skip ACL checks in such cases. Closes #2909.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Changelog for NeoFS Node
 - Control service's Drop call does not clean metabase (#2822)
 - It was impossible to specify memory amount as "1b" (one byte) in config, default was used instead (#2899)
 - Container session token's lifetime was not checked (#2898)
+- ACL checks for split objects could be forced by a node than might lack access (#2909) 
 
 ### Changed
 - neofs-cli allows several objects deletion at a time (#2774)
--- a/cmd/neofs-node/object.go
+++ b/cmd/neofs-node/object.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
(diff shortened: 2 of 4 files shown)


carpawell:
Checked with the initial [testcase](https://github.com/nspcc-dev/neofs-s3-gw/issues/969) for this issue: now it fails with a different problem (: But objects are PUT and it does not seem to be this issue anymore.

roman-khimov (approved):

PR #2915, opened 2024-08-13
Set up account alongside pub key in eacl rules in CLI

### LINKED ISSUE #2914: Support account-based EACLs in CLI
I'm always frustrated when we can't make something with NeoFS CLI. We can have account-based EACLs after #2883, but we can't manage them via CLI which blocks #823.

Where key can be used accounts should be allowed too.

smallhive:
Closes #2914.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Changelog for NeoFS Node
 - Support for 0.20.0+ neofs-contract archive format (#2872)
 - `neofs-cli control object status` command (#2886)
 - Check the account alongside the public key in ACL (#2883)
+- Allow addresses to be used in EACLs created from CLI (#2914)
 
 ### Fixed
 - Control service's Drop call does not clean metabase (#2822)
--- a/cmd/neofs-cli/modules/acl/extended/create.go
+++ b/cmd/neofs-cli/modules/acl/extended/create.go
@@ -41,7 +41,8 @@ Target is
   'user' for container owner, 
   'system' for Storage nodes in container and Inner Ring nodes,
   'others' for all other request senders, 
-  'pubkey:<key1>,<key2>,...' for exact request sender, where <key> is a hex-encoded 33-byte public key.
+  'pubkey:<key1>,<key2>,...' for exact request sender, where <key> is a hex-encoded 33-byte public key,
+  'address:<adr1>,<adr2>,...' for exact request sender, where <adr> is a base58 25-byte address. Example: NSiVJYZej4XsxG5CUpdwn7VRQk8iiiDMPM.
 
 When both '--rule' and '--file' arguments are used, '--rule' records will be placed higher in resulting extended ACL table.
 `,
--- a/cmd/neofs-cli/modules/util/acl.go
+++ b/cmd/neofs-cli/modules/util/acl.go
@@ -14,6 +14,7 @@ import (
 	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
 	"github.com/nspcc-dev/neofs-sdk-go/container/acl"
 	"github.com/nspcc-dev/neofs-sdk-go/eacl"
+	"github.com/nspcc-dev/neofs-sdk-go/user"
 	"github.com/olekukonko/tablewriter"
 	"github.com/spf13/cobra"
 )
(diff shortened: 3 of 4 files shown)


cmd/neofs-cli/modules/acl/extended/create.go
```diff
@@ -41,7 +41,8 @@ Target is
   'user' for container owner, 
   'system' for Storage nodes in container and Inner Ring nodes,
   'others' for all other request senders, 
-  'pubkey:<key1>,<key2>,...' for exact request sender, where <key> is a hex-encoded 33-byte public key.
+  'pubkey:<key1>,<key2>,...' for exact request sender, where <key> is a hex-encoded 33-byte public key,
+  'account:<acc1>,<acc2>,...' for exact request sender, where <acc> is a base58 25-byte account.
```
carpawell, line 45: i think we need to note that it is our account somehow (mb reference to user ID or say that is it neo-go's string address)
@roman-khimov

cmd/neofs-cli/modules/util/acl.go
```diff
@@ -276,7 +277,23 @@ func parseEACLRecord(args []string) (*eacl.Record, error) {
 			}
 
 			eacl.AddFormedTarget(r, role, pubs...)
+		case "account": // targets
+			var (
+				err      error
+				accounts []user.ID
+			)
 
+			if len(ss) == 2 {
```
carpawell, line 286: what if len is not 2? the whole `case` is senseless then (in contrast, the previous may miss the second arg)

cmd/neofs-cli/modules/acl/extended/create.go
roman-khimov: Stringified they're the same, we can in fact just say that it's an address.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Changelog for NeoFS Node
 - Support for 0.20.0+ neofs-contract archive format (#2872)
 - `neofs-cli control object status` command (#2886)
 - Check the account alongside the public key in ACL (#2883)
+- Set up the account alongside the public key in CLI eacl rules (#2914)
```
roman-khimov, line 12: Allow addresses to be used in EACLs created from CLI

cmd/neofs-cli/modules/acl/extended/create.go
roman-khimov: And it'd be preferable, `NSiVJYZej4XsxG5CUpdwn7VRQk8iiiDMPM` is widely known as _address_.

cmd/neofs-cli/modules/util/acl.go
```diff
@@ -276,7 +277,23 @@ func parseEACLRecord(args []string) (*eacl.Record, error) {
 			}
 
 			eacl.AddFormedTarget(r, role, pubs...)
+		case "account": // targets
```
roman-khimov, line 280: address

CHANGELOG.md
smallhive: Replaced

cmd/neofs-cli/modules/acl/extended/create.go
smallhive: Replaced account with address. Added example

cmd/neofs-cli/modules/util/acl.go
smallhive: Done

cmd/neofs-cli/modules/util/acl.go
smallhive: ss can be length 1 (if just `deny getrange address` without address and `:`). It is an invalid record. Changed the condition for error, in this case.

Otherwise ss have length 2, because `strings.SplitN(args[i], ":", 2)`
Even we have a record like `address:NWcSSzMkw5Vuq3gFmcih5yfCNxfXjWdKz8:NWcSSzMkw5Vuq3gFmcih5yfCNxfXjWdKz8` with two `:`. The result length is still 2:
* address
* NWcSSzMkw5Vuq3gFmcih5yfCNxfXjWdKz8:NWcSSzMkw5Vuq3gFmcih5yfCNxfXjWdKz8

This is just invalid format

roman-khimov: But it needs to be handled.

roman-khimov (approved):

carpawell (approved):

PR #2919, opened 2024-08-19
.github: add missing xk6 binaries to system tests

evgeniiz321:


.github/workflows/run-tests.yml
```diff
@@ -94,6 +94,20 @@ jobs:
...
         working-directory: neofs-testcases
 
+      - name: Checkout xk6-neofs repository
+        uses: actions/checkout@v4
+        with:
+          repository: nspcc-dev/xk6-neofs
+          ref: master
+          path: xk6-neofs
+
+      - name: Build xk6-neofs
```
roman-khimov, line 104: Why building it? Why https://github.com/nspcc-dev/xk6-neofs/releases/tag/v0.1.2 is not sufficient?

evgeniiz321: because we run tests on macos, the binary above is for amd64. I can add if/else, but it seems simpler just to always build it.

roman-khimov: OK, got it, to be fixed with the next release.

roman-khimov (approved):

PR #2926, opened 2024-08-30
Upd/go 1.21

carpawell:
Tried to find a balance b/w memory management and readability. Feel free to ask for changes if some of `slices.Clone`s, etc are not suitable for you.


cmd/neofs-adm/internal/modules/morph/config.go
```diff
@@ -69,9 +70,7 @@ func dumpNetworkConfig(cmd *cobra.Command, _ []string) error {
 			netmapEigenTrustIterationsKey,
 			netmapEpochKey, netmapInnerRingCandidateFeeKey,
 			netmapMaxObjectSizeKey, netmapWithdrawFeeKey:
-			nbuf := make([]byte, 8)
-			copy(nbuf, v)
-			n := binary.LittleEndian.Uint64(nbuf)
+			n := binary.LittleEndian.Uint64(slices.Clone(v[:8]))
```
roman-khimov, line 73: But why is it copied at all?

pkg/network/group.go
```diff
@@ -108,12 +109,7 @@ func (x *AddressGroup) FromIterator(iter MultiAddressIterator) error {
 		return errors.New("missing network addresses")
 	}
 
-	if cap(as) >= addrNum {
-		as = as[:0]
-	} else {
-		as = make(AddressGroup, 0, addrNum)
-	}
-
+	as = slices.Grow(as, addrNum)[:0]
```
roman-khimov, line 112: I think it should be `slices.Grow(as[:0], addrNum)` then. But it's an interesting pattern as well since it allows to handle `nil`.

pkg/services/object/put/distributed.go
```diff
@@ -231,11 +232,7 @@ func (x placementIterator) iterateNodesForObject(obj oid.ID, f func(nodeDesc) er
 				}
 				return errIncompletePut{singleErr: err}
 			}
-			if uint(cap(nextNodeGroupKeys)) < replRem {
-				nextNodeGroupKeys = make([]string, 0, replRem)
-			} else {
-				nextNodeGroupKeys = nextNodeGroupKeys[:0]
-			}
+			nextNodeGroupKeys = slices.Grow(nextNodeGroupKeys, int(replRem))[:0]
```
roman-khimov, line 235: Ditto.

pkg/morph/client/netmap/netmap.go
```diff
@@ -119,7 +120,7 @@ func decodeNodeInfo(dst *netmap.NodeInfo, itemNode stackitem.Item) error {
 
 	switch node.State.Int64() {
 	default:
-		return fmt.Errorf("unsupported state %v", node.State)
+		return fmt.Errorf("%w: unsupported state %v", errors.ErrUnsupported, node.State)
```
roman-khimov, line 123: unsupported unsupported

pkg/morph/event/netmap/update_peer.go
```diff
@@ -46,7 +47,7 @@ func (s UpdatePeer) NotaryRequest() *payload.P2PNotaryRequest {
 func (s *UpdatePeer) decodeState(state int64) error {
 	switch s.state = nodestate.Type(state); s.state {
 	default:
-		return fmt.Errorf("unsupported node state %d", state)
+		return fmt.Errorf("%w: unsupported node state %d", errors.ErrUnsupported, state)
```
roman-khimov, line 50: Ditto

roman-khimov:
`slices.Clone()` is benign to me, btw. It won't differ from make/copy (but preserves `nil`).

cmd/neofs-adm/internal/modules/morph/verified_domains.go
```diff
@@ -190,11 +191,9 @@ func verifiedNodesDomainSetAccessList(cmd *cobra.Command, _ []string) error {
...
 loop:
 	for i := range records {
-		for j := range additionalRecords {
-			if additionalRecords[j] == records[i] {
-				mAlreadySetIndices[i] = struct{}{}
-				continue loop
-			}
+		if slices.Contains(additionalRecords, records[i]) {
+			mAlreadySetIndices[i] = struct{}{}
+			continue loop
```
cthulhu-rider, line 196: i'd drop no longer needed label

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Changelog for NeoFS Node
 ### Removed
 
 ### Updated
+- Go to 1.21 version (#2517)
```
cthulhu-rider, line 17: is this commit related to go upgrade?

cthulhu-rider (approved):

cmd/neofs-adm/internal/modules/morph/config.go
roman-khimov: Likely because it can be _less_ than 8 bytes in which case current code will panic.

carpawell:
> slices.Clone() is benign to me, btw

@roman-khimov, what would you say about https://github.com/nspcc-dev/neofs-node/blob/35d3c67c36177def07b6af9817fe2f507a53e6b2/pkg/local_object_storage/blobstor/fstree/fstree.go#L390-L391? I would be extremely happy to always copy slices as `slices.Clone` but it wastes some memory so i am not sure.

roman-khimov:
Notice that we don't have capacity specified here, so `data` can have some additional as well. But in this particular case `slices.Clone()` doesn't fit since we know we'll read more of data into the same buffer and we know its size exactly.

pkg/network/group.go
carpawell: so what will we do here? it was not a mistake, in my head it is "grow slice capacity, make its length zero". do not say the inverted version is bad just wanted to do a different thing

cmd/neofs-adm/internal/modules/morph/config.go
carpawell: ok, reverted this change as the minimal possible solution here (it would be strange if we have less than 8 bytes here though)

pkg/morph/event/netmap/update_peer.go
carpawell: thought about it, the full version is "unsupported operation: unsupported node state %d" now. cannot say i would be sad if i saw it, it may fit my expectation about error chain construction logic but don't mind dropping

carpawell:
@roman-khimov, sorry, fast search failed. https://github.com/nspcc-dev/neofs-node/blob/35d3c67c36177def07b6af9817fe2f507a53e6b2/pkg/local_object_storage/pilorama/forest.go#L164-L165 can be considered. I have changed it in this PR since we will drop the tree anyway hehe, but the question is still the same. Obviously there is no need to have more `Node` as a reserve in memory at all, there will not be any more nodes. The operation is quite often too.

carpawell:
> Notice that we don't have capacity specified here, so data can have some additional as well.

BTW, @roman-khimov, not sure it is true: https://go.dev/ref/spec#Making_slices_maps_and_channels. I always thought that missing cap equals len.

pkg/network/group.go
roman-khimov: Well, they're the same in effect since `Grow` doesn't change length. Whatever you prefer.

roman-khimov:
>   Error: SA1019: elliptic.Marshal has been deprecated since Go 1.21: for ECDH, use the crypto/ecdh package. This function returns an encoding equivalent to that of PublicKey.Bytes in crypto/ecdh.  (staticcheck)

carpawell:
@roman-khimov, hm, new linter started to see it. Fixed (described in the commit what a "fix" was done).

pkg/util/keyer/dashboard.go
```diff
@@ -45,7 +44,7 @@ func (d Dashboard) PrettyPrint(uncompressed, useHex bool) {
 
 	if d.pubKey != nil {
 		if uncompressed {
```
roman-khimov, line 46: Frankly, this branch doesn't have a lot of reasons to exist. But let's have it for now.

roman-khimov (approved):

Issue #2876, opened 2024-06-27
Exchange meta signatures during initial replication


roman-khimov:
## Is your feature request related to a problem? Please describe.
I'm always frustrated when object metadata is not signed properly. 

## Describe the solution you'd like
Once https://github.com/nspcc-dev/neofs-sdk-go/issues/589 is in we can exchange signatures of payloads like:
```
{
    "cid": "...",
    "oid": "...",
    "size": "...",
    "deletion": ["oid"],
    "lock": ["oid"],
    "network": number,
    "validuntil": number,
}
```
Exact format to be discussed.

carpawell:
@roman-khimov, @cthulhu-rider, about the message to sign: I see three main questions here:
1. Encoding of this message (it should be stable, should have implementations for other languages, and should be easy to understand)
2. Be able to easily validate information on contract side (i am not sure what contract/node should do exactly when it finds that an object was replicated _but_ meta info is not valid and was not accepted; though i am sure it will have applications and handling cases)
3. Versioning; LOCKs were not always with us, they were added after the main object operations, some other additional fields may appear in the future, and they will have to be added to the scheme then.

I have considered some options and taking all 3 restrictions made me hope our [ordered JSON](https://github.com/nspcc-dev/go-ordered-json) should be the best one. It is already maintained and used by us, so it is kinda controllable evil (TBH, usual (non-forked, RFC-specified) JSON with an array of objects like `{"key_name": "key_value"}` is a more preferable option to me but i cannot evaluate how serious such a suggestion is, need help here).

However, none of my suggestions (the ordered JSON not an exception) practically explains how we can satisfy the inevitable `3.` question. Once a new field appears, every node that stores objects should be able to understand a new scheme for new objects but also be ready to sign (or verify a signature) without the new field. Nodes container list also may be mixed (the new ones and the old ones), and it will be technically correct for them to sign different messages but also technically incorrect for an initial node to approve such mixed signatures. _But if backward compatibility_ is not a problem for us, we may extend API with a version number instead of a [simple flag](https://github.com/nspcc-dev/neofs-api/blob/01229b49e9165e693a48e2b96ccad528df34b55e/object/service.proto#L729) and simply drop unknown fields when an old node replicates an object. Saying there is no versioning at all and a new feature is a new feature and everybody should be suitable and ready for it is also a nice option as usual. Need some comments here.

roman-khimov:
[...]

PR #2928, opened 2024-09-04
feat/exchange object meta's signatures 

carpawell:


carpawell:
Format, ` "network": number,`, valid interval should be discussed. @roman-khimov, @cthulhu-rider

@AnnaShaleva, could you, please, also take a peek at what we are planning to use as a standard message? (see https://github.com/nspcc-dev/neofs-node/issues/2876 also)

pkg/services/object/put/distributed.go
```diff
@@ -138,11 +152,30 @@ func (t *distributedTarget) sendObject(node nodeDesc) error {
...
+			return fmt.Errorf("%w: public key differs", apistatus.ErrSignatureVerification)
+		}
+
+		fmt.Println("DEBUG: trying to verify signatures")
+
+		if !sig.Verify(t.objSharedMeta) {
+			return apistatus.ErrSignatureVerification
+		}
+
+		fmt.Println("DEBUG: verification was OK")
```
roman-khimov, line 176: :eyes:

pkg/core/object/replicate.go
```diff
@@ -0,0 +1,68 @@
...
+//
+// This (ordered) format is used (keys are strings):
+//
+//	"cid": _raw_ container ID (32 bytes)
+//	"oid": _raw_ object ID (32 bytes)
+//	"size": payload size
+//	"deleted": array of _raw_ object IDs
+//	"locked": array of _raw_ object IDs
+//	"validuntil": last valid epoch number for meta information
+//
```
roman-khimov, line 32: In some sense `network` is already present in `cid` the way it's derived now, but I'm still not sure about potential replays. Also, we could use this index for large objects, so some chain data can be useful.

roman-khimov (commented):
In general it's OK for the first version, we just need to catch the basic relations provided by NeoFS.

pkg/network/transport/object/grpc/replication.go
```diff
@@ -178,7 +180,20 @@ func (s *Server) Replicate(_ context.Context, req *objectGRPC.ReplicateRequest)
 		}}, nil
 	}
 
-	return new(objectGRPC.ReplicateResponse), nil
+	resp := new(objectGRPC.ReplicateResponse)
+	if req.GetSignObject() {
```
cthulhu-rider, line 184: do u plan to add tests for these cases? there are few of them now

pkg/services/object/put/distributed.go
```diff
@@ -138,11 +152,30 @@ func (t *distributedTarget) sendObject(node nodeDesc) error {
...
+
+	if !node.local {
+		if !bytes.Equal(sig.PublicKeyBytes(), node.info.PublicKey()) {
+			return fmt.Errorf("%w: public key differs", apistatus.ErrSignatureVerification)
+		}
+
+		fmt.Println("DEBUG: trying to verify signatures")
+
+		if !sig.Verify(t.objSharedMeta) {
+			return apistatus.ErrSignatureVerification
```
cthulhu-rider, line 173: quite a few signatures are checked during communication, so I would add context what _exact_ verification has failed

pkg/services/object/put/distributed.go
```diff
@@ -127,6 +130,17 @@ func (t *distributedTarget) Close() (oid.ID, error) {
...
+	var lockedObjs []oid.ID
+	switch t.objMeta.Type() {
+	case objectSDK.TypeTombstone:
+		deletedObjs = t.objMeta.Objects()
+	case objectSDK.TypeLock:
+		lockedObjs = t.objMeta.Objects()
+	default:
+	}
+
+	t.objSharedMeta = object.MetaInfo(t.obj.GetContainerID(), t.obj.GetID(), t.obj.PayloadSize(), deletedObjs, lockedObjs, t.obj.CreationEpoch())
```
cthulhu-rider, line 143: `t.objSharedMeta` is used conditionally while calculated unconditionally. Lets sync

pkg/services/object/put/distributed.go
```diff
@@ -138,11 +152,30 @@ func (t *distributedTarget) sendObject(node nodeDesc) error {
...
 		return fmt.Errorf("could not write header: %w", err)
-	} else if _, err := target.Close(); err != nil {
+	}
+
+	_, sig, err := target.Close()
+	if err != nil {
 		return fmt.Errorf("could not close object stream: %w", err)
 	}
+
+	if !node.local {
```
cthulhu-rider, line 165: shouldnt this be
```suggestion
	if t.localNodeInContainer && !node.local {
```
? Seems like ur proposal would fail when request goes through a non-container node

cthulhu-rider (commented):
quite a large amount of work has been done, I like how it fits into the old concepts

in the commit desc u indicate that the signatures are only partially used for now, do we have any pledged proposals for further use? I would also refer to them in order to confirm the motivation for the preliminary changes

pkg/network/transport/object/grpc/replication.go
```diff
@@ -190,3 +205,41 @@ func objectFromMessage(gMsg *objectGRPC.Object) (*object.Object, error) {
...
+
+		locked = make([]oid.ID, l.NumberOfMembers())
+		l.ReadMembers(locked)
+	default:
+	}
+
+	metaInfo := objectcore.MetaInfo(o.GetContainerID(), o.GetID(), o.PayloadSize(), deleted, locked, o.CreationEpoch())
+
+	var sig neofscrypto.Signature
+	err := sig.Calculate(s.signer, metaInfo)
```
cthulhu-rider, line 236: this differs from the original protocol [definition](https://github.com/nspcc-dev/neofs-api/blob/01229b49e9165e693a48e2b96ccad528df34b55e/object/service.proto#L725-L728)

do I understand correctly that we decided to limit the amount of information being signed, and this approach will eventually be backported into the protocol?

carpawell: yes, things have changed with time (or i did not get them from the beginning). we have discussed it with @roman-khimov, a custom message should be provided (see the 2. in the [comment](https://github.com/nspcc-dev/neofs-node/issues/2876#issuecomment-2319693139) mostly)

> and this approach will eventually be "backported" into the protocol?

yes, that is right. not release was done since so it should not be a problem

pkg/services/object/put/distributed.go
carpawell: hm, that's interesting. what a non-contianer node should do if it faces an object whose meta signatures it should check? @roman-khimov

pkg/services/object/put/distributed.go
carpawell: can you elaborate a little? this is calculated before any concurrent code, but is read concurrently, yes. do you mean some comments about it? or you want to see a lock here?

pkg/network/transport/object/grpc/replication.go
```diff
@@ -190,3 +205,41 @@ func objectFromMessage(gMsg *objectGRPC.Object) (*object.Object, error) {
...
+	case object.TypeLock:
+		var l object.Lock
+		err := l.Unmarshal(o.Payload())
+		if err != nil {
+			return nil, fmt.Errorf("reading locked objects: %w", err)
+		}
+
+		locked = make([]oid.ID, l.NumberOfMembers())
+		l.ReadMembers(locked)
+	default:
```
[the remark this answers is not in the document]
carpawell: no, why? if we are replicating a lock/tomb object, we add IDs, but if we are adding a regular object, nothing should be done in this `switch`
or what you suggest here?

pkg/core/object/replicate.go
```diff
@@ -0,0 +1,68 @@
...
+// Last valid epoch is object's creation epoch + 10.
+func MetaInfo(cID cid.ID, oID oid.ID, pSize uint64, deleted []oid.ID, locked []oid.ID, createdAt uint64) []byte {
+	kvs := make([]stackitem.MapElement, 0, currentVersion)
+	kvs = append(kvs, kv(cidKey, cID[:]))
```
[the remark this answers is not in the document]
carpawell: ok

pkg/core/object/replicate.go
```diff
@@ -0,0 +1,68 @@
...
+//
+//	"cid": _raw_ container ID (32 bytes)
+//	"oid": _raw_ object ID (32 bytes)
+//	"size": payload size
+//	"deleted": array of _raw_ object IDs
+//	"locked": array of _raw_ object IDs
+//	"validuntil": last valid epoch number for meta information
+//
+// Last valid epoch is object's creation epoch + 10.
+func MetaInfo(cID cid.ID, oID oid.ID, pSize uint64, deleted []oid.ID, locked []oid.ID, createdAt uint64) []byte {
```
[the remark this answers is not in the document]
carpawell: dont mind

pkg/core/object/replicate.go
carpawell: ok, added network magic

pkg/services/object/put/distributed.go
carpawell: sry, fixed

carpawell:
> in the commit desc u indicate that the signatures are only partially used for now, do we have any pledged proposals for further use?

Yes, there should be some event [contracts](https://github.com/nspcc-dev/neofs-contract/issues/414) and contract-side [validation](https://github.com/nspcc-dev/neofs-contract/issues/413). This is kinda the first iteration, all the other work will be done soon after this PR is discussed and approved.

> I would also refer to them in order to confirm the motivation for the preliminary changes

Added contract issues to the commit desc.

pkg/services/object/put/distributed.go
roman-khimov: Why would it be involved in replication?

pkg/network/transport/object/grpc/replication.go
carpawell: added some

pkg/services/object/put/distributed.go
carpawell: what does a node do when it is asked to sign with session key objects of a container it does not belong to?

roman-khimov: It's signing and passing the object to container node.

carpawell: oh, i have overthought some things in this code but yes, `t.localNodeInContainer` is enough, added. (also added nil signature check)

carpawell:
Full report: https://rest.fs.neo.org/HXSaMJXk2g8C14ht8HSi7BBaiYZ1HeWh2xnWPGQCg4H6/1872-1726003656/index.html

roman-khimov (approved):

Issue #2890, opened 2024-07-10
Cobra command runners skip `defer`ed funcs and postruns on failures


cthulhu-rider:
some [`Cobra`](https://pkg.go.dev/github.com/spf13/cobra)-based programs exit instantly on particular errors:
* https://github.com/nspcc-dev/neofs-node/blob/450c4e5ed85f77dc77a13625337d71f7d97fa9ef/cmd/neofs-cli/internal/common/exit.go#L22
* https://github.com/nspcc-dev/neofs-node/blob/450c4e5ed85f77dc77a13625337d71f7d97fa9ef/cmd/neofs-lens/internal/errors.go#L20

they call `os.Exit` due to which neither [cmd](https://pkg.go.dev/github.com/spf13/cobra#Command) postruns, nor [finalizers](https://pkg.go.dev/github.com/spf13/cobra#OnFinalize), nor even exec `defer` funcs are called. The latter are usually responsible for allocating resources such as network connections, files, etc.

## Steps to Reproduce
1. add `cmd/neofs-test/main.go`
```go
func main() {
	(&cobra.Command{
		Use: "test",
		Run: func(cmd *cobra.Command, _ []string) {
			cobra.OnFinalize(func() {
				cmd.Println("in finalizer")
			})
			defer func() {
				cmd.Println("in defer")
			}()
			fmt.Fprintf(os.Stderr, "err: %v\n", errors.New("any error"))
			os.Exit(1)
		},
		PostRun: func(cmd *cobra.Command, args []string) {
			cmd.Println("in PostRun")
		},
		PersistentPostRun: func(cmd *cobra.Command, args []string) {
			cmd.Println("in PersistentPostRun")
		},
	}).Execute()
}

```
3. `make bin/neofs-test`
4. `./bin/neofs-test

## Expected Behavior
```
$ ./bin/neofs-test
in defer
in PostRun
in PersistentPostRun
in finalizer
err: any error
$ echo $?
1
```

## Current Behavior
```
$ ./bin/neofs-test
err: any error
$ echo $?
1
```

### 1. Finalizing closure in `Run`
the most clean solution to me in terms of code and command lifetime. At the same time, requires a lot of changes, code will be pretty "spaghetti"
```go
func main() {
	err := (&cobra.Command{
		Use:           "test",
		SilenceErrors: true,
		SilenceUsage:  true,
		Run: func(cmd *cobra.Command, _ []string) {
			var code int
			var err error
[...]

PR #2933, opened 2024-09-10
Feat/control service object list

### LINKED ISSUE #2853: List objects via control service
I'm always frustrated when I can't get a list of objects node knows about. Lens can tell it (#2852), but it requires stopping a node (at least with peapods) which is not very convenient.

Add some listing command to the control service and CLI.

End-rey:
Closes #2853.


pkg/services/control/service.proto
```diff
@@ -20,6 +20,9 @@ service ControlService {
     // Returns list that contains information about all shards of a node.
     rpc ListShards (ListShardsRequest) returns (ListShardsResponse);
 
+    // Returns list that contains information about all objects in a node.
+    rpc ListObjects (ListObjectsRequest) returns (ListObjectsResponse);
```
carpawell, line 24: have you considered streaming RPC? there can be millions or even billions objects, each of them at least is 64 bytes long

pkg/services/control/types.proto
```diff
@@ -141,6 +141,10 @@ message ShardInfo {
     string pilorama_path = 7 [json_name = "piloramaPath"];
 }
 
+message ObjectInfo {
+    string address = 1 [json_name = "object address"];
```
carpawell, line 145: dont mind `string` in general but a raw byte slice can be a little bit more efficient here

cmd/neofs-cli/modules/control/object_list.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/nspcc-dev/neofs-node/pkg/services/control"
+	"github.com/spf13/cobra"
+)
+
+var listObjectsCmd = &cobra.Command{
+	Use:   "list",
+	Short: "List of all objects in the storage node",
+	Long:  "List of all objects in the storage node",
+	Args:  cobra.NoArgs,
+	Run:   listObjects,
```
carpawell, line 17: we are trying to move to `RunE` (i hope, https://github.com/nspcc-dev/neofs-node/issues/2890), can you, please, start to use it here too?

End-rey: I was based on this #623, but I don't think I understood.

Makefile
```diff
@@ -69,7 +69,7 @@ dep:
 protoc:
 	@GOPRIVATE=github.com/nspcc-dev go mod vendor
 	# Install specific version for protobuf lib
-	@go list -f '{{.Path}}/...@{{.Version}}' -m  github.com/golang/protobuf | xargs go install -v
+	@go list -f '{{.Path}}/...@{{.Version}}' -m  google.golang.org/protobuf | xargs go install -v
```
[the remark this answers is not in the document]
End-rey: Done with streaming RPC.

cmd/neofs-cli/modules/control/object_list.go
carpawell: ok, since it is not discussed, let it be `Run` then

cmd/neofs-cli/modules/control/object_list.go
```diff
@@ -0,0 +1,54 @@
...
+	pk := key.Get(cmd)
+
+	req := &control.ListObjectsRequest{
+		Body: &control.ListObjectsRequest_Body{},
+	}
+	signRequest(cmd, pk, req)
+
+	cli := getClient(ctx, cmd)
+
+	var resp []*control.ListObjectsResponse
```
carpawell, line 37: my stream request was mostly about memory consumption and user experience: if we received some info, we can print it and drop it from memory. if we are about to download e.g. 100mb it will take seconds of no output for a user. can you try not to hold every streamed message in memory?

End-rey: I thought about it, but I couldn't find a good solution. Now I've done it using the function inside.

pkg/services/control/service.proto
```diff
@@ -167,6 +170,34 @@ message ListShardsResponse {
...
+    // Body signature.
+    Signature signature = 2;
+}
+
+// ListObjects response.
+message ListObjectsResponse {
+    // Response body structure.
+    message Body {
+        // Address of the object.
+        bytes object_address = 1;
```
cthulhu-rider, line 191: per-object message would be very inefficient, especially with each message signing. Suggest to make it `repeated`. WIth this, element limit should be contolled. See similar [implementation](https://github.com/nspcc-dev/neofs-node/blob/0a6e88a30c8e07605cd151cbabada1818b2e9e8f/pkg/services/object/transport_splitter.go#L151)

pkg/services/control/server/list_objects.go
```diff
@@ -0,0 +1,45 @@
...
+		return status.Error(codes.PermissionDenied, err.Error())
+	}
+
+	// check availability
+	err := s.ready()
+	if err != nil {
+		return status.Error(codes.Internal, err.Error())
+	}
+
+	objectsIds, err := engine.List(s.storage, 0)
```
cthulhu-rider, line 22: huge number of objects can overload the server unnecessarily. Consider https://github.com/nspcc-dev/neofs-node/blob/1a9784c11ecc9f41193a4b99c5171ad470afb016/pkg/local_object_storage/engine/list.go#L55-L62

pay attention to `Does not include inhumed objects` statement

cmd/neofs-cli/modules/control/object_list.go
```diff
@@ -0,0 +1,47 @@
...
+	req := &control.ListObjectsRequest{
+		Body: &control.ListObjectsRequest_Body{},
+	}
+	signRequest(cmd, pk, req)
+
+	cli := getClient(ctx, cmd)
+
+	var err error
+	err = cli.ExecRaw(func(client *rawclient.Client) error {
+		err = control.ListObjects(client, req, func(r *control.ListObjectsResponse) {
```
cthulhu-rider, line 39: better to just `return` w/o assignment to closured var

cmd/neofs-cli/modules/control/object_list.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/common"
+	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/commonflags"
+	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/key"
+	"github.com/nspcc-dev/neofs-node/pkg/services/control"
+	"github.com/spf13/cobra"
+)
+
+var listObjectsCmd = &cobra.Command{
+	Use:   "list",
+	Short: "List of all objects in the storage node",
```
cthulhu-rider, line 14:
```suggestion
	Short: "List all objects in the storage node",
```
or
```suggestion
	Short: "Get list of all objects in the storage node",
```

pkg/services/control/service.proto
End-rey: Done with repeated field.

pkg/services/control/server/list_objects.go
End-rey: Done with `ListWithCursor`. Set maximum amount of addresses from func to 1024, but not sure about this number.

pkg/services/control/server/list_objects.go
```diff
@@ -0,0 +1,59 @@
...
+
+		addresses := res.AddressList()
+		objectsAddresses := make([][]byte, 0, len(addresses))
+		for _, objectId := range addresses {
+			objectsAddresses = append(objectsAddresses, []byte(objectId.Address.EncodeToString()))
+		}
+
+		resp := &control.ListObjectsResponse{
+			Body: &control.ListObjectsResponse_Body{
+				ObjectAddress: objectsAddresses,
```
cthulhu-rider, line 44: now u propose to transmit 1024 of 89B addresses per message, which is ~89K. It's still far away from available 4MB limit, so we gonna sign and send more messages then we could

lets reserve 64K for service bytes and future fields, and fit soft limit `=(4M-64K)/89=46390` of addresses. Ofc write a comment how this number is chosen

the next optimization could be packing addresses into the message w/o intermediate `addresses := res.AddressList()` memory buffer. `engine.ListWithCursorPrm` could accept functional address handler to make it possible. I suggest to leave an issue for this

End-rey: Done.

cthulhu-rider:
some code became deprecated with 5bbbf048ab64440ffca5d2a580f55289c7eb4e55, fix pls

cthulhu-rider (approved):

roman-khimov (approved):

PR #2934, opened 2024-09-11
IR additional health status

### LINKED ISSUE #2923: Health status: HEALTH_STATUS_UNDEFINED
while v43 process looks healthy (reports metrics, no errors in the log) health status is always `HEALTH_STATUS_UNDEFINED`

```
# neofs-cli control healthcheck -c /etc/neofs/ir/control.yml --ir
Health status: HEALTH_STATUS_UNDEFINED
```

this breaks ansible and update process

End-rey:
Closes #2923.

I'm not sure I've found the right name for the status, and I was also thinking that I could add an additional status before updating contracts, since this is the longest process.
I made a metric type like the storage node (number from enum), but I also thought about the string parameter.


pkg/services/control/ir/types.proto
```diff
@@ -18,12 +18,15 @@ enum HealthStatus {
     // Undefined status, default value.
     HEALTH_STATUS_UNDEFINED = 0;
 
+    // Create IR server
+    CREATE_SERVER = 1;
```
roman-khimov, line 22: So it something in-between the IR's `New` and `Start`, right? What happens there? All of blockchain init?

End-rey: As I understand it, all initialization of blockchain, clients, sidechain, configurations and others takes place in `New` func of ir, so after this function it just goes `Strart`.

pkg/services/control/ir/types.proto
```diff
@@ -18,12 +18,15 @@ enum HealthStatus {
...
+
     // IR application is starting.
-    STARTING = 1;
+    STARTING = 2;
 
     // IR application is started and serves all services.
-    READY = 2;
+    READY = 3;
```
carpawell, line 28: i think you have broken the older version of cli that expect `2` as READY: https://github.com/nspcc-dev/neofs-node/blob/4a1bc798d2c5cbfd366981528745132c20b09f42/cmd/neofs-cli/modules/control/healthcheck.go#L70-L72

it is internal API but still i think it is not worth it, @roman-khimov

carpawell: also, FIY, if we do such things we usually add something to our CHANGELOG to prepare our users for their catastrophic update, e.g.: https://github.com/nspcc-dev/neofs-node/blob/4a1bc798d2c5cbfd366981528745132c20b09f42/CHANGELOG.md?plain=1#L154-L170

pkg/innerring/innerring.go
```diff
@@ -308,12 +308,17 @@ func (s *Server) registerCloser(f func() error) {
...
 }
 
-// New creates instance of inner ring sever structure.
+// New creates instance of inner ring server structure.
 func New(ctx context.Context, log *zap.Logger, cfg *viper.Viper, errChan chan<- error) (*Server, error) {
 	var err error
 	server := &Server{log: log}
 
-	server.setHealthStatus(control.HealthStatus_HEALTH_STATUS_UNDEFINED)
+	server.setHealthStatus(control.HealthStatus_CREATE_SERVER)
```
carpawell, line 316: i think `HealthStatus_CREATE_SERVER` was not a problem at all. you started a server, it only created `&Server{log: log}`, so it is really undefined now. i guess the initial issue was mostly about extending codes: you set it like "STARTING_BLOCKCHAIN" and then call `server.bc.Run(ctx)`; you set it like "DEPLOYING_NETWORK" and then you call `deploy.Deploy(ctx, deployPrm)`, etc. it allows understand more about what is happening, starting IR may take minutes now, and all an admin got is minutes of "undefined" that goes immediately to READY then

carpawell: the rule of thumb may be: any operation that is _not_ about local CPU work but about any I/O that potentially may block and that an admin should know about

pkg/services/control/ir/types.proto
roman-khimov: Compatibility better be kept, it doesn't cost a lot.

End-rey: Added new statuses after the existing ones, therefore, it no longer breaks.

pkg/innerring/innerring.go
End-rey: Added new statuses that seemed to work with I/O and can bloc func, but I'm not sure about them.

pkg/services/control/ir/types.proto
```diff
@@ -26,4 +26,18 @@ enum HealthStatus {
...
+    // Create wallet from file.
+    CREATING_WALLET = 5;
+
+    STARTING_BLOCKCHAIN = 6;
+
+    // Deploying Neo network by blockchain
+    DEPLOYING_NETWORK = 7;
+
+    // Initialize NeoFS contracts
+    INIT_CONTRACTS = 8;
```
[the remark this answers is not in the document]
End-rey: The list is now like this:
0. HEALTH_STATUS_UNDEFINED 
1. STARTING 
2. READY
3. SHUTTING_DOWN
4. INITIALIZING_NETWORK (for compatibility with last version of cli, I added it at the end)

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ Changelog for NeoFS Node
 
 ### Added
 - More effective FSTree writer for HDDs, new configuration options for it (#2814)
+- New health statuses in inner ring (#2934)
```
cthulhu-rider, line 8: lets add this record with the corresponding commit

cthulhu-rider (approved):
idk why lint action failed

roman-khimov (approved):

PR #2937, opened 2024-09-12
Fix/TS expiration handling without network

carpawell:


CHANGELOG.md
```diff
@@ -40,11 +40,17 @@ Changelog for NeoFS Node
 - `ObjectService`'s `Search` and `Replicate` RPC handlers cache up to 1000 lists of container nodes (#2892)
 - Default max_traceable_blocks Morph setting lowered to 17280 from 2102400 (#2897)
 - `ObjectService`'s `Get`/`Head`/`GetRange` RPC handlers cache up to 10K lists of per-object sorted container nodes (#2896)
+- Metabase graveyard scheme (#2929)
```
roman-khimov, line 44: What? 0.43.0 is there already.

CHANGELOG.md
```diff
@@ -40,11 +40,17 @@ Changelog for NeoFS Node
...
 
 ### Updated
 - neofs-contract dependency to 0.20.0 (#2872)
 - NeoGo dependency to 0.106.3 (#2872)
 
+### Updating from v0.42.1
+It is required to resynchronize metabases due to changed metabase scheme; any
```
roman-khimov, line 51: Can you handle it during update? Current epoch + configured tombstone lifetime, maybe?

carpawell: you mean rewrite every record on the first start? we can but do we want it?

CHANGELOG.md
carpawell: missed the target

CHANGELOG.md
roman-khimov: I think it's perfectly fine. It really simplifies upgrade.

carpawell: added automigration

pkg/local_object_storage/metabase/version.go
```diff
@@ -49,7 +58,7 @@ func checkVersion(tx *bbolt.Tx, initialized bool) error {
 	return nil
 }
 
-func updateVersion(tx *bbolt.Tx, version uint64) error {
+func (db *DB) updateVersion(tx *bbolt.Tx, version uint64) error {
```
roman-khimov, line 61: Why making it a method if `db` is never used?

pkg/local_object_storage/metabase/graveyard.go
```diff
@@ -261,6 +262,52 @@ func (db *DB) DropGraves(tss []TombstonedObject) error {
...
 
+// DropExpiredTSMarks run through the graveyard and drops tombstone marks with
+// tombstones whose expiration is _less_ than provided epoch.
+// Returns number of marks dropped.
+func (db *DB) DropExpiredTSMarks(epoch uint64) (int, error) {
+	db.modeMtx.RLock()
+	defer db.modeMtx.RUnlock()
+
+	if db.mode.NoMetabase() {
+		return -1, ErrDegradedMode
```
cthulhu-rider, line 256: dont see any sense in negative return, can be zero?

pkg/local_object_storage/metabase/graveyard.go
```diff
@@ -261,6 +262,52 @@ func (db *DB) DropGraves(tss []TombstonedObject) error {
...
+	err := db.boltDB.Update(func(tx *bbolt.Tx) error {
+		bkt := tx.Bucket(graveyardBucketName)
+		c := bkt.Cursor()
+		k, v := c.First()
+
+		for k != nil {
+			if binary.LittleEndian.Uint64(v[addressKeySize:]) < epoch {
+				err := c.Delete()
+				if err != nil {
+					return fmt.Errorf("cleared %d TS marks in %d epoch and got error: %w", counter, epoch, err)
```
cthulhu-rider, line 272: id add context to the error outside db handler

pkg/local_object_storage/metabase/graveyard.go
```diff
@@ -261,6 +262,52 @@ func (db *DB) DropGraves(tss []TombstonedObject) error {
...
+
+	var counter int
+
+	err := db.boltDB.Update(func(tx *bbolt.Tx) error {
+		bkt := tx.Bucket(graveyardBucketName)
+		c := bkt.Cursor()
+		k, v := c.First()
+
+		for k != nil {
+			if binary.LittleEndian.Uint64(v[addressKeySize:]) < epoch {
```
cthulhu-rider, line 269: i'd still check `v` length explicitly to avoid unclear panic, especially since the suffix was not always there

pkg/local_object_storage/metabase/graveyard.go
```diff
@@ -261,6 +262,52 @@ func (db *DB) DropGraves(tss []TombstonedObject) error {
...
+	if db.mode.NoMetabase() {
+		return -1, ErrDegradedMode
+	} else if db.mode.ReadOnly() {
+		return -1, ErrReadOnlyMode
+	}
+
+	var counter int
+
+	err := db.boltDB.Update(func(tx *bbolt.Tx) error {
+		bkt := tx.Bucket(graveyardBucketName)
```
cthulhu-rider, line 264: what `if bkt == nil`?

pkg/local_object_storage/metabase/graveyard.go
```diff
@@ -220,10 +228,18 @@ func garbageFromKV(k []byte) (res GarbageObject, err error) {
...
+	if err != nil {
+		return res, fmt.Errorf("decode tombstone target from key: %w", err)
+	}
+
+	err = decodeAddressFromKey(&res.tomb, v[:addressKeySize])
+	if err != nil {
+		return res, fmt.Errorf("decode tombstone address from value: %w", err)
+	}
+
+	if len(v) == addressKeySize+8 {
```
cthulhu-rider, line 241: `>` case should be either erroneous or no-problem, silence is bad

pkg/local_object_storage/metabase/version.go
```diff
@@ -71,3 +80,26 @@ func getVersion(tx *bbolt.Tx) uint64 {
...
 	return 0
 }
+
+var migrateFrom = map[uint64]func(*DB, *bbolt.Tx) error{
+	2: migrateFrom2Version,
+}
+
+func migrateFrom2Version(db *DB, tx *bbolt.Tx) error {
+	tsExpiration := db.epochState.CurrentEpoch() + objectconfig.DefaultTombstoneLifetime
+	bkt := tx.Bucket(graveyardBucketName)
```
cthulhu-rider, line 90: again can be nil

pkg/local_object_storage/metabase/version.go
```diff
@@ -71,3 +80,26 @@ func getVersion(tx *bbolt.Tx) uint64 {
...
+	c := bkt.Cursor()
+
+	for k, v := c.First(); k != nil; k, v = c.Next() {
+		newVal := make([]byte, addressKeySize, addressKeySize+8)
+		copy(newVal, v)
```
cthulhu-rider, line 95: imo we should assert `len(v) == addressKeySize`

pkg/local_object_storage/metabase/version.go
```diff
@@ -71,3 +80,26 @@ func getVersion(tx *bbolt.Tx) uint64 {
 
 	return 0
 }
+
+var migrateFrom = map[uint64]func(*DB, *bbolt.Tx) error{
+	2: migrateFrom2Version,
```
cthulhu-rider, line 85: won't it be easier to just make migration steps one-by-one? iiuc following version funcs will have to call previous ones, right?

cthulhu-rider (commented):
i dont think it's good to separate commit dropping old implementation from the one changing the approach. Although red diff is pretty, transient state with dead code makes no sense

pkg/local_object_storage/metabase/version.go
roman-khimov: It should one-by-one, but we may strip older handlers (like we don't have handlers for 0 and 1 already). We don't have a lot of versions, so 10 `nil` pointers are not that big of a problem, but a map works fine as well (check N, check N+1, etc).

cthulhu-rider: ok will see in practice

carpawell: there is a single migration for now, not sure how it will look further, can be changed in the future

pkg/local_object_storage/metabase/version.go
carpawell: added but i dont know, it is kinda we need a check in every code that takes _any_ value from db that was _not created by this exact code_ (if fact any start except a start from scratch)
returning an error now, not sure what else can be done here

pkg/local_object_storage/metabase/version.go
carpawell: all the code added in this PR works on either explicitly inited database or on the database with a known version (that has also been inited sometime ago). in this case added a check since this code works on a database that _may_ be corrupted by some different code but a user can do nothing with it and it is not a situation that can be handled somehow

pkg/local_object_storage/metabase/graveyard.go
carpawell: also not expected on an inited metabase: https://github.com/nspcc-dev/neofs-node/blob/b135a0efe9c6e38ce69f53f9dd1c191d62f218bd/pkg/local_object_storage/metabase/control.go#L104

pkg/local_object_storage/metabase/graveyard.go
carpawell: > to avoid unclear panic

and what should such code do then? err? warn print? skip? since we have migration now and db versioning, this situation unexpected and never is a normal situation

pkg/local_object_storage/metabase/graveyard.go
carpawell: moved but dont see any improvements/disadvantages

pkg/local_object_storage/metabase/graveyard.go
carpawell: just to highlight the error? changed but don't see a problem here

pkg/local_object_storage/metabase/graveyard.go
cthulhu-rider: panic with clear report

pkg/local_object_storage/metabase/graveyard.go
cthulhu-rider: clear panic

pkg/local_object_storage/metabase/version.go
cthulhu-rider: > returning an error now, not sure what else can be done here

some error is at least better than silence. When we expect strict value format, and it's unintentionally violated in fact, app should react on its own - remove the shard from the game or completely shutdown. Otherwise, further runtime may be unsafe in relation to stored data

pkg/local_object_storage/metabase/graveyard.go
carpawell: this operation is a process of `Inhume` and none of the methods in `inhume.go` does this check and panic. this is important to me cause it is literally "put" and then "get" with checking if the result is not nil, not sure bbolt code should be like this; and if we want this to be paniced with a desc, we have to add a desc in many more places

pkg/local_object_storage/metabase/graveyard.go
carpawell: should such code (and many more) have clear panic too: https://github.com/nspcc-dev/neofs-node/blob/b135a0efe9c6e38ce69f53f9dd1c191d62f218bd/pkg/local_object_storage/metabase/util.go#L248-L258
?

pkg/local_object_storage/metabase/version.go
carpawell: > How exactly node behaves on metabase panic currently?

this is done in the main routine, it stops

pkg/local_object_storage/metabase/graveyard.go
cthulhu-rider: any panic should be as much informative as possible preferably

pkg/local_object_storage/metabase/graveyard.go
cthulhu-rider: as i said in https://github.com/nspcc-dev/neofs-node/pull/2937#discussion_r1771277638, we need to dedicate a new issue for this then. Could u pls create it?

cthulhu-rider (approved):

roman-khimov:
>> transient state with dead code makes no sense
>
> It separates functional changing and unused code deletion

OK both ways to me (since the node works fine between commits anyway). Maybe with slight preference to deletion in the same commit as the code becomes redundant.

pkg/local_object_storage/metabase/graveyard.go
carpawell: can you describe the problem so I can create it? as i have already said, panic is not expected _at all_. data is versioned and/or inited explicitly, any changed database data should be considered as a strange and not-possible-to-handle situation. panic with trace or panic with trace and two more words almost don't differ to me, dev will have to understand the whole situation anyway, and user will understand nothing. if values from bbolt should be expected corrupted we have to changed _many many_ places. considering all the said above, what do you want to be attached to a new issue?

roman-khimov: This can only happen with corrupted DB. Corrupted meta DB -> resync meta please. But making it a proper error in all cases won't be easy.

roman-khimov (approved):

PR #2942, opened 2024-09-19
cli: Don't print help and don't skip `defer`ed funcs on failures

### LINKED ISSUE #623: cli: Do not print help on any failure

End-rey:
Closes #623, #2890.

In `neofs-cli` I add func `common.WrapError` that wrap error to `custom_errors.ExitErr` and return it on every error, But I doubt it. Maybe it is necessary to wrap only the first appearance of this error, or vice versa, only the last one.


cmd/neofs-cli/internal/common/exit.go
```diff
@@ -3,25 +3,32 @@ package common
...
 
-// ExitOnErr prints error and exits with a code depending on the error type
+// WrapError wrap error to custom_errors.ExitErr, if it not already, and add a code depending on the error type
 //
 //	0 if nil
 //	1 if [sdkstatus.ErrServerInternal] or untyped
 //	2 if [sdkstatus.ErrObjectAccessDenied]
 //	3 if [ErrAwaitTimeout]
-func ExitOnErr(cmd *cobra.Command, errFmt string, err error) {
+func WrapError(errFmt string, err error) *custom_errors.ExitErr {
```
roman-khimov, line 21: I'd probably do it as `WrapError(err error)` and leave `fmt.Errorf()` to the caller if and when needed. You still can unwrap internally to get the code, but formatting would be easier, I think. You can have cases like `"smth (additional data: %d): %w"` and this won't fit the current interface.

CHANGELOG.md
```diff
@@ -9,9 +9,11 @@ Changelog for NeoFS Node
 - Expose health status of inner ring via Prometheus (#2934)
 
 ### Fixed
+- `defer` func is not skipped in cobra-based programs (#2942)
```
roman-khimov, line 12: Not sure it makes sense for outside reader, but don't have immediate suggestions either.

roman-khimov (commented):
Conflicts, rebase pls. But looks nice in general. Probably we can add some tests for it as well.

cmd/neofs-adm/internal/modules/storagecfg/root.go
```diff
@@ -382,15 +455,20 @@ loop:
 			cmd.Print(".")
 		case <-timer.C:
 			cmd.Printf("\nTimeout while waiting for transaction to persist.\n")
-			if getConfirmation(false, "Continue configuration? yes/[no]: ") {
-				return
+			if flag, err := getConfirmation(false, "Continue configuration? yes/[no]: "); flag {
+				if err != nil {
```
cthulhu-rider, line 459: this never happens for `flag==true`, i think the condition should be
```go
if flag, err := getConfirmation(false, "Continue configuration? yes/[no]: "); flag || err != nil {
	return err
}
```

cmd/neofs-cli/internal/common/exit.go
cthulhu-rider, line 21: it should not return pointers to be compliant with `ExitOnErr` making
```go
var e ExitErr
if !errors.As(err, &e) {
	e.Code = 1
}
```
easy to check
```go
func TestExit(t *testing.T) {
	custom_errors.ExitOnErr(common.WrapError("context; %w", common.ErrAwaitTimeout))
}
```
expected `3`, got `1`

cmd/neofs-cli/internal/client/sdk.go
```diff
@@ -19,12 +18,12 @@ var errInvalidEndpoint = errors.New("provided RPC endpoint is incorrect")
 
 // GetSDKClientByFlag returns default neofs-sdk-go client using the specified flag for the address.
 // On error, outputs to stderr of cmd and exits with non-zero code.
-func GetSDKClientByFlag(ctx context.Context, cmd *cobra.Command, endpointFlag string) *client.Client {
+func GetSDKClientByFlag(ctx context.Context, endpointFlag string) (*client.Client, error) {
 	cli, err := getSDKClientByFlag(ctx, endpointFlag)
 	if err != nil {
-		common.ExitOnErr(cmd, "can't create API client: %w", err)
+		return nil, common.WrapError("can't create API client: %w", err)
```
cthulhu-rider, line 24: i dont think this is a good place to throw `ExitErr`, this is the responsibility of `RunE`s. So, just `return nil, fmt.Errorf(...)`

cmd/neofs-cli/internal/common/exit.go
```diff
@@ -3,25 +3,32 @@ package common
...
 	if err == nil {
-		return
+		return nil
+	}
+
+	var exitErr *custom_errors.ExitErr
+	if errors.As(err, &exitErr) {
```
cthulhu-rider, line 27: having this means that code does not care who makes `ExitErr` while it should. Imo normally this should never be true. The most stable approach is to wrap by `RunE` functions only

cmd/neofs-cli/internal/common/eacl.go
```diff
@@ -14,33 +14,34 @@ import (
 var errUnsupportedEACLFormat = fmt.Errorf("%w: unsupported eACL format", errors.ErrUnsupported)
 
 // ReadEACL reads extended ACL table from eaclPath.
-func ReadEACL(cmd *cobra.Command, eaclPath string) eacl.Table {
+func ReadEACL(cmd *cobra.Command, eaclPath string) (eacl.Table, error) {
 	_, err := os.Stat(eaclPath) // check if `eaclPath` is an existing file
 	if err != nil {
-		ExitOnErr(cmd, "", errors.New("incorrect path to file with EACL"))
+		return eacl.Table{}, WrapError("", errors.New("incorrect path to file with EACL"))
```
cthulhu-rider, line 20: here too enough to `return eacl.Table{}, errors.New("incorrect path to file with EACL")` and let caller how to wrap it. With this, u wont need to wrap all function returns

cmd/neofs-lens/internal/meta/get.go
```diff
@@ -16,28 +16,35 @@ var getCMD = &cobra.Command{
...
 }
 
-func getFunc(cmd *cobra.Command, _ []string) {
+func getFunc(cmd *cobra.Command, _ []string) error {
 	var addr oid.Address
 
 	err := addr.DecodeString(vAddress)
-	common.ExitOnErr(cmd, common.Errf("invalid address argument: %w", err))
+	if err != nil {
+		return common.Errf("invalid address argument: %w", err)
```
cthulhu-rider, line 32: `common.Errf` made sense with `ExitOnErr` func since it allowed to avoid `if` statement. Now we should use `fmt.Errorf` (and purge `Errf`)

cthulhu-rider (changes requested):
in total, lux gut

afaik #623 can be resolved purely independent, so its better to dedicate a separate small (last) commit for it to not mix with heavy #2890.

also tag intermediate commits with `Refs #2890.`

End-rey:
Now, in the `neofs-cli`, func `common.WrapError` is used after receiving the `cobra` command error in `main`.

cthulhu-rider (approved):
still suggest to ref first 3 commits to #2890 https://github.com/nspcc-dev/neofs-node/pull/2942#pullrequestreview-2317317501

cmd/neofs-cli/modules/object/hash.go
```diff
@@ -48,40 +47,54 @@ func initObjectHashCmd() {
...
 	salt, err := hex.DecodeString(strSalt)
-	common.ExitOnErr(cmd, "could not decode salt: %w", err)
+	if err != nil {
+		return fmt.Errorf("could not decode salt: %w", err)
+	}
 
-	pk := key.GetOrGenerate(cmd)
-	cli := internalclient.GetSDKClientByFlag(ctx, cmd, commonflags.RPC)
+	pk, _ := key.GetOrGenerate(cmd)
+	cli, _ := internalclient.GetSDKClientByFlag(ctx, commonflags.RPC)
```
roman-khimov, line 79: Not handled.

cmd/neofs-cli/modules/object/head.go
```diff
@@ -69,15 +84,23 @@ func getObjectHeader(cmd *cobra.Command, _ []string) {
 
 	res, err := internalclient.HeadObject(ctx, prm)
 	if err != nil {
-		if ok := printSplitInfoErr(cmd, err); ok {
-			return
+		if ok, err := printSplitInfoErr(cmd, err); ok {
+			if err != nil {
```
roman-khimov, line 88: Looks like `return err` for both cases.

cmd/neofs-cli/modules/object/range.go
```diff
@@ -94,36 +110,48 @@ func getObjectRange(cmd *cobra.Command, _ []string) {
 
 	_, err = internalclient.PayloadRange(ctx, prm)
 	if err != nil {
-		if ok := printSplitInfoErr(cmd, err); ok {
-			return
+		if ok, err := printSplitInfoErr(cmd, err); ok {
+			if err != nil {
+				return err
```
roman-khimov, line 115: Just `return err`, no need for `nil` check, you don't wrap it anyway.

roman-khimov (approved):

PR #2946, opened 2024-09-23
shard: drop more of context from local stores

### LINKED ISSUE #1911: Add `contextcheck` linter
Carefully check each case, I think many of them are worth fixing.

```
dzeta@wpc ~/r/neofs-node (add-linters)> golangci-lint run ./...
pkg/services/tree/replicator.go:76:3        contextcheck  Function `replicationWorker->replicationWorker$1` should pass the context parameter
pkg/network/cache/multi.go:109:16           contextcheck  Function `createForAddress->Dial->Balance->SendUnary->Init->setNeoFSAPIServer` should pass the context parameter
pkg/services/object/get/v2/util.go:438:35   contextcheck  Function `HeadObject->SendUnary->Init` should pass the context parameter
pkg/services/policer/check.go:81:44         contextcheck  Function `Inhume->Inhume$1` should pass the context parameter
pkg/services/policer/check.go:125:17        contextcheck  Function `processNodes->isClientErrMaintenance->processNodes$1` should pass the context parameter
cmd/neofs-node/config.go:903:57             contextcheck  Function `Reload->addShard->Init->init->updateObjectCounter->hasWriteCache->needRefillMetabase->GetMode->createShard->removeShards->calculateShardID` should pass the context parameter
pkg/services/object/get/get.go:81:14        contextcheck  Function `execute->analyzeStatus->executeLocal` should pass the context parameter
cmd/neofs-cli/internal/client/sdk.go:82:24  contextcheck  Function `Dial->Balance->SendUnary->Init->setNeoFSAPIServer` should pass the context parameter
pkg/services/control/server/gc.go:44:23     contextcheck  Function `Delete->Delete$1` should pass the context parameter
pkg/services/object/search/search.go:21:14  contextcheck  Function `execute->analyzeStatus` should pass the context parameter
```

roman-khimov:
Similar to edc447f530cd040ab4c2e79e06235b8271680183, we can't have any meaninful context here and current context checks are broken. Fixes #1911, the last case of

  Error in `cmd/neofs-node/config.go`: `Function `Reload->Init->init->listenEvents` should pass the context parameter`.

goes away with this.

### DIFF
--- a/pkg/local_object_storage/engine/inhume.go
+++ b/pkg/local_object_storage/engine/inhume.go
@@ -1,7 +1,6 @@
 package engine
 
 import (
-	"context"
 	"errors"
 
 	meta "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/metabase"
@@ -324,7 +323,7 @@ func (e *StorageEngine) isLocked(addr oid.Address) (bool, error) {
 	return locked, outErr
 }
 
-func (e *StorageEngine) processExpiredObjects(_ context.Context, addrs []oid.Address) {
+func (e *StorageEngine) processExpiredObjects(addrs []oid.Address) {
 	var prm InhumePrm
 	prm.MarkAsGarbage(addrs...)
 
@@ -334,17 +333,10 @@ func (e *StorageEngine) processExpiredObjects(_ context.Context, addrs []oid.Add
 	}
 }
 
-func (e *StorageEngine) processExpiredLocks(ctx context.Context, lockers []oid.Address) {
+func (e *StorageEngine) processExpiredLocks(lockers []oid.Address) {
 	e.iterateOverUnsortedShards(func(sh hashedShard) (stop bool) {
 		sh.HandleExpiredLocks(lockers)
-
-		select {
-		case <-ctx.Done():
-			e.log.Info("interrupt processing the expired locks by context")
-			return true
-		default:
-			return false
-		}
+		return false
 	})
 }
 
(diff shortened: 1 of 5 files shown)


cthulhu-rider (approved):

pkg/local_object_storage/engine/inhume.go
```diff
@@ -324,7 +323,7 @@ func (e *StorageEngine) isLocked(addr oid.Address) (bool, error) {
 	return locked, outErr
 }
 
-func (e *StorageEngine) processExpiredObjects(_ context.Context, addrs []oid.Address) {
```
carpawell, line 327: tbh, i am not sure it _should never be used there_. expired object handling _can_ be interrupted, and an expired now object will still be expired in every next epoch, so if we are about to stop working, such operation is safe to stop immediately. however, none of the current "iteration" storage engine operations supports context so ok

pkg/local_object_storage/shard/gc.go
```diff
@@ -117,16 +117,13 @@ func (gc *gc) listenEvents() {
 		v.cancelFunc()
```
carpawell, line 117: we do not need this func now then cause it can not work now. again, it is more like a disadvantage for me, cause i do not see any context restrictions here and the whole code looks like a context-based one, despite all the operations ignore contexts

roman-khimov: FS operations aren't context-bound. (local) DB operations aren't context-bound. Local object storage works with FS and DBs.

carpawell (approved):

PR #2955, opened 2024-10-01
Fix/object exist operation logs

carpawell:


pkg/local_object_storage/engine/inhume_test.go
```diff
@@ -153,4 +154,21 @@ func TestStorageEngine_Inhume(t *testing.T) {
 		_, err = wrongShard.Get(getPrm)
 		require.ErrorAs(t, err, new(apistatus.ObjectNotFound))
 	})
+
+	t.Run("inhuming object twice", func(t *testing.T) {
```
cthulhu-rider, line 158: is it related to the 404 change? i see that it helped to find the reason, but afaik it does not assert the introduced fix

pkg/local_object_storage/shard/errors.go
```diff
@@ -10,19 +10,19 @@ import (
 // IsErrNotFound checks if error returned by Shard Get/Head/GetRange method
 // corresponds to missing object.
 func IsErrNotFound(err error) bool {
-	return errors.As(err, new(apistatus.ObjectNotFound))
+	return errors.Is(err, apistatus.ErrObjectNotFound)
```
cthulhu-rider, line 13: how about unwrap them now?

the only thing that could stop me is that returning status errors is pretty weird for local layer, and it would be worth replacing them with more generic ones. These functions would reduce the amount of changes in this case

roman-khimov (approved):

pkg/local_object_storage/engine/inhume_test.go
carpawell: > is it related to the 404 change

it is the situation that led to a confusing log on our machines. since it is not an error and all the problems are only about logs, i decided to add it just to be sure it is not a problem (but it really looked like a problem if you do not know this test passes and if https://github.com/nspcc-dev/neofs-node/issues/2956 is not in your mind)

pkg/local_object_storage/shard/errors.go
carpawell: sorry, did not get you. what exactly to unwrap?

cthulhu-rider: funcs in this file. But i saw an issue (urs as i remember) which proposes to replace API errors with more generic ones. Sad i cannot find it anymore. With it, it's better to keep the funcs for now

cthulhu-rider (approved):

Issue #2947, opened 2024-09-23
Inconsistent behavior with a container session token signature


evgeniiz321:
1. Generate container session token with the owner wallet and the user wallet as a session wallet
2. Sign generated container session token with some 3rd wallet, e.g. the stranger wallet

```
[MainThread] 2024-09-23 15:17:25 [INFO] Command: ./neofs-cli --config /Users/evgeniiz/Projects/neofs-testcases-nspcc/neofs-testcases/wallet_config.yml util sign session-token --wallet '/Users/evgeniiz/Projects/neofs-testcases-nspcc/neofs-testcases/TemporaryDir-ccb1e798-597d-48f8-b286-0ae0c2b7c685/cfaa9e9e-0a06-4817-b231-92320d84ae30.json' --from '/Users/evgeniiz/Projects/neofs-testcases-nspcc/neofs-testcases/TemporaryDir-ccb1e798-597d-48f8-b286-0ae0c2b7c685/52c78400-091c-42f6-a3a7-c6db1254c6f7' --to '/Users/evgeniiz/Projects/neofs-testcases-nspcc/neofs-testcases/TemporaryDir-ccb1e798-597d-48f8-b286-0ae0c2b7c685/TestFilesDir/ede3fe1a-8eda-418b-9157-957548110d94'
Success:
return code: 0 
Output: signed session token saved in /Users/evgeniiz/Projects/neofs-testcases-nspcc/neofs-testcases/TemporaryDir-ccb1e798-597d-48f8-b286-0ae0c2b7c685/TestFilesDir/ede3fe1a-8eda-418b-9157-957548110d94
```

3. Create container using the user wallet and the session token signed by the stranger wallet
4. I expect to see an error, because container session token is created for a user wallet but signed with a stranger wallet, but no errors, container is created successfully:

```
[...]

PR #2958, opened 2024-10-02
Fix unauthorized container ops with sessions

cthulhu-rider:
CLI change is under discussion, see https://github.com/nspcc-dev/neofs-node/issues/2947#issuecomment-2388521485

### DIFF
--- a/cmd/neofs-cli/modules/util/sign_session.go
+++ b/cmd/neofs-cli/modules/util/sign_session.go
@@ -9,9 +9,16 @@ import (
 	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/common"
 	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/commonflags"
 	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/key"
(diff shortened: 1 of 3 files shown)


roman-khimov (approved):

cmd/neofs-cli/modules/util/sign_session.go
```diff
@@ -71,7 +80,15 @@ func signSessionToken(cmd *cobra.Command, _ []string) error {
...
 
-	err = stok.Sign(user.NewAutoIDSignerRFC6979(*pk))
+	forceIss, err := cmd.Flags().GetBool(forceIssuerFlag)
+	if err != nil {
+		return err
+	}
+	if forceIss {
+		err = stok.Sign(user.NewAutoIDSignerRFC6979(*pk))
+	} else {
+		err = stok.SetSignature(neofsecdsa.SignerRFC6979(*pk))
```
carpawell, line 90: check already attached issuer? warn/err if does not match/missing?

cthulhu-rider: i wouldnt add one more key<->user relation. And error is bad here to me cuz this utility should help, not limit. Dont think warning will be useful too: if other issuer is a problem - user will get his error later anyway, if not - there is nothing to worry bout

PR #2960, opened 2024-10-03
ir: remove `contracts.alphabet.amount` from config

### LINKED ISSUE #2292: Remove contracts.alphabet.amount
As discussed in #2277, `contracts.alphabet.amount` seems to be redundant, we always know the number of CNs and we always have a corresponding number of alphabet nodes (even though we may have a bigger number of IR nodes). So we can drop it.

End-rey:
Closes #2292.

I found the only place where this option is mentioned. Do I need to count the number of alphabet contracts from the config or take this number from `morph`?


carpawell (commented):
I think all the "threshold" code [below](https://github.com/nspcc-dev/neofs-node/blob/7365c7c5e6afb4f4a29fd75cfc17dfb6bb971951/pkg/innerring/contracts.go#L116-L122) is also not needed, it is only about the cases when we were limited by config but it is impossible now.
Some conflicts also.

cthulhu-rider (commented):
merge commits or ref 2nd->1st. Tag the issue in the 1st

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ attribute, which is used for container domain name in NNS contracts (#2954)
 
 ### Removed
 - Support for node.key configuration (#2959)
+- Remove `contracts.alphabet.amount` from inner ring config (#2960)
```
roman-khimov, line 25: `Remove` doesn't make much sense in `Removed` section.

roman-khimov (approved):

PR #2963, opened 2024-10-07
node, ir: set default dial timeout to 1 minute

### LINKED ISSUE #2891: default timeouts are extremely aggressive
1. connection timeouts less than 10s are extremely aggressive and leads to connection breaks
2. connection timeouts should be the same across all services
3. i'd recommend to set all network timeouts to 1 minute

End-rey:
Closes #2891.

Question about these timeouts:
- in node:
   - `apiclient.stream_timeout`
   - `policer.head_timeout`
- in inner ring:
   - `audit.timeout.*`

Should I do the same for them?


roman-khimov (commented):
Stream timeout should be OK to raise. Policer is a bit more tricky, some 10-15s is OK for it, but 1m seems to be too much. Audit better be left as is.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ attribute, which is used for container domain name in NNS contracts (#2954)
 - Metabase graveyard scheme (#2929)
 - When an error is returned, no additional help output is displayed in cobra-based programs (#2942)
 - Use org-wide linter (#2943)
+- Default dial timeout to one minute (#2963)
```
carpawell, line 24: can we also have it changeloged in `Updating from` section?

carpawell (commented):
> Stream timeout should be OK to raise

@roman-khimov the only problem I know about such timeouts raising is bad failover performance: once a node is down too many routines are waiting for too long before they think the node is dead.

Head/audit timeouts are also ok to me to be kept

roman-khimov:
> raising is bad failover performance

Sure. Can be adjusted for fast networks. Mainnet/testnet are likely to be OK with 1m.

CHANGELOG.md
```diff
@@ -40,6 +41,11 @@ introduced in version 0.22.3 and support for binary keys was removed from
 other components in 0.33.0 and 0.37.0. Please migrate to wallets (see 0.37.0
 notes) if you've not done it previously.
 
+Increace default timeout for dialing connections in node: `morph.dial_timeout`,
```
carpawell, line 44:
```suggestion
Increase default timeout for dialing connections in node: `morph.dial_timeout`, 
```

carpawell (approved):
Otherwise ok

roman-khimov (approved):

PR #2966, opened 2024-10-09
metabase: save last epoch when metabase was resynchronized

### LINKED ISSUE #2887: Store time at which metabase was resynced
I'm always frustrated when we miss some details about meta when it [behaves badly](https://github.com/nspcc-dev/neofs-node/issues/2844).

Store some timestamp that indicates the last metabase resynchronization. Do not think absolute time is a good idea, so epoch can be a better option.

(:

## Additional context
Every time we debug metabase is an "additional context".

End-rey:
Closes #2887.

### DIFF
--- a/pkg/local_object_storage/metabase/control.go
+++ b/pkg/local_object_storage/metabase/control.go
@@ -133,26 +133,36 @@ func (db *DB) init(reset bool) error {
 			}
 		}
 
-		if !reset {
+		if reset {
+			err = tx.ForEach(func(name []byte, b *bbolt.Bucket) error {
+				if _, ok := mStaticBuckets[string(name)]; !ok {
+					return tx.DeleteBucket(name)
+				}
+
+				return nil
+			})
+			if err != nil {
+				return err
+			}
+
+			err = updateLastResyncEpoch(tx, db.epochState.CurrentEpoch())
+			if err != nil {
+				return err
+			}
+			err = updateVersion(tx, version)
+			if err != nil {
+				return err
+			}
+		} else {
 			err = syncCounter(tx, false)
 			if err != nil {
 				return fmt.Errorf("could not sync object counter: %w", err)
 			}
-
-			return nil
 		}
 
-		err = tx.ForEach(func(name []byte, b *bbolt.Bucket) error {
-			if _, ok := mStaticBuckets[string(name)]; !ok {
-				return tx.DeleteBucket(name)
-			}
+		db.info.LastResyncEpoch = getLastResyncEpoch(tx)
 
-			return nil
-		})
-		if err != nil {
-			return err
-		}
-		return updateVersion(tx, version)
+		return nil
 	})
 }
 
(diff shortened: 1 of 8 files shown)


pkg/local_object_storage/metabase/db.go
```diff
@@ -290,6 +292,14 @@ func (db *DB) SetLogger(l *zap.Logger) {
 	db.log = l
 }
 
+// SaveLastResyncEpoch saves the current epoch as the last one
+// when metabase was re-synchronized, to DB and to Info.
+func (db *DB) SaveLastResyncEpoch() {
```
roman-khimov, line 297: Why a separate method, why not a part of `Reset()`?

roman-khimov (commented):
I think we also need some lens command that will show it, otherwise what's the point. And this command should be aware of different mb versions.

pkg/local_object_storage/metabase/db.go
```diff
@@ -290,6 +292,14 @@ func (db *DB) SetLogger(l *zap.Logger) {
...
+	currEpoch := db.epochState.CurrentEpoch()
+	db.lastResyncEpoch = currEpoch
+	db.info.LastResyncEpoch = currEpoch
```
carpawell, line 300: how can we use it? every time we restart application this field is lost. i planned to extend metabase storage with a number that is filled every reset and is read every start

carpawell (changes requested):

carpawell:
> I think we also need some lens command that will show it

But there is nothing it will show us in the current PR...

End-rey:
Now, the last epoch when the metabase was resynchronized is saved at db in the `shardInfoBucket` with the key `last_resync_epoch`, when metabase is `init` with `reset == true` flag and it is read to `info` field every `init`.

Also, make `lens` command, but don't understand what does it means.
> this command should be aware of different mb versions.

pkg/local_object_storage/metabase/control.go
```diff
@@ -138,21 +138,31 @@ func (db *DB) init(reset bool) error {
 			if err != nil {
 				return fmt.Errorf("could not sync object counter: %w", err)
 			}
+		} else {
```
carpawell, line 141: just a little suggestion: if now we have both branches (if-else) i would prefer not to use negation (`!reset`), i think it would be read simpler

pkg/local_object_storage/metabase/last_resync_epoch_test.go
```diff
@@ -0,0 +1,55 @@
...
+		require.NoError(t, err)
+		require.Equal(t, epoch, gotEpoch)
+		require.Equal(t, epoch, db.DumpInfo().LastResyncEpoch)
+	}
+
+	// Clean db without resynchronizations, so it is 0.
+	checkEpoch(t, 0)
+
+	// After Reset, last resync epoch == current epoch.
+	require.NoError(t, db.Reset())
```
carpawell, line 40: can we add a single non-zero test case, please?
1. empty base
2. empty to non-empty
3. non-empty to non-empty
4. non-empty reload with no changes

i gues that is the full list of use cases and we are ready to be confident

carpawell (approved):

roman-khimov (approved):
Migrated DBs will just return 0, right?

PR #2968, opened 2024-10-17
Revive dead object via control service

### LINKED ISSUE #1450: Allow to revive dead object via control service
1. Check if the object was removed (probably via `neofs-lens`)
2. Allow to forcefully revive the object via `neofs-cli control` service.

cc @cthulhu-rider

End-rey:
Closes #1450.

### DIFF
--- a/pkg/local_object_storage/engine/revive.go
+++ b/pkg/local_object_storage/engine/revive.go
@@ -1,0 +1,41 @@
+package engine
+
+import (
+	meta "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/metabase"
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	"go.uber.org/zap"
+)
+
+// ReviveShardStatus contains the Status of the object's revival in the Shard and Shard ID.
+type ReviveShardStatus struct {
+	ID     string
+	Status meta.ReviveStatus
+}
+
+// ReviveStatus represents the status of the object's revival in the StorageEngine.
+type ReviveStatus struct {
+	Shards []ReviveShardStatus
+}
+
+// ReviveObject forcefully revives object by oid.Address in the StorageEngine.
+// Iterate over all shards despite errors and purge all removal marks from all metabases.
+func (e *StorageEngine) ReviveObject(address oid.Address) (res ReviveStatus, err error) {
+	e.iterateOverUnsortedShards(func(sh hashedShard) (stop bool) {
+		reviveStatus, err := sh.ReviveObject(address)
+		id := *sh.ID()
+		res.Shards = append(res.Shards, ReviveShardStatus{
+			ID:     id.String(),
+			Status: reviveStatus,
+		})
+		if err != nil {
+			e.log.Warn("failed to revive object in shard",
+				zap.String("shard", id.String()),
+				zap.String("address", address.EncodeToString()),
+				zap.Error(err),
+			)
+		}
+
+		return false
+	})
+	return
+}
--- a/pkg/local_object_storage/metabase/revive.go
+++ b/pkg/local_object_storage/metabase/revive.go
@@ -1,0 +1,147 @@
+package meta
+
+import (
+	"fmt"
+
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/util/logicerr"
+	"github.com/nspcc-dev/neofs-sdk-go/object"
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	"go.etcd.io/bbolt"
+)
+
+// ErrObjectWasNotRemoved is returned when object neither in the graveyard nor was marked with GC mark.
+var ErrObjectWasNotRemoved = logicerr.New("object neither in the graveyard nor was marked with GC mark")
+
+// ErrReviveFromContainerGarbage is returned when the object is in the container that marked with GC mark.
+var ErrReviveFromContainerGarbage = logicerr.New("revive from container marked with GC mark")
+
+type reviveStatusType int
+
+const (
+	// ReviveStatusGraveyard is the type of revival status of an object from a graveyard.
+	ReviveStatusGraveyard reviveStatusType = iota
+	// ReviveStatusGarbage is the type of revival status of an object from the garbage bucket.
+	ReviveStatusGarbage
+	// ReviveStatusError is the type of status when an error occurs during revive.
+	ReviveStatusError
+)
+
+// ReviveStatus groups the resulting values of ReviveObject operation.
+// Contains the type of revival status and message for details.
+type ReviveStatus struct {
+	statusType reviveStatusType
+	message    string
+}
+
+// Message returns message of status.
+func (s *ReviveStatus) Message() string {
+	return s.message
+}
+
+// StatusType returns the type of revival status.
+func (s *ReviveStatus) StatusType() reviveStatusType {
+	return s.statusType
+}
+
+func (s *ReviveStatus) setStatusGraveyard(tomb string) {
+	s.statusType = ReviveStatusGraveyard
+	s.message = fmt.Sprintf("successful revival from graveyard, tomb: %s", tomb)
+}
+
+func (s *ReviveStatus) setStatusGarbage() {
+	s.statusType = ReviveStatusGarbage
+	s.message = "successful revival from garbage bucket"
+}
+
+func (s *ReviveStatus) setStatusError(err error) {
+	s.statusType = ReviveStatusError
+	s.message = fmt.Sprintf("didn't revive, err: %v", err)
+}
+
+// ReviveObject revives object by oid.Address. Removes GCMark/Tombstone records in the corresponding buckets
+// and restore metrics.
+func (db *DB) ReviveObject(addr oid.Address) (res ReviveStatus, err error) {
+	db.modeMtx.RLock()
+	defer db.modeMtx.RUnlock()
+
+	if db.mode.ReadOnly() {
+		res.setStatusError(ErrReadOnlyMode)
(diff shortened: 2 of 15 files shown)


pkg/local_object_storage/shard/mode/mode.go
```diff
@@ -48,12 +48,12 @@ func (m Mode) String() string {
...
 }
 
-// NoMetabase returns true iff m is operating without the metabase.
+// NoMetabase returns true if m is operating without the metabase.
 func (m Mode) NoMetabase() bool {
 	return m&Degraded != 0
 }
 
-// ReadOnly returns true iff m prohibits modifying operations with shard.
+// ReadOnly returns true if m prohibits modifying operations with shard.
```
carpawell, line 56: "iff" usually means "if and only if", do not think it is a mistake here

pkg/local_object_storage/engine/revive.go
```diff
@@ -0,0 +1,39 @@
...
+	e.iterateOverUnsortedShards(func(sh hashedShard) (stop bool) {
+		reviveStatus, err := sh.ReviveObject(address)
+		id := *sh.ID()
+		res.Shards = append(res.Shards, ReviveShardStatus{
+			ID:     id.String(),
+			Status: reviveStatus,
+			Error:  err,
+		})
+		if err != nil {
+			e.log.Debug("failed to revive object in shard", zap.String("shard", id.String()), zap.Error(err))
```
carpawell, line 33: maybe WARN? but not sure

carpawell, line 33: can we attach the object address to simplify problem debugging?

pkg/local_object_storage/metabase/revive.go
```diff
@@ -0,0 +1,155 @@
...
+
+// ErrReviveFromContainerGarbage is returned when the object is in the container that marked with GC mark.
+var ErrReviveFromContainerGarbage = logicerr.New("revive from container marked with GC")
+
+type reviveStatusType int
+
+const (
+	// ReviveStatusGraveyard is the type of revival status of an object from a graveyard.
+	ReviveStatusGraveyard reviveStatusType = iota
+	// ReviveStatusGC is the type of revival status of an object from GC.
```
carpawell, line 23: "from GC" confuses me a little. "from the garbage bucket"? of explicitly say that GC mark was removed

pkg/local_object_storage/metabase/revive.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+func (s *ReviveStatus) setStatusGC() {
+	s.statusType = ReviveStatusGC
+	s.message = "successful revive object with GC mark"
+}
+
+func (s *ReviveStatus) setStatusError(err error) {
+	s.statusType = ReviveStatusError
+	s.message = fmt.Sprintf("don't revive, err: %v", err)
```
carpawell, line 58: "didn't"?

pkg/local_object_storage/metabase/revive.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+func (s *ReviveStatus) setStatusGraveyard(tomb string) {
+	s.statusType = ReviveStatusGraveyard
+	s.message = fmt.Sprintf("successful revival from graveyard, tomb: %s", tomb)
+}
+
+func (s *ReviveStatus) setStatusGC() {
+	s.statusType = ReviveStatusGC
+	s.message = "successful revive object with GC mark"
```
carpawell, line 53: revival/revive difference in messages in the same context

pkg/local_object_storage/metabase/revive.go
```diff
@@ -0,0 +1,155 @@
...
+		targetKey := addressKey(addr, buf)
+
+		if graveyardBKT == nil {
+			// incorrect metabase state, does not make
+			// sense to check garbage bucket
+			return ErrObjectWasNotRemoved
+		}
+
+		val := graveyardBKT.Get(targetKey)
+		if val != nil {
```
carpawell, line 93: if an object is in graveyard, do we skip garbage bucket then?

End-rey: After removal from the graveyard, the next operation I remove from the garbage bucket. But I noticed a duplicate code, so I've refractored it a little now, now after all the checks, anyway it removes it from the garbage bucket.

carpawell (approved):

pkg/services/control/server/object_revive.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	// check availability
+	err = s.ready()
+	if err != nil {
+		return nil, err
+	}
+
+	var addr oid.Address
+	err = addr.DecodeString(request.GetBody().GetObjectAddress())
```
roman-khimov, line 25: Why a string representation here? I'd not expect this in a protobuf protocol, you can avoid decoding.

pkg/local_object_storage/engine/revive.go
```diff
@@ -0,0 +1,43 @@
...
+	meta "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/metabase"
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	"go.uber.org/zap"
+)
+
+// ReviveShardStatus contains the Status of the object's revival in the Shard, Shard ID and Error.
+type ReviveShardStatus struct {
+	ID     string
+	Status meta.ReviveStatus
+	Error  error
```
roman-khimov, line 13: Error here...

pkg/local_object_storage/engine/revive.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// ReviveStatus represents the status of the object's revival in the StorageEngine.
+type ReviveStatus struct {
+	Shards []ReviveShardStatus
+}
+
+// ReviveObject forcefully revives object by oid.Address in the StorageEngine.
+// Iterate over all shards despite errors and purge all removal marks from all metabases.
+func (e *StorageEngine) ReviveObject(address oid.Address) (res ReviveStatus, err error) {
```
roman-khimov, line 23: ... error there. Where is the error, really?

pkg/services/control/server/object_revive.go
End-rey: I don't really understand. Do I need to transmit via bytes? I did the same as the [status](https://github.com/nspcc-dev/neofs-node/blob/5c5c42c1c275eb40c33fdd80869b10f07c1d7779/pkg/services/control/server/object_status.go#L27).

pkg/local_object_storage/engine/revive.go
End-rey: I thought it could be used somewhere, but in the end I didn't use it. Therefore, I removed it from `ReviveShardStatus`.

pkg/services/control/server/object_revive.go
roman-khimov: Whoa. Really unexpected. Well, it's a good reference, but I'd say it's a bug of `status` implementation. We can either divert here and make it better (it simplifies things) or follow established practice. Not critical then.

carpawell: i thought i was expired by some another example for `status` but i do not find any string implementation (it is strange, cause i am for raw bytes too) so let it be raw bytes then cause we also have it in control API: https://github.com/nspcc-dev/neofs-node/blob/5c5c42c1c275eb40c33fdd80869b10f07c1d7779/pkg/services/control/service.proto#L117-L119

status can be fixed separately

End-rey: Can you please explain where I should use `bytes` and where `string`?
For example, here:
https://github.com/nspcc-dev/neofs-node/blob/1c786839073c7e4a08ab753d3d6a5e98d8d8be7c/pkg/services/control/service.proto#L454-L456
I see that `shard_id` can also be `bytes`, as shown here:
https://github.com/nspcc-dev/neofs-node/blob/f858a81cbd1c9116042c0e92adad72405c8741d7/pkg/services/control/types.proto#L123

carpawell: i think shard ID is ok to be `string` cause there is no fixed public rule for bytes -> string conversion (for addresses we know it), but for things from our public API it is ok to be bytes and cli can convert is inside according to SDK
@roman-khimov

roman-khimov: Shard ID can be a string, it's deeply internal to the node, no one knows how to convert it properly into something human-oriented.

pkg/services/control/server/object_revive.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	// check availability
+	err = s.ready()
+	if err != nil {
+		return nil, err
+	}
+
+	var addr oid.Address
+	err = addr.DecodeString(string(request.GetBody().GetObjectAddress()))
```
roman-khimov, line 25: Wait, but the idea was exactly to avoid `DecodeString`.

End-rey: But I don't really understand how to avoid it. I did it like [drop object](https://github.com/nspcc-dev/neofs-node/blob/77404cc7f103c82e3ebcf93e8dee61b8270e523d/pkg/services/control/server/gc.go#L36), that @carpawell provided as example:
https://github.com/nspcc-dev/neofs-node/blob/5567e5fef477bbf9085124a6fb5e44a32fa42e35/pkg/services/control/service.proto#L123-L125
And I can't find a method to convert bytes to an address.

roman-khimov: Looks like our control API is broken all over the place wrt this problem, so we can just follow the way it currently works and solve the problem for all requests in #2980.

roman-khimov (approved):

PR #2977, opened 2024-10-22
Fix/metis2 dead

carpawell:


pkg/local_object_storage/writecache/flush.go
```diff
@@ -270,9 +270,27 @@ func (c *cache) flushObject(obj *object.Object, data []byte) error {
...
+				err = fmt.Errorf("trying to restore missing object in metabase: %w", err)
+			}
+
+			return err
+		}
+
 		c.reportFlushError("can't update object storage ID",
 			addr.EncodeToString(), err)
 	}
+
```
roman-khimov, line 293: Ouch.

pkg/local_object_storage/writecache/flush.go
```diff
@@ -270,9 +270,27 @@ func (c *cache) flushObject(obj *object.Object, data []byte) error {
 
 	_, err = c.metabase.UpdateStorageID(updPrm)
 	if err != nil {
+		if errors.Is(err, apistatus.ErrObjectNotFound) {
+			// we have the object and we just successfully put it so all the
+			// information for restoring the object is here; meta can be
+			// currupted, resynced, etc, just trying our best
```
roman-khimov, line 276: corrupted

pkg/local_object_storage/writecache/flush.go
```diff
@@ -270,9 +270,27 @@ func (c *cache) flushObject(obj *object.Object, data []byte) error {
...
+			var prmMeta meta.PutPrm
+			prmMeta.SetObject(obj)
+			prmMeta.SetStorageID(res.StorageID)
+
+			_, err = c.metabase.Put(prmMeta)
+			if err != nil {
+				err = fmt.Errorf("trying to restore missing object in metabase: %w", err)
+			}
+
+			return err
```
roman-khimov, line 287: Still makes me wonder _how_ this can happen. Some log is also good to have (failed to update but restored).

pkg/local_object_storage/shard/control.go
```diff
@@ -161,82 +161,93 @@ func (s *Shard) resyncMetabase() error {
...
-	err = blobstor.IterateBinaryObjects(s.blobStor, func(addr oid.Address, data []byte, descriptor []byte) error {
-		if err := obj.Unmarshal(data); err != nil {
-			s.log.Warn("could not unmarshal object",
-				zap.Stringer("address", addr),
-				zap.String("err", err.Error()))
-			return nil
+	if s.writeCache != nil {
+		// ensure there will not be any raсes in write-cache -> blobstor object
+		// backgroung flushing while iterating blobstor
+		err = s.writeCache.Flush(true)
```
roman-khimov, line 167: We're asking for a metabase resync and getting a cache flush. It's somewhat unobvious. Can we have caches/stores RO and perform resync as is?

pkg/local_object_storage/writecache/flush.go
carpawell: any metabase resync with objects that are in WC but not in blobstor yet will show such behavior; also objects PUT in degraded mode will have the same problem. anyway, suggested code is correct: we want to update storage ID but we do not have object, so why not put it to metabase?

CHANGELOG.md
```diff
@@ -67,6 +67,7 @@ notes) if you've not done it previously.
 - `ObjectService`'s `Search` and `Replicate` RPC handlers cache up to 1000 lists of container nodes (#2892)
 - Default max_traceable_blocks Morph setting lowered to 17280 from 2102400 (#2897)
 - `ObjectService`'s `Get`/`Head`/`GetRange` RPC handlers cache up to 10K lists of per-object sorted container nodes (#2896)
+- Pprof and metrics services stop at the end of SN's application lifecycle (#2976)
```
roman-khimov, line 71: resync behavior change is not listed

roman-khimov (approved):

Issue #1950, opened 2022-10-21
Need to make process of verify node config more intelligent


anikeev-yadro:
I have case with mistake in config, which lead to part of config was not used.
Config with mistake: 
```
storage:
    shard:
        0:
            blobstor:
            -   path: /srv/neofs/data0/blobovnicza
                type: blobovnicza
            -   path: /srv/neofs/data0
                type: fstree
            default:
                blobstor:
                -   depth: 1
                    opened_cache_capacity: 32
                    perm: '0644'
                    size: 512mb
                    width: 4
                -   depth: 4
                    perm: '0644'
                compress: true
                compression_exclude_content_types:
                - audio/*
                - video/*
                gc:
                    remover_batch_size: 1000
                    remover_sleep_interval: 10s
                metabase:
                    max_batch_delay: 20ms
                    max_batch_size: 1000
                    perm: '0644'
                pilorama:
                    max_batch_delay: 5ms
                    max_batch_size: 500
                    no_sync: false
                resync_metabase: true
                small_object_size: 128 kb
                writecache:
                    capacity: 3gb
                    enabled: true
                    max_object_size: 8mb
                    memcache_capacity: 1gb
                    small_object_size: 128kb
                    workers_number: 30
            metabase:
                path: /srv/neofs/meta0/metabase0.db
            pilorama:
                path: /srv/neofs/meta0/pilorama0.db
            writecache:
                path: /srv/neofs/meta0/write_cache0
```
In this case ```default:``` section was in the ```shard.0``` section. But it should be at the same level. This mistake have lead to ``default``` section was ignored by node.

We should make the process of verify node config more intelligent.

roman-khimov:
Looks a lot like https://github.com/nspcc-dev/neofs-node/issues/2663.

PR #2981, opened 2024-10-25
Reject configuration with unknown fields

### LINKED ISSUE #2663: Reject configuration with unknown fields
https://github.com/nspcc-dev/neo-go/issues/3128 is very cool and saved us time several times already, you immediately see obsolete/wrong configuration and can fix it easily (instead of waiting for the service misbehavior). I want this for IR and SN, but I'm not sure it's gonna be easy.

End-rey:
Closes #2663, #1950.

The problems I have encountered:
- node config: `grpc` section with list of structures and `node` section with unlimited number of attributes.
- `.env` files are not being processed, because of the work of the `viper` package. Should I do it? It might be more difficult.

### DIFF
--- a/cmd/internal/configvalidator/validate.go
+++ b/cmd/internal/configvalidator/validate.go
@@ -1,0 +1,106 @@
+package configvalidator
+
+import (
+	"errors"
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+// ErrUnknowField returns when an unknown field appears in the config.
+var ErrUnknowField = errors.New("unknown field")
+
+// CheckForUnknownFields validates the config struct for unknown fields with the config map.
+// If the field in the config is a map of a key and some value, need to use `mapstructure:,remain` and
+// `prefix` tag with the key prefix, even it is empty string.
+func CheckForUnknownFields(configMap map[string]any, config any) error {
+	return checkForUnknownFields(configMap, config, "")
+}
+
+func checkForUnknownFields(configMap map[string]any, config any, currentPath string) error {
+	expectedFields := getFieldsFromStruct(config)
+
+	for key, val := range configMap {
+		fullPath := key
+		if currentPath != "" {
+			fullPath = currentPath + "." + key
+		}
+
+		_, other := expectedFields[",remain"]
+		_, exists := expectedFields[key]
+		if !exists && !other {
+			return fmt.Errorf("%w: %s", ErrUnknowField, fullPath)
+		}
+
+		var fieldVal reflect.Value
+		if other && !exists {
+			fieldVal = reflect.ValueOf(config).
+				FieldByName(getStructFieldByTag(config, "prefix", key, strings.HasPrefix))
+		} else {
+			fieldVal = reflect.ValueOf(config).
+				FieldByName(getStructFieldByTag(config, "mapstructure", key, func(a, b string) bool {
+					return a == b
+				}))
+		}
+
+		switch fieldVal.Kind() {
+		case reflect.Slice:
+			if fieldVal.Len() == 0 {
+				return nil
+			}
+			fieldVal = fieldVal.Index(0)
+		case reflect.Map:
+			fieldVal = fieldVal.MapIndex(reflect.ValueOf(key))
+		default:
+		}
+
+		if !fieldVal.IsValid() {
+			return fmt.Errorf("%w: %s", ErrUnknowField, fullPath)
+		}
+
+		nestedMap, okMap := val.(map[string]any)
+		nestedSlice, okSlice := val.([]any)
+		if (okMap || okSlice) && fieldVal.Kind() == reflect.Struct {
+			if okSlice {
+				for _, slice := range nestedSlice {
+					if nestedMap, okMap = slice.(map[string]any); okMap {
+						if err := checkForUnknownFields(nestedMap, fieldVal.Interface(), fullPath); err != nil {
+							return err
+						}
+					}
+				}
+			} else if err := checkForUnknownFields(nestedMap, fieldVal.Interface(), fullPath); err != nil {
+				return err
+			}
+		} else if okMap != (fieldVal.Kind() == reflect.Struct) {
+			return fmt.Errorf("%w: %s", ErrUnknowField, fullPath)
+		}
+	}
+
+	return nil
+}
+
+func getFieldsFromStruct(config any) map[string]struct{} {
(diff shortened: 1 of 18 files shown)


End-rey:
There was a problem that the default value of the list was a slice with a length of 0, so I added a check for this. And now my devenv works.
But it seems like automated system tests didn't pass because the `ir` or `node` didn't run well. How can I find out what the error was that caused them to fail? Maybe there is a bad configuration?

cmd/internal/configvalidator/validate.go
```diff
@@ -0,0 +1,102 @@
...
+	"errors"
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+// ErrUnknowField returns when an unknown field appears in the config.
+var ErrUnknowField = errors.New("unknown field")
+
+// CheckForUnknownFields validate the config struct for unknown fields with the config map.
```
roman-khimov, line 13: validates

cmd/neofs-ir/internal/validate/validate_test.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"path/filepath"
+	"testing"
+
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCheckForUnknownFieldsExample(t *testing.T) {
```
roman-khimov, line 11: Negative test scenarios?

cmd/neofs-node/config/config.go
```diff
@@ -54,6 +55,13 @@ func New(_ Prm, opts ...Option) *Config {
 		}
 	}
 
+	if o.validate {
```
roman-khimov, line 58: Why conditional?

cmd/neofs-ir/internal/validate/config.go
```diff
@@ -0,0 +1,213 @@
+package validate
+
+import "time"
+
+type validConfig struct {
+	Logger struct {
+		Level string `mapstructure:"level"`
```
roman-khimov, line 7: Can we just name it `level` and not have any `mapstructure` tags?

cmd/neofs-ir/defaults.go
```diff
@@ -60,7 +69,6 @@ func defaultConfiguration(cfg *viper.Viper) {
 	cfg.SetDefault("wallet.address", "")  // account address
 	cfg.SetDefault("wallet.password", "") // password
 
-	cfg.SetDefault("timers.emit", "0")
```
roman-khimov, line 63: There has to be some test for default configuration only.

cmd/internal/configvalidator/validate.go
```diff
@@ -0,0 +1,102 @@
...
+// If the field in the config is a map of a key and some value, need to use `mapstructure:,remain` and
+// `prefix` tag with the key prefix, even it is empty string.
+func CheckForUnknownFields(configMap map[string]any, config any, currentPath string) error {
```
roman-khimov, line 16: You better not expose `currentPath` here, it's an implementation detail.

cmd/internal/configvalidator/validate.go
```diff
@@ -0,0 +1,102 @@
...
+		if jsonTag := field.Tag.Get("mapstructure"); jsonTag != "" {
+			fields[jsonTag] = struct{}{}
+		} else {
+			fields[field.Name] = struct{}{}
+		}
+	}
+	return fields
+}
+
+func getStructFieldByTagComparation(config any, tagKey, tagValue string, comparation func(a, b string) bool) string {
```
roman-khimov, line 93: `getStructFieldByTag` is sufficient and it's `comparison`.

cmd/internal/configvalidator/validate.go
```diff
@@ -0,0 +1,102 @@
...
+// ErrUnknowField returns when an unknown field appears in the config.
+var ErrUnknowField = errors.New("unknown field")
+
+// CheckForUnknownFields validate the config struct for unknown fields with the config map.
+// If the field in the config is a map of a key and some value, need to use `mapstructure:,remain` and
+// `prefix` tag with the key prefix, even it is empty string.
+func CheckForUnknownFields(configMap map[string]any, config any, currentPath string) error {
+	expectedFields := getFieldsFromStruct(config)
+
+	for key, val := range configMap {
```
roman-khimov, line 19: Seems to be _very_ inefficient since you traverse the same items for every key. Maybe it's not a problem given the size of our configuration files, but I'd try flattening (into a slice) and sorting `configMap` which allows to walk by prefixes more effectively then. Still, what's the time we spend here for example config?

cmd/neofs-ir/internal/validate/config.go
```diff
@@ -0,0 +1,213 @@
+package validate
+
+import "time"
+
+type validConfig struct {
```
roman-khimov, line 5: What bugs me is that we have this structure used just for validation while in fact we also have similar structures in various subcomponents. Can those be reused here to avoid duplication? I mean we can _always_ parse into this config and then operate with components (just like NeoGo does).

roman-khimov (commented):
Can this be solved with https://pkg.go.dev/github.com/spf13/viper#UnmarshalExact?

cmd/neofs-ir/internal/validate/config.go
End-rey: We can't because the fields of the structure must be exported for `Unmarshal`.

End-rey:
> Can this be solved with https://pkg.go.dev/github.com/spf13/viper#UnmarshalExact?

Partially it can, but there are such fields in the config, for example, `node.attribute_*` and there can be as many as you like, so the tag `,remain` is used, because of which any incorrect fields will fall into this field and then validation will be successful, although in fact there will be an error. I tried this method and it didn't pass my tests.

cmd/internal/configvalidator/validate.go
End-rey: For `ir` example config, the execution time of this function is 192 microseconds, for `node` - 197 microseconds. 
It's going to be a little difficult considering `,remain` case and not so necessary now.

cmd/neofs-node/config/config.go
End-rey: It's made for `call_test` and `cast_test` tests that check the correct parsing of values, and there is separate configuration for tests.

roman-khimov (approved):

PR #2983, opened 2024-10-28
docs: add docs about all cli commands

### LINKED ISSUE #1413: Add README.md files to cli commands
After #1074 is finished we can generate documentation with `neofs-cli gendoc` and move generated files to proper folders.
It would be nice if visiting the package in github displayed the README.md file for the command.

Related #1396.

End-rey:
Generate docs with `neofs-cli gendoc ./docs/cli-commands`.

Closes #1413.

So, if we add a new command, should we generate docs now? Should we also add some checks so that we don't forget to do this?

### DIFF
--- a/.github/workflows/cli-docs-update.yml
+++ b/.github/workflows/cli-docs-update.yml
@@ -1,0 +1,24 @@
+name: CLI commands docs check
+
+on:
+  pull_request:
+    branches:
+      - master
+      - support/**
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: cli-docs-check
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Generate new CLI doc files
+        run: make cli-gendoc
+
+      - name: Fail if CLI commands files are changed but the documentation is not updated
+        run: |
+          git add -N docs/cli-commands
+          git diff --exit-code -- docs/cli-commands/
--- a/docs/cli-commands/neofs-cli.md
+++ b/docs/cli-commands/neofs-cli.md
@@ -1,0 +1,42 @@
+## neofs-cli
+
+Command Line Tool to work with NeoFS
+
+### Synopsis
+
+NeoFS CLI provides all basic interactions with NeoFS and it's services.
+
+It contains commands for interaction with NeoFS nodes using different versions
+of neofs-api and some useful utilities for compiling ACL rules from JSON
+notation, managing container access through protocol gates, querying network map
+and much more!
+
+```
+neofs-cli [flags]
+```
+
+### Options
+
+```
+  -c, --config string   Config file (default is $HOME/.config/neofs-cli/config.yaml)
+  -h, --help            help for neofs-cli
+  -v, --verbose         Verbose output
+      --version         Application version and NeoFS API compatibility
+```
+
+### SEE ALSO
+
+* [neofs-cli accounting](neofs-cli_accounting.md)	 - Operations with accounts and balances
+* [neofs-cli acl](neofs-cli_acl.md)	 - Operations with Access Control Lists
+* [neofs-cli bearer](neofs-cli_bearer.md)	 - Operations with bearer token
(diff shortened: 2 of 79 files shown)


carpawell:
> Should we also add some checks so that we don't forget to do this?

I think it will be useful, the same way we check CHANGELOG and configuration files. I am sure I will forget regenerating docs.

.github/workflows/cli-docs-update.yml
```diff
@@ -0,0 +1,36 @@
...
+            cmd/neofs-cli/**
+
+      - name: Get changed CLI doc files
+        id: docs-diff
+        uses: tj-actions/changed-files@v42
+        with:
+          files: docs/cli-commands/**
+
+      - name: Fail if CLI files are changed but the documentation is not updated
+        if: steps.cli-diff.outputs.any_changed == 'true' && steps.docs-diff.outputs.any_changed == 'false'
```
roman-khimov, line 32: I can change a lot of things in `cmd/neofs-cli` that won't affect documentation in any way. Better regenerate (and maybe even have some `Makefile` target for this) and check for diffs.

docs/cli-commands/neofs-cli.md
```diff
@@ -0,0 +1,43 @@
...
+* [neofs-cli control](neofs-cli_control.md)	 - Operations with storage node
+* [neofs-cli gendoc](neofs-cli_gendoc.md)	 - Generate documentation for this command
+* [neofs-cli netmap](neofs-cli_netmap.md)	 - Operations with Network Map
+* [neofs-cli object](neofs-cli_object.md)	 - Operations with Objects
+* [neofs-cli session](neofs-cli_session.md)	 - Operations with session token
+* [neofs-cli storagegroup](neofs-cli_storagegroup.md)	 - Operations with Storage Groups
+* [neofs-cli tree](neofs-cli_tree.md)	 - Operations with the Tree service
+* [neofs-cli util](neofs-cli_util.md)	 - Utility operations
+
+###### Auto generated by spf13/cobra on 28-Oct-2024
```
roman-khimov, line 43: Although this one will trigger every day, can it be removed (at least the date, nothing against `spf13/cobra`, but date is problematic)?

pkg/util/gendoc/gendoc.go
```diff
@@ -14,7 +14,8 @@ import (
 )
 
 const (
-	gendocTypeFlag = "type"
+	gendocDisableAutoGenTagFlag = "disable-auto-gen-tag"
```
carpawell, line 17: not sure we need it, but dont mind, mb @roman-khimov has some thoughts

.github/workflows/cli-docs-update.yml
```diff
@@ -0,0 +1,33 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: cli-docs-check
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Save current CLI docs files
```
carpawell, line 18: can we just try to generate docs and see if `git diff`/\*smth else\* says something new was done?

pkg/util/gendoc/gendoc.go
roman-khimov: I think we do because otherwise we can't have consistent output each time we run the generator.

roman-khimov (approved):

pkg/util/gendoc/gendoc.go
carpawell: i meant if we need it to be possible to change via separate command flag

.github/workflows/cli-docs-update.yml
```diff
@@ -0,0 +1,24 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: Generate new CLI doc files
+        run: make cli-gendoc
+
+      - name: Fail if CLI commands files are changed but the documentation is not updated
+        run: |
+          git add -N docs/cli-commands
+          git diff --exit-code -- docs/cli-commands/
```
carpawell, line 24: can you add new line at the end of file, please?

carpawell (approved):

PR #2984, opened 2024-10-28
cli: add a default range for `object hash` to cli `help`

### LINKED ISSUE #1693: There is some default range for `neofs-cli object hash`
When I run `neofs-cli object hash` with no range specified, I get some hash
```
$ neofs-cli object hash --rpc-endpoint s01.neofs.devenv:8080 --wallet wallets/wallet.json --cid <cid> --oid <oid>
Enter password > 
0d29545d312dfa77ffe7a94415955db74cc8fbdb9a1ca5b6d54342efd2b2de31
```
while it's not clear from the help message what is the range
```
...
--range string          Range to take hash from in the form offset1:length1,...
...
```
## Expected Behavior
I'd expect either the exact reasonable default mentioned in the help message OR the `--range` key to be obligatory.
By the way, in the `neofs-cli object range` the `--range` key is obligatory
```
$ neofs-cli object range --rpc-endpoint s01.neofs.devenv:8080 --wallet wallets/wallet.json --cid <cid> --oid <oid>
exactly one range must be specified, got: 0
$ echo $?
1
```

End-rey:
Closes #1693.

The `object hash` command has logic when `len(ranges) == 0`, so I don't think it should be made obligatory.

### DIFF
--- a/cmd/neofs-cli/modules/object/hash.go
+++ b/cmd/neofs-cli/modules/object/hash.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	internalclient "github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/client"
+	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/common"
 	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/commonflags"
 	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/key"
 	"github.com/nspcc-dev/neofs-sdk-go/checksum"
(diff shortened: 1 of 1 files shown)


cmd/neofs-cli/modules/object/hash.go
```diff
@@ -42,7 +42,7 @@ func initObjectHashCmd() {
 	flags.String(commonflags.OIDFlag, "", commonflags.OIDFlagUsage)
 	_ = objectHashCmd.MarkFlagRequired(commonflags.OIDFlag)
 
-	flags.String("range", "", "Range to take hash from in the form offset1:length1,...")
+	flags.String("range", "", "Range to take hash from in the form offset1:length1,... (by default, gets a hash of the full object payload")
```
roman-khimov, line 45: `full object payload length if not specified`?

cmd/neofs-cli/modules/object/hash.go
```diff
@@ -42,7 +42,7 @@ func initObjectHashCmd() {
 	flags.String(commonflags.OIDFlag, "", commonflags.OIDFlagUsage)
 	_ = objectHashCmd.MarkFlagRequired(commonflags.OIDFlag)
 
-	flags.String("range", "", "Range to take hash from in the form offset1:length1,...")
+	flags.String("range", "", "Range to take hash from in the form offset1:length1,... Full object payload length if not specified")
```
carpawell, line 45: can we print a verbose message in this case? (we have such practice in neofs-cli)

cthulhu-rider (approved):

roman-khimov (approved):

carpawell (approved):

PR #2985, opened 2024-10-28
cli: fix overriding of the `Timestamp` container attribute

### LINKED ISSUE #1339: Undefined behaviour on duplicated container attributes when using neofs-cli
It's not allowed to create a container with duplicated attributes.
When using neofs-cli, I noticed that it reports about this constraint from time to time. It depends on the attribute kind.

## Current Behavior
Below I try to create two containers with duplicated attributes. In the first snippet, the attributes are custom, and I immediately get an error. In the second snippet, the attribute `Timestamp` is added by the neofs-cli "under the hood". I believe it is also the case of attributes duplications, but neofs-cli sends a container creation request into the node. (In this case the command failed after 60sec timeout.)
```
$ neofs-cli --rpc-endpoint s01.neofs.devenv:8080 --wallet L58r7pn31wvBAECxPPo96AM4fH3JHfUrfGRyo9pQoa13KMfP3QEJ \
container create --policy "REP 2 IN X CBF 1 SELECT 4 FROM * AS X"  --attributes Size=small, Size=big  --await

invalid container attribute
```
```
$ neofs-cli --rpc-endpoint s01.neofs.devenv:8080 --wallet L58r7pn31wvBAECxPPo96AM4fH3JHfUrfGRyo9pQoa13KMfP3QEJ \
container create --policy "REP 2 IN X CBF 1 SELECT 4 FROM * AS X"  --attributes Timestamp=new  --await

container ID: HaEweDDwUWZbcC5t97dUk34svexgr8q27c6PHTgsKSNQ
awaiting...
```
## Expected Behavior
I'm not sure what is the expected behaviour because it is undefined. If a user duplicates a "well-known" attribute, like `Timestamp`, it will be redefined OR it is considered as an error?

## Your Environment
```
$ neofs-cli --version
Version: v0.27.5-229-g606dfa34 
Build: 2022-04-12T20:08:21+00:00 
Debug: false
```

End-rey:
Closes #1339.

### DIFF
--- a/cmd/neofs-cli/modules/container/create.go
+++ b/cmd/neofs-cli/modules/container/create.go
@@ -229,6 +229,15 @@ func parseAttributes(dst *container.Container, attributes []string) error {
 			return errors.New("invalid container attribute")
 		}
 
+		if kvPair[0] == "Timestamp" && !containerNoTimestamp {
+			return errors.New("can't override default timestamp attribute, use '--disable-timestamp' flag")
+		}
+
+		if kvPair[0] == "Name" && containerName != "" && kvPair[1] != containerName {
+			return errors.New("name from the '--name' flag and the 'Name' attribute are not equal, " +
+				"you need to use one of them or make them equal")
+		}
+
 		dst.SetAttribute(kvPair[0], kvPair[1])
 	}
 
--- a/cmd/neofs-cli/modules/container/create.go
+++ b/cmd/neofs-cli/modules/container/create.go
@@ -238,6 +238,12 @@ func parseAttributes(dst *container.Container, attributes []string) error {
 				"you need to use one of them or make them equal")
 		}
 
+		if kvPair[0] == "" {
+			return errors.New("empty attribute key")
+		} else if kvPair[1] == "" {
+			return fmt.Errorf("empty attribute value for key %s", kvPair[0])
+		}
+
 		dst.SetAttribute(kvPair[0], kvPair[1])
 	}
 
--- a/cmd/neofs-cli/modules/object/put.go
+++ b/cmd/neofs-cli/modules/object/put.go
@@ -225,6 +225,12 @@ func parseObjectAttrs(cmd *cobra.Command, ctx context.Context) ([]object.Attribu
 			}
 		}
 
+		if kv[0] == "" {
+			return nil, errors.New("empty attribute key")
+		} else if kv[1] == "" {
+			return nil, fmt.Errorf("empty attribute value for key %s", kv[0])
+		}
+
 		attrs[i].SetKey(kv[0])
 		attrs[i].SetValue(kv[1])
 	}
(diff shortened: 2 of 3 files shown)


cmd/neofs-cli/modules/container/create.go
```diff
@@ -229,6 +229,16 @@ func parseAttributes(dst *container.Container, attributes []string) error {
 			return errors.New("invalid container attribute")
 		}
 
+		if kvPair[0] == "" {
+			return errors.New("empty attribute key")
+		} else if kvPair[1] == "" {
+			return errors.New("empty attribute value")
```
carpawell, line 235: can we add key with empty attribute to the error msg?

carpawell (approved):

cmd/neofs-cli/modules/container/create.go
```diff
@@ -229,6 +229,16 @@ func parseAttributes(dst *container.Container, attributes []string) error {
 			return errors.New("invalid container attribute")
 		}
 
+		if kvPair[0] == "" {
```
cthulhu-rider, line 232: emptiness checks aint related to the timestamp so deserve a separate commit

cthulhu-rider (changes requested):
there are also `--name` and `--global-name` attributty flags, they should be treated the same

also https://github.com/nspcc-dev/neofs-node/issues/1339#issuecomment-1110650798 suggests to cover object attributes similarly within this issue, the current one is not enough to close the issue

cmd/neofs-cli/modules/object/put.go
```diff
@@ -219,11 +223,25 @@ func parseObjectAttrs(cmd *cobra.Command) ([]object.Attribute, error) {
...
 		}
+
+		if kv[0] == object.AttributeTimestamp && !disableTime {
+			return nil, errors.New("can't override default timestamp attribute, use '--disable-timestamp' flag")
+		}
+
+		if kv[0] == object.AttributeFileName && !disableFilename {
+			return nil, errors.New("can't override default filename attribute, use '--disable-filename' flag")
+		}
+
```
cthulhu-rider, line 234: there are also lifetime flags `-e` or `-l` setting `__NEOFS__EXPIRATION_EPOCH` which can be `kv[0]`. They also must not cause duplication or conflict

cmd/neofs-cli/modules/container/create.go
```diff
@@ -229,6 +229,15 @@ func parseAttributes(dst *container.Container, attributes []string) error {
 			return errors.New("invalid container attribute")
 		}
 
+		if kvPair[0] == "Timestamp" && !containerNoTimestamp {
+			return errors.New("can't override default timestamp attribute, use '--disable-timestamp' flag")
+		}
+
+		if kvPair[0] == "Name" && containerName != "" && kvPair[1] != containerName {
+			return errors.New("name from the '--name' flag and the 'Name' attribute are not equal, " +
+				"you need to use one of them or make it equal")
```
roman-khimov, line 238: make them equal

roman-khimov (approved):

cthulhu-rider (approved):
@End-rey resolve conflicts and go

PR #2987, opened 2024-10-29
node: Fix false negative connection to FS chain in multi-endpoint setup

### LINKED ISSUE #2986: Node fails to connect to secondary RPCs during startup
## Expected Behavior
Nodes going over all of the `morph.endpoints` list for connection attempts.

## Current Behavior
Looks like only the first one is tried during startup, because with a configuration of
```
morph:
  endpoints:
    - ws://192.168.43.87:30333/ws
    - ws://192.168.43.88:30333/ws
    - ws://192.168.43.89:30333/ws
    - ws://192.168.43.90:30333/ws
    - ws://192.168.43.91:30333/ws
    - ws://192.168.43.85:30333/ws
    - ws://192.168.43.86:30333/ws
  dial_timeout: 20s
```
and 100 nodes starting (with 7 blockchain nodes available) only 60 are able to connect to the first one while all the others just fail with
```
2024-10-28T21:26:08.752Z	info	neofs-node/config.go:665	failed to create neo RPC client	{"endpoints": ["ws://192.168.43.87:30333/ws", "ws://192.168.43.88:30333/ws", "ws://192.168.43.89:30333/ws", "ws://192.168.43.90:30333/ws", "ws://192.168.43.91:30333/ws", "ws://192.168.43.85:30333/ws", "ws://192.168.43.86:30333/ws"], "error": "could not create RPC client: WS client creation: Internal error (-32603) - websocket users limit reached"}
```

## Possible Solution
Restore the expected behavior, do not fail immediately, walk through the list.

## Regression
No.

## Your Environment
* Version used: 0.43.0
* Operating System and version (`uname -a`): Ubuntu stable

cthulhu-rider:
Defect of 0d08b51afe772444a5bf280b2c6750bb1d001d48. Previously, node unconditionally visited all configured FS chain endpoints, and the result of the connection to the last address was final. In the case where the previous nodes (at least one) were available but the last one was not, a general init error occurred.

Now node breaks on first successful dial.

Fixes #2986.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ attribute, which is used for container domain name in NNS contracts (#2954)
 - Do not search for tombstones when handling their expiration, use local indexes instead (#2929)
 - Unathorized container ops accepted by the IR (#2947)
 - Structure table in the SN-configuration document (#2974)
+- False negative connection to NeoFS chain in multi-endpoint setup with at least one live node (#2986)
 
 ### Changed
 - `ObjectService`'s `Put` RPC handler caches up to 10K lists of per-object sorted container nodes (#2901)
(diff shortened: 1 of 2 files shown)


roman-khimov (approved):
Legit, changelog please.

PR #2996, opened 2024-11-01
Make RPC client try more

roman-khimov:


pkg/morph/client/client.go
```diff
@@ -493,28 +451,26 @@ func (c *Client) TxHalt(h util.Uint256) (res bool, err error) {
 
 // TxHeight returns true if transaction has been successfully executed and persisted.
 func (c *Client) TxHeight(h util.Uint256) (res uint32, err error) {
-	c.switchLock.RLock()
-	defer c.switchLock.RUnlock()
+	var conn = c.conn.Load()
```
carpawell, line 454: so many `var`s at the first line of the func...

pkg/morph/client/constructor.go
```diff
@@ -106,101 +106,90 @@ func New(key *keys.PrivateKey, opts ...Option) (*Client, error) {
...
-
-			curNotifyChan: make(chan *state.ContainedNotificationEvent),
-			curBlockChan:  make(chan *block.Block),
-			curNotaryChan: make(chan *result.NotaryRequestEvent),
-		},
-	}
-
-	var err error
-	var act *actor.Actor
+	var (
```
carpawell, line 109: even here? more parentheses, more nesting, more lines

carpawell (approved):
I do not mind the code but not sure I understand why this PR is so big, "Reconnect indefinitely." could be solved in a single line change while the other functional changes just made this client less flexible (this client is used in SN, IR and even used to be used in external client or maybe still).

roman-khimov:
The other change here is that the client will fail requests easily instead of blocking when it's disconnected. It's harder to predict the effect, but it seems to be more fair (associated requests will fail quickly instead of waiting for unknown time).

pkg/morph/client/client.go
roman-khimov: Love Pascal.

pkg/morph/client/constructor.go
roman-khimov: I can make add even more of them into this `var ()` block for you. But hey, we're defining some variables, `var ()` is exactly for that and it's preferred over a series of `var ... var ... var ...`.

cthulhu-rider (approved):
woth to be changeloged

pkg/morph/client/constructor.go
carpawell: as always, i prefer `:=`

PR #2998, opened 2024-11-02
morph: support reloading morph endpoints with sighup

### LINKED ISSUE #1871: Support reloading morph endpoints with SIGHUP

End-rey:
Closes #1871.

### DIFF
--- a/docs/sighup.md
+++ b/docs/sighup.md
@@ -10,8 +10,6 @@ Available for reconfiguration fields:
 
 ```yml
   head_timeout:
-  cache_size:
-  cache_time:
   replication_cooldown:
   object_batch_size:
   max_workers:
--- a/docs/sighup.md
+++ b/docs/sighup.md
@@ -32,3 +32,9 @@ comparing paths from `shard.blobstor` section. After this we have 3 sets:
 | Changed section | Actions                                                                                                              |
 |-----------------|----------------------------------------------------------------------------------------------------------------------|
 | `path`          | If `path` is different, metabase is closed and opened with a new path. All other configuration will also be updated. |
+
+### Morph
+
+| Changed section | Actions                                                                                                                                                                                                 |
+|-----------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `endpoints`     | Updates N3 endpoints.<br/>If new `endpoints` do not contain the endpoint client is connected to, it will reconnect to another endpoint from the new list. Node service can be interrupted in this case. |
--- a/pkg/morph/client/constructor.go
+++ b/pkg/morph/client/constructor.go
@@ -139,7 +139,6 @@ func New(key *keys.PrivateKey, opts ...Option) (*Client, error) {
 			return nil, errors.New("no endpoints were provided")
 		}
 
-		cli.endpoints = cfg.endpoints
 		conn = cli.connEndpoints()
 		if conn == nil {
 			err = errors.New("could not establish Neo RPC connection")
--- a/pkg/morph/client/constructor.go
+++ b/pkg/morph/client/constructor.go
@@ -40,7 +40,8 @@ type cfg struct {
 	autoSidechainScope bool
 	signer             *transaction.Signer
 
-	endpoints []string
+	endpointsLock *sync.RWMutex
+	endpoints     []string
 
 	singleCli *rpcclient.WSClient // neo-go client for single client mode
 
@@ -63,6 +64,7 @@ func defaultConfig() *cfg {
 		signer: &transaction.Signer{
 			Scopes: transaction.CalledByEntry,
 		},
+		endpointsLock:       &sync.RWMutex{},
 		reconnectionDelay:   5 * time.Second,
 		reconnectionRetries: 5,
 	}
--- a/pkg/morph/client/reload.go
+++ b/pkg/morph/client/reload.go
@@ -1,0 +1,29 @@
+package client
+
+import "slices"
+
+// Reload allows runtime reconfiguration for WithEndpoints parameter.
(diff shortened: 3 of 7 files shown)


pkg/morph/client/reload.go
```diff
@@ -0,0 +1,25 @@
...
+	c.endpoints = cfg.endpoints
+
+	var conn = c.conn.Load()
+	if conn == nil {
+		return
+	}
+	currentEndpointID := conn.client.Endpoint()
+
+	if slices.Index(c.endpoints, currentEndpointID) == -1 {
+		conn.client.Close()
```
roman-khimov, line 23: Brutal. I'd rather document that we're keeping current connection.

End-rey: Added to a `Reload` function description.

roman-khimov: I mean dropping this code, RPC reconnections are costly for node, so we can avoid them and use current node while it's alive.

pkg/morph/client/reload.go
```diff
@@ -0,0 +1,27 @@
...
+// Keeping the current connection alive, if in the config there is an endpoint that the client is connected to.
+func (c *Client) Reload(opts ...Option) {
+	cfg := new(cfg)
+	for _, o := range opts {
+		o(cfg)
+	}
+
+	c.cfg.endpoints = cfg.endpoints
+
+	c.endpoints = cfg.endpoints
```
roman-khimov, line 16: This needs to be protected wrt other threads, btw.

End-rey: [Here](https://github.com/nspcc-dev/neofs-node/pull/2996/commits/5ec3d5a1762e3dabc287fdd0d535306c9b38d278#) `switchLock *sync.RWMutex` has been deleted. Do I need to add it again or is there some other way?

roman-khimov: It was created for a different purpose. You can introduce som endpoints RWLock now.

docs/sighup.md
```diff
@@ -10,8 +10,6 @@ Available for reconfiguration fields:
 
 ```yml
   head_timeout:
-  cache_size:
```
carpawell, line 13: a little bit confused why this happened in this PR. you just noted that it is outdated? a separate commit then?

pkg/morph/client/reload.go
```diff
@@ -0,0 +1,31 @@
...
+// Reload allows runtime reconfiguration for WithEndpoints parameter.
+func (c *Client) Reload(opts ...Option) {
+	cfg := new(cfg)
+	for _, o := range opts {
+		o(cfg)
+	}
+
+	c.endpointsLock.Lock()
+
+	c.cfg.endpoints = cfg.endpoints
```
carpawell, line 14: hm, maybe not this PR's zone but why we need two field updates?

pkg/morph/client/reload.go
carpawell: @roman-khimov, this happens only if we are connected to a node that should not be used anymore according to the updated nodes list

pkg/morph/client/reload.go
roman-khimov: Yeah, I wanted to fix this immediately as well. But would be nice if @End-rey untangles it.

pkg/morph/client/reload.go
roman-khimov: My intention is not to drop this connection anyway. And I'm not sure you can rely on these comparisons.

End-rey: Ah, now I get it. That is, even if the new config does not have this endpoint, keep in touch with it?

roman-khimov: Yes.

carpawell: you are connected to node X, then you updated to a list that does not have node X, what is your expectation as an admin?

roman-khimov: Depends on documentation.

End-rey: I'm also for switching, it seems logical to me. But if this behavior is documented, then fine.

roman-khimov: Let's ask @532910.

docs/sighup.md
```diff
@@ -32,3 +32,9 @@ comparing paths from `shard.blobstor` section. After this we have 3 sets:
 | Changed section | Actions                                                                                                              |
 |-----------------|----------------------------------------------------------------------------------------------------------------------|
 | `path`          | If `path` is different, metabase is closed and opened with a new path. All other configuration will also be updated. |
+
+### Morph
+
+| Changed section | Actions                                                                                                                                                                                                |
+|-----------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `endpoints`     | Updates N3 endpoints.<br/>If in the `endpoints` there is no endpoint that the client is connected to, it will attempt to connect to another endpoint N3. Node service can be interrupted in this case. |
```
roman-khimov, line 40: If new `endpoints` do not contain the endpoint client is connected to, it will reconnect to another endpoint from the new list

Smth like that (other sentences are ok)

pkg/morph/client/constructor.go
```diff
@@ -40,7 +40,8 @@ type cfg struct {
 	autoSidechainScope bool
 	signer             *transaction.Signer
 
-	endpoints []string
+	endpoints     []string
+	endpointsLock *sync.RWMutex
```
roman-khimov, line 44: Usually it's written like
```
	endpointsLock *sync.RWMutex
	endpoints     []string
```
Lock first, then thing(s) it protects.

carpawell (approved):

PR #3001, opened 2024-11-06
metabase: drop almost all of Prm/Res structures

roman-khimov:
They:
 * are present in some methods and not in others
 * add substantial cognitive overhead for anyone looking
 * require more code to handle trivial things
 * can allocate more and more of microobjects creating GC pressure

Test wrappers in the same package perfectly suggest what everyone wanted to have in the first place. So this patch is a pure refactoring and doesn't change any behavior, but it drops almost all structures. "Almost" because there are cases where we pass a number of things (mostly as a result) and having some enclosing structure makes some sense there.

### DIFF
--- a/cmd/neofs-lens/internal/meta/list.go
+++ b/cmd/neofs-lens/internal/meta/list.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 
 	common "github.com/nspcc-dev/neofs-node/cmd/neofs-lens/internal"
-	meta "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/metabase"
 	"github.com/spf13/cobra"
 )
 
@@ -40,15 +39,12 @@ func listFunc(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("%s flag must be positive", limitFlagName)
 	}
 
-	var prm meta.ListPrm
-	prm.SetCount(vLimit)
-
-	res, err := db.ListWithCursor(prm)
+	addrs, _, err := db.ListWithCursor(int(vLimit), nil)
 	if err != nil {
 		return fmt.Errorf("metabase's `ListWithCursor`: %w", err)
 	}
 
-	for _, addressWithType := range res.AddressList() {
+	for _, addressWithType := range addrs {
 		cmd.Printf("%s, Type: %s\n", addressWithType.Address, addressWithType.Type)
 	}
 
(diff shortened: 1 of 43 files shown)


cmd/neofs-lens/internal/meta/list.go
```diff
@@ -40,15 +39,12 @@ func listFunc(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("%s flag must be positive", limitFlagName)
 	}
 
-	var prm meta.ListPrm
-	prm.SetCount(vLimit)
-
-	res, err := db.ListWithCursor(prm)
+	addrs, _, err := db.ListWithCursor(int(vLimit), nil)
```
carpawell, line 42: why int?

pkg/local_object_storage/metabase/delete.go
```diff
@@ -14,50 +14,25 @@ import (
...
+	// allowed, it claims a missing object.
+	Sizes []uint64
 }
 
-// Delete removed object records from metabase indexes.
+// Delete removes object records from metabase indexes.
 // Does not stop on an error if there are more objects to handle requested;
 // returns the first error appeared with a number of deleted objects wrapped.
-func (db *DB) Delete(prm DeletePrm) (DeleteRes, error) {
+func (db *DB) Delete(addrs []oid.Address) (DeleteRes, error) {
```
carpawell, line 35: it was variadic. not sure api has become better. but also does not claim it as a much worse one

pkg/local_object_storage/metabase/movable.go
```diff
@@ -7,97 +7,54 @@ import (
...
+	return db.boltDB.Update(func(tx *bbolt.Tx) error {
 		toMoveIt := tx.Bucket(toMoveItBucketName)
 		return toMoveIt.Put(key, zeroValue)
 	})
-
-	return
 }
 
 // DoNotMove removes `MoveIt` mark from the object.
-func (db *DB) DoNotMove(prm DoNotMovePrm) (res DoNotMoveRes, err error) {
```
carpawell, line 73: wow, i had no idea we had this method

carpawell (approved):

cmd/neofs-lens/internal/meta/list.go
roman-khimov: Because that's the type `ListPrm` had for `count`.

pkg/local_object_storage/metabase/delete.go
roman-khimov: From what I see real users have arrays anyway, so this works fine.

PR #3009, opened 2024-11-12
Makefile: use `buildx` component to build images with BuildKit

### LINKED ISSUE #3006: legacy builder is deprecated, install the buildx component to build images with BuildKit
```
% make images 
⇒ Build NeoFS storage docker image 
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            Install the buildx component to build images with BuildKit:
            https://docs.docker.com/go/buildx/
```

End-rey:
The GH workflow uses the [build-push-action](https://github.com/docker/build-push-action), where images are built with `buildx`. So I made it like this in our `Makefile`.
Also, fix warnings in Dockerfiles.

Closes #3006.

### DIFF
--- a/.docker/Dockerfile.adm
+++ b/.docker/Dockerfile.adm
@@ -1,4 +1,4 @@
-FROM golang:1.23 as builder
+FROM golang:1.23 AS builder
 ARG BUILD=now
 ARG VERSION=dev
 ARG REPO=repository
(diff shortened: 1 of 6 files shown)


roman-khimov (approved):

carpawell (commented):
Are you sure it is a solution? As I understand, the client just asks to install some additional plugin, not sure Makefile is wrong here and we are doing incorrect things.

End-rey:
> Are you sure it is a solution? As I understand, the client just asks to install some additional plugin, not sure Makefile is wrong here and we are doing incorrect things.

I didn't have any mistake initially. But I set the value `DOCKER_BUILDKIT=0` and then the message started to appear. What I did in the commit fixed the situation.

cthulhu-rider (approved):

PR #3013, opened 2024-11-13
Deprecate peapod, provide migration to fstree

### LINKED ISSUE #2924: Deprecate peapod, provide migration to fstree
I'm always frustrated when we keep code we no longer need. After #2814 peapods can be considered obsolete since fstree has better performance in all cases (and can be tweaked in many ways). We also:
 * have #2813 that is peapod-specific
 * have #1951 that peapods are susceptible to
 * have #2107 that is not peapod-specific, but likely can be handled much easier without yet another BoltDB on the disk
 * have #2497 where we can finally eliminate the concept of substorages, one shard --- one blobstor is much easier to handle (not within the scope of current issue, but in future, see #2480 also)

The proposal is to discontinue them in the same way we did for bbcz:
 * declare them as deprecated and to be removed
 * create a migration tool that moves data and fixes configuration file
 * roll it all out, keep peapods supported for one release, migrate
 * drop support for peapods in subsequent releases

Keeping peapods, but why?

End-rey:
Closes #2924.

Is there anything else I should do in this pr?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ attribute, which is used for container domain name in NNS contracts (#2954)
 - Docs files for cli commands to the `docs/cli-commands` folder (#2983)
 - `logger.encoding` config option (#2999)
 - Reloading morph endpoints with SIGHUP (#2998)
+- New `peapod-to-fstree` tool providing peapod-to-fstree data migration (#3013)
 
 ### Fixed
 - Do not search for tombstones when handling their expiration, use local indexes instead (#2929)
@@ -43,6 +44,7 @@ attribute, which is used for container domain name in NNS contracts (#2954)
 - Reject configuration with unknown fields (#2981)
 - Log sampling is disabled by default now (#3011)
 - EACL is no longer considered for system role (#2972)
+- Deprecate peapod substorage (#3013)
 
 ### Removed
 - Support for node.key configuration (#2959)
@@ -60,6 +62,22 @@ Binary keys are no longer supported by storage node, NEP-6 wallet support was
 introduced in version 0.22.3 and support for binary keys was removed from
 other components in 0.33.0 and 0.37.0. Please migrate to wallets (see 0.37.0
 notes) if you've not done it previously.
+
+To migrate data from Peapods to FSTree:
+```shell
+$ peapod-to-fstree -config </path/to/storage/node/config>
+```
+For any shard, the data from the configured Peapod is copied into
+a FSTree that must be already configured.
+Notice that peapod DB is not deleted during migration. Configuration is also
+updated, for example, `/etc/neofs/config.yaml` -> `/etc/neofs/config.yaml.migrated`.
+WARN: carefully review the updated config before using it in the application!
+
+Now there is an FSTree in which files of any size will be more efficient and easier to store.
+We support both `fstree` and `peapod` sub-storages,
+but `peapod` can be removed in future versions. We recommend using `fstree`.
+If you want to use only `fstree` and storage node already stores some data,
+don't forget to make data migration described above.
 
 ## [0.43.0] - 2024-08-20 - Jukdo
 
--- a/cmd/peapod-to-fstree/main.go
+++ b/cmd/peapod-to-fstree/main.go
@@ -1,0 +1,327 @@
+package main
+
+import (
+	"bytes"
+	"errors"
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"path/filepath"
+	"slices"
+	"time"
+
+	"github.com/nspcc-dev/neofs-node/cmd/neofs-node/config"
+	engineconfig "github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/engine"
+	shardconfig "github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/engine/shard"
+	fstreeconfig "github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/engine/shard/blobstor/fstree"
+	objectcore "github.com/nspcc-dev/neofs-node/pkg/core/object"
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor/common"
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor/compression"
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor/fstree"
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor/peapod"
+	meta "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/metabase"
+	"go.etcd.io/bbolt"
+	"gopkg.in/yaml.v3"
+)
+
+type epochState struct{}
+
+func (s epochState) CurrentEpoch() uint64 {
+	return 0
+}
+
+func main() {
+	nodeCfgPath := flag.String("config", "", "Path to storage node's YAML configuration file")
+
+	flag.Parse()
+
+	if *nodeCfgPath == "" {
+		log.Fatal("missing storage node config flag")
+	}
+
+	srcPath := *nodeCfgPath
+
+	dstPath := srcPath + ".migrated"
+
+	log.Printf("migrating configuration to '%s' file...\n", dstPath)
+
+	err := migrateConfigToFstree(dstPath, srcPath)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	log.Println("configuration successfully migrated, migrating data in shards...")
+
+	appCfg := config.New(config.Prm{}, config.WithConfigFile(*nodeCfgPath))
+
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -59,6 +61,23 @@ introduced in version 0.22.3 and support for binary keys was removed from
...
+For any shard, the data from the configured Peapod is copied into
+a FSTree that must be already configured, if it is not, into
+a created FSTree dir named `blobstore` in the directory where the tree is
+located. For example, `/neofs/data/peapod.db` -> `/neofs/data/blobstore`.
+Notice that existing Peapod are untouched. Configuration is also
+updated, for example, `/etc/neofs/config.yaml` -> `/etc/neofs/config_fstree.yaml`.
+WARN: carefully review the updated config before using it in the application!
+
+Now there is an FSTree in which files of any size will be more efficient and easier to store.
+Use only `fstree` sub-storage in `blobstor` config section.
```
roman-khimov, line 77: I'd say that currently we support both, but peapod can be removed in future versions. And we recommend using fstree.

CHANGELOG.md
```diff
@@ -59,6 +61,23 @@ introduced in version 0.22.3 and support for binary keys was removed from
...
 
+To migrate data from Peapods to FSTree:
+```shell
+$ peapod-to-fstree -config </path/to/storage/node/config>
+```
+For any shard, the data from the configured Peapod is copied into
+a FSTree that must be already configured, if it is not, into
+a created FSTree dir named `blobstore` in the directory where the tree is
+located. For example, `/neofs/data/peapod.db` -> `/neofs/data/blobstore`.
+Notice that existing Peapod are untouched. Configuration is also
```
roman-khimov, line 72: peapod DB is not deleted during migration

pkg/local_object_storage/blobstor/peapod/peapod.go
```diff
@@ -39,6 +39,8 @@ type batch struct {
 // Peapod provides storage for relatively small NeoFS binary object (peas).
 // Peapod is a single low-level key/value database optimized to work with big
 // number of stored units.
+//
+// Deprecated: Use fstree.FSTree instead.
```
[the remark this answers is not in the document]
End-rey: What about `peapod.New`? Maybe I should leave him too?

cmd/peapod-to-fstree/main.go
```diff
@@ -0,0 +1,291 @@
...
+			return nil
+		}
+
+		ppdPath := ppd.Path()
+		if !filepath.IsAbs(ppdPath) {
+			log.Fatalf("Peapod path '%s' is not absolute, make it like this in the config file first\n", ppdPath)
+		}
+
+		if fstr == nil {
+			fstr = fstree.New(
```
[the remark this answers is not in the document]
roman-khimov: Right now the limits imposed on config (2 substorages min) are suggesting that we can't have a blobstor without fstree. And I don't like this magic in general, we can legitimately fail in this case, just say "please configure some fstree for this shard" and let the user figure it out. This won't be a problem in real life.

pkg/local_object_storage/blobstor/peapod/peapod.go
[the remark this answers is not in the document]
carpawell: dont we (and this exact commit) need to print REALLY HUGE LETTERS IN LOGS PLEASE DO NOT USE PEAPOD ANYMORE?

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ attribute, which is used for container domain name in NNS contracts (#2954)
 - Docs files for cli commands to the `docs/cli-commands` folder (#2983)
 - `logger.encoding` config option (#2999)
 - Reloading morph endpoints with SIGHUP (#2998)
+- New `peapod-to-fstree` tool providing peapod-to-fstree data migration (#3013)
```
carpawell, line 22: this commit's message was ruined but some squash? should be fixed

cmd/peapod-to-fstree/main.go
```diff
@@ -0,0 +1,269 @@
...
+					storagesCfg[i].Perm(),
+					storagesCfg[i].FlushInterval(),
+				)
+			default:
+				return fmt.Errorf("invalid storage type: %s", storagesCfg[i].Type())
+			}
+		}
+
+		if ppd == nil {
+			log.Println("Peapod is not configured for the current shard, going to next one...")
```
carpawell, line 69: for what shard?

cmd/peapod-to-fstree/main.go
```diff
@@ -0,0 +1,269 @@
...
+	flag.Parse()
+
+	if *nodeCfgPath == "" {
+		log.Fatal("missing storage node config flag")
+	}
+
+	appCfg := config.New(config.Prm{}, config.WithConfigFile(*nodeCfgPath))
+
+	err := engineconfig.IterateShards(appCfg, false, func(sc *shardconfig.Config) error {
+		log.Println("processing shard...")
```
carpawell, line 36: what shard?

carpawell (changes requested):
I am almost sure we also want to migrate some metabase indexes _in this util_, otherwise i do not know who should do it.
See https://github.com/nspcc-dev/neofs-node/blob/ba1d63952ef555989171bb3e017fe054f3abceb2/pkg/local_object_storage/blobstor/get.go#L19-L32, https://github.com/nspcc-dev/neofs-node/blob/ba1d63952ef555989171bb3e017fe054f3abceb2/pkg/local_object_storage/blobstor/peapod/peapod.go#L422-L425 and https://github.com/nspcc-dev/neofs-node/blob/ba1d63952ef555989171bb3e017fe054f3abceb2/pkg/local_object_storage/blobstor/fstree/fstree.go#L308

carpawell:
Linter feels bad.

pkg/local_object_storage/blobstor/peapod/peapod.go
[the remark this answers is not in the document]
roman-khimov: No. Not yet.

cmd/peapod-to-fstree/main.go
```diff
@@ -0,0 +1,326 @@
...
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor/peapod"
+	meta "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/metabase"
+	"go.etcd.io/bbolt"
+	"gopkg.in/yaml.v3"
+)
+
+type epochState struct{}
+
+func (s epochState) CurrentEpoch() uint64 {
+	return 0
```
roman-khimov, line 32: Not sure what can be affected by this.

End-rey: How can I easily get the epoch state in another way? The metabase requires it to be set.

roman-khimov: Only via network interaction. Maybe it's not a problem in our case (updating storage ID doesn't require epoch IIRC). Or maybe we can get some latest one from meta or local state.

End-rey: I can get LastResyncEpoch from metabase:
https://github.com/nspcc-dev/neofs-node/blob/45802c65709a8be508ccd8b949efaf85933ddd6f/pkg/local_object_storage/metabase/last_resync_epoch.go#L12-L14

roman-khimov: It's somewhat different in meaning and it's a new one, 0.43.0 DB doesn't have it.

carpawell: why do you need a precise epoch? just pick an epoch that does not claim anything about object expiration and allows GET and PUT freely, that is `0`. @roman-khimov, what bothers you?

> How can I easily get the epoch state in another way? The metabase requires it to be set.

i did this and i did not know there would be so many utils that just open metabases and look at them. normal metabase always has to know the current epoch (via internal state or via explicit parameter to every operation)

roman-khimov (approved):

cmd/peapod-to-fstree/main.go
```diff
@@ -0,0 +1,330 @@
...
+	nodeCfgPath := flag.String("config", "", "Path to storage node's YAML configuration file")
+
+	flag.Parse()
+
+	if *nodeCfgPath == "" {
+		log.Fatal("missing storage node config flag")
+	}
+
+	srcPath := *nodeCfgPath
+	ss := strings.Split(srcPath, ".")
```
carpawell, line 45: i would recomment smth like
```go
extension := filepath.Ext(srcPath)
name := srcPath[0:len(srcPath)-len(extension)]
```

dots are not prohibited in file names

carpawell (approved):

cmd/peapod-to-fstree/main.go
roman-khimov: In general, you can just do `+ ".migrated"` and that's it. `srcPath` can be anything, with many dots, without dots, whatever.

PR #3015, opened 2024-11-13
acl: do not check EACL for system role

### LINKED ISSUE #2972: HEAD operation leads to RPC requests
## Expected Behavior
HEAD operation returning result using local data only.

## Current Behavior
```
read object header from NeoFS: status: code = 1024 message = connection to the RPC node has been lost"
```
Why some RPC request?

## Possible Solution
Change something.

Not known.

## Context
Mainnet.

## Regression
Doesn't look like.

## Your Environment
* Version used: 0.43.0
* Operating System and version (`uname -a`): Debian stable

roman-khimov:
EACL can not have any rules for system role since 0.38.0 (ab909a371b069c9fcc728f617b1e29aeaf135ab2), so performing these checks is not very helpful. Of course one can still ban the node by key, but that would make a lot of regular operations fail and broken container is not very helpful.

This fixes #2972 as much as possible (containers can be cached for a longer period of time).

### DIFF
--- a/pkg/services/object/acl/acl.go
+++ b/pkg/services/object/acl/acl.go
@@ -129,6 +129,22 @@ func (c *Checker) CheckEACL(msg any, reqInfo v2.RequestInfo) error {
 		return nil
 	}
 
+	var eaclRole eaclSDK.Role
+	switch op := reqInfo.RequestRole(); op {
+	default:
+		eaclRole = eaclSDK.Role(op)
+	case acl.RoleOwner:
+		eaclRole = eaclSDK.RoleUser
+	case acl.RoleInnerRing, acl.RoleContainer:
+		eaclRole = eaclSDK.RoleSystem
+	case acl.RoleOthers:
+		eaclRole = eaclSDK.RoleOthers
+	}
+
+	if eaclRole == eaclSDK.RoleSystem {
+		return nil // Controlled by BasicACL, EACL can not contain any rules for system role since 0.38.0.
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/services/object/acl/acl.go
```diff
@@ -129,6 +129,22 @@ func (c *Checker) CheckEACL(msg any, reqInfo v2.RequestInfo) error {
...
+	case acl.RoleOwner:
+		eaclRole = eaclSDK.RoleUser
+	case acl.RoleInnerRing, acl.RoleContainer:
+		eaclRole = eaclSDK.RoleSystem
+	case acl.RoleOthers:
+		eaclRole = eaclSDK.RoleOthers
+	}
+
+	if eaclRole == eaclSDK.RoleSystem {
+		return nil // Controlled by BasicACL, EACL can not contain any rules for system role since 0.38.0.
```
cthulhu-rider, line 145: no need to make a separate condition for this, we already have the `case`

cthulhu-rider (approved):

pkg/services/object/acl/acl.go
carpawell: also did not get it but maybe @roman-khimov has some thoughts on it

carpawell (approved):

pkg/services/object/acl/acl.go
roman-khimov: That `switch` is so nice in its request role->acl role mapping that I didn't wont to interrupt the flow there. Can be changed at any time.

PR #3016, opened 2024-11-14
metrics, pprof: support reloading services with SIGHUP

### LINKED ISSUE #1868: Reload pprof/metrics services on SIGHUP

End-rey:
Closes #1868.

Is it right that we should reload services, even if the config is not updated?

There is also such a linter error:
```contextcheck  Function `preRunAndLog->preRunAndLog$2->Shutdown` should pass the context parameter```

Do I need to pass the context honestly or is there some another way?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ attribute, which is used for container domain name in NNS contracts (#2954)
 - New `peapod-to-fstree` tool providing peapod-to-fstree data migration (#3013)
 - Reloading node attributes with SIGHUP (#3005)
 - Reloading pool sizes (#3018)
+- Reloading pprof/metrics services with SIGHUP (#3016)
 
 ### Fixed
 - Do not search for tombstones when handling their expiration, use local indexes instead (#2929)
--- a/cmd/neofs-node/config.go
+++ b/cmd/neofs-node/config.go
@@ -24,7 +24,6 @@ import (
 	shardconfig "github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/engine/shard"
 	fstreeconfig "github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/engine/shard/blobstor/fstree"
 	loggerconfig "github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/logger"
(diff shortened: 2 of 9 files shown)


roman-khimov:
> Is it right that we should reload services, even if the config is not updated?

This is a service interruption and we can easily avoid it.

> Do I need to pass the context honestly or is there some another way?

In this case it can be suppressed, `configWatcher` is misusing `context.Context`, so passing it over is not very helpful.

End-rey:
Do I understand correctly that it is also necessary to overwrite this [variable](https://github.com/nspcc-dev/neofs-node/blob/45802c65709a8be508ccd8b949efaf85933ddd6f/cmd/neofs-node/config.go#L377) and all its derivatives if enabled changes?

roman-khimov:
Likely so, it should be possible to enable/disable the service with SIGHUP.

End-rey:
I have made sure that the metrics are completely reloaded in all services where they are used. But I'm not at all sure that I did it right.

roman-khimov (commented):
Looking at the amount changes I suspect we're doing something wrong. We're using global metric registry and yet we have some service that is collecting all metrics into a single place. This makes zero sense to me. In general, we still want to use the default registry since it adds Go/process-specific things that we want to export. This means each package can define metrics it wants _locally_ and add data to them irrespectively of settings (or just with some local flag). And then services would just be HTTP windows to things provided by packages, it'd be trivial to restart them.

So my suggestion is to refactor `pkg/metrics` out of the way completely (moving respective metrics to appropriate packages) and then solve the restart problem easily. 

@carpawell, @cthulhu-rider?

carpawell:
> each package can define metrics it wants locally

I feel like it is a simpler way in general but still, the idea of local metrics for every package scares me a little just because it will be hard to control them (i mean you will never see all the metrics in a single place then). My internal feeling is close to what @End-rey did in this PR. However, @roman-khimov, I agree that we are fighting against the lib. Dont have a strict opinion here. Maybe if some decision is required, then I have to say that KISS should win here and local metrics in every service should be a better choice overall.

End-rey:
Made "always collect" mode so node only reloads metrics server with SIGHUP.

roman-khimov (approved):

Issue #1770, opened 2022-09-08
Reload config on SIGHUP


fyrchik:
- [x] Set logging level (research if it is possible) [#1832]
- [x] #1868
- [x] #1869
- [x]  Blobstor parameters, see #1707 
- [ ] Pool sizes
- [x] #1870
- [x] #1871

See also https://github.com/nspcc-dev/neofs-http-gw/issues/200

PR #3019, opened 2024-11-20
support reloading pool sizes with SIGHUP

End-rey:
Get new pool sizes from these fields in config:
- `object.put.pool_size_remote`
- `replicator.pool_size` (but I noticed that it is immediately overwritten by this [function](https://github.com/nspcc-dev/neofs-node/blob/40762d3e0743bae8edf5efd6894be7b83f5a6c2f/pkg/services/policer/process.go#L93), maybe I should also reload `policer.max_workers`?)
- `storage.shard_pool_size`

Refs #1770.


pkg/local_object_storage/engine/control.go
```diff
@@ -191,6 +191,14 @@ func (rCfg *ReConfiguration) AddShard(id string, opts []shard.Option) {
 
 // Reload reloads StorageEngine's configuration in runtime.
 func (e *StorageEngine) Reload(rcfg ReConfiguration) error {
+	e.mtx.Lock()
+	e.shardPoolSize = rcfg.shardPoolSize
```
roman-khimov, line 195: Not sure it needs to be updated at all, nothing really uses it.

pkg/local_object_storage/engine/control.go
```diff
@@ -191,6 +191,14 @@ func (rCfg *ReConfiguration) AddShard(id string, opts []shard.Option) {
...
+
+	for _, pool := range e.shardPools {
```
roman-khimov, line 197: And here RLock() is sufficient. RLock is taken below anyway.

cmd/neofs-node/config.go
```diff
@@ -770,6 +771,21 @@ func initObjectPool(cfg *config.Config) (pool cfgObjectRoutines) {
...
+	c.cfgObject.pool.putRemoteCapacity = objectconfig.Put(c.cfgReader).PoolSizeRemote()
+	c.cfgObject.pool.putRemote.Tune(c.cfgObject.pool.putRemoteCapacity)
+
+	c.cfgObject.pool.replicatorPoolSize = replicatorconfig.PoolSize(c.cfgReader)
+	if c.cfgObject.pool.replicatorPoolSize <= 0 {
+		c.cfgObject.pool.replicatorPoolSize = c.cfgObject.pool.putRemoteCapacity
+	}
+	c.cfgObject.pool.replication.Tune(c.cfgObject.pool.replicatorPoolSize)
+
+	c.cfgObject.poolLock.Unlock()
```
roman-khimov, line 786: Better `defer Unlock()`.

cmd/neofs-node/config.go
```diff
@@ -830,6 +846,8 @@ func (c *cfg) needBootstrap() bool {
 // It is calculated as size/capacity ratio of "remote object put" worker.
 // Returns float value between 0.0 and 1.0.
 func (c *cfg) ObjectServiceLoad() float64 {
+	c.cfgObject.poolLock.RLock()
+	defer c.cfgObject.poolLock.RUnlock()
```
roman-khimov, line 850: You can avoid the lock by replacing `putRemoteCapacity` with https://pkg.go.dev/github.com/panjf2000/ants/v2#Pool.Cap

pkg/local_object_storage/engine/control.go
End-rey: Below in the [function](https://github.com/nspcc-dev/neofs-node/blob/50e3ff4ec6546823e430df924d27e752380e3484/pkg/local_object_storage/engine/control.go#L268), when adding shards, it is used.

https://github.com/nspcc-dev/neofs-node/blob/f1b69823ac8a97b70f4f1981ebb82b30a99511c5/pkg/local_object_storage/engine/shards.go#L113-L117

roman-khimov: That's init mostly, but we can also manage shards at runtime, so probably yes, it can be used.

roman-khimov (approved):

pkg/local_object_storage/engine/control.go
carpawell: > but we can also manage shards at runtime

it is supported

pkg/util/worker_pool.go
```diff
@@ -21,6 +21,9 @@ type WorkerPool interface {
 	// functions have returned so synchronization must be achieved
 	// via other means (e.g. sync.WaitGroup).
 	Release()
+
+	// Tune changes the capacity of this pool.
+	Tune(int)
```
carpawell, line 26: does this change solve something?

cmd/neofs-node/config.go
```diff
@@ -860,6 +875,10 @@ func (c *cfg) configWatcher(ctx context.Context) {
 
 			c.shared.policer.Reload(c.policerOpts()...)
 
+			// Pool
+
+			c.reloadObjectPoolSizes()
```
carpawell, line 880: can you, please, put it before `// Policer` or even `// Logger`? imo it would be more convenient: first some internal thins and then a series of `Reload`

carpawell (approved):

roman-khimov:
Conflicts.

pkg/util/worker_pool.go
End-rey: https://github.com/nspcc-dev/neofs-node/blob/b13fa5af7eb07fc94909e7e3b8ea9331114956d7/pkg/local_object_storage/engine/control.go#L204-L206
`e.shardPools` are `WorkerPools`, and I need to `Tune`.

PR #3022, opened 2024-11-21
node: add metrics for shard's capacity

### LINKED ISSUE #1794: Add metrics for the amount of taken space on each shard

End-rey:
Closes #1794.

### DIFF
--- a/cmd/neofs-node/metrics.go
+++ b/cmd/neofs-node/metrics.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/nspcc-dev/neofs-node/cmd/neofs-node/config"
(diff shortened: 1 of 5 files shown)


cmd/neofs-node/metrics.go
```diff
@@ -25,3 +31,34 @@ func initMetrics(c *cfg) *httputil.Server {
 
 	return srv
 }
+
+func (c *cfg) setShardsCapacity() error {
```
roman-khimov, line 35: You obviously can share the result of `writeSystemAttributes` with this. Or at least refactor out some common code. DRY.

pkg/metrics/engine.go
```diff
@@ -121,6 +122,13 @@ func newEngineMetrics() engineMetrics {
 			Name:      "payload_size",
 			Help:      "Accumulated size of all objects in a shard",
 		}, []string{shardIDLabelKey})
+
+		capacitySize = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Namespace: storageNodeNameSpace,
+			Subsystem: engineSubsystem,
+			Name:      "capacity",
+			Help:      "Contains the shard's capacity, GB",
```
roman-khimov, line 130: And btw, I doubt we're using GB as the base unit for size in metrics. It should follow other values like `payload_size`.

roman-khimov (approved):

carpawell (approved):

PR #3005, opened 2024-11-08
node: support reloading node attributes with SIGHUP

End-rey:
Closes #1870.

Also fix bug from #2998, that incorrectly checked when to reconnect.

PR #3023, opened 2024-11-21
Update to the latest SDK

End-rey:
Refs #3005.

Wait for fixing SDK problems.


roman-khimov:
Updated and included into https://github.com/nspcc-dev/neofs-node/pull/3103.

Issue #2993, opened 2024-10-31
Deleting unnecessary maps in the config


End-rey:
## Is your feature request related to a problem? Please describe.

I'm always frustrated when there are fields in the config that are difficult to validate, particularly in the form of a map.
In the node config, there are fields:
- `node.attribute_*`
- `storage.shard.*`

## Describe the solution you'd like
Make these fields lists.

## Describe alternatives you've considered

Leave everything as it is, but then complex validation will remain and it will not be easy to reuse the config.

## Additional context

Refs https://github.com/nspcc-dev/neofs-node/pull/2981#issuecomment-2447033994.
After this, we can use [viper.UnmarshalExact](https://pkg.go.dev/github.com/spf13/viper#UnmarshalExact) for validation and reuse the config in the code.

PR #3037, opened 2024-11-29
node: add check for unmarshalling config to validate

### LINKED ISSUE #3034: neofs-node breaks after update to 0.44.0
```
systemd[1]: Started neofs-storage.service - NeoFS Storage node.
neofs-node[182252]: panic: unable to decode config: 1 error(s) decoding:
neofs-node[182252]: * 'node[<interface {} Value>]' expected type 'string', got unconvertible type 'map[string]interface {}', value: 'map[exit_zero:false]'
neofs-node[182252]: goroutine 1 [running]:
neofs-node[182252]: github.com/nspcc-dev/neofs-node/cmd/neofs-node/config.New({}, {0xc00021ff08, 0x1, 0x5?})
neofs-node[182252]:         github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/config.go:61 +0x273
neofs-node[182252]: main.main()
neofs-node[182252]:         github.com/nspcc-dev/neofs-node/cmd/neofs-node/main.go:47 +0x27b
systemd[1]: neofs-storage.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
systemd[1]: neofs-storage.service: Failed with result 'exit-code'.
```

End-rey:
This is a check in case there is a ",remain" tag in the configuration field, and if there is an incorrect data format, then the value of the field was ignored, since then there will be a check to see if this field is correct or not. This makes some errors clearer after unmarshalling.

Closes #3034.

Actually, I don't like this fix. I want the attribute field to be a list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Changelog for NeoFS Node
 ### Added
 
 ### Fixed
+- Fail gracefully on error from config validation (#3037)
 
 ### Changed
 
--- a/cmd/neofs-node/config/config.go
+++ b/cmd/neofs-node/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"log"
 	"strings"
 
 	"github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/internal"
@@ -58,7 +59,7 @@ func New(_ Prm, opts ...Option) *Config {
 	if o.validate {
 		err := validate.ValidateStruct(v)
 		if err != nil {
-			panic(err)
+			log.Fatalf("failed config validation: %v", err)
 		}
 	}
 


roman-khimov (commented):
Far from being pretty, but looks like unavoidable at the same time.

Changes to `go.mod`?

Can we also not throw a panic from SN and fail gracefully?

roman-khimov:
At the same time, if we're to do https://github.com/nspcc-dev/neofs-node/issues/2993, this won't be a problem, right?

End-rey:
> Changes to `go.mod`?

No, `mapstructure` is a part of the `viper` package and it's been in `go.mod` for a long time, if that's what you mean.


> At the same time, if we're to do https://github.com/nspcc-dev/neofs-node/issues/2993, this won't be a problem, right?

I think yes.

End-rey:
> leave the rest for #2993

So, remove the check for unmarshalling?

roman-khimov:
Yes.

roman-khimov (commented):
Changelog?

roman-khimov (approved):

PR #3039, opened 2024-12-02
Blobstor: drop the last prm and res

roman-khimov:


roman-khimov:
https://github.com/nspcc-dev/neofs-node/actions/runs/12126755105

roman-khimov:
Buggy, to be fixed.

> info	log/log.go:12	local object storage operation	{"shard_id": "CAXqYYjUEAbdr8gcLFgrpa", "address": "56b9fmqewnWFfShD9SmKF5GzB4tpmGdfR9acgCftuejk/8yZJAhrohFoNY1DTyHSzhCYTZozP1TzaPwgjokvk2zZu", "op": "metabase PUT"}
2024/12/02 20:01:33 init shard CAXqYYjUEAbdr8gcLFgrpa: could not initialize *shard.metabaseSynchronizer: could not put objects to the meta from blobstor: blobstor iterator failure: exec read-only BoltDB transaction: logical error: no split info on parent object

roman-khimov:
Waiting for https://github.com/nspcc-dev/neofs-node/actions/runs/12139756512

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -304,9 +304,8 @@ func (t *FSTree) Put(prm common.PutPrm) (common.PutRes, error) {
 	if err := util.MkdirAllX(filepath.Dir(p), t.Permissions); err != nil {
 		return common.PutRes{}, fmt.Errorf("mkdirall for %q: %w", p, err)
 	}
-	if !prm.DontCompress {
-		prm.RawData = t.Compress(prm.RawData)
-	}
+	prm.RawData = t.Compress(prm.RawData)
```
carpawell: hm, a confusing change imo. suggest renaming to `CompressIfEnabled`

roman-khimov: That's the way it's named now, I'd not bother.

carpawell: previously it was near `if !prm.DontCompress {`, now it just always compresses

carpawell (approved):

pkg/local_object_storage/blobstor/fstree/fstree.go
roman-khimov: That didn't change anything in fact, `prm.DontCompress` was an _external_ signal, while `t.Compress` can always do whatever it wants to internally.

One thing that bugs me is that our `blobstor` is not really a BLOB storage system, it handles compression at the storage level (why?), it deals with objects for whatever reason. Would be much nicer to see some `[]byte` in/out here and handle compression at engine level. But that's all a different story.

PR #3043, opened 2024-12-06
Fix/new epoch handling

carpawell:


roman-khimov (approved):

carpawell:
Did not add intentionally since was not sure it was a "user" information but let them know about notification changes at least.

cthulhu-rider (approved):

PR #3044, opened 2024-12-10
Feat/push object meta to contracts

carpawell:


pkg/innerring/processors/netmap/process_epoch.go
```diff
@@ -65,14 +68,64 @@ func (np *Processor) processNewEpoch(ev netmapEvent.NewEpoch) {
...
 }
 
+func (np *Processor) updatePlacementInContract(nm netmap.NetMap, l *zap.Logger) error {
+	// TODO: https://github.com/nspcc-dev/neofs-node/issues/3045
+	cids, err := np.containerWrp.List(nil)
+	if err != nil {
+		return fmt.Errorf("can't get containers list: %w", err)
+	}
+
+	for _, cID := range cids {
```
carpawell, line 94: @roman-khimov, i think this is the only thing that we may discuss: IMO if netmap is changed, many containers change their placement and it is easier to recalculate them all once in a single epoch (and not touch them for many-many epochs) than to cache the whole map from the previous epoch OR to fetch it, calculate and compare every vector in every container to see if anything has changed

pkg/innerring/processors/netmap/cleanup_table.go
```diff
@@ -64,6 +68,14 @@ func (c *cleanupTable) update(snapshot netmap.NetMap, now uint64) {
 	}
 
 	c.lastAccess = newMap
+
+	// order is expected to be the same from epoch to epoch
```
cthulhu-rider, line 72: can u pls remind what ensures this?

pkg/innerring/processors/netmap/process_epoch.go
```diff
@@ -65,14 +69,88 @@ func (np *Processor) processNewEpoch(ev netmapEvent.NewEpoch) {
...
+		l := l.With(zap.String("cid", cID.String()))
```
cthulhu-rider, line 95: Stringer

pkg/innerring/processors/netmap/process_epoch.go
```diff
@@ -65,14 +69,88 @@ func (np *Processor) processNewEpoch(ev netmapEvent.NewEpoch) {
...
+			continue
+		}
+
+		err = np.putPlacementVectors(cID, policy, vectors)
+		if err != nil {
+			l.Error("can't put placement vectors to Container contract", zap.Error(err))
+			continue
+		}
+
+		l.Debug("updated container placement in Container contract")
```
cthulhu-rider, line 123: at this point, only the request has been sent, are we sure that it will succeed?

pkg/innerring/processors/netmap/cleanup_table.go
```diff
@@ -64,6 +68,14 @@ func (c *cleanupTable) update(snapshot netmap.NetMap, now uint64) {
...
+	mapChanged := !slices.EqualFunc(c.prev.Nodes(), nmNodes, func(i1 netmap.NodeInfo, i2 netmap.NodeInfo) bool {
+		return bytes.Equal(i1.PublicKey(), i2.PublicKey())
+	})
+	c.prev = snapshot
```
cthulhu-rider, line 76: only nodes are needed

pkg/morph/client/container/meta.go
```diff
@@ -0,0 +1,27 @@
...
+	"fmt"
+
+	"github.com/nspcc-dev/neofs-node/pkg/morph/client"
+)
+
+// SubmitObjectPut puts object meta information.
+//
+// Returns any error encountered that caused the saving to interrupt.
+func (c *Client) SubmitObjectPut(meta []byte, sigs [][]byte) error {
+	if len(meta) == 0 || len(sigs) == 0 {
```
cthulhu-rider, line 14: so `sigs` required? https://github.com/nspcc-dev/neofs-contract/pull/448#discussion_r1864201887

pkg/services/object/put/distributed.go
```diff
@@ -10,6 +10,7 @@ import (
 
 	"github.com/nspcc-dev/neofs-node/pkg/core/client"
 	"github.com/nspcc-dev/neofs-node/pkg/core/object"
+	chaintcontainer "github.com/nspcc-dev/neofs-node/pkg/morph/client/container"
```
cthulhu-rider, line 13: chaint?

pkg/services/object/put/distributed.go
```diff
@@ -143,7 +149,22 @@ func (t *distributedTarget) Close() (oid.ID, error) {
...
 		lockedObjs, t.obj.CreationEpoch(), t.networkMagicNumber)
 	id := t.obj.GetID()
-	return id, t.placementIterator.iterateNodesForObject(id, t.sendObject)
+	err := t.placementIterator.iterateNodesForObject(id, t.sendObject)
+	if err != nil {
+		return oid.ID{}, err
+	}
+
+	if t.localNodeInContainer {
+		t.metaMtx.RLock()
```
cthulhu-rider, line 159: there is no more concurrent access at this point, isn't it?

pkg/services/object/put/distributed.go
```diff
@@ -143,7 +149,22 @@ func (t *distributedTarget) Close() (oid.ID, error) {
...
+		defer t.metaMtx.RUnlock()
+
+		err = t.cnrClient.SubmitObjectPut(t.objSharedMeta, t.collectedSignatures)
+		if err != nil {
+			t.placementIterator.log.Info("submitting meta information", zap.Stringer("oid", id), zap.Error(err))
```
cthulhu-rider, line 173: `failed to submit`

pkg/services/object/put/distributed.go
```diff
@@ -154,7 +155,7 @@ func (t *distributedTarget) Close() (oid.ID, error) {
 		return oid.ID{}, err
 	}
 
-	if t.localNodeInContainer {
+	if t.localNodeInContainer && t.metainfoConsistencyAttr != "" {
```
cthulhu-rider, line 158: could u pls remind is `t.metainfoConsistencyAttr != ""` == enabled meta-on-chain?

pkg/services/object/put/distributed.go
```diff
@@ -33,7 +33,8 @@ type distributedTarget struct {
 	objMeta            object.ContentMeta
 	networkMagicNumber uint32
 
-	cnrClient *chaintcontainer.Client
+	cnrClient               *chaintcontainer.Client
+	metainfoConsistencyAttr string
```
cthulhu-rider, line 37: now it's handled as boolean, will behavior ever depend on the particular value?

cthulhu-rider (commented):
sequence of commits is inconvenient for review. Some subsequent commits answer questions about the previous ones, and sometimes it seems they even fix them. And looks like the whole desired behavior achieved only by having them all at once

anyway, i'll leave all the comments accumulated by the middle just in case

pkg/morph/client/container/nodes.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"fmt"
+
+	"github.com/nspcc-dev/neofs-node/pkg/morph/client"
+)
+
+// AddNextEpochNodes registers public keys as a container's placement vector
+// with specified index. Registration must be finished with final
+// [Client.CommitContainerListUpdate] call. Always sends a notary request with
+// Alphabet multi-signature.
```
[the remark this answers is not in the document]
carpawell: and is always signed by one of the Alphabet node in fact. i added this like because we have `AsAlphabet` and smth like this option for the client. this line highlights that it is always sent by the alphabet node, SN and any other client should not so this

pkg/morph/client/container/nodes.go
```diff
@@ -0,0 +1,50 @@
...
+func (c *Client) AddNextEpochNodes(cid []byte, placementIndex int, nodesKeys [][]byte) error {
```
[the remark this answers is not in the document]
carpawell: no specific idea. since it is a real _go_ code, i wanted indexes to be `int` while REPs in our SDK are `uint32`. i wanted it to be something special but there are no types in neo VM. i even had to change something on the contract side just because it did not compile
about this thread: it was not a mistake, made intentionally, if you do not agree, ping me one more time and i will change

pkg/innerring/processors/netmap/cleanup_table.go
carpawell: netmap are candidates that are stored in storage by their pub key, so it is sorted for sure. then it is `std.Serialize`. @roman-khimov, is array serialization in vm stable?

pkg/innerring/processors/netmap/process_epoch.go
```diff
@@ -65,14 +69,88 @@ func (np *Processor) processNewEpoch(ev netmapEvent.NewEpoch) {
...
+	for i, vector := range vectors {
+		replicas = append(replicas, p.ReplicaNumberByIndex(i))
+
+		err := np.containerWrp.AddNextEpochNodes(cID[:], i, pubKeys(vector))
+		if err != nil {
+			return fmt.Errorf("can't add %d placement vector to Container contract: %w", i, err)
+		}
+	}
+
+	err := np.containerWrp.CommitContainerListUpdate(cID[:], replicas)
```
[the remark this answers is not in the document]
carpawell: no, there was not, that is why after internal discussion, there is a single TX for it. Whether everything ok, or not

pkg/innerring/processors/netmap/process_epoch.go
carpawell: idk, it is our common approach. i do not remember that we are waiting for TX in any of our handlers

> are we sure that it will succeed?

it is now easy to wait. not sure we need it. it is a potential blocker. alphabet node did everything it can

pkg/morph/client/container/meta.go
carpawell: your link is a unit test for contract API. it was tested with not-a-real container with no real REP values, so it did not require any signatures (kinda ok: you said you wanted no REP, the contract proved that at least 0 REP is validated)

on the other hand, this is a client side that wants to prove that somebody stores an object (non-zero REP, otherwise -- why?). it is senseless to provide no signatures. if REP > 0 and you send no signatures, it will fail

pkg/innerring/processors/netmap/cleanup_table.go
carpawell: imo not a problem. it is only an epoch, a single `int` more. snapshot with epoch is even better. theoretically, we can log what is the difference, b/w what epochs

pkg/services/object/put/distributed.go
carpawell: that is right but still, mutex saves _exactly this field_ so lets read it safely too, if you don't mind. mb, it will save us in future changes

pkg/services/object/put/distributed.go
carpawell: in practice yes. but non-zero also varies so i added a `switch` later

pkg/services/object/put/distributed.go
carpawell: yes, in the latter commit. hm, i see, lets squash some commits

carpawell:
> sequence of commits is inconvenient for review

IMO, it is the opposite, i tired to highlight step by step what should be changed to be possible it to work. Squashed few commits not to repeat some things

> And looks like the whole desired behavior achieved only by having them all at once

Some extend morph client without behavior changes. Some prepare IR. Some prepare SN.

pkg/innerring/processors/container/process_container.go
```diff
@@ -181,6 +210,14 @@ func (cp *Processor) approveDeleteContainer(e *containerEvent.Delete) {
 		cp.log.Error("could not approve delete container",
 			zap.Error(err),
 		)
+		return
+	}
+
+	cID := cid.ID(e.ContainerID())
+	err = processors.UpdatePlacementVectors(cID, cp.cnrClient, nil, nil)
```
roman-khimov: Why? It should be deleted by contract right with the container itself.

pkg/innerring/processors/netmap/cleanup_table.go
roman-khimov: VM is stable in every aspect of it given that inputs/outputs are the same.

pkg/innerring/processors/netmap/process_epoch.go
roman-khimov: Works for now.

pkg/innerring/processors/container/process_container.go
carpawell: well, this should be done then on the contract side, will do

carpawell:
Added meta validity based on blocks, not epochs.

pkg/innerring/processors/container/process_container.go
carpawell: https://github.com/nspcc-dev/neofs-contract/pull/456

pkg/services/object/put/distributed.go
```diff
@@ -199,36 +204,86 @@ func (t *distributedTarget) sendObject(node nodeDesc) error {
...
 
-		if !bytes.Equal(sig.PublicKeyBytes(), node.info.PublicKey()) {
-			l.Info("public key differs in object meta signature")
+		for i, sig := range sigs {
+			if !bytes.Equal(sig.PublicKeyBytes(), node.info.PublicKey()) {
+				l.Info("public key differs in object meta signature", zap.Int("signature index", i))
+				continue
+			}
+
+			if !sig.Verify(t.objSharedMeta) {
```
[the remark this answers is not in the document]
carpawell: @roman-khimov, hm, that is right. but i am not sure what else can be done here, we cannot collect two different signatures since we cannot send half signatures for the "outdated" nodes and half for the "actual" one, server-side always needs to get a single version for vub (or we need contract updates again). i can suggest sending three signatures and the server can always try `currentEpoch + 1` (btw, this is even better, since we can get an object in the last block of the epoch, and then handling may break in the next block on the contract side), which handles every kind of epoch divergence that is not bigger than 1 epoch

pkg/innerring/processors/container/process_container.go
```diff
@@ -110,6 +114,31 @@ func (cp *Processor) approvePutContainer(ctx *putContainerContext) {
...
+		cp.log.Error("could not build placement for Container contract update", zap.Stringer("cid", ctx.cID), zap.Error(err))
+		return
+	}
+
+	replicas := make([]uint32, 0, policy.NumberOfReplicas())
+	for i := range vectors {
+		replicas = append(replicas, policy.ReplicaNumberByIndex(i))
+	}
+
+	err = processors.UpdatePlacementVectors(ctx.cID, cp.cnrClient, vectors, replicas)
```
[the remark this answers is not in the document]
carpawell: hm, added waiting for the container first, i think this is the best we can do here

pkg/morph/client/container/nodes.go
```diff
@@ -49,3 +54,27 @@ func (c *Client) CommitContainerListUpdate(cid cid.ID, replicas []uint32) error
...
+
+func pubKeys(nodes []netmap.NodeInfo) [][]byte {
+	res := make([][]byte, 0, len(nodes))
+	for _, node := range nodes {
+		res = append(res, node.PublicKey())
+	}
+
+	// arrays take parts in transaction that should be multi-singed, so order
+	// is important to be the same
+	slices.SortFunc(res, func(a, b []byte) int {
```
roman-khimov, line 66: `slices.SortFunc(res, bytes.Compare)`

roman-khimov (approved):

PR #3048, opened 2024-12-11
meta: fix upgrade to v3

roman-khimov:


carpawell:
For now, the only idea of how this could happen is restarting before migration is completed.

roman-khimov:
But it all happens in a single transaction, either it's in or it's out.

carpawell (approved):
Hm, this explains a lot.

roman-khimov:
The key is 2addebfe172d3a036eb3259bc652c9d609e4cde5, yes. And that was present in the original patch as well.

pkg/local_object_storage/metabase/version_test.go
```diff
@@ -307,16 +308,20 @@ func TestMigrate2to3(t *testing.T) {
...
-
-		return migrateFrom2Version(db, tx)
+		return updateVersion(tx, 2)
 	})
 	require.NoError(t, err)
 
+	db.mode = mode.DegradedReadOnly // Force reload.
+	ok, err := db.Reload(WithPath(db.info.Path), WithEpochState(epochState{}))
+	require.True(t, ok)
+	require.NoError(t, err)
```
cthulhu-rider, line 320: better to check before `ok` to know what exactly failed if anything

cthulhu-rider (commented):
@roman-khimov d34687cecfe016caae3b37b93a7c37f2a3f4f63a worth changelog to me, very admin-friendly change

cthulhu-rider (approved):

roman-khimov:
> @roman-khimov [d34687c](https://github.com/nspcc-dev/neofs-node/commit/d34687cecfe016caae3b37b93a7c37f2a3f4f63a) worth changelog to me, very admin-friendly change

I don't think it changes anything for the admin, especially now. This would be a (really minor) issue for us when migration from 3 to 4 is to appear, old structure can handle this too, but it's somewhat less convenient.

pkg/local_object_storage/metabase/version_test.go
roman-khimov: Yes, that's the way I got to the right parameters, but it somehow changed before hitting GH. Fixed.

Issue #1866, opened 2022-10-10
Add IR control service call for node removal


realloc:
For testing purposes and for on-prem installations it may be useful to quickly remove a storage node from Netmap.
This must be done via Control API call and produce a notary type Tx invocation of corresponding contract method.

Technically it could be a temporary in-memory list of nodes marked for removal on each IR node so that required IR majority could collect the multisig.

carpawell:
Why not `neofs-adm`?

realloc:
This feature has to work without access to all IR keys. For example, as a reaction from some external monitoring system.
Possible use case:
- Some IR companion daemon detects a Storage Node misbehaving or dead
- If it makes no sense to wait for the timeout, it adds that node to the removal list using local IR control API
- Subjected node gets removed. Maybe with optional Epoch change.

carpawell:
What that Call is expected to do? Just send a notary request? So we do not know all the alphabet keys but are able to send n*2/3+1 control calls to the Alphabet nodes (in practice -- with using their keys), right?

realloc:
In practice this should be a collective effort. Each IR manages its local list and then the IR majority can do the force removal.

roman-khimov:
Similar to #2661 in many ways.

roman-khimov:
Depends on #2088.

roman-khimov:
Done

PR #3059, opened 2024-12-19
control: add network settings management via ir control

### LINKED ISSUE #2088: Allow IR to change network parameters
Possible solution:
1. Initiate notary request via control API
2. Other nodes can list and accept pending requests via the same API.

End-rey:
Closes #2088, #1866.


End-rey:
I have implemented the following methods: `newEpoch`, `setConfig` and `removeNode`. What else needs to be done?
 
When calling the `list` command, only hashes are given. Is it necessary to display the information in more detail?

pkg/morph/client/notary.go
```diff
@@ -460,21 +462,21 @@ func (c *Client) notaryInvoke(committee, invokedByAlpha bool, contract util.Uint
...
 	}
 
 	nAct, err := notary.NewActor(conn.client, cosigners, c.acc)
 	if err != nil {
-		return err
+		return "", err
 	}
 
-	mainH, fbH, untilActual, err := nAct.Notarize(nAct.MakeTunedCall(contract, method, nil, func(r *result.Invoke, t *transaction.Transaction) error {
+	tx, err := nAct.MakeTunedCall(contract, method, nil, func(r *result.Invoke, t *transaction.Transaction) error {
```
carpawell, line 479: why not just use `mainH`? no need in change as i see

pkg/services/control/ir/service.proto
```diff
@@ -41,3 +50,89 @@ message HealthCheckResponse {
...
+
+    // Body signature.
+    Signature signature = 2;
+}
+
+// NotaryRequest request.
+message NotaryRequestRequest {
+    // Request body structure.
+    message Body {
+        string method = 1;
```
carpawell, line 86: maybe it would be better to have a more generic `contractHash, method, args` or maybe not @roman-khimov, @cthulhu-rider

pkg/services/control/ir/service.proto
```diff
@@ -41,3 +50,89 @@ message HealthCheckResponse {
...
+    // Body signature.
+    Signature signature = 2;
+}
+
+// NotarySign request.
+message NotarySignRequest {
+    // Request body structure.
+    message Body {
+        // Hash of transaction that need to be signed.
+        string hash = 1;
```
carpawell, line 117: i think raw representation would be better as we usually prefer in our protobuf protocol cc @roman-khimov

pkg/services/control/ir/types.proto
```diff
@@ -30,3 +30,12 @@ enum HealthStatus {
...
 }
+
+// Info about transaction.
+message TransactionInfo {
+    // Script that contains in the transaction.
+    bytes script = 1;
+
+    // Hash of transaction.
+    string hash = 2;
+}
```
carpawell, line 41: add empty line before EOF, please

pkg/services/control/ir/types.proto
```diff
@@ -30,3 +30,12 @@ enum HealthStatus {
     // Initializing Neo network
     INITIALIZING_NETWORK = 4;
 }
+
+// Info about transaction.
+message TransactionInfo {
+    // Script that contains in the transaction.
+    bytes script = 1;
```
carpawell, line 37: it is not used now, right? do we need it?

pkg/services/control/ir/types.proto
```diff
@@ -30,3 +30,12 @@ enum HealthStatus {
...
+
+    // Hash of transaction.
+    string hash = 2;
```
carpawell, line 40: i would also prefer raw hash here

carpawell:
> Is it necessary to display the information in more detail?

I think it will be useful in the future, we may leave an issue or do this now. [notaryPreparator](https://github.com/nspcc-dev/neofs-node/blob/373a6e046ce93fa76863b63bedbe6c3e371856f6/pkg/morph/event/notary_preparator.go#L84) can simplify it significantly.

pkg/morph/client/notary.go
```diff
@@ -305,13 +305,13 @@ func (c *Client) UpdateNeoFSAlphabetList(alphas keys.PublicKeys, txHash util.Uin
...
 
 // NotaryInvoke invokes contract method by sending tx to notary contract in
-// blockchain. Fallback tx is a `RET`. If Notary support is not enabled
-// it fallbacks to a simple `Invoke()`.
+// blockchain and returns the hash of tx. Fallback tx is a `RET`. If Notary support is not enabled
+// it fallbacks to a simple `Invoke()`, but doesn't return the hash of tx.
 //
 // `nonce` and `vub` are used only if notary is enabled.
-func (c *Client) NotaryInvoke(contract util.Uint160, fee fixedn.Fixed8, nonce uint32, vub *uint32, method string, args ...any) error {
+func (c *Client) NotaryInvoke(contract util.Uint160, fee fixedn.Fixed8, nonce uint32, vub *uint32, method string, args ...any) (string, error) {
```
roman-khimov, line 312: Why are you returning this hash as a string? It's `util.Uint256` naturally.

pkg/services/control/ir/server/calls.go
```diff
@@ -32,3 +33,90 @@ func (s *Server) HealthCheck(_ context.Context, req *control.HealthCheckRequest)
...
+	if err := s.isValidRequest(req); err != nil {
+		return nil, status.Error(codes.PermissionDenied, err.Error())
+	}
+
+	resp := new(control.NotarySignResponse)
+
+	body := new(control.NotarySignResponse_Body)
+	resp.SetBody(body)
+
+	hash, err := util.Uint256DecodeStringBE(string(req.GetBody().GetHash()))
```
roman-khimov, line 108: Pass it as binary, don't mess with strings in a binary protocol.

pkg/innerring/network.go
```diff
@@ -0,0 +1,181 @@
...
+
+func (s *Server) ListNotaryRequests() ([]string, error) {
+	hashesMap, err := s.netmapClient.Morph().GetRawNotaryPool()
+	if err != nil {
+		return nil, err
+	}
+
+	result := make([]string, 0, len(hashesMap.Hashes))
+	for k := range hashesMap.Hashes {
+		result = append(result, k.String())
```
roman-khimov, line 37: Why stringified? `[]util.Uint256` that is.

pkg/services/control/ir/service.proto
roman-khimov: Let's keep it simple for now, can be extended in future.

pkg/innerring/network.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+	return result, nil
+}
+
+func (s *Server) RequestNotary(method string, args ...string) (hash string, err error) {
```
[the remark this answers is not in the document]
cthulhu-rider: write docs to know that this string is instead

pkg/innerring/network.go
```diff
@@ -0,0 +1,175 @@
...
+	"strconv"
+
+	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
+	"github.com/nspcc-dev/neo-go/pkg/util"
+	"github.com/nspcc-dev/neofs-contract/rpc/netmap"
+	netmapcore "github.com/nspcc-dev/neofs-node/pkg/morph/client/netmap"
+	"go.uber.org/zap"
+)
+
+func (s *Server) ListNotaryRequests() ([]util.Uint256, error) {
```
roman-khimov, line 15: Exported methods should be documented (all of them here).

cmd/neofs-cli/modules/control/notary_request.go
```diff
@@ -0,0 +1,122 @@
...
+		if len(args) > 0 {
+			cmd.Println("method 'newEpoch', but the args provided, they will be ignored")
+		}
+	case "setConfig":
+		if len(args) != 1 {
+			return fmt.Errorf("invalid number of args provided for 'setConfig', expected 1, got %d", len(args))
+		}
+
+		bodyArgs := make([]string, 0, len(args)*2)
+		for _, arg := range args {
```
roman-khimov, line 67: Looks like `len(args) == 1` above, this can be simplified.

cmd/neofs-cli/modules/control/notary_request.go
```diff
@@ -0,0 +1,122 @@
...
+			}
+
+			bodyArgs = append(bodyArgs, k, fmt.Sprint(v))
+		}
+		body.SetArgs(bodyArgs)
+	case "removeNode":
+		if len(args) != 1 {
+			return fmt.Errorf("invalid number of args provided for 'removeNode', expected 1, got %d", len(args))
+		}
+		body.SetArgs(args)
```
roman-khimov, line 80: Check (deserialize) the key and push it in binary, please.

pkg/innerring/network.go
```diff
@@ -0,0 +1,175 @@
...
+		if len(args) == 0 {
+			return util.Uint256{}, errors.New("empty node key")
+		}
+		if len(args) > 1 {
+			s.log.Warn("1 public key expected, all but the first one will be ignored",
+				zap.String("method", method),
+				zap.Strings("args", args))
+		}
+
+		nodePubKey, err := keys.NewPublicKeyFromString(args[0])
```
roman-khimov, line 82: Because working with strings here is not the best thing to do.

cmd/neofs-cli/modules/control/notary_request.go
End-rey: So, do I need to make `args` in the request in bytes for this?

roman-khimov: You mean protobuf-wise? Maybe, I'm not sure what's the protobuf difference between string and bytes. Anyway we don't want to pass hex values over this interface.

cthulhu-rider: the only diff is that `string` must be UTF-8 data https://protobuf.dev/programming-guides/encoding/#cheat-sheet-key. And `string` is encoded exactly like `bytes` https://protobuf.dev/programming-guides/encoding/#structure (https://github.com/nspcc-dev/neofs-sdk-go/blob/5f693d22cfe6350060e546e383c293317253352a/internal/proto/encoding.go#L163-L166)

roman-khimov: Kinda as expected. Then `bytes` that is.

roman-khimov (approved):

PR #3063, opened 2024-12-23
Feat/add big objects support for meta on chain

carpawell:
~~Blocked by #3044.~~

### DIFF
--- a/pkg/core/object/replicate.go
+++ b/pkg/core/object/replicate.go
@@ -19,7 +19,7 @@ const (
 	sizeKey         = "size"
 	deletedKey      = "deleted"
 	lockedKey       = "locked"
-	validUntilKey   = "validuntil"
+	validUntilKey   = "validUntil"
 )
 
 // EncodeReplicationMetaInfo uses NEO's map (strict order) serialized format as a raw
@@ -33,7 +33,7 @@ const (
 //	"size": payload size
 //	"deleted": array of _raw_ object IDs
 //	"locked": array of _raw_ object IDs
-//	"validuntil": last valid block number for meta information
+//	"validUntil": last valid block number for meta information
 //
 // Last valid epoch is object's creation epoch + 10.
 func EncodeReplicationMetaInfo(cID cid.ID, oID oid.ID, pSize uint64,
--- a/pkg/core/object/replicate.go
+++ b/pkg/core/object/replicate.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/nspcc-dev/neo-go/pkg/vm/stackitem"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
+	objectsdk "github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 )
 
@@ -21,6 +22,7 @@ const (
 	previousPartKey = "previousPart"
 	deletedKey      = "deleted"
 	lockedKey       = "locked"
+	typeKey         = "type"
 )
 
 // EncodeReplicationMetaInfo uses NEO's map (strict order) serialized format as a raw
@@ -37,7 +39,8 @@ const (
 //	"previousPart": [OPTIONAL] _raw_ object ID (32 bytes)
 //	"deleted": [OPTIONAL] array of _raw_ object IDs
 //	"locked": [OPTIONAL] array of _raw_ object IDs
-func EncodeReplicationMetaInfo(cID cid.ID, oID, firstPart, previousPart oid.ID, pSize uint64,
+//	"type": [OPTIONAL] object type enumeration
+func EncodeReplicationMetaInfo(cID cid.ID, oID, firstPart, previousPart oid.ID, pSize uint64, typ objectsdk.Type,
 	deleted, locked []oid.ID, vub uint64, magicNumber uint32) []byte {
 	kvs := []stackitem.MapElement{
 		kv(cidKey, cID[:]),
@@ -58,6 +61,9 @@ func EncodeReplicationMetaInfo(cID cid.ID, oID, firstPart, previousPart oid.ID, 
 	}
 	if len(locked) > 0 {
 		kvs = append(kvs, oidsKV(lockedKey, locked))
+	}
+	if typ != objectsdk.TypeRegular {
+		kvs = append(kvs, kv(typeKey, uint32(typ)))
 	}
 
 	result, err := stackitem.Serialize(stackitem.NewMapWithValue(kvs))
--- a/pkg/core/object/replicate.go
+++ b/pkg/core/object/replicate.go
@@ -9,17 +9,16 @@ import (
 )
 
 const (
-	currentVersion = 7 // it is also a number of fields
-)
-
-const (
-	networkMagicKey = "network"
+	// required fields.
 	cidKey          = "cid"
 	oidKey          = "oid"
 	sizeKey         = "size"
-	deletedKey      = "deleted"
-	lockedKey       = "locked"
 	validUntilKey   = "validUntil"
+	networkMagicKey = "network"
+
+	// optional fields.
+	deletedKey = "deleted"
+	lockedKey  = "locked"
 )
 
 // EncodeReplicationMetaInfo uses NEO's map (strict order) serialized format as a raw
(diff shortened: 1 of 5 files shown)


pkg/network/transport/object/grpc/replication.go
```diff
@@ -206,6 +206,12 @@ func objectFromMessage(gMsg *objectGRPC.Object) (*object.Object, error) {
 }
 
 func (s *Server) metaInfoSignature(o object.Object) ([]byte, error) {
+	firstObj, _ := o.FirstID()
+	if firstObj.IsZero() && o.HasParent() && o.SplitID() == nil {
```
cthulhu-rider: last operand is inobvious to me. How about create a helper func for this? U check this condition twice here

pkg/core/object/replicate.go
```diff
@@ -38,14 +38,46 @@ const (
 //	"validuntil": last valid block number for meta information
 //
 // Last valid epoch is object's creation epoch + 10.
-func EncodeReplicationMetaInfo(cID cid.ID, oID, firstPart oid.ID, pSize uint64,
-	deleted, locked []oid.ID, vub uint64, magicNumber uint32) []byte {
+func EncodeReplicationMetaInfo(o object.Object, vub uint64, magicNumber uint32) ([]byte, error) {
```
cthulhu-rider, line 41: callers aint happy about this change. One was completely okay w/o an error, while the second forced to do multiple decoding. If u wanna decrease prm num, typedef a structure

pkg/network/transport/object/grpc/replication.go
```diff
@@ -206,50 +205,24 @@ func objectFromMessage(gMsg *objectGRPC.Object) (*object.Object, error) {
...
 	thirdBlock := secondBlock + currentEpochDuration
 
-	firstMeta := objectcore.EncodeReplicationMetaInfo(o.GetContainerID(), o.GetID(), firstObj, o.PayloadSize(), deleted, locked, firstBlock, s.mNumber)
-	secondMeta := objectcore.EncodeReplicationMetaInfo(o.GetContainerID(), o.GetID(), firstObj, o.PayloadSize(), deleted, locked, secondBlock, s.mNumber)
-	thirdMeta := objectcore.EncodeReplicationMetaInfo(o.GetContainerID(), o.GetID(), firstObj, o.PayloadSize(), deleted, locked, thirdBlock, s.mNumber)
+	firstMeta, err := objectcore.EncodeReplicationMetaInfo(o, firstBlock, s.mNumber)
+	if err != nil {
+		return nil, fmt.Errorf("encoding metadata")
+	}
+	secondMeta, _ := objectcore.EncodeReplicationMetaInfo(o, secondBlock, s.mNumber)
```
cthulhu-rider, line 218: one more proof of the refactoring dubiousness

pkg/core/object/replicate.go
```diff
@@ -30,18 +31,20 @@ const (
...
 //
 // Last valid epoch is object's creation epoch + 10.
-func EncodeReplicationMetaInfo(cID cid.ID, oID oid.ID, pSize uint64,
+func EncodeReplicationMetaInfo(cID cid.ID, oID, firstPart oid.ID, pSize uint64,
 	deleted, locked []oid.ID, vub uint64, magicNumber uint32) []byte {
 	kvs := []stackitem.MapElement{
 		kv(networkMagicKey, magicNumber),
 		kv(cidKey, cID[:]),
 		kv(oidKey, oID[:]),
+		kv(firstPartKey, firstPart[:]),
```
cthulhu-rider: i dont suggest to push missing `firstPart` as zero ID, it's a reserved invalid value. Push emptiness instead: this will match the zero length condition from ur contract change https://github.com/nspcc-dev/neofs-contract/pull/457/commits/416936f4c6344c5745ae863e25fb0ee5c23e708c

pkg/network/transport/object/grpc/replication.go
carpawell: this is just a context of this func: you need to repeat the same encoding 3 times, error happens for all or does not happen at all

ok, it is easier to drop this commit

pkg/core/object/replicate.go
carpawell: > typedef a structure

object.Object is such a struct, why not?

pkg/core/object/replicate.go
carpawell: added optional fields support. `firstPart` is now one of them

pkg/core/object/replicate.go
```diff
@@ -9,41 +9,45 @@ import (
...
 // This (ordered) format is used (keys are strings):
 //
-//	"network": network magic
 //	"cid": _raw_ container ID (32 bytes)
 //	"oid": _raw_ object ID (32 bytes)
 //	"size": payload size
-//	"deleted": array of _raw_ object IDs
-//	"locked": array of _raw_ object IDs
 //	"validUntil": last valid block number for meta information
+//	"network": network magic
```
roman-khimov, line 37: Better keep them alphabetically sorted (in optional/non-optional groups).

pkg/core/object/replicate.go
```diff
@@ -9,41 +9,45 @@ import (
...
+//	"deleted": [OPTIONAL] array of _raw_ object IDs
+//	"locked": [OPTIONAL] array of _raw_ object IDs
 func EncodeReplicationMetaInfo(cID cid.ID, oID oid.ID, pSize uint64,
 	deleted, locked []oid.ID, vub uint64, magicNumber uint32) []byte {
 	kvs := []stackitem.MapElement{
-		kv(networkMagicKey, magicNumber),
```
roman-khimov, line 42: Why moving it?

pkg/core/object/replicate.go
```diff
@@ -59,6 +62,9 @@ func EncodeReplicationMetaInfo(cID cid.ID, oID, firstPart, previousPart oid.ID,
 	if len(locked) > 0 {
 		kvs = append(kvs, oidsKV(lockedKey, locked))
 	}
+	if typ != objectsdk.TypeRegular {
+		kvs = append(kvs, kv(typeKey, uint32(typ)))
```
roman-khimov, line 66: Notice that tombstone/lock can be detected by deleted/locked already. But this covers more things, so OK.

pkg/network/transport/object/grpc/replication.go
cthulhu-rider: check is OK, i just suggest to share and doc it. Again, if i'd need to check whether object is first or not, i'd miss split ID check

pkg/core/object/replicate.go
cthulhu-rider: object has no `deleted, locked []oid.ID` fields

pkg/core/object/replicate.go
```diff
@@ -9,41 +9,45 @@ import (
...
-const (
-	networkMagicKey = "network"
+	// required fields.
 	cidKey          = "cid"
 	oidKey          = "oid"
 	sizeKey         = "size"
-	deletedKey      = "deleted"
-	lockedKey       = "locked"
 	validUntilKey   = "validUntil"
+	networkMagicKey = "network"
```
cthulhu-rider, line 18: motivated movement?

pkg/core/object/replicate.go
cthulhu-rider: > in optional/non-optional groups)

:+1:

>  alphabetically sorted

:-1:. Excessive load to the dev brain "where to place new key?". And i like how they are grouped by type currently

pkg/core/object/replicate.go
```diff
@@ -9,41 +9,45 @@ import (
...
 		kv(cidKey, cID[:]),
 		kv(oidKey, oID[:]),
 		kv(sizeKey, pSize),
-		oidsKV(deletedKey, deleted),
-		oidsKV(lockedKey, locked),
 		kv(validUntilKey, vub),
+		kv(networkMagicKey, magicNumber),
+	}
+
+	if len(deleted) > 0 {
```
cthulhu-rider, line 59: i'd split unit tests into `min` and `full` cases. And include fields added in the following commits

pkg/core/object/replicate.go
```diff
@@ -43,6 +47,12 @@ func EncodeReplicationMetaInfo(cID cid.ID, oID oid.ID, pSize uint64,
 		kv(networkMagicKey, magicNumber),
 	}
 
+	if !firstPart.IsZero() {
```
cthulhu-rider, line 53: and test these too

pkg/core/object/replicate.go
cthulhu-rider, line 66: although `typ` is `uint32`, lets to cast to `int32`. It will be `int32` soon (im also unhappy bout this, but wcyd). We'll most likely forget to change conversion later

carpawell: ok

pkg/core/object/replicate.go
carpawell: done

pkg/core/object/replicate.go
carpawell: the same as in https://github.com/nspcc-dev/neofs-node/pull/3063#discussion_r1897966594

ok-ok, waste of time, more code to review, etc. but i feel better this way

pkg/core/object/replicate.go
carpawell: you mean lock type with some non-empy `deleted` should be an error (panic in fact) in this func?

pkg/core/object/replicate.go
carpawell: we had no release yet, and ordering only matters in this func so i wanted it to be the same as in docs. as soon as we agree with the thread above, the order still means something: it will always be like this in VM code, so why not

pkg/core/object/replicate.go
carpawell: > Better keep them alphabetically sorted

well, my suggestion is not a random order, it was an intention. i grouped it like "logically" or as @cthulhu-rider said, currently, it is by type too. do you have any pros as to why it is better to sort them? new value will be harder to add, and alphabet sorting has no advantages (or change my mind)

pkg/core/object/replicate.go
carpawell: > object has no deleted, locked []oid.ID fields

yes, but `object.Object` is enough to get this information and it will be always gotten like this

this commit was already dropped anyway

pkg/network/transport/object/grpc/replication.go
carpawell: > i'd miss split ID check

as is said, this check is the only thing that differs v1 from v2. but if drop code is requested i usually support it, dropped

pkg/network/transport/object/grpc/replication.go
```diff
@@ -206,9 +206,17 @@ func objectFromMessage(gMsg *objectGRPC.Object) (*object.Object, error) {
 }
 
 func (s *Server) metaInfoSignature(o object.Object) ([]byte, error) {
+	firstObj := o.GetFirstID()
+	if o.HasParent() && !firstObj.IsZero() {
```
roman-khimov, line 210: Why `!`?

carpawell: oh, sorry, fast "fix" while rebasing, it was correct before

roman-khimov (approved):

PR #3080, opened 2025-01-13
Store object metadata in metabases for SearchV2 service

cthulhu-rider:
- [x] store metadata on Put
- [x] unfiltered
- [x] test scenarios for filtered
- [x] filtered
- [x] attributes
- [x] test sorting
- [ ] removal
- [ ] migration

### DIFF
--- a/pkg/local_object_storage/metabase/metadata.go
+++ b/pkg/local_object_storage/metabase/metadata.go
@@ -1,0 +1,796 @@
+package meta
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"math/big"
+	"slices"
+
+	"github.com/google/uuid"
+	"github.com/mr-tron/base58"
+	"github.com/nspcc-dev/neofs-sdk-go/client"
+	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
+	"github.com/nspcc-dev/neofs-sdk-go/object"
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	"github.com/nspcc-dev/neofs-sdk-go/user"
+	"github.com/nspcc-dev/neofs-sdk-go/version"
+	"github.com/nspcc-dev/tzhash/tz"
+	"go.etcd.io/bbolt"
+)
+
+const (
+	metaPrefixID = byte(iota)
+	metaPrefixAttrIDInt
+	metaPrefixAttrIDPlain
+	metaPrefixIDAttr
+)
+
+const (
+	intValLen      = 33                              // prefix byte for sign + fixed256 in metaPrefixAttrIDInt
+	attrIDFixedLen = 1 + oid.Size + utf8DelimiterLen // prefix first
+)
+
+var (
+	maxUint256 = new(big.Int).SetBytes([]byte{255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255,
+		255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255})
+	maxUint256Neg = new(big.Int).Neg(maxUint256)
+)
+
+func invalidMetaBucketKeyErr(key []byte, cause error) error {
+	return fmt.Errorf("invalid meta bucket key (prefix 0x%X): %w", key[0], cause)
+}
+
+// TODO: fill on migration.
+// TODO: ROOT and PHY props.
+// TODO: cleaning on obj removal.
+func putMetadata(tx *bbolt.Tx, cnr cid.ID, id oid.ID, ver version.Version, owner user.ID, typ object.Type, creationEpoch uint64,
+	payloadLen uint64, pldHash, pldHmmHash, splitID []byte, parentID, firstID oid.ID, attrs []object.Attribute) error {
+	metaBkt, err := tx.CreateBucketIfNotExists(metaBucketKey(cnr))
+	if err != nil {
+		return fmt.Errorf("create meta bucket for container: %w", err)
+	}
+	idk := [1 + oid.Size]byte{metaPrefixID}
+	copy(idk[1:], id[:])
+	if err := metaBkt.Put(idk[:], nil); err != nil {
+		return fmt.Errorf("put object ID to container's meta bucket: %w", err)
+	}
+
+	var keyBuf keyBuffer
+	if err = putPlainAttribute(metaBkt, &keyBuf, id, object.FilterVersion, ver.String()); err != nil {
+		return err
+	}
+	if err = putPlainAttribute(metaBkt, &keyBuf, id, object.FilterOwnerID, string(owner[:])); err != nil {
+		return err
+	}
+	if err = putPlainAttribute(metaBkt, &keyBuf, id, object.FilterType, typ.String()); err != nil {
+		return err
+	}
+	if err = putIntAttribute(metaBkt, &keyBuf, id, object.FilterCreationEpoch, new(big.Int).SetUint64(creationEpoch)); err != nil {
+		return err
+	}
+	if err = putIntAttribute(metaBkt, &keyBuf, id, object.FilterPayloadSize, new(big.Int).SetUint64(payloadLen)); err != nil {
+		return err
+	}
+	if err = putPlainAttribute(metaBkt, &keyBuf, id, object.FilterPayloadChecksum, string(pldHash)); err != nil {
+		return err
+	}
+	if err = putPlainAttribute(metaBkt, &keyBuf, id, object.FilterPayloadHomomorphicHash, string(pldHmmHash)); err != nil {
+		return err
+	}
+	if len(splitID) > 0 {
+		if err = putPlainAttribute(metaBkt, &keyBuf, id, object.FilterSplitID, string(splitID)); err != nil {
+			return err
+		}
+	}
+	if !firstID.IsZero() {
+		if err = putPlainAttribute(metaBkt, &keyBuf, id, object.FilterFirstSplitObject, string(firstID[:])); err != nil {
+			return err
+		}
+	}
+	if !parentID.IsZero() {
+		if err = putPlainAttribute(metaBkt, &keyBuf, id, object.FilterParentID, string(parentID[:])); err != nil {
+			return err
+		}
+	}
+	for i := range attrs {
+		ak, av := attrs[i].Key(), attrs[i].Value()
+		if n, isInt := parseInt(av); isInt && n.Cmp(maxUint256Neg) >= 0 && n.Cmp(maxUint256) <= 0 {
+			err = putIntAttribute(metaBkt, &keyBuf, id, ak, n)
+		} else {
+			err = putPlainAttribute(metaBkt, &keyBuf, id, ak, av)
+		}
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// SearchCursor is a cursor used for continuous search in the DB.
+type SearchCursor struct {
+	Key      []byte // PREFIX_ATTR_DELIM_VAL_ID, prefix is unset
+	ValIDOff int
+}
+
+// NewSearchCursorFromString decodes cursor from the string according to the
+// primary attribute.
+func NewSearchCursorFromString(s, primAttr string) (*SearchCursor, error) {
+	if s == "" {
+		return nil, nil
+	}
+	b := make([]byte, 1+base64.StdEncoding.DecodedLen(len(s)))
+	n, err := base64.StdEncoding.Decode(b[1:], []byte(s))
+	if err != nil {
+		return nil, fmt.Errorf("decode cursor from Base64: %w", err)
+	}
+	b = b[:1+n]
+	if primAttr == "" {
+		if n != oid.Size {
+			return nil, fmt.Errorf("wrong OID cursor len %d", n)
+		}
+		return &SearchCursor{Key: b}, nil
+	}
+	if n > object.MaxHeaderLen {
+		return nil, fmt.Errorf("cursor len %d exceeds the limit %d", n, object.MaxHeaderLen)
+	}
+	ind := bytes.Index(b[1:], utf8Delimiter) // 1st is prefix
+	if ind < 0 {
+		return nil, errors.New("missing delimiter")
+	}
+	if !bytes.Equal(b[1:1+ind], []byte(primAttr)) {
+		return nil, errors.New("wrong attribute")
+	}
+	var res SearchCursor
+	res.ValIDOff = 1 + len(primAttr) + len(utf8Delimiter)
+	if len(b[res.ValIDOff:]) <= oid.Size {
+		return nil, errors.New("missing value")
+	}
+	res.Key = b
+	return &res, nil
+}
+
+// Search selects up to count container's objects from the given container
+// matching the specified filters.
+func (db *DB) Search(cnr cid.ID, fs object.SearchFilters, attrs []string, cursor *SearchCursor, count uint16) ([]client.SearchResultItem, *SearchCursor, error) {
+	if blindlyProcess(fs) {
+		return nil, nil, nil
+	}
+	var res []client.SearchResultItem
+	var newCursor *SearchCursor
+	var err error
+	if len(fs) == 0 {
+		res, newCursor, err = db.searchUnfiltered(cnr, cursor, count)
+	} else {
+		res, newCursor, err = db.search(cnr, fs, attrs, cursor, count)
+	}
+	if err != nil {
+		return nil, nil, err
+	}
+	return res, newCursor, nil
+}
+
+func (db *DB) search(cnr cid.ID, fs object.SearchFilters, attrs []string, cursor *SearchCursor, count uint16) ([]client.SearchResultItem, *SearchCursor, error) {
+	var res []client.SearchResultItem
+	var newCursor *SearchCursor
+	err := db.boltDB.View(func(tx *bbolt.Tx) error {
+		mb := tx.Bucket(metaBucketKey(cnr))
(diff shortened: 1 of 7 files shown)


pkg/local_object_storage/metabase/util.go
```diff
@@ -4,13 +4,17 @@ import (
...
+	"math/big"
 
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"go.etcd.io/bbolt"
 )
 
+// UTF-8 data separator used in DB.
+var utf8Delimiter = []byte("\xc0\x80") // U+0000
```
roman-khimov, line 16: Have you tried `[]byte{0}`?

cthulhu-rider: zero byte is a utf8 char, so we can miss sep b/w key and val after their concat

roman-khimov: I thought it's an invalid one.

cthulhu-rider: do u think we need LV format, or this one is OK?

cthulhu-rider: lets start with 0xFF - shorter, and cannot be a part of attribute

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -0,0 +1,237 @@
...
+
+const (
+	metaPrefixA  = byte(iota)
+	metaPrefixBI // integer attributes
+	metaPrefixBS // all other attributes
+	metaPrefixC
+)
+
+var (
+	maxUint256 = new(big.Int).SetBytes([]byte{255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255,
```
roman-khimov, line 28: Isn't it `127, 255, 255...`? At least VM has _signed_ 256-bit integers.

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -0,0 +1,237 @@
...
+		255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255})
+	maxUint256Neg = new(big.Int).Neg(maxUint256)
+)
+
+// TODO: fill on Init
```
roman-khimov, line 33: That's for migration of the old DB, right?

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -0,0 +1,237 @@
...
+	"github.com/nspcc-dev/neofs-sdk-go/user"
+	"github.com/nspcc-dev/neofs-sdk-go/version"
+	"go.etcd.io/bbolt"
+)
+
+const (
+	metaPrefixA  = byte(iota)
+	metaPrefixBI // integer attributes
+	metaPrefixBS // all other attributes
+	metaPrefixC
```
roman-khimov, line 24: These better be named semantically.

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -0,0 +1,237 @@
...
+	idk := [1 + oid.Size]byte{metaPrefixA}
+	copy(idk[1:], id[:])
+	if err := mb.Put(idk[:], nil); err != nil {
+		return fmt.Errorf("put object ID to container's meta bucket: %w", err)
+	}
+
+	var k []byte
+	// TODO: move to global funcs
+	makeKeyB := func(prefix byte, attr string, valLen int) (int, int) {
+		ln := 1 + oid.Size + len(attr) + valLen + len(utf8Delimiter)*2 // TODO: constantize some stuff
```
roman-khimov, line 50: `slices.Concat()`?

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -0,0 +1,237 @@
...
+		kLn, vOff = makeKeyC(attr, len(val))
+		copy(k[vOff:], val)
+		if err := mb.Put(k[:kLn], nil); err != nil {
+			return fmt.Errorf("put object attribute %q to container's meta bucket: %w", attr, err) // TODO: distinguishable context
+		}
+		return nil
+	}
+	putInt := func(attr string, n *big.Int) error {
+		// TODO: check uint256 overflow
+		kLn, vOff := makeKeyB(metaPrefixBI, attr, 1+32) // sign + 256-bit. Sign makes some sensible order
```
roman-khimov, line 90: I see you're using `sign+value` encoding here (not 256 fixed), this is related to 256 signed note. Let's think a bit about. A separate byte can be easier to handle instead of sign bit, but some numbers can be incompatible with VM. Does it matter? Not sure.

pkg/local_object_storage/metabase/metadata.go
cthulhu-rider: exactly

pkg/local_object_storage/metabase/metadata.go
cthulhu-rider: for sure, will do this finally. But for, now I'm used to naming from the task

pkg/local_object_storage/metabase/metadata.go
cthulhu-rider: yeah i thought about this. Will make code improvement after functionaly stabilization if u dont mind

pkg/local_object_storage/metabase/metadata.go
roman-khimov: BTW, https://github.com/holiman/uint256 can be useful here, it's much faster than elastic `big.Int` (refs https://github.com/nspcc-dev/neo-go/issues/1581).

pkg/local_object_storage/metabase/metadata.go
cthulhu-rider: added key buffering, concatenation does not fit him

pkg/local_object_storage/metabase/metadata.go
cthulhu-rider: done

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -0,0 +1,742 @@
...
+		}
+		primSeekPrefix = primSeekKey[:1+len(primAttr)+len(utf8Delimiter)]
+		valID := primSeekKey[len(primSeekPrefix):]
+		if len(valID) <= oid.Size {
+			return nil, nil, fmt.Errorf("%w: too small VAL_OID len %d", errInvalidCursor, len(valID))
+		}
+		prevResPrimVal, prevResOID = valID[:len(valID)-oid.Size], valID[len(valID)-oid.Size:]
+	} else {
+		primSeekKey = slices.Concat(primSeekKey[:1], []byte(primAttr), utf8Delimiter)
+		primSeekPrefix = primSeekKey
```
roman-khimov, line 225: What if it's `val < 100`? Won't it be looking at pref_val_100 and next (bigger) ones?

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -0,0 +1,742 @@
...
+		}
+	}
+	if !parentID.IsZero() {
+		if err = putPlainAttribute(metaBkt, &keyBuf, id, object.FilterParentID, string(parentID[:])); err != nil {
+			return err
+		}
+	}
+	for i := range attrs {
+		ak, av := attrs[i].Key(), attrs[i].Value()
+		if n, isInt := parseInt(av); isInt && n.Cmp(maxUint256Neg) >= 0 && n.Cmp(maxUint256) <= 0 {
```
carpawell, line 104: so we just do not index ints that overflows us? isn't it an error? should we support such cases?

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -0,0 +1,742 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Search selects up to count container's objects from the given container
+// matching the specified filters.
+func (db *DB) Search(cnr cid.ID, fs object.SearchFilters, attrs []string, cursor string, count uint16) ([]client.SearchResultItem, string, error) {
```
carpawell, line 119: should `meta` package use SDK's `client` package?

pkg/local_object_storage/metabase/metadata.go
cthulhu-rider: yes, this is OK, fixed256 is an enormous value covering any usecase nowadays

pkg/local_object_storage/metabase/metadata.go
cthulhu-rider: SDK is a basest in the node

pkg/local_object_storage/metabase/metadata.go
cthulhu-rider: note that seek key has no val, so here it'll peek the lowest `pref_attr_` value

but ur question got me thinking: if matcher is, in opposite, `>` or `>=`, why dont we peek `prefix_attr_val` right away? All less values will be skipped anyway, no need to filter them out manually. Worth todo

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -0,0 +1,742 @@
...
+		if attr == object.FilterContainerID || attr == object.FilterID {
+			return nil, "", fmt.Errorf("invalid filter #%d: prohibited attribute %s", i, attr)
+		}
+	}
+	for i := range attrs {
+		if attrs[i] == "" {
+			return nil, "", fmt.Errorf("empty attribute #%d", i)
+		}
+		if attrs[i] == object.FilterContainerID || attrs[i] == object.FilterID {
+			return nil, "", fmt.Errorf("prohibited attribute %s", attrs[i])
```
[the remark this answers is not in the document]
cthulhu-rider: dropped

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -0,0 +1,742 @@
...
+			val, err := attrSkr.get(id, attrs[i])
+			if err != nil {
+				return nil, nil, err
+			}
+			if collected[i], err = attrSkr.restoreVal(id, attrs[i], val); err != nil {
+				return nil, nil, err
+			}
+		}
+		if n == count {
+			more = true
```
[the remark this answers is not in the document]
cthulhu-rider: yep, except when attributes are not requested

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -0,0 +1,742 @@
...
+		if ind < 0 {
+			return nil, nil, fmt.Errorf("%w: missing delimiter", errInvalidCursor)
+		}
+		if !bytes.Equal(primSeekKey[1:1+ind], []byte(primAttr)) {
+			return nil, nil, fmt.Errorf("%w: wrong primary attribute", errInvalidCursor)
+		}
+		primSeekPrefix = primSeekKey[:1+len(primAttr)+len(utf8Delimiter)]
+		valID := primSeekKey[len(primSeekPrefix):]
+		if len(valID) <= oid.Size {
+			return nil, nil, fmt.Errorf("%w: too small VAL_OID len %d", errInvalidCursor, len(valID))
```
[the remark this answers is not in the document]
cthulhu-rider: possible with https://github.com/nspcc-dev/neofs-node/pull/3080#discussion_r1946717969. Will do

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -0,0 +1,742 @@
...
+		return err
+	}
+	if err = putPlainAttribute(metaBkt, &keyBuf, id, object.FilterPayloadChecksum, string(pldHash)); err != nil {
+		return err
+	}
+	if err = putPlainAttribute(metaBkt, &keyBuf, id, object.FilterPayloadHomomorphicHash, string(pldHmmHash)); err != nil {
+		return err
+	}
+	if len(splitID) > 0 {
+		// TODO: deprecated, maybe not index?
```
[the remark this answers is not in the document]
cthulhu-rider: dropped

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -0,0 +1,742 @@
...
+	var err error
+	if len(fs) == 0 {
+		res, resCursor, err = db.searchUnfiltered(cnr, []byte(cursor), count)
+	} else {
+		res, resCursor, err = db.search(cnr, fs, attrs, []byte(cursor), count)
+	}
+	if err != nil {
+		return nil, "", err
+	}
+	return res, base64.StdEncoding.EncodeToString(resCursor), nil
```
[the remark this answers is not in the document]
cthulhu-rider: good to deduplicate processing b/w shards, done

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -0,0 +1,742 @@
...
+}
+
+func (db *DB) search(cnr cid.ID, fs object.SearchFilters, attrs []string, cursor []byte, count uint16) ([]client.SearchResultItem, []byte, error) {
+	seekKey := make([]byte, 1+base64.StdEncoding.DecodedLen(len(cursor))) // TODO: limit?
```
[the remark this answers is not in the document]
cthulhu-rider: done

pkg/local_object_storage/metabase/metadata.go
cthulhu-rider: done

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -0,0 +1,796 @@
...
+		if dbVal, err = base58.Decode(fltVal); err != nil {
+			return nil, nil, fmt.Errorf("decode %q attribute value from Base58: %w", attr, err)
+		}
+	case object.FilterPayloadChecksum, object.FilterPayloadHomomorphicHash:
+		var err error
+		if dbVal, err = hex.DecodeString(fltVal); err != nil {
+			return nil, nil, fmt.Errorf("decode %q attribute value from HEX: %w", attr, err)
+		}
+	case object.FilterSplitID:
+		uid, err := uuid.Parse(fltVal)
```
roman-khimov, line 505: Not sure this works for `MatchCommonPrefix`.

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -0,0 +1,796 @@
...
+		}
+		primSeekPrefix = primSeekKey[:cursor.ValIDOff]
+		valID := cursor.Key[cursor.ValIDOff:]
+		prevResPrimVal, prevResOID = valID[:len(valID)-oid.Size], valID[len(valID)-oid.Size:]
+	} else {
+		if primMatcher == object.MatchStringEqual || primMatcher == object.MatchCommonPrefix ||
+			primMatcher == object.MatchNumGT || primMatcher == object.MatchNumGE {
+			var err error
+			if primSeekKey, primSeekPrefix, err = seekKeyForAttribute(primAttr, fs[0].Value()); err != nil {
+				return nil, nil, fmt.Errorf("invalid primary filter value: %w", err)
```
roman-khimov, line 227: Shouldn't this be checked _before_ going into `searchInBucket`?

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -0,0 +1,796 @@
...
+			}
+			if collected[i], err = attrSkr.restoreVal(id, attrs[i], val); err != nil {
+				return nil, nil, err
+			}
+		}
+		if n == count {
+			more = true
+			if len(attrs) > 0 {
+				break
+			} // else "later" objects may have "less" ID, so we should continue
```
roman-khimov, line 415: But how? OIDs are sorted naturally as well and if have enough results then subsequent IDs should be probed on subsequent (continuation) search queries.

pkg/local_object_storage/metabase/metadata.go
cthulhu-rider: can be, dont think this is sufficient for now

pkg/local_object_storage/metabase/metadata.go
cthulhu-rider: yeah, to support them we'd need to stringify DB values, which i didnt wanna do. So, i decided to just not support prefixing split IDs:
1. they are deprecated
2. cant imagine who can need to prefix any IDs

pkg/local_object_storage/metabase/metadata.go
cthulhu-rider: consider objects with ascending index and descending IDs. If no attributes are requested, we should sort by IDs. We can state that first _filtered_ attribute takes order priority in this case. If so, we can always break here

roman-khimov: OK, let's leave it as is for now.

roman-khimov (approved):

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -0,0 +1,796 @@
...
+	ValIDOff int
+}
+
+// NewSearchCursorFromString decodes cursor from the string according to the
+// primary attribute.
+func NewSearchCursorFromString(s, primAttr string) (*SearchCursor, error) {
+	if s == "" {
+		return nil, nil
+	}
+	b := make([]byte, 1+base64.StdEncoding.DecodedLen(len(s)))
```
carpawell, line 126: isnt `DecodeString` about the same but shorter?

pkg/local_object_storage/metabase/metadata.go
carpawell: sdk in general -- agree, but the `client` package -- not sure

carpawell (approved):

pkg/local_object_storage/metabase/metadata.go
cthulhu-rider: `DecodeString` creates new buffer on his own, this isnt desired here

PR #3079, opened 2025-01-10
node: Use `github.com/nspcc-dev/neofs-api-go/v2` module less

cthulhu-rider:
one more cut from #3057, continues #3078

PR #3057, opened 2024-12-18
Do not use `github.com/nspcc-dev/neofs-api-go/v2` where unnecessary

cthulhu-rider:
self-reminder: check refs in commit messages before merging

a lot of changes, but this will maximally amortize coming update to the latest SDK

PR #3081, opened 2025-01-14
Simplify recurrent request handling (except `Object` service)

cthulhu-rider:
one more cut from #3057, continues #3079


cmd/neofs-node/config.go
```diff
@@ -418,21 +417,22 @@ type cfg struct {
...
 	}
 
-	val.(netmap.NetMap).WriteToV2(msg)
-
-	return nil
+	return val.(netmap.NetMap), nil
 }
 
+// CurrentEpoch returns the latest cached epoch.
+func (c *cfg) CurrentEpoch() uint64 { return c.networkState.CurrentEpoch() }
```
carpawell, line 434: why `GetNetworkMap` but `CurrentEpoch`?

cmd/neofs-node/container.go
```diff
@@ -8,8 +8,13 @@ import (
 	"errors"
 	"fmt"
 
-	containerV2 "github.com/nspcc-dev/neofs-api-go/v2/container"
-	containerGRPC "github.com/nspcc-dev/neofs-api-go/v2/container/grpc"
+	apicontainer "github.com/nspcc-dev/neofs-api-go/v2/container"
+	protocontainer "github.com/nspcc-dev/neofs-api-go/v2/container/grpc"
+	apirefs "github.com/nspcc-dev/neofs-api-go/v2/refs"
+	refs "github.com/nspcc-dev/neofs-api-go/v2/refs/grpc"
```
carpawell, line 14: should this be `protorefs`? just to make it meaningful according to the other aliases

carpawell (approved):

roman-khimov (approved):

cmd/neofs-node/container.go
cthulhu-rider: there is no `refs` package in SDK root level, so we can do w/o alias (considering that apigo will leave)

cmd/neofs-node/config.go
cthulhu-rider: this is a simple getter while `GetNetworkMap` can fail. I didnt think about this much tbh

PR #3083, opened 2025-01-16
services/object: Inline metric handler

cthulhu-rider:
one more cut from #3057, continues #3081

to be continued


pkg/services/object/server.go
```diff
@@ -138,6 +168,7 @@ func (s *server) Put(gStream protoobject.ObjectService_PutServer) error {
 		if err := stream.Send(putReq); err != nil {
 			if errors.Is(err, util.ErrAbortStream) {
 				resp, err := stream.CloseAndRecv()
+				s.pushOpExecResult(stat.MethodObjectPut, err, t)
```
roman-khimov, line 171: No stat if `!errors.Is(err, util.ErrAbortStream)`?

pkg/services/object/server.go
```diff
@@ -95,31 +113,39 @@ type server struct {
...
 		return err
 	}
 
+	t := time.Now()
 	for {
 		req, err := gStream.Recv()
 		if err != nil {
 			if errors.Is(err, io.EOF) {
 				resp, err := stream.CloseAndRecv()
+				s.pushOpExecResult(stat.MethodObjectPut, err, t)
```
roman-khimov, line 148: And here for !io.EOF?

pkg/services/object/server.go
```diff
@@ -95,31 +113,39 @@ type server struct {
...
+func (s *server) pushOpExecResult(op stat.Method, err error, startedAt time.Time) {
+	s.metrics.HandleOpExecResult(op, err == nil, time.Since(startedAt))
+}
+
 // Put opens internal Object service Put stream and overtakes data from gRPC stream to it.
 func (s *server) Put(gStream protoobject.ObjectService_PutServer) error {
 	stream, err := s.srv.Put(gStream.Context())
 	if err != nil {
 		return err
 	}
```
roman-khimov, line 141: Maybe some defer would be more handy for this one.

pkg/services/object/server.go
cthulhu-rider: fixed

pkg/services/object/server.go
cthulhu-rider: fixed

roman-khimov (approved):

PR #3084, opened 2025-01-20
services/object: Inline signing service

cthulhu-rider:
Continues 1e40321b88d3d125de85fa4a3be0db67924951b3. Now sign service is gone.


pkg/services/object/server.go
```diff
@@ -182,13 +214,27 @@ func (s *server) Delete(ctx context.Context, req *protoobject.DeleteRequest) (*p
...
+		return s.makeStatusDeleteResponse(t, err)
 	}
 
-	return resp.ToGRPCMessage().(*protoobject.DeleteResponse), nil
+	return s.signDeleteResponse(t, util.StatusOKErr, resp.ToGRPCMessage().(*protoobject.DeleteResponse))
+}
+
+func (s *server) signHeadResponse(err error, startTime time.Time, resp *protoobject.HeadResponse) (*protoobject.HeadResponse, error) {
+	resp = util.SignResponse(&s.signer, resp, v2object.HeadResponse{})
+	s.pushOpExecResult(stat.MethodObjectHead, err, startTime)
```
roman-khimov, line 230: Not something I'd expect from `sign...` method.

pkg/services/object/server.go
```diff
@@ -182,13 +214,27 @@ func (s *server) Delete(ctx context.Context, req *protoobject.DeleteRequest) (*p
...
+	return resp, nil
+}
+
+func (s *server) makeStatusHeadResponse(startTime time.Time, err error) (*protoobject.HeadResponse, error) {
```
roman-khimov, line 234: Can we have a generic version of it with two types?

pkg/services/util/sign.go
```diff
@@ -229,7 +58,11 @@ var (
 // ToStatus unwraps the deepest error from err and converts it into the response
 // status.
 func ToStatus(err error) *protostatus.Status {
-	return statusFromErr(err).ToGRPCMessage().(*protostatus.Status)
+	// unwrap error
+	for e := errors.Unwrap(err); e != nil; e = errors.Unwrap(err) {
```
roman-khimov, line 62: I see it's the old code inlined, but this doesn't seem to be correct. It implies that the most inner error is `apistatus.*` one, but this assumption can be wrong. The way `ErrorToV2()` is implemented (using `errors.As()`) I'd say this is not needed.

cthulhu-rider: we have #2744 bout this, or i changed the behavior?

pkg/services/object/server.go
cthulhu-rider: didnt get what do u mean by "two types". What signature would u expect?

pkg/services/object/server.go
cthulhu-rider: yeah, this is some kind of `finExec`. Ok will refactor

pkg/services/object/server.go
roman-khimov: Likely not possible without an additional `InitFromMetaHeader` for each `protoobject.*`. The idea is to have `*protoobject.Smth` as a generic type, but it also requires `v2object.Smth` companion. Which is handled in `SignResponse` already with R, RV2 and RV2PTR, so our hypothetical `makeGenericStatusResponse` would require all of them plus one more type for `*protoobject.Smth`. Can be somewhat messy.

cthulhu-rider: generic struct fields would also be helpful, but Go does not support them. If it becomes, and when v2 leaves, we can simplify this code

Issue #3045, opened 2024-12-11
Use iterators for container listings


carpawell:
https://github.com/nspcc-dev/neofs-contract/pull/300

## Expected Behavior
New container listing with no result length restriction is used.

## Current Behavior
New API is not used. [List](https://github.com/nspcc-dev/neofs-node/blob/3129bdeadc87017c95d4877b82c57e64db5539f8/cmd/neofs-node/cache.go#L322) is the only option for now.

## Possible Solution
Add new method to `morph` package, use it.

## Steps to Reproduce (for bugs)
Take a look at `morph` pkg or search for "containersOf" const.

## Context
https://github.com/nspcc-dev/neofs-contract/pull/300

## Regression
No.

## Your Environment
0.44.0

PR #3085, opened 2025-01-20
feat/Process object notification from chain

carpawell:


pkg/services/meta/meta.go
```diff
@@ -0,0 +1,125 @@
...
+	cnrEv   chan *state.ContainedNotificationEvent
+	epochEv chan *state.ContainedNotificationEvent
+
+	m        sync.RWMutex
+	storages map[cid.ID]*containerStorage
+}
+
+// New makes [Meta].
+func New(l *zap.Logger, cLister ContainerLister, ws *rpcclient.WSClient, containerH, nmH util.Uint160, rootPath string) (*Meta, error) {
+	storagesFS, err := os.ReadDir(rootPath)
```
roman-khimov, line 56: This limits us to less than 1024 containers by default, btw.

pkg/services/meta/meta.go
```diff
@@ -0,0 +1,125 @@
...
+		bCh:      make(chan *block.Header),
+		objEv:    make(chan *state.ContainedNotificationEvent),
+		cnrEv:    make(chan *state.ContainedNotificationEvent),
+		epochEv:  make(chan *state.ContainedNotificationEvent),
+		storages: storages}, nil
+}
+
+// Run starts notification handling. Must be called only on instances created
+// with [New]. Blocked until context is done.
+func (m *Meta) Run(ctx context.Context) error {
```
roman-khimov, line 118: Why using context here?

pkg/services/meta/notifications.go
```diff
@@ -0,0 +1,497 @@
...
+
+				m.m.RLock()
+				_, ok := m.storages[ev.cID]
+				m.m.RUnlock()
+				if !ok {
+					l.Debug("skipping object notification", zap.Stringer("inactual container", ev.cID))
+					continue
+				}
+
+				go func() {
```
roman-khimov, line 87: Unordered. Potentially dangerous resource-wise.

pkg/services/meta/notifications.go
```diff
@@ -0,0 +1,497 @@
...
+				return fmt.Errorf("put batch for %d block to %q storage: %w", ind, p, err)
+			}
+
+			st.opsBatch = nil
+		}
+
+		st.m.Unlock()
+
+		st.mpt.Flush(ind)
+		st.mpt.Collapse(collapseDepth)
```
roman-khimov, line 173: Collapsing for every block is not very efficient.

pkg/services/meta/notifications.go
```diff
@@ -0,0 +1,497 @@
...
+
+	cID, ok := arr[0].Value().([]byte)
+	if !ok {
+		return res, fmt.Errorf("unexpected container ID stack item: %T", arr[0].Value())
+	}
+	oID, ok := arr[1].Value().([]byte)
+	if !ok {
+		return res, fmt.Errorf("unexpected object ID stack item: %T", arr[1].Value())
+	}
+	meta, ok := arr[2].Value().([]stackitem.MapElement)
```
roman-khimov, line 284: Why not handling it as a proper map (yeah, it only works via `Index()` API now, but still)? You don't care about element order. I think depending on order is too fragile here.

pkg/services/meta/notifications.go
```diff
@@ -0,0 +1,497 @@
...
+	st.m.Lock()
+	defer st.m.Unlock()
+
+	if st.opsBatch == nil {
+		st.opsBatch = make(map[string][]byte)
+	}
+
+	commPrefix := e.oID[:]
+
+	st.opsBatch[string(append([]byte{oidIndex}, commPrefix...))] = []byte{1}
```
roman-khimov, line 370: You can have `[]byte{}`

pkg/services/meta/notifications.go
carpawell: > Why not handling it as a proper map (yeah, it only works via Index() API now, but still)?

if you are talking about calling `Index` and then pick this item in _slice_, i do not see any difference. i would like to have `Get`/`Pick` as the opposite to the existing `Add` and know nothing about slice but there is no such method. i do not know why

> I think depending on order is too fragile here.

i do not agree. while reading this code, yes, it may seem strange. but we are reading node's code that has received notification. it means every node signed a correct order and number of meta fields. so either the order correct and the node can work with it, or nothing happens at all

anyway, i really would like to have `Get` from the map type, and agree to use it orderless. but i dont

pkg/services/meta/notifications.go
carpawell: can you suggest a rule? connect it to real-time? 30 seconds? minutes?

pkg/services/meta/notifications.go
carpawell: > Unordered

can this be bad? eventually, we plan to receive every notification in a block at once. now the order and new block handling is also not been ordered and will not be

> Potentially dangerous resource-wise.

we have been discussing notification handling from time to time for so long. we _have_ to read notifications always as says neo-go client (IMO, it is the client's bad restriction and it is a mistake) and we cannot do any other calls if we are blocked on notification handling. considering all the things i know and all the discussions i had, i may only say that using all the resources we have (routines) is the only way to prevent deadlock, otherwise, there is always a number of notifications that can be received via a channel and their number exceeds any buffer size and it leads to the lock

pkg/services/meta/meta.go
carpawell: https://github.com/nspcc-dev/neofs-node/blob/d7407d117b3fba12088c6b803f1e2a9894294fbe/cmd/neofs-node/worker.go#L15-L23 the only our practice reason. i do not like it

pkg/services/meta/meta.go
carpawell: can you elaborate? what exactly limits us? we can add a tree structure if needed. but i am not sure i understood our point

pkg/services/meta/container.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+
+		if len(root) != util.Uint256Size {
+			return nil, fmt.Errorf("root hash from db is %d bytes long, expect %d", len(root), util.Uint256Size)
+		}
+
+		prevRootNode = mpt.NewHashNode([util.Uint256Size]byte(root))
+	}
+
+	mTrie := mpt.NewTrie(prevRootNode, mpt.ModeGC, storage.NewMemCachedStore(st))
```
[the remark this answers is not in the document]
carpawell: isnt `ModeGC` a flag that includes everything and it is also a `ModeLatest` flag? or you mean `ModeLatest` does not need GC routine at all and everything will be cleaned and nothing is lost ever?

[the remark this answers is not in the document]
roman-khimov: Latest is like KeepOnlyLatest in NeoGo, only the latest and greatest MPT is stored, old elements are deleted immediately. It does not require GC.

pkg/services/meta/meta.go
roman-khimov: `ulimit -n`. It can be changed, not something set in stone, but still.

pkg/services/meta/notifications.go
roman-khimov: No hard rules here. It needs to be collapsed each time you think it's too big. Go figure.

pkg/services/meta/notifications.go
roman-khimov: We have `FromStackItem` in RPC bindings, probably it'll help us here.

pkg/services/meta/notifications.go
roman-khimov: 1. We care about ordering here. That's exactly the feature we need from this chain, ordered events.
2. Client is limited by a single pipe transmitting all data, it's like HOL blocking issue if you like. But we can build a reasonable enough pipeline with buffers that will work in most cases and will just reset the connection if it's overflown. Then this boils down to synchronizing with the chain from cold start which is the issue we have to solve anyway.

carpawell:
There are still some TODOs that require additional attention. Maybe later, maybe now. Some more tests will be added, currently, they are broken because of new updates.

pkg/services/meta/notifications.go
```diff
@@ -0,0 +1,571 @@
...
+
+	epoch, ok := arr[0].Value().(*big.Int)
+	if !ok {
+		return 0, fmt.Errorf("unexpected epoch stack item: %T", arr[0].Value())
+	}
+
+	return epoch.Int64(), nil
+}
+
+func (m *Meta) handleEpochNotification(e int64) error {
```
roman-khimov, line 533: Expired objects are not deleted.

roman-khimov (commented):
s/MTP/MPT/

cmd/neofs-node/config.go
```diff
@@ -698,6 +715,13 @@ func initBasics(c *cfg, key *keys.PrivateKey, stateStorage *state.PersistentStor
 		fatalOnErr(err)
 	}
 
+	c.shared.wsCli, err = rpcclient.NewWS(c.ctx, addresses[0], rpcclient.WSOptions{
```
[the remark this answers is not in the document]
carpawell: whoops, rebase problems. this field is not used and it is kinda no-op code, multi endpoints are handled inside meta service now

pkg/services/meta/notifications.go
```diff
@@ -0,0 +1,571 @@
...
+				return res, fmt.Errorf("unexpected locked objects type: %T", meta[lockedIndex].Value.Value())
+			}
+		}
+	}
+
+	return res, nil
+}
+
+func (m *Meta) handleObjectNotification(e objEvent) error {
+	// TODO check network magic?
```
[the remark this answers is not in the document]
carpawell: will do. i could not answer if it is possible to misuse networks. but if we have this field from our chain, i guess, it should be done

pkg/services/meta/notifications.go
```diff
@@ -0,0 +1,571 @@
...
+	}
+	if len(e.lockedObjects) > 0 {
+		st.opsBatch[string(append([]byte{lockedIndex}, commPrefix...))] = e.lockedObjects
+	}
+	if e.typ != objectsdk.TypeRegular {
+		st.opsBatch[string(append([]byte{typeIndex}, commPrefix...))] = []byte{byte(e.typ)}
+	}
+
+	// TODO: receive full object header and store its non-MTP parts in
+	//  persistent object storage
```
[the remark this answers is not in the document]
carpawell: you mean every KV that i put into MPT should also be placed to the raw storage?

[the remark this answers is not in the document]
roman-khimov: Yes.

pkg/services/meta/notifications.go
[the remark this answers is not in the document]
roman-khimov: It's mostly for replay protection, but we have it, can be checked.

pkg/services/meta/notifications.go
carpawell: hm, we do not have expiration epoch in meta-from-chain, so we cannot find such cases without object HEADing object. or we have to add it to the contract

pkg/services/meta/notifications.go
carpawell: what exact struct do you suggest to use from bindings?

pkg/services/meta/notifications.go
roman-khimov: Can be postponed.

pkg/services/meta/container.go
```diff
@@ -0,0 +1,129 @@
...
+	// TODO: receive full object header and store its non-MPT parts in
+	//  persistent object storage
+
+	return nil
+}
+
+// mptToStoreBatch drops the first byte from every key in the map.
+func mptToStoreBatch(b map[string][]byte) map[string][]byte {
+	for k, v := range b {
+		b[k[1:]] = v
```
cthulhu-rider, line 92: this is unsafe. https://go.dev/ref/spec#For_range states
> If a map entry is created during iteration, that entry may be produced during the iteration or may be skipped.

and u dont drop `k` elements. It's easier to fill a new map in general. But since u have fixed-len keys, this is doable:
```go
	for k, v := range b {
		if len(k) == 34 {
			b[k[1:]] = v
		}
	}
	maps.DeleteFunc(b, func(k string, _ []byte) bool { return len(k) == 34 })
```

since it's very specific, i'd inline this into the only usage place

pkg/services/meta/container.go
```diff
@@ -0,0 +1,129 @@
...
+func (s *containerStorage) putObject(e objEvent) error {
+	s.m.Lock()
+	defer s.m.Unlock()
+
+	newKVs := make(map[string][]byte)
+	commsuffix := e.oID[:]
+
+	// batching that is implemented for MPT ignores key's first byte
+
+	newKVs[string(append([]byte{0, oidIndex}, commsuffix...))] = []byte{}
```
cthulhu-rider, line 55:
```suggestion
	newKVs[string(slices.Concat([]byte{0, oidIndex}, commsuffix))] = []byte{}
```
but i'd just concat strings

pkg/services/meta/container.go
```diff
@@ -0,0 +1,129 @@
...
+			return nil, fmt.Errorf("root hash from db is %d bytes long, expect %d", len(root), util.Uint256Size)
+		}
+
+		prevRootNode = mpt.NewHashNode([util.Uint256Size]byte(root))
+	}
+
+	mTrie := mpt.NewTrie(prevRootNode, mpt.ModeLatest, storage.NewMemCachedStore(st))
+
+	return &containerStorage{
+		m:    sync.RWMutex{},
```
cthulhu-rider, line 124: no need

pkg/services/meta/container.go
```diff
@@ -0,0 +1,129 @@
...
+func mptToStoreBatch(b map[string][]byte) map[string][]byte {
+	for k, v := range b {
+		b[k[1:]] = v
+	}
+
+	return b
+}
+
+func storageForContainer(rootPath string, cID cid.ID) (*containerStorage, error) {
+	cIDString := cID.EncodeToString()
```
cthulhu-rider, line 99: can be inlined

pkg/services/meta/container.go
```diff
@@ -0,0 +1,129 @@
...
+		}
+
+		if len(root) != util.Uint256Size {
+			return nil, fmt.Errorf("root hash from db is %d bytes long, expect %d", len(root), util.Uint256Size)
+		}
+
+		prevRootNode = mpt.NewHashNode([util.Uint256Size]byte(root))
+	}
+
+	mTrie := mpt.NewTrie(prevRootNode, mpt.ModeLatest, storage.NewMemCachedStore(st))
```
cthulhu-rider, line 121: this too

pkg/services/meta/meta.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/nspcc-dev/neo-go/pkg/rpcclient"
+	"github.com/nspcc-dev/neo-go/pkg/util"
+	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
+	"go.uber.org/zap"
+	"golang.org/x/exp/maps"
+	"golang.org/x/sync/errgroup"
+)
+
+const (
+	// raw storage prefixes.
```
cthulhu-rider, line 21: desc of the whole `const` group to me

cmd/neofs-node/meta.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("read network map at the current epoch #%d: %w", curEpoch, err)
+	}
+
+	c.m.RLock()
+	if c.prevNetMap != nil {
+		cnrsSame := slices.EqualFunc(c.prevCnrs, actualContainers, func(cID1, cID2 cid.ID) bool {
+			return bytes.Equal(cID1[:], cID2[:])
+		})
+		netmapSame := slices.EqualFunc(c.prevNetMap.Nodes(), networkMap.Nodes(), func(n1 netmapsdk.NodeInfo, n2 netmapsdk.NodeInfo) bool {
```
cthulhu-rider, line 81: seems like no need to calc this if `!cnrsSame`

cmd/neofs-node/meta.go
```diff
@@ -0,0 +1,133 @@
...
+		return nil, fmt.Errorf("read current NeoFS epoch: %w", err)
+	}
+	networkMap, err := c.network.GetNetMapByEpoch(curEpoch)
+	if err != nil {
+		return nil, fmt.Errorf("read network map at the current epoch #%d: %w", curEpoch, err)
+	}
+
+	c.m.RLock()
+	if c.prevNetMap != nil {
+		cnrsSame := slices.EqualFunc(c.prevCnrs, actualContainers, func(cID1, cID2 cid.ID) bool {
```
cthulhu-rider, line 78:
```suggestion
		cnrsSame := slices.Equal(c.prevCnrs, actualContainers)
```

cmd/neofs-node/meta.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("read current NeoFS epoch: %w", err)
+	}
+	networkMap, err := c.network.GetNetMapByEpoch(curEpoch)
+	if err != nil {
+		return nil, fmt.Errorf("read network map at the current epoch #%d: %w", curEpoch, err)
+	}
+
+	c.m.RLock()
+	if c.prevNetMap != nil {
```
cthulhu-rider, line 77:
```suggestion
	if c.prevNetMap != nil && c.prevRes != nil {
```
?

cmd/neofs-node/meta.go
```diff
@@ -0,0 +1,133 @@
...
+	network    netmap.Source
+
+	m          sync.RWMutex
+	prevCnrs   []cid.ID
+	prevNetMap *netmapsdk.NetMap
+	prevRes    map[cid.ID]struct{}
+}
+
+func (c *containerListener) List() (map[cid.ID]struct{}, error) {
+	actualContainers, err := c.cnrClient.List(nil)
```
cthulhu-rider, line 63: have u considered https://github.com/nspcc-dev/neofs-contract/blob/27ef3cc95fd1b601845effe6b5b6f36b30744836/contracts/container/contract.go#L578 for this?

pkg/services/meta/meta.go
```diff
@@ -0,0 +1,213 @@
...
+	epochEv     chan *state.ContainedNotificationEvent
+
+	objNotificationBuff chan *state.ContainedNotificationEvent
+}
+
+const objsBufferSize = 1024
+
+// New makes [Meta].
+func New(l *zap.Logger, cLister ContainerLister, timeout time.Duration, endpoints []string, containerH, nmH util.Uint160, rootPath string) (*Meta, error) {
+	storagesFS, err := os.ReadDir(rootPath)
```
cthulhu-rider, line 66: have u considered https://pkg.go.dev/io/fs#WalkDir for this?

pkg/services/meta/meta.go
```diff
@@ -0,0 +1,213 @@
...
+	storagesRead := make(map[cid.ID]*containerStorage)
+	for _, s := range storagesFS {
+		sName := s.Name()
+		cID, err := cid.DecodeString(sName)
+		if err != nil {
+			l.Warn("skip unknown container storage entity", zap.String("name", sName), zap.Error(err))
+			continue
+		}
+
+		storagesRead[cID], err = storageForContainer(rootPath, cID)
```
cthulhu-rider, line 79: if `err != nil`, `storagesRead[cID] == nil`, guess further code is not protected from this. Suggest to not add a map entry on failure

pkg/services/meta/meta.go
```diff
@@ -0,0 +1,213 @@
...
+			err = storagesRead[cID].drop()
+			if err != nil {
+				l.Warn("could not drop container storage", zap.Stringer("cID", cID), zap.Error(err))
+			}
+
+			delete(storagesRead, cID)
+		}
+	}
+
+	storages := storagesRead
```
cthulhu-rider, line 101: looks like a simple renaming, no need imo

pkg/services/meta/meta.go
```diff
@@ -0,0 +1,213 @@
...
+		storagesRead[cID], err = storageForContainer(rootPath, cID)
+		if err != nil {
+			l.Warn("skip container storage that cannot be read", zap.String("name", sName), zap.Error(err))
+			continue
+		}
+	}
+
+	cnrsNetwork, err := cLister.List()
+	if err != nil {
+		return nil, fmt.Errorf("listing node's containers: %w", err)
```
cthulhu-rider, line 88: shouldn't we free the `storagesRead` in this case? Also further

cthulhu-rider (commented):
gr8 work is done, viewed few random parts only

pkg/services/meta/meta.go
```diff
@@ -0,0 +1,213 @@
...
+
+	v, err := m.ws.GetVersion()
+	if err != nil {
+		return fmt.Errorf("get version: %w", err)
+	}
+	m.magicNumber = uint32(v.Protocol.Network)
+
+	err = m.subscribeForMeta()
+	if err != nil {
+		m.ws.Close()
```
cthulhu-rider, line 159: done for this but not for other errors, pay attention pls

pkg/services/meta/container.go
```diff
@@ -0,0 +1,129 @@
...
+	commsuffix := e.oID[:]
+
+	// batching that is implemented for MPT ignores key's first byte
+
+	newKVs[string(append([]byte{0, oidIndex}, commsuffix...))] = []byte{}
+	newKVs[string(append([]byte{0, sizeIndex}, commsuffix...))] = e.size.Bytes()
+	if len(e.firstObject) == sha256.Size {
+		newKVs[string(append([]byte{0, firstPartIndex}, commsuffix...))] = e.firstObject
+	}
+	if len(e.prevObject) == sha256.Size {
```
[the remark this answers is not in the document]
carpawell: > what else size can it be except 0?

none. contract checks for it. i do not know what could be done if it is like 5 long. do you have a suggestion?

> oid.Size would be more clear

done

pkg/services/meta/container.go
carpawell: tbh, i did smth strange here, thought about smth different. this is a total mistake, fixed

pkg/services/meta/container.go
carpawell: but why? i usually use `slices.Concat` when there are more than two slices otherwise it is useless for me

pkg/services/meta/container.go
carpawell: i prefer explicit struct initiation but searched for this in our code and there is not a single second example like this so ok, dropped

pkg/services/meta/meta.go
carpawell: it is. i thought and think that there will be more when we add more features

cmd/neofs-node/meta.go
carpawell: ok but required one more `m.Unlock` which i usually do not like

cmd/neofs-node/meta.go
carpawell: yes, but i think it should be fixed once for all callers: https://github.com/nspcc-dev/neofs-node/issues/3045

pkg/services/meta/meta.go
carpawell: it is. in fact, i thought and think that there will be more such

pkg/services/meta/meta.go
carpawell: it is. prefer not to change `*Read` var

pkg/services/meta/meta.go
carpawell: you meant closing dbs? done

pkg/services/meta/meta.go
```diff
@@ -0,0 +1,213 @@
...
+			for _, st := range m.storages {
+				wg.Go(func() error {
+					st.m.Lock()
+					defer st.m.Unlock()
+
+					st.mpt.Collapse(collapseDepth)
+
+					_, err := st.mpt.Store.PersistSync()
+					if err != nil {
+						return fmt.Errorf("persisting %q storage: %w", st.path, err)
```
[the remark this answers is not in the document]
carpawell: it is a big question for me in fact. i feel like either everything should be in sync and ok, or we should stop
@roman-khimov

pkg/services/meta/container.go
cthulhu-rider: func is still the same, but correct now

pkg/services/meta/container.go
cthulhu-rider: dont insist

pkg/services/meta/meta.go
carpawell: i prefer not to use iterators when it is not necessary cause it adds nesting IMO. here, i think it is possible to read everything at once without problems

pkg/services/meta/container.go
carpawell: now it is used twice

pkg/services/meta/container.go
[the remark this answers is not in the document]
carpawell: oh, i got you. no, `e` (`objEvent`) is already checked and parsed thing from the contract. it just could be missing, or `oid.Size`ed. changed to `> 0`

pkg/services/meta/meta.go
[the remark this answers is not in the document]
roman-khimov: Should be OK for the current stage. In general other containers should continue to work, but that will be solved automatically when we have proper chain sharding.

roman-khimov (approved):

PR #3088, opened 2025-01-22
Netmap with nodev2

### LINKED ISSUE #1914: Send `UpdateState` instead of `AddPeer` as node's heartbeat
In current implementation storage nodes periodically call `AddPeer` method of `Netmap` contract. Seems like we can call `UpdateState` instead. This will show a clear difference between bootstrap and heartbeat.

I usually stick to dedicated methods for each need: this makes it easier to control the variability of behavior. Maybe it's worth to add `Touch` (`Ping`, `Pulsate`, `Heartbear) contract method?

I also propose to write docs about bootstrap procedure within the issue.

### LINKED ISSUE #3047: Use new netmap method after contract update
https://github.com/nspcc-dev/neofs-contract/pull/445

It says we need to use both approaches for some releases. I am almost sure the next minor release should include new contract update, so we need to call the new method starting from 0.45.0.

\-

## Additional context
https://github.com/nspcc-dev/neofs-contract/pull/445/commits/71fc8a9e61dec7700690f2e6cad04326b80e1e13
https://github.com/nspcc-dev/neofs-contract/issues/298
https://github.com/nspcc-dev/neofs-contract/issues/297

roman-khimov:
Fixes #3047, fixes #1914. The logic is rather simple:
 * we update to 0.45.0, SNs push data to both lists
 * the switch is turned on and we're using new lists now
 * old methods and data to be removed afterwards

Related to https://github.com/nspcc-dev/neofs-contract/pull/461, once it's there the default will be to use v2 for new installations.


pkg/innerring/processors/netmap/process_peers.go
```diff
@@ -68,6 +95,35 @@ func (np *Processor) processAddPeer(ev netmapEvent.AddPeer) {
 	}
 }
 
+// Check the new node and allow/reject adding it to netmap.
+func (np *Processor) processAddNode(ev netmapEvent.AddNode) {
```
carpawell, line 77: why did you decide to put it here? all the previous flow was about parsing and than handling, it was convenient IMO. now we have a dozen of the old parsers/handlers and the new one that executes script and returns some contract thing as an event, not a known sdk/core structure

pkg/morph/event/notary_preparator.go
```diff
@@ -295,7 +295,7 @@ func validateNestedArgs(expArgLen int64, ops []Op) error {
 			}
 
 			expArgLen++
-		case currentCode == opcode.PACK:
+		case currentCode == opcode.PACK || currentCode == opcode.PACKSTRUCT:
```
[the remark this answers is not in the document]
roman-khimov: It's a valid opcode that can be used just like any other opcode. It's just that all of the tools we have for entry scripts now never use it.

pkg/innerring/processors/netmap/process_peers.go
roman-khimov: What's the difference with `processAddPeer`? It's the same thing.

cmd/neofs-adm/internal/modules/fschain/netmap_candidates.go
```diff
@@ -29,22 +31,76 @@ func listNetmapCandidatesNodes(cmd *cobra.Command, _ []string) error {
...
-		return fmt.Errorf("can't fetch list of network config keys from the netmap contract: %w", err)
+		return fmt.Errorf("can't list candidates: %w", err)
 	}
-	if res.State != "HALT" {
-		return fmt.Errorf("netmap contract returned unexpected exception: %s", res.FaultException)
+	defer func() {
+		_ = inv.TerminateSession(sess)
+	}()
+	items, err := inv.TraverseIterator(sess, &iter, 0)
+	for err == nil && len(items) > 0 {
```
carpawell, line 67: here too

pkg/innerring/processors/netmap/process_peers.go
carpawell: i meant, i would call `netmapEvent.Node2Info(&ev.Node)` inside the parser, otherwise, i do not see why we need to parse smth and why we have separated every notification flow into "parser" and "handler". but in the `processAddPeer` it is just bytes so maybe your version is not that different

carpawell (approved):

pkg/innerring/processors/netmap/process_peers.go
roman-khimov: Well, parser still parses, but if you're to take a look at what happens in `processAddPeer` it accepts a byte string from the parser which it then unmarshals. So it takes whatever is the native contract/invocation thing as an input. Same thing here, it takes (parsed) contract structure as an input and does whatever it needs to with it (including conversion into SDK type).

PR #3089, opened 2025-01-23
Fix bugs with removing split objects

### LINKED ISSUE #3026: Expired lock test fails occasionaly
## Expected Behavior
Green.

## Current Behavior
https://rest.fs.neo.org/HXSaMJXk2g8C14ht8HSi7BBaiYZ1HeWh2xnWPGQCg4H6/2136-1732220712/index.html#suites/295ed2e000f8fa6f3ade59cc13b98615/6deb00ef86b7dc66/
https://rest.fs.neo.org/HXSaMJXk2g8C14ht8HSi7BBaiYZ1HeWh2xnWPGQCg4H6/2136-1732221029/index.html#suites/295ed2e000f8fa6f3ade59cc13b98615/c49306c8063b4d85/

It fails at the waiting stage, all iterations returns the object.

## Possible Solution
Needs to be investigated, either it's a test issue (not giving enough time for things to settle) or it's a bug. From the first glance it looks as a bug.

Run tests.object.test_object_lock.TestObjectLockWithGrpc#test_expired_object_should_be_deleted_after_locks_are_expired 

## Your Environment
* Version used: any recent master
* Server setup and configuration:
* Operating System and version (`uname -a`):

End-rey:
Fix #3026.

For bug about expired objects that aren't deleted after locks expire, there's also a simpler solution - [here](https://github.com/nspcc-dev/neofs-node/blob/8b65833d3651185fd9078e36b5acac73b1eaf762/pkg/local_object_storage/shard/gc.go#L121) make it consistently: handle expired locks and then handle expired objects. But really I don't like this, because the parallelism in the calculations is lost, and the logic for deleting the locks may still be wrong.

### DIFF
--- a/pkg/local_object_storage/engine/inhume.go
+++ b/pkg/local_object_storage/engine/inhume.go
@@ -264,14 +264,40 @@ func (e *StorageEngine) processExpiredObjects(addrs []oid.Address) {
 }
 
 func (e *StorageEngine) processExpiredLocks(lockers []oid.Address) {
+	var unlocked, expired []oid.Address
 	for _, sh := range e.unsortedShards() {
-		sh.HandleExpiredLocks(lockers)
+		unlocked = sh.HandleExpiredLocks(lockers)
+		for _, sh2 := range e.unsortedShards() {
+			expired = append(expired, sh2.FilterExpired(unlocked)...)
+		}
+	}
+	expired = removeDuplicateAddresses(expired)
+	e.log.Debug("expired objects after locks expired",
+		zap.Stringers("addrs", expired),
+		zap.Stringers("locks", lockers))
+
+	err := e.inhume(expired, false, nil, 0)
+	if err != nil {
+		e.log.Warn("handling expired locks", zap.Error(err))
 	}
 }
 
 func (e *StorageEngine) processDeletedLocks(lockers []oid.Address) {
+	var unlocked, expired []oid.Address
 	for _, sh := range e.unsortedShards() {
-		sh.HandleDeletedLocks(lockers)
+		unlocked = sh.HandleDeletedLocks(lockers)
+		for _, sh2 := range e.unsortedShards() {
+			expired = append(expired, sh2.FilterExpired(unlocked)...)
+		}
+	}
+	expired = removeDuplicateAddresses(expired)
+	e.log.Debug("expired objects after locks are deleted",
+		zap.Stringers("addrs", expired),
+		zap.Stringers("locks", lockers))
+
+	err := e.inhume(expired, false, nil, 0)
+	if err != nil {
+		e.log.Warn("handling deleted locks", zap.Error(err))
 	}
 }
 
@@ -298,5 +324,18 @@ func oIDsToAddresses(cID cid.ID, oo []oid.ID) []oid.Address {
 		res = append(res, addr)
 	}
 
+	return res
+}
+
+func removeDuplicateAddresses(addrs []oid.Address) []oid.Address {
+	uniqueMap := make(map[oid.Address]struct{})
+	res := make([]oid.Address, 0, len(addrs))
+
+	for _, addr := range addrs {
+		if _, ok := uniqueMap[addr]; !ok {
+			uniqueMap[addr] = struct{}{}
+			res = append(res, addr)
+		}
+	}
 	return res
 }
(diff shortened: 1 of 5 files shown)


pkg/local_object_storage/engine/inhume.go
```diff
@@ -264,14 +265,24 @@ func (e *StorageEngine) processExpiredObjects(addrs []oid.Address) {
 }
 
 func (e *StorageEngine) processExpiredLocks(lockers []oid.Address) {
+	var unlocked []oid.Address
 	for _, sh := range e.unsortedShards() {
-		sh.HandleExpiredLocks(lockers)
+		unlocked = append(unlocked, sh.HandleExpiredLocks(lockers)...)
```
roman-khimov, line 270: Can be a huge list IIUC.

pkg/local_object_storage/engine/inhume.go
```diff
@@ -300,3 +311,28 @@ func oIDsToAddresses(cID cid.ID, oo []oid.ID) []oid.Address {
...
+
+// handleUnlockedObjects filters out unlocked objects by lockers for expiration
+// and marks them for removal.
+func (e *StorageEngine) handleUnlockedObjects(unlocked []oid.Address) error {
+	var expired []oid.Address
+	for _, sh := range e.unsortedShards() {
+		expired = append(expired, sh.FilterExpired(unlocked)...)
+	}
+
+	addrs := util.SliceToMap(expired)
```
roman-khimov, line 323: Why do you need a map here?

pkg/local_object_storage/engine/inhume.go
```diff
@@ -300,3 +311,28 @@ func oIDsToAddresses(cID cid.ID, oo []oid.ID) []oid.Address {
...
+
+	for addr := range addrs {
+		ok, err := e.inhumeAddr(addr, false, nil, 0)
+		if err != nil {
+			e.log.Warn("could not inhume object in shard", zap.Error(err))
+			return err
```
roman-khimov, line 329: Save error, continue looping?

roman-khimov (commented):
Looks legit.

pkg/local_object_storage/engine/inhume.go
End-rey: I thought there would be repetitions.

pkg/local_object_storage/engine/inhume.go
End-rey: So, do we need to filter this right away? I can do an internal loop on the shards for filtering. Can there be many shards?

pkg/local_object_storage/engine/inhume.go
End-rey: I did the same like here:
https://github.com/nspcc-dev/neofs-node/blob/8b65833d3651185fd9078e36b5acac73b1eaf762/pkg/local_object_storage/engine/inhume.go#L54-L60

pkg/local_object_storage/engine/inhume.go
roman-khimov: Shouldn't be, I think. But you can sort it in-place and skip duplicates as well, should be a little cheaper than creating a new map.

carpawell: in fact i looked at `lockSingle` and it really looks like repetitions should appear

pkg/local_object_storage/engine/inhume.go
```diff
@@ -264,14 +264,35 @@ func (e *StorageEngine) processExpiredObjects(addrs []oid.Address) {
...
 	for _, sh := range e.unsortedShards() {
-		sh.HandleExpiredLocks(lockers)
+		unlocked = sh.HandleExpiredLocks(lockers)
+		for _, sh2 := range e.unsortedShards() {
+			expired = append(expired, sh2.FilterExpired(unlocked)...)
+		}
+	}
+	ss := []string{}
+	for _, addr := range expired {
+		ss = append(ss, addr.String())
```
carpawell, line 276: `zap.Stringers`?

pkg/local_object_storage/engine/inhume.go
```diff
@@ -264,14 +264,35 @@ func (e *StorageEngine) processExpiredObjects(addrs []oid.Address) {
...
+	}
+	e.log.Debug("expired objects", zap.Strings("addrs", ss))
```
carpawell, line 278: can we extend the message a little? i would like to know that we are handling "new" expired objects after some locks were dropped

End-rey:
Improved the debug message, added the removal of repetitions.

carpawell (approved):

PR #3091, opened 2025-01-27
Drop BoltDB from write-cache

### LINKED ISSUE #3076: Drop BoltDB from write cache
I'm always frustrated when we have something we don't need. BoltDB is absolutely useless for write caching given that the cache uses SSD (see https://github.com/nspcc-dev/neofs-node/issues/2337#issuecomment-2489310539 also).

Drop it from the structure completely. Provide migration, flush all objects from it to the storage and then forever drop the DB, don't use it for object storage.

End-rey:
Closes #3076.

Do I understand correctly that the workers were only for parallel processing from the BoltDB?


pkg/local_object_storage/writecache/flush.go
```diff
@@ -207,9 +207,6 @@ func (c *cache) flushFSTree(ignoreErrors bool) error {
 
 		err = c.flushObject(&obj, data)
 		if err != nil {
-			if ignoreErrors {
```
carpawell, line 210: can you explain this change, please? what led to "the error should be returned as with Bolt" argument?

btw commit message has "Although there _ARE_ `ignoreErrors` flag"

End-rey: I did this for this test: https://github.com/nspcc-dev/neofs-node/blob/8b65833d3651185fd9078e36b5acac73b1eaf762/pkg/local_object_storage/writecache/flush_test.go#L95-L96
It used to work because only the bolt was used and the error was not ignored there when flushing:
https://github.com/nspcc-dev/neofs-node/blob/8b65833d3651185fd9078e36b5acac73b1eaf762/pkg/local_object_storage/writecache/flush.go#L322-L324

roman-khimov: Tricky one. The behavior of ignoring the error comes from b64b14eb54cac04a27049e9eb0ef6fc414a85d59 which is made quite some time after the 0b4c867ef1381a947b2821d9759e4985e12a5249. Semantically, I'd say that if we're to `ignoreErrors`, we should ignore all of them, flush is not guaranteed in this case (best-effort). Treating is a "ignore read errors only" is a possibility too, but maybe it'd be more problematic in real life. The flag itself is suspicious to me, but while we have it we probably better keep the behavior as it is now and then the test should be adjusted accordingly (try flushing with false/true to check for the difference?)

pkg/local_object_storage/writecache/writecache.go
```diff
@@ -142,12 +138,28 @@ func (c *cache) Open(readOnly bool) error {
...
+	if readOnly {
+		c.mode = mode.ReadOnly
+	} else {
+		c.mode = mode.ReadWrite
+	}
+	c.modeMtx.Unlock()
+
+	err = c.migrate()
+	if err != nil {
+		c.log.Warn("could not migrate cache", zap.Error(err))
```
roman-khimov, line 151: Why not returning this error? Objects can be lost here.

pkg/local_object_storage/writecache/writecache.go
```diff
@@ -169,12 +181,64 @@ func (c *cache) Close() error {
...
-		if err != nil {
-			c.db = nil
+	return nil
+}
+
+func (c *cache) migrate() error {
+	path := filepath.Join(c.path, dbName)
+	_, err := os.Stat(path)
+	if errors.Is(err, os.ErrNotExist) {
+		return errors.New("there is no database file")
```
roman-khimov, line 191: Which is perfectly fine after migration, so it's not an error to me.

pkg/local_object_storage/writecache/writecache.go
```diff
@@ -169,12 +181,64 @@ func (c *cache) Close() error {
...
+	}
+
+	db, err := bbolt.Open(path, os.ModePerm, &bbolt.Options{
+		NoFreelistSync: true,
+		ReadOnly:       false,
```
roman-khimov, line 196: Can be `true`, you're not modifying anything.

pkg/local_object_storage/writecache/writecache.go
```diff
@@ -169,12 +181,64 @@ func (c *cache) Close() error {
...
+		Timeout:        time.Second,
+	})
+	if err != nil {
+		return fmt.Errorf("could not open database: %w", err)
+	}
+	defer db.Close()
```
roman-khimov, line 202: Close is manually and delete file after the flush.

pkg/local_object_storage/writecache/writecache.go
```diff
@@ -169,12 +181,64 @@ func (c *cache) Close() error {
...
+			}
+
+			return nil
+		})
+	})
+	if err != nil {
+		return fmt.Errorf("could not migrate default bucket: %w", err)
+	}
+
+	err = db.Batch(func(tx *bbolt.Tx) error {
```
roman-khimov, line 236: This is not needed, no file --- no problem.

pkg/local_object_storage/writecache/flush.go
End-rey: Here in [`SetMode`](https://github.com/nspcc-dev/neofs-node/blob/b3e19e2c00dedfd5d6620a62859567ac7885458c/pkg/local_object_storage/writecache/mode.go#L17), if I change `ignoreErrors` to `false`:
https://github.com/nspcc-dev/neofs-node/blob/b3e19e2c00dedfd5d6620a62859567ac7885458c/pkg/local_object_storage/writecache/mode.go#L21-L26
All tests will be passed without this change.

End-rey:
An error is currently being returned after migration. Can you check which errors may appear, specifically I am concerned when there are no default bucket or other errors inside the database `View`.

pkg/local_object_storage/writecache/writecache.go
```diff
@@ -163,18 +180,71 @@ func (c *cache) Close() error {
 
 	if c.closeCh != nil {
 		close(c.closeCh)
-	}
-	c.wg.Wait()
```
roman-khimov, line 167: You're not waiting for the flusher thread to finish, please do it (via channel close for example).

pkg/local_object_storage/writecache/flush.go
```diff
@@ -1,165 +1,45 @@
...
 const (
-	// flushBatchSize is amount of keys which will be read from cache to be flushed
-	// to the main storage. It is used to reduce contention between cache put
-	// and cache persist.
-	flushBatchSize = 512
-	// defaultFlushWorkersCount is number of workers for putting objects in main storage.
-	defaultFlushWorkersCount = 20
 	// defaultFlushInterval is default time interval between successive flushes.
-	defaultFlushInterval = time.Second
+	defaultFlushInterval = 10 * time.Second
```
roman-khimov, line 19: This change should be reflected in documentation.

pkg/local_object_storage/shard/shutdown_test.go
```diff
@@ -28,8 +27,7 @@ func TestWriteCacheObjectLoss(t *testing.T) {
 
 	dir := t.TempDir()
 	wcOpts := []writecache.Option{
-		writecache.WithSmallObjectSize(smallSize),
-		writecache.WithMaxObjectSize(smallSize * 2)}
+		writecache.WithMaxObjectSize(size * 2)}
```
roman-khimov, line 30: Just declare `maxObjSize = whatever` and use it.

pkg/local_object_storage/writecache/writecache.go
```diff
@@ -163,18 +180,71 @@ func (c *cache) Close() error {
...
+}
+
+func (c *cache) migrate() error {
+	path := filepath.Join(c.path, dbName)
+	_, err := os.Stat(path)
+	if errors.Is(err, os.ErrNotExist) {
+		c.log.Debug("no migration needed, there is no database file")
+		return nil
+	}
+
```
roman-khimov, line 196: I think we need some Info message when migration starts, it can take some time.

pkg/local_object_storage/writecache/writecache.go
```diff
@@ -163,18 +180,71 @@ func (c *cache) Close() error {
...
+	if err != nil {
+		return fmt.Errorf("could not close database: %w", err)
+	}
+
+	err = os.Remove(path)
+	if err != nil {
+		return fmt.Errorf("could not remove database file: %w", err)
+	}
+
+	c.log.Debug("successfully migrated and removed database file")
```
roman-khimov, line 248: Info, it's important.

pkg/local_object_storage/writecache/writecache.go
```diff
@@ -163,18 +180,71 @@ func (c *cache) Close() error {
...
+
+			if err := c.flushObject(&obj, v); err != nil {
+				return err
+			}
+
+			return nil
+		})
+	})
+	if err != nil {
+		return fmt.Errorf("could not migrate default bucket: %w", err)
```
roman-khimov, line 235: You need to `db.Close()` before doing this `return`.

pkg/local_object_storage/writecache/flush.go
roman-khimov: These flushers are very tricky in fact. The node should continue funtioning, but ignoring errors silently is bad. The most common flush error is "out of disk space" (the other is "your disk is broken forever"), continuing flushes in this case is not very productive.

pkg/local_object_storage/writecache/flush.go
End-rey: I did't find where in the documentation it is specified about the flush interval for the write-cache. It's not set up anywhere.

pkg/local_object_storage/writecache/writecache.go
End-rey: I didn't really understand what channel you were referring to, but I understood that the `WaitGroup` was needed.

roman-khimov: WG works too, but usually for single threads channels are being used to synchronize exits.

pkg/local_object_storage/writecache/flush.go
carpawell: > The flag itself is suspicious to me

as i remember, it was provided mostly for degraded cases when we think that disk is dying and we are trying to save (persist) as many objects as possible, so we ignore all problems about object X to be able to continue to object X+1

> Treating is a "ignore read errors only" is a possibility too

i guess this is the closest interpretation to what was planned to do

IMO, it is a good discussion for a separate issue, we also have a shard dump routine that also has some `ignoreError` things. can be fixed/discussed at a time

pkg/local_object_storage/writecache/writecache.go
```diff
@@ -142,12 +136,35 @@ func (c *cache) Open(readOnly bool) error {
...
+	c.modeMtx.Lock()
+	if readOnly {
+		c.mode = mode.ReadOnly
+	} else {
+		c.mode = mode.ReadWrite
+	}
+	c.modeMtx.Unlock()
+
+	// Migration part
+	if !readOnly {
```
carpawell, line 148: can we prevent `!` and just invert branches?

pkg/local_object_storage/writecache/writecache.go
```diff
@@ -163,18 +180,71 @@ func (c *cache) Close() error {
...
 
-	var err error
-	if c.db != nil {
-		err = c.db.Close()
-		if err != nil {
-			c.db = nil
+	return nil
+}
+
+func (c *cache) migrate() error {
```
carpawell, line 189: can we have some simple migrate test?

CHANGELOG.md
```diff
@@ -15,12 +15,17 @@ Changelog for NeoFS Node
...
 
 ### Removed
+- BoltDB from write-cache (#3091)
 
 ### Updated
 
 ### Updating from v0.44.2
+`small_object_size`, `workers_number`, `max_batch_size` and `max_batch_delay`
+paramteters are removed from `writecache` config. These parameters are related
+to the BoltDB part of the write-cache, which is dropped from the code.
```
carpawell, line 28: add, please, a few words that migration will be performed automatically

cmd/neofs-lens/internal/writecache/root.go
```diff
@@ -26,11 +25,14 @@ func init() {
 	Root.AddCommand(getCMD)
 }
 
-func openWC() (*bbolt.DB, error) {
-	db, err := writecache.OpenDB(vPath, true)
+// openWC opens and returns read-only writecache.Cache located in vPath.
+func openWC() (writecache.Cache, error) {
+	ws := writecache.New(writecache.WithPath(vPath))
```
carpawell, line 30:
```suggestion
	wc := writecache.New(writecache.WithPath(vPath))
```

pkg/local_object_storage/writecache/options.go
```diff
@@ -93,48 +83,13 @@ func WithMaxObjectSize(sz uint64) Option {
...
-func WithMaxBatchDelay(d time.Duration) Option {
-	return func(o *options) {
-		if d > 0 {
-			o.maxBatchDelay = d
-		}
-	}
-}
-
 // WithNoSync sets an option to allow returning to caller on PUT before write is persisted.
 // Note, that we use this flag for FSTree only and DO NOT use it for a bolt DB because
```
carpawell, line 94: this line should be dropped

carpawell (commented):
> Do I understand correctly that the workers were only for parallel processing from the BoltDB?

TBH, I do not understand what is the key difference b/w db and FSTree if we are talking about flushing. I think it is possible to speed up flushing if there are 2+ workers. We do spend some time on object reading from WC cache and its unmarshaling. @roman-khimov

Also, we store raw objects now, I think closing this issue should be done with more precise WC size handling: https://github.com/nspcc-dev/neofs-node/blob/9bc61c5d636d9b2acda0befa70e00c74c8bf5e55/pkg/local_object_storage/writecache/state.go#L11-L13
@roman-khimov

roman-khimov:
Estimations and workers are separate things to be done in separate issues, please open them if they're not here yet. Eventually I'd like wc to keep it's object list in memory (with sizes) and yes, it needs better flushing mechanism. But this doesn't change anything anything here, Bolt is to be removed and it is removed.

pkg/local_object_storage/writecache/flush.go
carpawell: https://github.com/nspcc-dev/neofs-node/issues/3099

pkg/local_object_storage/writecache/writecache.go
End-rey: It wasn't easy with the channel because of read-only. It's clearer and easier for me to use WG.

carpawell:
> Estimations and workers are separate things to be done in separate issues, please open them if they're not here yet.

Do not agree. It used to be parallel, it is not now, it is a degradation. It used to have incorrect estimations only because of the existing bbolt base, now we do not have it but still, our expectation about WC size is numOfObjs*theBiggestPossibleObj. I see no better place for these fixes except this PR.

Opened: https://github.com/nspcc-dev/neofs-node/issues/3100, https://github.com/nspcc-dev/neofs-node/issues/3101.

roman-khimov:
> It used to be parallel

Bolt-only. FSTree is all the same.

roman-khimov (approved):

PR #3097, opened 2025-01-29
Object inlines

roman-khimov:
The last chunk from #3057. Less api-go overall, should be the same functionally.


pkg/services/object/server.go
```diff
@@ -733,20 +734,152 @@ func (s *server) Search(req *protoobject.SearchRequest, gStream protoobject.Obje
...
+		addrs := node.AddressGroup()
+		for i := range addrs {
+			res, err := searchOnRemoteNode(ctx, c, addrs[i], nodePub, req)
+			if err == nil {
+				return res, nil
+			}
+			if firstErr == nil {
+				firstErr = err
+			}
+			// TODO: log error
```
carpawell, line 1739: todo thing

pkg/services/object/server.go
```diff
@@ -102,13 +105,22 @@ type FSChain interface {
 	LocalNodeUnderMaintenance() bool
 }
 
-// Storage groups ops of the node's object storage required to serve NeoFS API
-// Object service.
+type sessions interface {
```
carpawell, line 108: do not see why it should be a separate private interface

carpawell (approved):
Many many many many TODOs. But it seems to be in a working condition.

pkg/services/object/server.go
roman-khimov: https://github.com/nspcc-dev/neofs-node/issues/3098

pkg/services/object/server.go
roman-khimov: Yeah, not really needed, but we can't delay SDK much.

PR #3054, opened 2024-12-16
internal: drop unused NeoFS SDK-related code

AnnaShaleva:
https://github.com/nspcc-dev/neofs-node/issues/2477 is closed long time ago, I suppose this code is irrelevant anymore.

Ref. https://github.com/nspcc-dev/neo-go/issues/3757#issuecomment-2545055792.

PR #3103, opened 2025-01-30
New SDK

roman-khimov:
Replaces #3023, #3054. Depends on #3097.


roman-khimov:
Full runs:
https://rest.fs.neo.org/HXSaMJXk2g8C14ht8HSi7BBaiYZ1HeWh2xnWPGQCg4H6/2381-1738258976/index.html
https://rest.fs.neo.org/HXSaMJXk2g8C14ht8HSi7BBaiYZ1HeWh2xnWPGQCg4H6/2381-1738258792/index.html

Only https://github.com/nspcc-dev/neofs-testcases/issues/944 is a problem, but it's the same with the master branch.

pkg/core/object/fmt.go
```diff
@@ -134,12 +134,7 @@ func (v *FormatValidator) Validate(obj *object.Object, unprepared bool) error {
 		} else {
 			// V2 split
 
-			if !firstSet {
-				// first part only
-				if obj.Parent() == nil {
-					return errors.New("v2 split: first object part does not have parent header")
-				}
-			} else {
+			if firstSet {
```
carpawell, line 137: why?

pkg/network/cache/multi.go
```diff
@@ -374,13 +357,23 @@ func (x *multiClient) RawForAddress(addr network.Address, f func(*grpc.ClientCon
...
 	if err != nil {
 		x.ReportError(err)
 	}
 	return err
 }
 
+func (x *multiClient) Conn() *grpc.ClientConn {
+	var cc *grpc.ClientConn
+
+	_ = x.iterateClients(context.TODO(), func(c clientcore.Client) error {
```
carpawell, line 370: will context be meaningful in the future?

pkg/core/object/fmt.go
roman-khimov: I've expected this one. IIUC if `obj.HasParent()` now it can never have `obj.Parent() == nil` and semantically this makes total sense. Which in turn means that the whole `if !firstSet` branch turns no-op. See the respective test, it doesn't work with or without this code.

pkg/network/cache/multi.go
roman-khimov: I don't think so. We need to get some connection, we're not making any real requests here.

cthulhu-rider (approved):

pkg/core/object/fmt.go
carpawell: forgot we are in `obj.HasParent()` branch

pkg/network/cache/multi.go
carpawell: ok, just do not like `context.TODO()` in any "release" code

carpawell (approved):

PR #3105, opened 2025-01-30
logger: add `timestamp` config option

### LINKED ISSUE #3067: Add an option to enable timestamps in node logs
Logs without timestamps can sometimes be hard to read. 
```
info	neofs-ir/main.go:156	prometheus is enabled
info	neofs-ir/main.go:156	pprof is enabled
info	core/blockchain.go:234	initial gas supply is not set or wrong, setting default value	{"InitialGASSupply": "52000000"}
```
Adding timestamps, e.g. `13:29:05.707384` to each log entry would significantly increase logs' usefulness.

End-rey:
Closes #3067.


cmd/neofs-ir/main.go
```diff
@@ -67,7 +67,7 @@ func main() {
 	c.Level = logLevel
 	c.Encoding = cfg.GetString("logger.encoding")
 	c.Sampling = nil
-	if term.IsTerminal(int(os.Stdout.Fd())) {
+	if term.IsTerminal(int(os.Stdout.Fd())) || cfg.GetBool("logger.timestamp") {
```
roman-khimov, line 70: I'd rather follow `logger.timestamp` if it's set. So that if it's explicitly `false` then tty detection logic shouldn't be used.

roman-khimov (approved):

PR #3106, opened 2025-01-31
Change `WC` size estimations

### LINKED ISSUE #3101: Fix WC size estimations
Currently, its size (to control cache size and not overflow) is defined as `numOfObjs*theBiggestPossibleObj`. It is not precise at all. It was done mostly because of uncontrollable bbolt db size for small objects. But there is no bbolt db in WC now: https://github.com/nspcc-dev/neofs-node/pull/3091

## Expected Behavior
Taken space is smth like a sum of every object that is currently stored in WC.

## Current Behavior
Taken space is a magic product: https://github.com/nspcc-dev/neofs-node/blob/25314f58cb6d1066b10a405a12240dd830587de2/pkg/local_object_storage/writecache/state.go#L11-L13.

## Possible Solution
Store the sum from `Expected Behavior` and change it accordingly with every PUT and DELETE.

Look at https://github.com/nspcc-dev/neofs-node/blob/25314f58cb6d1066b10a405a12240dd830587de2/pkg/local_object_storage/writecache/state.go#L11-L13.

## Context
https://github.com/nspcc-dev/neofs-node/pull/3091#pullrequestreview-2580795118

## Regression
No.

## Your Environment
v0.44.2.

End-rey:
Closes #3101.

> Keep object list in memory along with sizes (map address to size)

Why are we doing this? We don't use it anywhere, do we?

Also, I forgot to include tests for migration from #3091, so I did it here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Changelog for NeoFS Node
 
 ### Changed
 - Number of cuncurrenly handled notifications from the chain was increased from 10 to 300 for IR (#3068)
+- Write-cache size estimations (#3106)
 
 ### Removed
 - Drop creating new eacl tables with public keys (#3096)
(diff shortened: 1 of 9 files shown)


roman-khimov:
> Why are we doing this? We don't use it anywhere, do we?

We will. It allows to answer `exists` or `get` or `head` faster (remember that any shard operation has to check WC first) and it'll also be useful for flushing. The presumption is that write cache is a small but highly dynamic structure (objects come and go), this list is useful for it.

End-rey:
If I understand you correctly, I added a new `sizeHandler` to the `iterate` func to collect sizes.

roman-khimov (approved):

roman-khimov:
`iterate` is more and more hairy now, but this should do the job.

PR #3107, opened 2025-01-31
ir: allow only predefined known system attributes

### LINKED ISSUE #3090: Make __NEOFS__METAINFO_CONSISTENCY processing config-dependent
I'm always frustrated when we don't have any controls over net experimental features.

Add IR config to allow and process `__NEOFS__METAINFO_CONSISTENCY`. If it's not enabled --- it's forbidden.

## Additional context
https://github.com/nspcc-dev/neofs-node/issues/2878

carpawell:
This is a good feature in general, but also it closes #3090 in some way.


carpawell:
@roman-khimov, this is kinda always a welcome feature but not sure if it solves #3090. In fact yes, no so such containers will be accepted until we are ready but some `experimental` config section for our apps can also be useful (and it does not require app rebuild).

roman-khimov:
Isn't that's written in #3090 already? We need some configuration, yes.

carpawell:
> Isn't that's written in #3090 already?

I meant, I am not sure it should be done. It can be useful in general (for other features) but strict attributes feature check also stops us from incorrect containers. The only real reason for configuration is turning meta on at some magic time but that is not true, I think we just release it and update test/main networks, it will not be unexpected. I don't mind configuration, I even think I will add it but I do not see this as required in the current context. But configuration is configuration, it should added, changelogged, presented, maintained, explained to users, etc.

roman-khimov:
How would you test it if you don't have any configuration option?

carpawell:
Do not see real cases when we turn it off and on many times. Added experimental value. Have not added it to the changelog since I think we are the only ones who should know about it. Added to the config examples though.

roman-khimov (approved):

PR #3111, opened 2025-02-07
Serve `ObjectService.SearchV2` RPC

cthulhu-rider:
* based on and blocked by #3080

### DIFF
--- a/pkg/core/object/metadata.go
+++ b/pkg/core/object/metadata.go
@@ -1,0 +1,128 @@
+package object
+
+import (
+	"bytes"
+	"math/big"
+	"slices"
+	"strings"
+
+	"github.com/nspcc-dev/neofs-sdk-go/client"
+)
+
+// TODO: docs.
+func MergeSearchResults(lim uint16, withAttr bool, sets [][]client.SearchResultItem, mores []bool) ([]client.SearchResultItem, bool) {
+	if lim == 0 || len(sets) == 0 {
+		return nil, false
+	}
+	if len(sets) == 1 {
+		n := min(uint16(len(sets[0])), lim)
+		return sets[0][:n], n < lim || slices.Contains(mores, true)
+	}
+	lim = calcMaxUniqueSearchResults(lim, sets)
+	res := make([]client.SearchResultItem, 0, lim)
+	var more bool
+	var minInt *big.Int
+	for minInd := -1; ; minInd, minInt = -1, nil {
+		for i := range sets {
+			if len(sets[i]) == 0 {
+				continue
+			}
+			if minInd < 0 {
+				minInd = i
+				if withAttr {
+					minInt, _ = new(big.Int).SetString(sets[i][0].Attributes[0], 10)
+				}
+				continue
+			}
+			cmpID := bytes.Compare(sets[i][0].ID[:], sets[minInd][0].ID[:])
+			if cmpID == 0 {
+				continue
+			}
+			if withAttr {
+				var curInt *big.Int
+				if minInt != nil {
+					curInt, _ = new(big.Int).SetString(sets[i][0].Attributes[0], 10)
+				}
+				var cmpAttr int
+				if curInt != nil {
+					cmpAttr = curInt.Cmp(minInt)
+				} else {
+					cmpAttr = strings.Compare(sets[i][0].Attributes[0], sets[minInd][0].Attributes[0])
+				}
+				if cmpAttr != 0 {
+					if cmpAttr < 0 {
+						minInd = i
+						if minInt != nil {
+							minInt = curInt
+						} else {
+							minInt, _ = new(big.Int).SetString(sets[i][0].Attributes[0], 10)
+						}
+					}
+					continue
+				}
+			}
+			if cmpID < 0 {
+				minInd = i
+				if withAttr {
+					minInt, _ = new(big.Int).SetString(sets[minInd][0].Attributes[0], 10)
+				}
+			}
+		}
+		if minInd < 0 {
+			break
+		}
+		res = append(res, sets[minInd][0])
+		if uint16(len(res)) == lim {
+			if more = len(sets[minInd]) > 1 || slices.Contains(mores, true); !more {
+			loop:
+				for i := range sets {
+					if i == minInd {
+						continue
+					}
+					for j := range sets[i] {
+						if more = sets[i][j].ID != sets[minInd][0].ID; more {
+							break loop
+						}
+					}
+				}
+			}
+			break
+		}
+		for i := range sets {
+			if i == minInd {
+				continue
+			}
+			for j := range sets[i] {
+				if sets[i][j].ID == sets[minInd][0].ID {
+					sets[i] = sets[i][j+1:]
+					break
+				}
+			}
+		}
+		sets[minInd] = sets[minInd][1:]
+	}
+	return res, more
+}
+
+func calcMaxUniqueSearchResults(lim uint16, sets [][]client.SearchResultItem) uint16 {
+	n := uint16(len(sets[0]))
+	if n >= lim {
+		return lim
+	}
+	for i := 1; i < len(sets); i++ {
+	nextItem:
+		for j := range sets[i] {
+			for k := range i {
+				for l := range sets[k] {
+					if sets[k][l].ID == sets[i][j].ID {
+						continue nextItem
+					}
+				}
+			}
+			if n++; n == lim {
+				return n
+			}
+		}
+	}
+	return n
+}
--- a/pkg/local_object_storage/engine/select.go
+++ b/pkg/local_object_storage/engine/select.go
@@ -2,8 +2,11 @@ package engine
 
 import (
 	"errors"
+	"fmt"
 
 	objectcore "github.com/nspcc-dev/neofs-node/pkg/core/object"
+	meta "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/metabase"
+	"github.com/nspcc-dev/neofs-sdk-go/client"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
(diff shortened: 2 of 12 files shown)


cthulhu-rider:
at the moment, API is available from the SN

there are still a few cases to fix regarding the final cursor

pkg/core/object/metadata.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// NewSearchResultMerger constructs SearchResultMerger merging up to limit
+// results.
+func NewSearchResultMerger(limit uint16) *SearchResultMerger {
+	return &SearchResultMerger{limit: limit}
+}
+
+// Add accumulates the next result.
+func (x *SearchResultMerger) Add(withCursor bool, next []client.SearchResultItem) {
```
roman-khimov, line 29: This is not very efficient, you're iterating over the whole result set again and again replacing some elements. Operating over the whole set of results would be much more efficient, you just take whatever is the minimal one at each step until you have enough results. This requires a bit more memory, but given our limits it's ok.

pkg/core/object/metadata.go
```diff
@@ -0,0 +1,115 @@
...
+	if len(next) > int(x.limit) {
+		panic(fmt.Sprintf("limit of items exceeded %d > %d", len(next), x.limit))
+	}
+	if withCursor && len(next) < int(x.limit) {
+		panic(fmt.Sprintf("cursor is set with fewer items %d < %d", len(next), x.limit))
```
roman-khimov, line 34: So remote node sending 1 item and a cursor will kill us?

pkg/core/object/metadata.go
```diff
@@ -0,0 +1,115 @@
...
+// NewSearchResultMerger constructs SearchResultMerger merging up to limit
+// results.
+func NewSearchResultMerger(limit uint16) *SearchResultMerger {
+	return &SearchResultMerger{limit: limit}
+}
+
+// Add accumulates the next result.
+func (x *SearchResultMerger) Add(withCursor bool, next []client.SearchResultItem) {
+	if len(next) > int(x.limit) {
+		panic(fmt.Sprintf("limit of items exceeded %d > %d", len(next), x.limit))
```
roman-khimov, line 31: Remote node sending 1001 element makes us panic?

pkg/core/object/metadata.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	if withCursor && len(next) < int(x.limit) {
+		panic(fmt.Sprintf("cursor is set with fewer items %d < %d", len(next), x.limit))
+	}
+	if len(next) == 0 {
+		return
+	}
+	if x.buf == nil {
+		x.buf = slices.Grow(next, int(x.limit))[:x.limit]
```
roman-khimov, line 40: Not very efficient, you may have no other elements. Many searches end up with 1-3 elements and allocating a 1000 for them doesn't seem to be good.

pkg/local_object_storage/engine/select.go
```diff
@@ -49,3 +52,94 @@ func (e *StorageEngine) Select(cnr cid.ID, filters object.SearchFilters) ([]oid.
 
 	return addrList, nil
 }
+
+// List returns `limit` available physically storage object addresses in engine.
+// If limit is zero, then returns all available object addresses.
+//
+// Returns an error if executions are blocked (see BlockExecution).
+func (e *StorageEngine) List(limit uint64) ([]oid.Address, error) {
```
roman-khimov, line 60: Rebase artifact?

pkg/local_object_storage/engine/select.go
```diff
@@ -49,3 +52,94 @@ func (e *StorageEngine) Select(cnr cid.ID, filters object.SearchFilters) ([]oid.
...
+		e.reportShardError(shs[0], "could not select objects from shard", err)
+	}
+	if len(shs) == 1 {
+		return items, nextCursor, nil
+	}
+	shs = shs[1:]
+	mrg := objectcore.NewSearchResultMerger(count)
+	mrg.Add(nextCursor != nil, items)
+	for i := range shs {
+		// TODO: avoid allocation inside and pass pre-allocated buffer
```
roman-khimov, line 126: We can afford 1K elements buffer. Usually it's some KBs for memory. Likely you won't have 100 shards. It can be optimized a little with iterators where you'd picking elements one by one from respective shards, but let's keep it simple for now.

pkg/local_object_storage/engine/select.go
```diff
@@ -49,3 +52,94 @@ func (e *StorageEngine) Select(cnr cid.ID, filters object.SearchFilters) ([]oid.
...
+
+	return addrList, nil
+}
+
+// Search performs Search op on all underlying shards and returns merged result.
+//
+// Fails instantly if executions are blocked (see [StorageEngine.BlockExecution]).
+func (e *StorageEngine) Search(cnr cid.ID, fs object.SearchFilters, attrs []string, cursor *meta.SearchCursor, count uint16) ([]client.SearchResultItem, *meta.SearchCursor, error) {
+	if e.metrics != nil {
+		defer elapsed(e.metrics.AddSearchDuration)() // TODO: sep metric?
```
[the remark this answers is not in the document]
carpawell: this metric number is magic then

pkg/core/object/metadata.go
```diff
@@ -0,0 +1,115 @@
...
+
+// NewSearchResultMerger constructs SearchResultMerger merging up to limit
+// results.
+func NewSearchResultMerger(limit uint16) *SearchResultMerger {
+	return &SearchResultMerger{limit: limit}
+}
+
+// Add accumulates the next result.
+func (x *SearchResultMerger) Add(withCursor bool, next []client.SearchResultItem) {
+	if len(next) > int(x.limit) {
```
carpawell, line 30: i do not understand these panics: a merger is created with some limit and then it is filled with some number of `Add` calls. this number is undefined and is based on shard number. why is it not possible (not expected) to exceed it and why results must be exactly `limit` len?

cthulhu-rider:
TODO https://github.com/nspcc-dev/neofs-node/pull/3124#discussion_r1954077458

pkg/services/object/server.go
```diff
@@ -1848,8 +1868,287 @@ func (s *server) Replicate(_ context.Context, req *protoobject.ReplicateRequest)
...
+	}
+	if body.Count == 0 {
+		return nil, errors.New("zero count")
+	} else if body.Count > 1000 {
+		return nil, fmt.Errorf("number of attributes %d exceeds the limit 1000", body.Count)
+	}
+	if len(body.Filters) > 8 {
+		return nil, fmt.Errorf("number of filter %d exceeds the limit 8", len(body.Filters))
+	}
+	if len(body.Attributes) > 4 {
```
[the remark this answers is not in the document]
cthulhu-rider: (S3 GW)

kk

pkg/local_object_storage/engine/select.go
cthulhu-rider: droppped

pkg/local_object_storage/engine/select.go
cthulhu-rider: this was a nightmare

pkg/local_object_storage/engine/select.go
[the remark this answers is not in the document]
cthulhu-rider: leaving

cthulhu-rider:
at the moment, as far as i tested, everything works. Gonna add following optimizations:
* https://github.com/nspcc-dev/neofs-node/pull/3111#discussion_r1953276251
* https://github.com/nspcc-dev/neofs-node/pull/3111#discussion_r1953312366

pkg/core/object/metadata.go
cthulhu-rider: done

pkg/core/object/metadata.go
cthulhu-rider: gone

pkg/core/object/metadata.go
cthulhu-rider: gone

pkg/services/object/server.go
```diff
@@ -1848,8 +1868,287 @@ func (s *server) Replicate(_ context.Context, req *protoobject.ReplicateRequest)
...
+				if !signed {
+					req.MetaHeader = &protosession.RequestMetaHeader{Ttl: 1, Origin: req.MetaHeader}
+					var err error
+					if req.VerifyHeader, err = neofscrypto.SignRequestWithBuffer(neofsecdsa.Signer(s.signer), req, nil); err != nil {
+						signErr = fmt.Errorf("sign request: %w", err)
+						return false
+					}
+					signed = true
+				}
+				// TODO: consider parallelism
```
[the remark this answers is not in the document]
cthulhu-rider: with config similar to https://github.com/nspcc-dev/neofs-node/blob/eb90734ebbde81aa0a304e90bb9ccd029b02642f/config/example/node.yaml#L106, right?

pkg/core/object/metadata.go
cthulhu-rider: gone

pkg/services/object/server.go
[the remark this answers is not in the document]
roman-khimov: It can be used for amplification attack if not limited, but at the same time I don't like these pools either. We can follow the same approach for now just to have some limits already.

pkg/core/object/metadata.go
```diff
@@ -0,0 +1,128 @@
...
+	nextItem:
+		for j := range sets[i] {
+			for k := range i {
+				for l := range sets[k] {
+					if sets[k][l].ID == sets[i][j].ID {
+						continue nextItem
+					}
+				}
+			}
+			if n++; n == lim {
```
roman-khimov, line 122: Does it have to be perfect? IIUC it's only used for allocation optimization, we can just return the sum of sets (or `lim`). These cycles are somewhat frightening, even though `lim` should help in most cases.

pkg/core/object/metadata.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	if len(sets) == 1 {
+		n := min(uint16(len(sets[0])), lim)
+		return sets[0][:n], n < lim || slices.Contains(mores, true)
+	}
+	lim = calcMaxUniqueSearchResults(lim, sets)
+	res := make([]client.SearchResultItem, 0, lim)
+	var more bool
+	var minInt *big.Int
+	for minInd := -1; ; minInd, minInt = -1, nil {
```
roman-khimov, line 25: I think it'd be much easier with https://pkg.go.dev/container/heap, you only need `len(sets)` of elements to store there, push all of them into heap, then pick the lowest and push a new one from the same queue.

pkg/core/object/metadata.go
cthulhu-rider: case-by-case. With this, when merging N same (or slightly different) results, we can alloc ~N times smaller buffer

anyway, code is already written. If u care about performance costs (we spend CPU only here), let's create an issue

pkg/core/object/metadata.go
cthulhu-rider: i'd try this withinin a separate issue if u dont mind

pkg/services/object/server.go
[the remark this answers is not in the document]
cthulhu-rider: i added `ants.Pool` with static capacity `100`, and would discuss/add config in a separate issue if u dont mind

roman-khimov (approved):

PR #3123, opened 2025-02-12
Metadata bucket and GC

cthulhu-rider:


pkg/local_object_storage/metabase/metadata.go
```diff
@@ -111,6 +110,32 @@ func putMetadata(tx *bbolt.Tx, cnr cid.ID, id oid.ID, ver version.Version, owner
...
+func deleteMetadata(tx *bbolt.Tx, cnr cid.ID, id oid.ID) error {
+	metaBkt := tx.Bucket(metaBucketKey(cnr))
+	if metaBkt == nil {
+		return nil
+	}
+	// removed keys must be pre-collected according to BoltDB docs.
+	// TODO: consider calculating key set from the object header
+	ks := make([][]byte, 0, 10) // at least system attributes (approx)
+	c := metaBkt.Cursor()
+	for k, _ := c.First(); k != nil; k, _ = c.Next() {
```
roman-khimov, line 122: It's horribly expensive and redundant to me. You have `metaPrefixIDAttr` data that can be used to find all the other keys.

cthulhu-rider: sure, done

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -111,6 +110,44 @@ func putMetadata(tx *bbolt.Tx, cnr cid.ID, id oid.ID, ver version.Version, owner
...
+	for kIDAttr, _ := c.Seek(pref); bytes.HasPrefix(kIDAttr, pref); kIDAttr, _ = c.Next() {
+		sepInd := bytes.LastIndex(kIDAttr, utf8Delimiter)
+		if sepInd < 0 {
+			return fmt.Errorf("invalid key with prefix 0x%X in meta bucket: missing delimiter", kIDAttr[0])
+		}
+		kAttrID := slices.Clone(kIDAttr)
+		kAttrID[0] = metaPrefixAttrIDPlain
+		copy(kAttrID[1:], kIDAttr[1+oid.Size:])
+		copy(kAttrID[len(kAttrID)-oid.Size:], id[:])
+		if val := kIDAttr[sepInd+utf8DelimiterLen:]; len(val) == intValLen && val[0] <= 1 { // most likely integer
```
roman-khimov, line 135: And if it's not an int it'll be OK anyway? I'm a bit surprised that IDAttr stores numerics in this format as well. Although it should be OK since we're pretty strict on what is a number and treat requests in the same way attributes are treated.

roman-khimov (approved):

pkg/local_object_storage/metabase/metadata.go
cthulhu-rider: > And if it's not an int it'll be OK anyway?

yes cuz `kAttrID` is added anyway. If non-int, then `kAttrIDInt` does not exist and `Delete()` is nop

cthulhu-rider: > urprised that IDAttr stores numerics in this format

that's a good point. I thought about this again, and seems like it'd be more convenient to store in text format:
1. attributes are collected as text, so easier to get
2. secondary filters will be numerical less often
3. numbers approaching max uint64 will be quite rare in practice. Larger ones billion times rarer. Max64 is 20B len in text format while we take 33B for each integer

according to this, i think storing `metaPrefixIDAttr` attributes in text format gonna be more efficient. What do u think?

roman-khimov: The original intent was to keep attribute contents in IDAttr index as is. If this doesn't break anything then I'd prefer this. If we need to do `<>` numeric comparisons we can convert on the fly, but mostly it'd be about exact matches, `<>` is more relevant for the primary attribute.

cthulhu-rider: gonna do within 3130

PR #3124, opened 2025-02-12
node/metabase: Support ROOT and PHY search filters in new meta bucket

cthulhu-rider:


pkg/local_object_storage/metabase/metadata.go
```diff
@@ -794,3 +807,10 @@ func (x *metaAttributeSeeker) restoreVal(id []byte, attr string, stored []byte)
 	}
 	return string(stored), nil
 }
+
+func convertFilterValue(f object.SearchFilter) (object.SearchMatchType, string) {
+	if attr := f.Header(); attr == object.FilterRoot || attr == object.FilterPhysical {
+		return object.MatchStringEqual, binPropMarker
```
roman-khimov, line 813: Do we have a check anywhere for these restricted (per attribute) operations?

roman-khimov (approved):

pkg/local_object_storage/metabase/metadata.go
cthulhu-rider: no checks. U mean that ROOT and PHY aint used w/ other matchers?

roman-khimov: Yes, it should be a part of the request semantics check.

cthulhu-rider: gonna do in #3111

carpawell (approved):

PR #3125, opened 2025-02-12
Metabase migration for SearchV2

cthulhu-rider:
* based on and blocked by #3124

### DIFF
--- a/pkg/local_object_storage/metabase/version.go
+++ b/pkg/local_object_storage/metabase/version.go
@@ -1,17 +1,21 @@
 package meta
 
 import (
+	"bytes"
 	"encoding/binary"
 	"errors"
 	"fmt"
 
 	objectconfig "github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/object"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/util/logicerr"
+	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
+	"github.com/nspcc-dev/neofs-sdk-go/object"
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"go.etcd.io/bbolt"
 )
 
 // currentMetaVersion contains current metabase version.
-const currentMetaVersion = 3
+const currentMetaVersion = 4
 
 var versionKey = []byte("version")
 
@@ -74,6 +78,7 @@ func getVersion(tx *bbolt.Tx) (uint64, bool) {
 
 var migrateFrom = map[uint64]func(*DB, *bbolt.Tx) error{
 	2: migrateFrom2Version,
+	3: migrateFrom3Version,
 }
 
 func migrateFrom2Version(db *DB, tx *bbolt.Tx) error {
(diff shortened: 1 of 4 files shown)


pkg/local_object_storage/metabase/version.go
```diff
@@ -105,3 +110,50 @@ func migrateFrom2Version(db *DB, tx *bbolt.Tx) error {
 
 	return updateVersion(tx, 3)
 }
+
+func migrateFrom3Version(_ *DB, tx *bbolt.Tx) error {
+	c := tx.Cursor()
+	pref := []byte{metadataPrefix}
+	if k, _ := c.Seek(pref); bytes.HasPrefix(k, pref) {
+		return fmt.Errorf("key with prefix 0x%X detected", pref)
```
carpawell, line 118: could we add some info why it is bad for us? (looks like already migrated?)

pkg/local_object_storage/metabase/version.go
```diff
@@ -105,3 +110,50 @@ func migrateFrom2Version(db *DB, tx *bbolt.Tx) error {
...
+			if len(k) != oid.Size {
+				return fmt.Errorf("wrong OID key len %d", len(k))
+			}
+			id := oid.ID(k)
+			var hdr object.Object
+			if err := hdr.Unmarshal(v); err != nil {
+				return fmt.Errorf("decode header of object %s from bucket value: %w", id, err)
+			}
+			par := hdr.Parent()
+			if err := putMetadataForObject(tx, hdr, par == nil, true); err != nil {
```
carpawell, line 140: i am always wondering what is `root && phy == true` and seems like this code _may_ accept it (and `putMetadataForObject` in general)

carpawell (approved):

pkg/local_object_storage/metabase/version.go
cthulhu-rider: u mean to error context? or just a comment in the code

pkg/local_object_storage/metabase/version.go
cthulhu-rider: mb `false`? `true` is a singular user object (small)

pkg/local_object_storage/metabase/version.go
carpawell: to the test message, yes. to show that we are only planning to have these keys but somehow bd already has it

pkg/local_object_storage/metabase/version.go
carpawell: hm, always had problems with fixing these terms' meanings. well if `root == true` is ok for small objects (and it seems we explain it [unequivocally](https://github.com/nspcc-dev/neofs-api/blob/00bfea07e771ee57f267d2a6387705f300057fea/object/types.proto#L111-L119)) then ok

pkg/local_object_storage/metabase/version.go
cthulhu-rider: added potential reasons

roman-khimov (approved):
It's correct. But what's gonna happen with 1-2M of objects is an open question to me. But we can try with any DB from metis, if it doesn't take a day to complete then ok.

PR #3126, opened 2025-02-12
Add additional RPC settings to IR config

### LINKED ISSUE #3121: Expose additional RPC settings
I'm always frustrated when I can't configure RPC MaxWebSocketClients and SessionPoolSize parameters. For private networks we can have bigger settings than defaults.

Add them to IR configuration.

## Additional context
7 IR/CN nodes can serve more than 422 nodes and session pool size currently severely limits the number of iterators available to SNs (which can do some things pretty much simultaneously).

Related to https://github.com/neo-project/neo/issues/3429, we can avoid sessions in many cases in fact, but configs are needed anyway.

End-rey:
Closes #3121.

### DIFF
--- a/pkg/innerring/internal/blockchain/blockchain.go
+++ b/pkg/innerring/internal/blockchain/blockchain.go
@@ -98,6 +98,16 @@ type RPCConfig struct {
 	//
 	// Optional.
 	TLSConfig
+
+	// The maximum simultaneous websocket client connection number.
+	//
+	// Optional: defaults to 64. Must not be larger than math.MaxInt32.
+	MaxWebSocketClients uint
+
+	// The maximum number of concurrent iterator sessions.
+	//
+	// Optional: defaults to 20. Must not be larger than math.MaxInt32.
+	SessionPoolSize uint
 }
 
 // TLSConfig configures additional RPC serving over TLS.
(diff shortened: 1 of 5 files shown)


pkg/innerring/internal/blockchain/blockchain.go
```diff
@@ -408,6 +418,8 @@ func New(cfg Config) (res *Blockchain, err error) {
 
 	cfgBaseApp.RPC.Enabled = true
 	cfgBaseApp.RPC.Addresses = cfg.RPC.Addresses
+	cfgBaseApp.RPC.MaxWebSocketClients = int(cfg.RPC.MaxWebSocketClients)
```
carpawell, line 421: so we do not check for its value and claim that it defaults to 64 only because we know neo-go does it?

carpawell (approved):

pkg/innerring/internal/blockchain/blockchain.go
End-rey: Yes. Validate value before?

End-rey: I did the same value check as above.

roman-khimov (approved):

PR #3134, opened 2025-02-14
Fix error from `CalculateNonceAndVUB`

End-rey:
The alphabetic key change event comes from the main chain with its hash, and when calculating vub, an error is returned because there is no hash in the fs chain. Ignore the error and calculate the value in another way.

### DIFF
--- a/pkg/innerring/state.go
+++ b/pkg/innerring/state.go
@@ -123,15 +123,18 @@ func (s *Server) voteForFSChainValidator(validators keys.PublicKeys, trigger *ut
 		nonce uint32 = 1
 		vub   uint32
 		vubP  *uint32
+		hash  util.Uint256
 	)
 
 	if trigger != nil {
-		nonce, vub, err = s.morphClient.CalculateNonceAndVUB(*trigger)
-		if err != nil {
-			return fmt.Errorf("could not calculate nonce and `validUntilBlock` values: %w", err)
-		}
-		vubP = &vub
+		hash = *trigger
 	}
+
+	nonce, vub, err = s.morphClient.CalculateNonceAndVUB(hash)
+	if err != nil {
+		return fmt.Errorf("could not calculate nonce and `validUntilBlock` values: %w", err)
+	}
+	vubP = &vub
 
 	s.contracts.alphabet.iterate(func(ind int, contract util.Uint160) {
 		_, err := s.morphClient.NotaryInvoke(contract, 0, nonce, vubP, voteMethod, epoch, validators)
--- a/pkg/morph/client/notary.go
+++ b/pkg/morph/client/notary.go
@@ -767,6 +767,13 @@ func (c *Client) CalculateNonceAndVUB(hash util.Uint256) (nonce uint32, vub uint
 
 	nonce = binary.LittleEndian.Uint32(hash.BytesLE())
 
+	if hash.Equals(util.Uint256{}) {
+		vub, err = c.notaryTxValidationLimit(conn)
+		if err != nil {
+			return 0, 0, fmt.Errorf("could not get vub validation limit: %w", err)
+		}
+		return nonce, vub, nil
+	}
 	height, err := c.getTransactionHeight(conn, hash)
 	if err != nil {
 		return 0, 0, fmt.Errorf("could not get transaction height: %w", err)
--- a/pkg/morph/client/notary.go
+++ b/pkg/morph/client/notary.go
@@ -268,7 +268,7 @@ func (c *Client) UpdateNotaryList(notaries keys.PublicKeys, txHash util.Uint256)
 
 	nonce, vub, err := c.CalculateNonceAndVUB(txHash)
 	if err != nil {
-		return fmt.Errorf("could not calculate nonce and `valicUntilBlock` values: %w", err)
+		return fmt.Errorf("could not calculate nonce and `validUntilBlock` values: %w", err)
 	}
 
 	return c.notaryInvokeAsCommittee(
@@ -292,7 +292,7 @@ func (c *Client) UpdateNeoFSAlphabetList(alphas keys.PublicKeys, txHash util.Uin
 
 	nonce, vub, err := c.CalculateNonceAndVUB(txHash)
 	if err != nil {
-		return fmt.Errorf("could not calculate nonce and `valicUntilBlock` values: %w", err)
+		return fmt.Errorf("could not calculate nonce and `validUntilBlock` values: %w", err)
 	}
 
 	return c.notaryInvokeAsCommittee(
--- a/pkg/morph/event/rolemanagement/designate.go
+++ b/pkg/morph/event/rolemanagement/designate.go
@@ -40,6 +40,6 @@ func ParseDesignate(e *state.ContainedNotificationEvent) (event.Event, error) {
 
 	return Designate{
 		Role:   noderoles.Role(bi.Int64()),
-		TxHash: e.Container,
+		TxHash: util.Uint256{},
 	}, nil
 }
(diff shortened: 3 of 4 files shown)


pkg/morph/client/notary.go
```diff
@@ -769,7 +769,12 @@ func (c *Client) CalculateNonceAndVUB(hash util.Uint256) (nonce uint32, vub uint
 
 	height, err := c.getTransactionHeight(conn, hash)
 	if err != nil {
-		return 0, 0, fmt.Errorf("could not get transaction height: %w", err)
+		c.logger.Warn("could not get transaction height", zap.Error(err))
+		vub, err = c.notaryTxValidationLimit(conn)
```
roman-khimov, line 773: Tricky one when we're close to the epoch boundary, but we don't have a lot of choices currently.

pkg/morph/client/notary.go
```diff
@@ -769,7 +769,12 @@ func (c *Client) CalculateNonceAndVUB(hash util.Uint256) (nonce uint32, vub uint
 
 	height, err := c.getTransactionHeight(conn, hash)
 	if err != nil {
-		return 0, 0, fmt.Errorf("could not get transaction height: %w", err)
+		c.logger.Warn("could not get transaction height", zap.Error(err))
```
roman-khimov, line 772: Not a `warn` in case of main chain interaction.

roman-khimov (commented):
Can we somehow do this _only_ for this special main chain case? The problem is that in the vast majority of cases the original error is more appropriate, there has to be a transaction. Maybe we can pass zero hash in this case and then fallback to this scheme without transaction height query?

pkg/morph/client/notary.go
carpawell: what do you mean? it seems like an epochless algorithm

carpawell (commented):
Agree that we need to separate this logic for events from main and FS chains. For FS chains it is a real problem if no TX found while for main it should not be even tried. TBH I even cannot say if we have any non-notary notifications from FS chain but it should not change anything, they may appear at any time, `morph` client was meant to be for every chain (at least it is now, without additional chages i would keep this).

pkg/morph/client/notary.go
roman-khimov: Yes, ok.

pkg/innerring/state.go
```diff
@@ -128,7 +128,11 @@ func (s *Server) voteForFSChainValidator(validators keys.PublicKeys, trigger *ut
 	if trigger != nil {
 		nonce, vub, err = s.morphClient.CalculateNonceAndVUB(*trigger)
 		if err != nil {
-			return fmt.Errorf("could not calculate nonce and `validUntilBlock` values: %w", err)
+			s.log.Debug("could not calculate nonce and `validUntilBlock` values", zap.Error(err))
+			nonce, vub, err = s.morphClient.CalculateNonceAndVUB(util.Uint256{})
```
cthulhu-rider, line 132: commit states
> ... an error is returned because there is no hash in the FSchain. Ignore the error and calculate the value in another way

but we switch to zero hash on any failure. Shouldnt we check the exact case?

pkg/innerring/state.go
```diff
@@ -128,7 +128,11 @@ func (s *Server) voteForFSChainValidator(validators keys.PublicKeys, trigger *ut
 	if trigger != nil {
 		nonce, vub, err = s.morphClient.CalculateNonceAndVUB(*trigger)
 		if err != nil {
-			return fmt.Errorf("could not calculate nonce and `validUntilBlock` values: %w", err)
+			s.log.Debug("could not calculate nonce and `validUntilBlock` values", zap.Error(err))
```
cthulhu-rider, line 131: the hash can also be useful

pkg/innerring/state.go
```diff
@@ -128,7 +128,11 @@ func (s *Server) voteForFSChainValidator(validators keys.PublicKeys, trigger *ut
...
+			nonce, vub, err = s.morphClient.CalculateNonceAndVUB(util.Uint256{})
+			if err != nil {
+				return fmt.Errorf("could not calculate nonce and `validUntilBlock` values for zero hash: %w", err)
```
cthulhu-rider, line 134: how about wrapping both errors?

pkg/innerring/state.go
roman-khimov: IIUC we should _always_ call `CalculateNonceAndVUB` with `util.Uint256{}` in `voteForFSChainValidator` if `trigger == nil`, because the only user of this method (not counting validator predefined in config) is governance processor reacting to main chain events.

pkg/morph/client/notary.go
```diff
@@ -767,6 +775,14 @@ func (c *Client) CalculateNonceAndVUB(hash util.Uint256) (nonce uint32, vub uint
 
 	nonce = binary.LittleEndian.Uint32(hash.BytesLE())
 
+	if hash.Equals(util.Uint256{}) {
+		c.logger.Debug("get zero hash to calculate nonce and vub")
```
roman-khimov, line 779: Not sure this is meaningful for debug.

pkg/morph/client/notary.go
```diff
@@ -268,7 +268,11 @@ func (c *Client) UpdateNotaryList(notaries keys.PublicKeys, txHash util.Uint256)
 
 	nonce, vub, err := c.CalculateNonceAndVUB(txHash)
 	if err != nil {
-		return fmt.Errorf("could not calculate nonce and `valicUntilBlock` values: %w", err)
+		c.logger.Debug("could not calculate nonce and `validUntilBlock` values", zap.Error(err))
+		nonce, vub, err = c.CalculateNonceAndVUB(util.Uint256{})
```
roman-khimov, line 272: Same here, it's not a fallback in case of error, it's either we're doing it this way or the other.

pkg/morph/client/notary.go
```diff
@@ -786,3 +796,24 @@ func (c *Client) getTransactionHeight(conn *connection, h util.Uint256) (uint32,
 	c.cache.txHeights.Add(h, height)
 	return height, nil
 }
+
+// CalculateNonceAndVUB calculates nonce and ValidUntilBlock values
+// based on transaction hash, and if it cannot get transaction height,
+// then it calculates alternatively rounded.
+func (c *Client) CalculateNonceAndVUBWithFallback(hash util.Uint256) (nonce uint32, vub uint32, err error) {
```
carpawell, line 803: doesnt old `CalculateNonceAndVUB` with `util.Uint256{}` do the trick? do we need an additional method? as i understand you took both of our suggestions (mine and @roman-khimov's) while a single of them is enough

pkg/morph/client/notary.go
End-rey: So these functions only include hashes from the main chain? Is that why we just have to always call with a zero hash?

pkg/morph/client/notary.go
End-rey: I just didn't really understand about whether hashes can come to these functions from outside the main chain. If this is the case, then indeed the function is not needed.

carpawell: > whether hashes can come to these functions from outside the main chain

imo, it is simpler to think it is possible. i would keep simple `if hash.Equals(util.Uint256{})` in this PR and pass empty hash when a caller understands that TX came from mainchain but should be handled in FSchain

End-rey:
Designation event of the mainnet RoleManagement contract contains a hash from the main chain, so I just made it zero, so that later it calculated `vub` as for the zero hash. Did I get it right?

pkg/morph/event/rolemanagement/designate.go
```diff
@@ -40,6 +40,6 @@ func ParseDesignate(e *state.ContainedNotificationEvent) (event.Event, error) {
 
 	return Designate{
 		Role:   noderoles.Role(bi.Int64()),
-		TxHash: e.Container,
+		TxHash: util.Uint256{},
```
roman-khimov, line 43: I'm not excited about this one because I'd expect this "real->fake" translation at some other layer dealing with the event. But it should work.

roman-khimov (approved):

PR #3148, opened 2025-02-17
object/searchv2: Fix ordering on multi-shard/node merge

cthulhu-rider:


pkg/core/object/metadata.go
```diff
@@ -2,35 +2,47 @@ package object
...
 	"github.com/nspcc-dev/neofs-sdk-go/client"
+	"github.com/nspcc-dev/neofs-sdk-go/object"
 )
 
+// IsIntegerSearchOp reports whether given op matches integer attributes.
+func IsIntegerSearchOp(op object.SearchMatchType) bool {
+	return op == object.MatchNumGT || op == object.MatchNumGE || op == object.MatchNumLT || op == object.MatchNumLE
+}
+
 // TODO: docs.
```
roman-khimov, line 19: Ouch. Still no docs.

pkg/core/object/metadata.go
```diff
@@ -2,35 +2,47 @@ package object
...
-func MergeSearchResults(lim uint16, withAttr bool, sets [][]client.SearchResultItem, mores []bool) ([]client.SearchResultItem, bool) {
+func MergeSearchResults(lim uint16, withAttr, cmpInt bool, sets [][]client.SearchResultItem, mores []bool) ([]client.SearchResultItem, bool, error) {
```
roman-khimov, line 20: Maybe it'd be easier to pass primary search filter iself here and derive withAttr from nil comparison and cmpInt from IsIntegerSearchOp. But this works too.

roman-khimov (approved):

pkg/core/object/metadata.go
cthulhu-rider: added

PR #3150, opened 2025-02-17
Transfer CLI to `SearchV2`

cthulhu-rider:


cmd/neofs-cli/modules/container/list_objects.go
```diff
@@ -44,40 +48,52 @@ var listContainerObjectsCmd = &cobra.Command{
...
+		var res []client.SearchResultItem
+		var cursor string
+		for {
+			res, cursor, err = cli.SearchObjects(ctx, id, *filters, nil, cursor, (*neofsecdsa.Signer)(pk), opts)
+			if err != nil {
+				return fmt.Errorf("rpc error: %w", err)
+			}
+			objectIDs = slices.Grow(objectIDs, len(objectIDs)+len(res))
+			for i := range res {
+				objectIDs = append(objectIDs, res[i].ID)
```
roman-khimov, line 91: Can be huge, we better process (output) them chunk by chunk. It'd also be more user-friendly, no big delay before the first result seen.

cthulhu-rider: thought about this. This is a behavior change (we could do this w/ old search too, but did not), but useful, so will add above

roman-khimov (approved):

PR #3158, opened 2025-02-19
node: add retries to update node state on new epoch

### LINKED ISSUE #3128: Retry netmap updates
I'm always frustrated when I realize that a simple
```
error	neofs-node/netmap.go:187	could not update node state on new epoch	{"epoch": 13, "error": "Internal error (-32603) - max session capacity reached"}
```
can lead to node never updating its map and never sending a heartbeat. Of course some errors can make it impossible to update (`rpcclient.ErrWSConnLost`), but at the same time there are temporary ones as well, this can go away easily after some 200ms.

Retry. Add some retries with https://pkg.go.dev/github.com/cenkalti/backoff, we also want to avoid spikes in load, so some random small delays are perfectly fine.

## Additional context
While we're adding more settings with #3121 and we can tune nodes to have higher limits any limit can be reached in various circumstances, so code should be prepared to deal with this situation rather than fail miserably.

End-rey:
Closes #3128.

Perhaps we can discuss the parameters of `backoff`.

### DIFF
--- a/cmd/neofs-node/netmap.go
+++ b/cmd/neofs-node/netmap.go
@@ -6,7 +6,10 @@ import (
 	"fmt"
 	"math/big"
 	"sync/atomic"
+	"time"
 
+	"github.com/cenkalti/backoff/v4"
+	"github.com/nspcc-dev/neo-go/pkg/rpcclient"
 	netmaprpc "github.com/nspcc-dev/neofs-contract/rpc/netmap"
 	"github.com/nspcc-dev/neofs-node/pkg/metrics"
 	"github.com/nspcc-dev/neofs-node/pkg/morph/event"
@@ -182,10 +185,31 @@ func initNetmapService(c *cfg) {
 	addNewEpochAsyncNotificationHandler(c, func(ev event.Event) {
 		e := ev.(netmapEvent.NewEpoch).EpochNumber()
 
-		ni, err := c.netmapLocalNodeState(e)
+		var (
+			ni      *netmapSDK.NodeInfo
+			err     error
+			retries uint64
+		)
+
+		expBackoff := backoff.NewExponentialBackOff()
+
+		err = backoff.RetryNotify(
+			func() error {
+				retries++
+				ni, err = c.netmapLocalNodeState(e)
+				if errors.Is(err, rpcclient.ErrWSConnLost) {
+					return backoff.Permanent(err)
+				}
+				return err
+			},
+			expBackoff,
+			func(err error, d time.Duration) {
+				c.log.Info("retrying due to error", zap.Uint64("epoch", e), zap.Error(err), zap.Duration("retry-after", d))
+			})
 		if err != nil {
-			c.log.Error("could not update node state on new epoch",
+			c.log.Error("could not update node state on new epoch after retries",
 				zap.Uint64("epoch", e),
+				zap.Uint64("retries", retries),
 				zap.Error(err),
 			)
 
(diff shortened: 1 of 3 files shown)


cmd/neofs-node/netmap.go
```diff
@@ -182,9 +184,26 @@ func initNetmapService(c *cfg) {
...
+		expBackoff.MaxElapsedTime = 5 * time.Second
+
+		err = backoff.RetryNotify(
+			func() error {
+				ni, err = c.netmapLocalNodeState(e)
+				return err
+			},
+			expBackoff,
+			func(err error, d time.Duration) {
+				c.log.Warn("retrying due to error", zap.Uint64("epoch", e), zap.Error(err), zap.Duration("retry-after", d))
```
roman-khimov, line 203: Info?

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/nspcc-dev/neofs-node
 go 1.22
 
 require (
+	github.com/cenkalti/backoff/v4 v4.2.1
```
roman-khimov, line 6: go.sum update?

cmd/neofs-node/netmap.go
```diff
@@ -182,9 +184,26 @@ func initNetmapService(c *cfg) {
...
 
-		ni, err := c.netmapLocalNodeState(e)
+		var (
+			ni  *netmapSDK.NodeInfo
+			err error
+		)
+
+		expBackoff := backoff.NewExponentialBackOff()
+		expBackoff.InitialInterval = 200 * time.Millisecond
+		expBackoff.MaxElapsedTime = 5 * time.Second
```
roman-khimov, line 194: We can try the default ones as well.

cmd/neofs-node/netmap.go
```diff
@@ -182,9 +184,26 @@ func initNetmapService(c *cfg) {
...
+
+		err = backoff.RetryNotify(
```
roman-khimov, line 196: Can we separate disconnect (when retry will never succeed) from other errors?

cmd/neofs-node/netmap.go
```diff
@@ -182,9 +184,26 @@ func initNetmapService(c *cfg) {
...
+				ni, err = c.netmapLocalNodeState(e)
+				return err
+			},
+			expBackoff,
+			func(err error, d time.Duration) {
+				c.log.Warn("retrying due to error", zap.Uint64("epoch", e), zap.Error(err), zap.Duration("retry-after", d))
+			})
 		if err != nil {
-			c.log.Error("could not update node state on new epoch",
+			c.log.Error("could not update node state on new epoch after retries",
```
cthulhu-rider, line 206: is retry guaranteed to happen? Cuz if not, this wont be always true

we can count and write num of attempts instead

go.mod
End-rey: Already there
https://github.com/nspcc-dev/neofs-node/blob/80d596f006580b8ae165c3a7d5a885c3e90816fb/go.sum#L13-L14

roman-khimov (approved):

carpawell (approved):

cmd/neofs-node/netmap.go
carpawell: why? i does not feel it info at all

roman-khimov: It's a pretty normal thing. Node is busy, ok, retrying. Now if we have no success after N retries then it's problem, but no earlier than that.

PR #3162, opened 2025-02-20
Store integers in two indexes of meta bucket for SearchV2

cthulhu-rider:


roman-khimov (approved):
Not sure you have green tests after each commit.

cthulhu-rider:
> Not sure you have green tests after each commit.

im sure they are not. When bug exists, i prefer to add corresponding unit test, which fails 1st, and then add a fix which also makes the test green

roman-khimov:
Which is not bisect-friendly. Tests should be green at every stage.

cthulhu-rider:
to me, tests should pass when no bug and fail when bug. Then they are useful

PR #3163, opened 2025-02-20
Morph client buffers

roman-khimov:


pkg/innerring/deploy.go
```diff
@@ -64,12 +64,20 @@ func (x *fsChain) SubscribeToNewBlocks() (<-chan *block.Block, error) {
...
 	if err != nil {
 		return nil, fmt.Errorf("listen to new blocks over Neo RPC WebSocket multi-endpoint: %w", err)
 	}
 
-	_, ch, _ := x.client.Notifications()
+	_, headerCh, _ := x.client.Notifications()
+	ch := make(chan *block.Block)
+	go func() {
+		// Client closure will stop it
+		for h := range headerCh {
```
cthulhu-rider, line 76: i'd `close(ch)` after the loop to notify the callers

pkg/morph/client/constructor.go
```diff
@@ -54,6 +54,18 @@ type cfg struct {
 
 const (
 	defaultDialTimeout = 5 * time.Second
+
+	// Mostly it's a single event per transaction, with 1s blocks we're
+	// not likely to have more than 5K transactions in a block now.
+	// The default is 512 tx.
```
cthulhu-rider, line 60:
```suggestion
	// The default is up to 512 tx per-block.
```
iiuc

pkg/morph/client/constructor.go
```diff
@@ -54,6 +54,18 @@ type cfg struct {
...
+	notifyChanBuf = 5000
```
cthulhu-rider, line 61: `notifyBufSize` or `notifyChanCap` to not think about "buffer of channels"

pkg/innerring/deploy.go
roman-khimov: Makes sense, done. Obviously this code is temporary (https://github.com/nspcc-dev/neofs-contract/pull/471).

pkg/morph/client/constructor.go
roman-khimov: s/buf/cap/ then

cthulhu-rider (approved):

pkg/innerring/deploy.go
```diff
@@ -64,12 +64,21 @@ func (x *fsChain) SubscribeToNewBlocks() (<-chan *block.Block, error) {
 		return ch, nil
 	}
 
-	err := x.client.ReceiveBlocks()
+	err := x.client.ReceiveHeaders()
```
carpawell, line 67: imo, useless renaming, a neofs dev will always think of it as a source of blocks, while "header" may mean many things

here and all the "block" -> "header" in the PR

carpawell (approved):

PR #3164, opened 2025-02-20
node/meta: Use neo-go's new notification API

### LINKED ISSUE #3142: Use getblocknotifications API for meta handling
It's not very efficient to subscribe to every container and we need a perfect block granularity wrt actions performed. So use the new getblocknotifications API introduced in 0.108.0 for container contract events. Don't filter for contracts initially, grab all of them, filter for the source and type only.

carpawell:
Receive object notifications by block and process them in sync, not independently. This is important for a consistent MPT state. Strictly speaking, it was a bug before but it was a bug that was impossible to fix without neo-go's new `getblocknotifications` API available with 0.108.0 release. Closes #3142.


pkg/services/meta/block.go
```diff
@@ -0,0 +1,94 @@
...
+	"context"
+	"fmt"
+
+	"github.com/nspcc-dev/neo-go/pkg/core/block"
+	"github.com/nspcc-dev/neo-go/pkg/core/mpt"
+	"github.com/nspcc-dev/neo-go/pkg/neorpc"
+	"go.uber.org/zap"
+)
+
+func (m *Meta) blockFetcher(ctx context.Context, buff <-chan *block.Header) {
```
cthulhu-rider, line 13: could u pls move these methods to a separate source file in a separate patch? this would easify the review

pkg/services/meta/meta.go
```diff
@@ -32,6 +33,17 @@ type ContainerLister interface {
 	List() (map[cid.ID]struct{}, error)
 }
 
+// wsClient is for test purposes only.
+type wsClient interface {
+	GetBlockNotifications(blockHash util.Uint256, filters ...*neorpc.NotificationFilter) (*result.BlockNotifications, error)
```
cthulhu-rider, line 38: suggest to write docs incl. what behavior `Meta` expects for blocks w/o notifications - error of both nil?

pkg/services/meta/block.go
```diff
@@ -0,0 +1,94 @@
...
+	l.Debug("handling block")
+
+	evName := objPutEvName
+	res, err := m.ws.GetBlockNotifications(h, &neorpc.NotificationFilter{
+		Name: &evName,
+	})
+	if err != nil {
+		return fmt.Errorf("fetching %s block: %w", h, err)
+	}
+
```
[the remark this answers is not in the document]
carpawell: oh right, fixed

pkg/services/meta/block.go
```diff
@@ -0,0 +1,94 @@
...
+	m.m.RLock()
+	defer m.m.RUnlock()
+
+	// TODO: is order preserved? what if at least one object could not be handled?
```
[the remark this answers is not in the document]
carpawell: and what about `what if at least one object could not be handled?` this is my general question in meta feature: in this PR, in the past ones, in the future ones. what if for some reason one event cannot be handled for some container? not because of a logical error but because of some really unexpected ones. any other future work looks senseless then

pkg/services/meta/block.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func (m *Meta) handleBlock(b *block.Header) error {
+	h := b.Hash()
+	ind := b.Index
+	l := m.l.With(zap.Stringer("block hash", h), zap.Uint32("index", ind))
+	l.Debug("handling block")
+
+	evName := objPutEvName
+	res, err := m.ws.GetBlockNotifications(h, &neorpc.NotificationFilter{
```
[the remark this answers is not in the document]
carpawell: i also wanted but what did you mean by "Don't filter for contracts initially, grab all of them" in the issue?

pkg/services/meta/meta.go
carpawell: what docs? this is literally simplification for tests. if there is either a test http server that can be mocked or we do not use tests at all, this interface does not exist. this is literally `rpcclient.WSClient`. docs will not save us if neo-go decides to change behavior. i do not plan to have any self written implementation here ever

pkg/services/meta/block.go
carpawell: done

carpawell: btw, i cost me additional 30+ mins for rebasing, conflicting, fixing, and rechecking additional threads below. this PR changes logic in general so old functions become separate things (it is by blocks now), but still PR is 200- lines, imo it was not worth it but it is already done

pkg/services/meta/meta.go
cthulhu-rider: ik why it's needed, idk how `GetBlockNotifications` interface behaves in mentioned case and wanna highlight this

carpawell:
Also fixed races.

pkg/services/meta/block.go
cthulhu-rider: ive no idea how u could face any conflicts with this change, but thx anyway

pkg/services/meta/block.go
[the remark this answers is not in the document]
roman-khimov: What does "could not be handled" mean? You're guaranteed to have some correct input data. Local failure to process is just a reason for reattempt or MPT resync in future.

pkg/services/meta/block.go
carpawell: because i was moving parts that should be changed to fix other threads. change it once in the original code, and change it second time in the moving commit. moving commit cannot be first because there was no "block" logic before and imo it was a good change as a single commit that changed everything

pkg/services/meta/block.go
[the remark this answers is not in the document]
carpawell: > What does "could not be handled" mean? ... Local failure...

Exact a local failure.

> reason for reattempt or MPT resync in future

yes, that is why TODO was left cause there is no logic for it now and i am not sure how it should look. mb i need to look at how neo-go does it

pkg/services/meta/meta.go
carpawell: will some reference to `rpcclient.WSClient` be enough in the comment? cause i do not understand you. nothing should implement it and it declares nothing for no one, it is literally `rpcclient.WSClient` but it is hard to mock it for tests. and i prefer this interface not to exist, but it would decrease coverage. i believe we will come to `httptest` package usage but not now

cthulhu-rider: yeah, u may state that this is a part of `WSClient` interface consumed by `Meta`

pkg/services/meta/block.go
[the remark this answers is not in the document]
roman-khimov: But you have just one "container" contract now. That was when we considered a contract per container.

roman-khimov (approved):

PR #3165, opened 2025-02-21
Fix panic in meta bucket GC

cthulhu-rider:


pkg/local_object_storage/metabase/metadata_test.go
```diff
@@ -1597,6 +1597,8 @@ func TestDB_SearchObjects(t *testing.T) {
 		for i := range objs {
 			appendAttribute(&objs[i], attrPlain, valPlain)
 			appendAttribute(&objs[i], attrInt, valInt)
+			appendAttribute(&objs[i], "IntOverflow", "115792089237316195423570985008687907853269984665640564039457584007913129639936")
+			appendAttribute(&objs[i], "IntOverflowNeg", "-115792089237316195423570985008687907853269984665640564039457584007913129639936")
```
roman-khimov, line 1601: "Underflow" is also known

cthulhu-rider: will know

roman-khimov (approved):

PR #3169, opened 2025-02-21
Delete objects from wc after flushing

### LINKED ISSUE #3077: Fix write cache flushing loop to drop objects
I'm always frustrated when objects may be stored in writecache forever. That's not the way it's supposed to work.

Drop files from write cache when they're moved to the lower layer storage.

## Additional context
See https://github.com/nspcc-dev/neofs-node/issues/2337#issuecomment-2489310539.

End-rey:
Closes #3077.

IIUC, the LRU cache was needed since the boltDB was processed in parallel and it was not safe to delete objects directly from there, but now only fstree and therefore it is no longer needed?

### DIFF
--- a/pkg/local_object_storage/writecache/storage.go
+++ b/pkg/local_object_storage/writecache/storage.go
@@ -1,35 +1,12 @@
 package writecache
 
 import (
-	"errors"
 	"fmt"
 	"os"
 
-	lru "github.com/hashicorp/golang-lru/v2"
-	"github.com/hashicorp/golang-lru/v2/simplelru"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor/fstree"
-	storagelog "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/internal/log"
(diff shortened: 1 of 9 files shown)


pkg/local_object_storage/writecache/storage.go
```diff
@@ -1,34 +1,18 @@
...
-	"github.com/hashicorp/golang-lru/v2/simplelru"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor/fstree"
-	storagelog "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/internal/log"
 	"github.com/nspcc-dev/neofs-node/pkg/util"
-	apistatus "github.com/nspcc-dev/neofs-sdk-go/client/status"
-	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
-	"go.uber.org/zap"
 )
 
 // store represents persistent storage with in-memory LRU cache
```
roman-khimov, line 11: Still mentions LRU.

roman-khimov:
CHANGELOG conflict.

roman-khimov (approved):

PR #3170, opened 2025-02-21
Fix incorrect sorting of some sys attributes on SearchV2 merge

cthulhu-rider:


pkg/core/object/metadata.go
```diff
@@ -22,7 +22,7 @@ func IsIntegerSearchOp(op object.SearchMatchType) bool {
 // compared numerically. Otherwise, lexicographically. Additional booleans show
 // whether corresponding sets can be continued. If the merged set can be
 // continued itself, true is returned.
-func MergeSearchResults(lim uint16, withAttr, cmpInt bool, sets [][]client.SearchResultItem, mores []bool) ([]client.SearchResultItem, bool, error) {
+func MergeSearchResults(lim uint16, firstAttr string, cmpInt bool, sets [][]client.SearchResultItem, mores []bool) ([]client.SearchResultItem, bool, error) {
```
roman-khimov, line 27: I've already suggested passing the first filter here and drop `cmpInt` altogether.

cthulhu-rider: we dont need its value, and forming args wouldnt be much simpler (if condition anyway). Prefer to keep primitive prms for now

roman-khimov (approved):

PR #3179, opened 2025-02-27
Parallelize WC flush process

### LINKED ISSUE #3100: Parallelize WC flush process
I'm always frustrated when we flush objects one by one from WC. It was done like "read many at once, push in parallel by many routines" for bbolt db in WC, and in a single thread for FSTree. There is no bbolt since https://github.com/nspcc-dev/neofs-node/pull/3091, it is time to define how we flush the only one FSTree left.

I am pretty sure this process can benefit from a worker pool for this task.

Keep it simple one by one.

## Additional context
https://github.com/nspcc-dev/neofs-node/pull/3091#pullrequestreview-2580795118

End-rey:
Closes #3100.

```
Flush took 2.501508862s for 100 objects with 1 workers
Flush took 702.815834ms for 100 objects with 4 workers
Flush took 202.958347ms for 100 objects with 16 workers
Flush took 24.22012974s for 1000 objects with 1 workers
Flush took 6.108409863s for 1000 objects with 4 workers
Flush took 1.621941699s for 1000 objects with 16 worker
```

### DIFF
--- a/pkg/local_object_storage/writecache/options.go
+++ b/pkg/local_object_storage/writecache/options.go
@@ -38,6 +38,8 @@ type options struct {
 	// maxCacheSize is the maximum total size of all objects saved in cache.
 	// 1 GiB by default.
 	maxCacheSize uint64
+	// workersCount is the number of workers flushing objects in parallel.
+	workersCount int
 	// objCounters contains object list along with sizes and overall size of cache.
 	objCounters counters
 	// noSync is true iff FSTree allows unsynchronized writes.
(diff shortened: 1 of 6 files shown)


roman-khimov (approved):

pkg/local_object_storage/writecache/options.go
```diff
@@ -83,6 +85,15 @@ func WithMaxObjectSize(sz uint64) Option {
 	}
 }
 
+// WithFlushWorkersCount sets number of workers to flushing objects in parallel.
+func WithFlushWorkersCount(c int) Option {
```
cthulhu-rider, line 89: unused outside the package, so no need to export

cthulhu-rider (approved):

pkg/local_object_storage/writecache/options.go
End-rey: @roman-khimov do we need a config option to configure this parameter as it was before?

roman-khimov: I thought about it, but I consider this solution as temporary, so adding an option to take it back in the next release doesn't seem to be very productive. I don't mind this function being exported though.

PR #3180, opened 2025-02-27
node/meta: do not keep block subscription without meta-containers

### LINKED ISSUE #3175: Don't subscribe to headers in meta service if there is no need to
I'm always frustrated when I realize meta service always listens for new headers.

If we have no containers to serve just don't subscribe. When they come --- subscribe and process events. Makes things a bit more efficient.

carpawell:
Subscribe for blocks only if there are containers with meta-on-chain enabled AND the current node is included in the placement. Closes #3175.

### DIFF
--- a/pkg/services/meta/blocks.go
+++ b/pkg/services/meta/blocks.go
@@ -32,8 +32,8 @@ func (m *Meta) handleBlock(b *block.Header) error {
 		return nil
 	}
 
-	m.m.RLock()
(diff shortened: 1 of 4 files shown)


carpawell:
Full run if needed: https://github.com/nspcc-dev/neofs-node/actions/runs/13567245267

roman-khimov (approved):

PR #3193, opened 2025-03-03
Add retries to process update peer notification

### LINKED ISSUE #3190: Node doesn't retry heartbeats
## Expected Behavior
Node sending heartbeat successfully.

## Current Behavior
```
error	netmap/process_peers.go:133	can't invoke netmap.UpdatePeer	{"error": "failed to submit notary request: The memory pool is full and no more transactions can be sent (-502) - no space left in the memory pool"}
```

## Possible Solution
Retry. Similar to #3128. -502 is a temporary error.

Add >300 nodes to the cluster.

## Your Environment
* Version used: 0.45.0

End-rey:
Closes #3190.


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ Changelog for NeoFS Node
 ## [Unreleased]
 
 ### Added
+- Retries to process update peer notification (#3193)
```
roman-khimov, line 7: More like "fixed"

pkg/innerring/processors/netmap/process_peers.go
```diff
@@ -126,10 +130,27 @@ func (np *Processor) processUpdatePeer(ev netmapEvent.UpdatePeer) {
 		}
 	}
 
-	nr := ev.NotaryRequest()
-	err = np.netmapClient.Morph().NotarySignAndInvokeTX(nr.MainTransaction, false)
+	var retries uint64
+	expBackoff := backoff.NewExponentialBackOff()
```
roman-khimov, line 134: Maybe we should add a layer below? Other invocations can be affected as well.

End-rey: Added retries for sending notary request in `NotarySignAndInvokeTX` func. Is that what you meant?

roman-khimov: Yeah. This raises some questions, not all errors are -512, some won't change even if retried, but in general it's worth trying and maybe we'll tune it afterwards.

roman-khimov (approved):

PR #3196, opened 2025-03-03
Use iterators for container listings

### LINKED ISSUE #3045: Use iterators for container listings

End-rey:
Closes #3045.

### DIFF
--- a/pkg/morph/client/container/list.go
+++ b/pkg/morph/client/container/list.go
@@ -15,26 +15,18 @@ import (
 // Returns the identifiers of all NeoFS containers if pointer
 // to user identifier is nil.
 func (c *Client) List(idUser *user.ID) ([]cid.ID, error) {
-	var rawID []byte
+	// neo-go's stack elements default limit
+	// is 2048, make it less a little
+	const prefetchNumber = 2000
 
+	var rawIdUser []byte
 	if idUser != nil {
-		rawID = idUser[:]
+		rawIdUser = idUser[:]
 	}
 
-	prm := client.TestInvokePrm{}
-	prm.SetMethod(listMethod)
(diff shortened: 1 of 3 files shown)


roman-khimov (approved):

pkg/morph/client/container/list.go
```diff
@@ -15,26 +15,18 @@ import (
 // Returns the identifiers of all NeoFS containers if pointer
 // to user identifier is nil.
 func (c *Client) List(idUser *user.ID) ([]cid.ID, error) {
-	var rawID []byte
+	// neo-go's stack elements default limit
+	// is 2048, make it less a little
+	const prefetchNumber = 2000
 
+	var rawIdUser []byte
```
cthulhu-rider, line 22: why u need this renaming?

End-rey: Oh, I did this in another function first to distinguish it from another variable below. Should I rename it back?

cthulhu-rider: let it be just to not rerun actions

cthulhu-rider (approved):

Issue #2918, opened 2024-08-14
Go 1.23 upgrade check list


roman-khimov:
See https://github.com/nspcc-dev/neo-go/issues/3553

roman-khimov:
https://tip.golang.org/wiki/RangefuncExperiment can be useful for node iterators of various kinds (of which we have a lot here). It'll be the standard way to handle similar problems soon.

PR #3199, opened 2025-03-04
go.mod: Upgrade `github.com/nspcc-dev/neofs-sdk-go`

cthulhu-rider:
* based on and blocked by https://github.com/nspcc-dev/neofs-sdk-go/pull/692

asserted the fix with the "big" object

to apply it, we need to upgrade SDK. It pulls Go 1.23 requirement. Current module hasn't bumped the min Go yet #2918. It's currently scheduled for [v0.46.0](https://github.com/nspcc-dev/neofs-node/milestone/70). We can consider inclusion in [v0.45.1](https://github.com/nspcc-dev/neofs-node/milestone/71). @roman-khimov @carpawell 

other ways without updating the SDK seem practically impossible to me at the moment. This tie is generally inconvenient, although the node keeps up with the SDK


roman-khimov:
We don't need Go 1.23 for 0.45.1.

cthulhu-rider:
I dont care about go much here, I want this fix to be released asap

roman-khimov (commented):
But otherwise OK. Even though I'm not excited about these changes (they better be for 0.46.0) we are using newer Go anyway already and we need to fix the problem. Workflows will fail though, so some adjustments are needed there.

roman-khimov:
Looks like we're not ready for this one. Let's roll out 0.45.1 with this known bug and upgrade Go and SDK properly then.

cmd/neofs-adm/internal/modules/fschain/dump_names.go
```diff
@@ -70,7 +71,7 @@ func dumpNames(cmd *cobra.Command, _ []string) error {
 			return len(iParts) < len(jParts)
 		}
 		for k := len(iParts) - 1; k >= 0; k-- {
-			var c = strings.Compare(iParts[k], jParts[k])
+			var c = cmp.Compare(iParts[k], jParts[k])
```
roman-khimov, line 74: With 1.23+ `strings.Compare` is the same as `cmp.Compare` performance-wise. This can be omitted, less packages to import.

roman-khimov (approved):
New iterators look good in many cases, simplifying flow, especially for early returns and things like that. But https://github.com/nspcc-dev/neofs-sdk-go/issues/693 will certainly need to be fixed, current code around `IterateNetworkEndpoints` is far from perfect.

cmd/neofs-adm/internal/modules/fschain/dump_names.go
cthulhu-rider: dropped this commit

PR #3200, opened 2025-03-04
Add retries for `Invoke` function

### LINKED ISSUE #3198: SN can't send heartbeat in a big network
## Expected Behavior
No error.

## Possible Solution
Obviously the pool can be enlarged and it'd completely solve the problem. But we can also retry, same as #3190 and #3128. And this can be at the `func (s StaticClient) Invoke(prm InvokePrm) error` level. But we need to distinguish temporary errors from permanent ones more correctly. See https://github.com/neo-project/proposals/blob/ff39b57eb67981b1ec7f0b244ec053aacdb7139a/nep-23.mediawiki

## Context
Building big things.

## Your Environment
* Version used: 0.45.0+

End-rey:
Closes #3198.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,8 @@ Changelog for NeoFS Node
 - Not all the components needed for write-cache migration were initialized (#3186)
 - Homomorphic hash indexes with empty values in meta bucket (#3187)
 - Failure of 1st split-chain child processing during metabase migration (#3187)
-- Retries to sending notary request (#3193)
 - Initial epoch tick by a just-started IR (#3007)
+- Missing retries when sending transaction/notary request leading to request failure in case of temporary errors (#3193, #3200)
 
 ### Changed
 - Use iterators for container listings (#3196)
--- a/pkg/morph/client/static.go
+++ b/pkg/morph/client/static.go
@@ -1,11 +1,16 @@
 package client
 
 import (
+	"errors"
 	"fmt"
+	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/nspcc-dev/neo-go/pkg/encoding/fixedn"
+	"github.com/nspcc-dev/neo-go/pkg/neorpc"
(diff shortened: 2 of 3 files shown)


pkg/morph/client/static.go
```diff
@@ -145,20 +155,51 @@ func (s StaticClient) Invoke(prm InvokePrm) error {
...
+				return backoff.Permanent(err)
+			}
+			return nil
+		},
+		expBackoff,
+		func(err error, d time.Duration) {
+			s.client.logger.Info("retrying due to error", zap.Error(err), zap.Duration("retry-after", d))
+		})
+	if err != nil {
+		s.client.logger.Error("can't Invoke after retries",
```
roman-khimov, line 197: I think we had no messages here before. Usually this is logged in the upper layer because it has more context (like what exactly we're trying to do).

pkg/morph/client/static.go
```diff
@@ -145,20 +155,51 @@ func (s StaticClient) Invoke(prm InvokePrm) error {
...
+				if errors.Is(err, neorpc.ErrMempoolCapReached) {
+					return err
+				}
+				return backoff.Permanent(err)
+			}
+			return nil
+		},
+		expBackoff,
+		func(err error, d time.Duration) {
+			s.client.logger.Info("retrying due to error", zap.Error(err), zap.Duration("retry-after", d))
```
roman-khimov, line 194: Debug to me.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Changelog for NeoFS Node
 - Failure of 1st split-chain child processing during metabase migration (#3187)
 - Retries to sending notary request (#3193)
 - Initial epoch tick by a just-started IR (#3007)
+- Retries to `Invoke` function (#3200)
```
roman-khimov, line 16: No one can understand what this means, to users it's more like "Missing retries when sending transaction/notary request leading to request failure in case of temporary errors". Which can be unified with the "notary request" two lines above.

roman-khimov (approved):

PR #3205, opened 2025-03-05
Fix failing notary retries

roman-khimov:


carpawell (commented):
1. Chlog.
2. As I remember, I was told, it is really important to get the same FB hash for notary service work. cc @AnnaShaleva

pkg/morph/event/notary_preparator.go
```diff
@@ -87,7 +87,7 @@ func notaryPreparator(localAcc util.Uint160, alphaKeys client.AlphabetKeys, bc B
 
 	dummyInvocationScript := append([]byte{byte(opcode.PUSHDATA1), 64}, make([]byte, 64)...)
 
-	const txCacheSize = 1000
+	const txCacheSize = 50000
```
AnnaShaleva, line 90: > but we'll be using bigger ones on some networks

It looks like an argument to move it to the node's configuration.

AnnaShaleva (approved):

carpawell (approved):

PR #3209, opened 2025-03-06
Interruptible and continuous migration of metabase to metabucket

cthulhu-rider:


pkg/local_object_storage/metabase/version.go
```diff
@@ -25,32 +26,39 @@ var versionKey = []byte("version")
...
+		return nil
+	}); err != nil {
+		return err
+	}
+	if migrateFromV3 {
+		if err := migrateFrom3Version(db.log, db.boltDB, db.cfg.containers); err != nil {
+			return fmt.Errorf("migrating from meta version 3 failed, consider database resync: %w", err)
 		}
 	}
-
```
roman-khimov, line 53: :eyes:

pkg/local_object_storage/metabase/version.go
```diff
@@ -25,32 +26,39 @@ var versionKey = []byte("version")
...
-		err := migrate(db, tx)
-		if err != nil {
-			return fmt.Errorf("migrating from meta version %d failed, consider database resync: %w", i, err)
+		// Outdated, but can be migrated.
+		if stored == 2 {
+			if err := migrateFrom2Version(db, tx); err != nil {
+				return fmt.Errorf("migrating from meta version 2 failed, consider database resync: %w", err)
+			}
+		}
+		migrateFromV3 = true
```
roman-khimov, line 53: You better expose stored/current and handle it outside, this code will be wrong wtih v5 immediately, previous one was more generic. Or at least checked the stored immediately, if it's 4 already this is not needed.

The scheme with maps is still applicable, btw. Just split the getter from the changer, `migrateFrom2Version` would just be wrapped in another tx.

pkg/local_object_storage/metabase/version.go
```diff
@@ -112,83 +115,118 @@ func migrateFrom2Version(db *DB, tx *bbolt.Tx) error {
...
-	pref := []byte{metadataPrefix}
-	if k, _ := c.Seek(pref); bytes.HasPrefix(k, pref) {
-		return fmt.Errorf("key with prefix 0x%X detected, metadata space is occupied by unexpected data or the version has not been updated to #%d", pref, currentMetaVersion)
-	}
-	if err := migrateContainersToMetaBucket(db.log, db.cfg.containers, tx); err != nil {
-		return err
+func migrateFrom3Version(l *zap.Logger, db *bbolt.DB, cs Containers) error {
+	var fromBkt, afterObj []byte
+	for {
+		if err := db.Batch(func(tx *bbolt.Tx) error {
```
roman-khimov, line 127: `db.Update`, please, you don't have any concurrency here.

pkg/local_object_storage/metabase/version.go
```diff
@@ -112,83 +115,118 @@ func migrateFrom2Version(db *DB, tx *bbolt.Tx) error {
...
 	}
-	return nil
+	var done uint
+	for ; k != nil; k, v = c.Next() {
+		ok, err := migrateObjectToMetaBucket(l, tx, cnr, k, v)
+		if err != nil {
+			return 0, nil, err
+		}
+		if ok {
+			if done++; done == limit {
```
roman-khimov, line 199: done++
if done == limit

Do not obfuscate things.

pkg/local_object_storage/metabase/version.go
cthulhu-rider: yeah, this was just for testing, forgot to revert

pkg/local_object_storage/metabase/version.go
cthulhu-rider: this is a valid Go code

pkg/local_object_storage/metabase/version.go
cthulhu-rider: > this code will be wrong wtih v5 immediately

V4+ is caught above

pkg/local_object_storage/metabase/version.go
roman-khimov: A lot of strings can be valid Go code, it doesn't mean they're all good. Mixing conditions and state changes is a classic problem, too easy to overlook when doing something in the same codebase a year after.

pkg/local_object_storage/metabase/version.go
cthulhu-rider: > You better expose stored/current and handle it outside

u mean read version in one tx, then update one-by-one in separate ones? or what

roman-khimov: Yes, exactly. If we can't have a single transaction in every case we can just change the migration function signature.

pkg/local_object_storage/metabase/version.go
cthulhu-rider: > Mixing conditions and state changes is a classic problem

Go authors do not recommend using it?

pkg/local_object_storage/metabase/version.go
cthulhu-rider: ok, will do. I just wanted to leave previous behavior untouched as much as possible

roman-khimov: Looks like we have less lines changed overall now which is good, a wrapper for old migrator is not a problem.

pkg/local_object_storage/metabase/version.go
roman-khimov: It's not related to Go, this is relevant for any structural language. And yes `err := smth(); err !=nil` is not the best thing too, it's just it's so common in Go that we're used to it (`err` locality can tiny bit helpful sometimes as well).

But this has the point: https://www.teamten.com/lawrence/programming/keep-if-clauses-side-effect-free.html

And these guys: https://web.git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/coding-style.rst?h=v6.14-rc5
know something too:

> Avoid tricky expressions.

roman-khimov (approved):

Issue #3194, opened 2025-03-03
Support Neo N3 account auth


cthulhu-rider:
introduced in https://github.com/nspcc-dev/neofs-api/issues/305

all system parts must serve authentication and authorization via new scheme

cthulhu-rider:
Neo Go code snippet demonstrating examplary authorization through new N3 scheme with contract-side key management (thx to @AnnaShaleva && @roman-khimov):
```go
func TestGenerateUserMgmtWitness(t *testing.T) {
	chain, _, httpSrv := initServerWithInMemoryChain(t)

	c, err := rpcclient.New(context.Background(), httpSrv.URL, rpcclient.Options{})
	require.NoError(t, err)
	t.Cleanup(c.Close)
	require.NoError(t, c.Init())

	// TODO: fill these variables:
	var (
		currHeight int64 = int64(chain.BlockHeight()) // may be current Epoch or whatever UserMgmt contract needs it to be.
		sig              = make([]byte, keys.SignatureLen)
		userName         = []byte("Bob")
		mgmtH      util.Uint160
		method     = "verifyUserSig"     // the name of UserMgmt's verification method.
		callFlags  = callflag.ReadStates // depends on the permissions UserMgmt contract needs in its verification method.
	)

	// This part of Invocation script should be filled by SN.
	snInv := io.NewBufBinWriter()
	emit.Int(snInv.BinWriter, currHeight)

	// This part of the Invocation script will be filled by witness issuer.
	inv := io.NewBufBinWriter()
	emit.Bytes(inv.BinWriter, sig)

	// Verification script will be filled by witness issuer as far.
	vrf := io.NewBufBinWriter()
	emit.Bytes(inv.BinWriter, userName)
	emit.Int(vrf.BinWriter, 3)                                  // 3 is the number of UserMgmt's verification method arguments. In this script it contains of: height, sig and userName.
	emit.Opcodes(vrf.BinWriter, opcode.PACK)                    // Pack arguments to Array.
	emit.AppCallNoArgs(vrf.BinWriter, mgmtH, method, callFlags) // Call UserMgmt's verification method (assuming that this method pushes bool on stack as a result).

	// Construct the verification script and execute it via RPC:
	s := append(snInv.Bytes(), append(inv.Bytes(), vrf.Bytes()...)...)
	res, err := c.InvokeScript(s, []transaction.Signer{
		// TODO: put your account as a signer if the UserManagement account uses CheckWitness.
	})
[...]

Issue #2795, opened 2024-04-01
Create and use helpers for crypto verifications


cthulhu-rider:
_Originally posted in https://github.com/nspcc-dev/neofs-node/pull/2674#discussion_r1542662972_

crypto helpers can be added to the SDK and pulled into the node code. Such refactoring is expected to simplify server implementation

PR #3216, opened 2025-03-10
Refactor and improve verification of tokens' signatures

cthulhu-rider:
preparation for #3194. If we agree, i'll extend the approach to all signature verification cases in the system. With this, N3 account support will simply be an additional `case` in the `switch`

### DIFF
--- a/internal/crypto/errors.go
+++ b/internal/crypto/errors.go
@@ -1,0 +1,14 @@
+package crypto
+
+import (
+	"errors"
+	"fmt"
+
+	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
+)
+
+var errIssuerMismatch = errors.New("issuer mismatches signature")
+
+func schemeError(s neofscrypto.Scheme, cause error) error {
+	return fmt.Errorf("scheme %v: %w", s, cause)
+}
--- a/pkg/core/object/fmt.go
+++ b/pkg/core/object/fmt.go
@@ -2,19 +2,16 @@ package object
 
 import (
 	"context"
-	"crypto/ecdsa"
-	"crypto/elliptic"
 	"errors"
 	"fmt"
 	"strings"
 
-	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
+	icrypto "github.com/nspcc-dev/neofs-node/internal/crypto"
 	"github.com/nspcc-dev/neofs-node/pkg/core/netmap"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"github.com/nspcc-dev/neofs-sdk-go/storagegroup"
-	"github.com/nspcc-dev/neofs-sdk-go/user"
 )
 
 // FormatValidator represents an object format validator.
@@ -202,15 +199,7 @@ func (v *FormatValidator) validateSignatureKey(obj *object.Object) error {
 		return errors.New("session token is not for object's signer")
 	}
 
-	if !token.VerifySignature() {
-		return errors.New("incorrect session token signature")
-	}
-	tokenSig, _ := token.Signature() // non-zero, otherwise VerifySignature would fail
-	pub, err := keys.NewPublicKeyFromBytes(tokenSig.PublicKeyBytes(), elliptic.P256())
-	if err != nil {
-		return errors.New("incorrect session token signature")
-	}
-	if signer := user.NewFromECDSAPublicKey(ecdsa.PublicKey(*pub)); signer != token.Issuer() {
+	if err := icrypto.AuthenticateToken(token); err != nil {
 		return errors.New("incorrect session token signature")
 	}
 
--- a/pkg/core/object/fmt.go
+++ b/pkg/core/object/fmt.go
@@ -200,7 +200,7 @@ func (v *FormatValidator) validateSignatureKey(obj *object.Object) error {
 	}
 
 	if err := icrypto.AuthenticateToken(token); err != nil {
-		return errors.New("incorrect session token signature")
+		return fmt.Errorf("authenticate session token: %w", err)
 	}
 
 	if issuer, owner := token.Issuer(), obj.Owner(); issuer != owner { // nil check was performed above
--- a/pkg/core/object/fmt.go
+++ b/pkg/core/object/fmt.go
@@ -2,15 +2,19 @@ package object
 
 import (
 	"context"
+	"crypto/ecdsa"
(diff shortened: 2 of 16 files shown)


pkg/innerring/processors/container/common.go
```diff
@@ -66,16 +67,12 @@ func (cp *Processor) verifySignature(v signatureVerificationData) error {
...
 			return errors.New("invalid session token signature")
 		}
 
-		var signerPub neofsecdsa.PublicKeyRFC6979
-		if err = signerPub.Decode(tok.IssuerPublicKeyBytes()); err != nil {
-			return fmt.Errorf("invalid issuer public key: %w", err)
-		}
-
-		if user.NewFromECDSAPublicKey(ecdsa.PublicKey(signerPub)) != v.ownerContainer {
+		if user.NewFromECDSAPublicKey(*pub) != v.ownerContainer {
```
roman-khimov, line 75: It seems to be incompatible with verification scripts, although works at the current stage.

internal/crypto/verify.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func verifyECDSAWalletConnectSignature(pubBin, sig []byte, signedDataFn signedDataFunc) (*ecdsa.PublicKey, bool) {
+	return verifyECDSASignature(pubBin, sig, signedDataFn, func(pub *ecdsa.PublicKey) neofscrypto.PublicKey {
+		return (*neofsecdsa.PublicKeyWalletConnect)(pub)
+	})
+}
+
+// VerifyTokenSignature verifies t signature.
+func VerifyTokenSignature[T interface {
```
roman-khimov, line 46: https://github.com/nspcc-dev/neofs-node/issues/2795?

roman-khimov (approved):

pkg/innerring/processors/container/common.go
cthulhu-rider: I'll take another look, maybe I'll be able to move this check inside the function switching over schemes and unify it for all places

internal/crypto/verify.go
cthulhu-rider: yes, it relates, I'll note in the commit. For now this API is some kind of experimental, so I'm not ready to place this code in the SDK. And for now it's easier to work within current module only

pkg/innerring/processors/container/common.go
```diff
@@ -66,16 +67,12 @@ func (cp *Processor) verifySignature(v signatureVerificationData) error {
...
 			return errors.New("invalid session token signature")
 		}
 
-		var signerPub neofsecdsa.PublicKeyRFC6979
-		if err = signerPub.Decode(tok.IssuerPublicKeyBytes()); err != nil {
-			return fmt.Errorf("invalid issuer public key: %w", err)
-		}
-
-		if user.NewFromECDSAPublicKey(ecdsa.PublicKey(signerPub)) != v.ownerContainer {
+		if tokSigner != v.ownerContainer {
```
roman-khimov, line 75: Technically, this should be a token issuer, not signer. And then token verification routine needs to check that issuer is the same as signer. But it works at this stage.

pkg/services/container/server.go
```diff
@@ -98,9 +98,13 @@ func (s *server) getVerifiedSessionToken(req interface {
 		return nil, fmt.Errorf("decode: %w", err)
 	}
 
-	if !token.VerifySignature() {
+	signer, err := icrypto.VerifyTokenSignature(&token)
+	if err != nil {
 		return nil, errors.New("invalid signature")
 	}
+	if signer != token.Issuer() {
```
roman-khimov, line 105: Like here, this check can be embedded into `VerifyTokenSignature()`.

pkg/core/object/fmt.go
```diff
@@ -199,9 +199,13 @@ func (v *FormatValidator) validateSignatureKey(obj *object.Object) error {
 		return errors.New("session token is not for object's signer")
 	}
 
-	if _, err := icrypto.VerifyTokenSignature(token); err != nil {
+	tokenSigner, err := icrypto.VerifyTokenSignature(token)
+	if err != nil {
 		return fmt.Errorf("verify session token signature: %w", err)
 	}
+	if tokenSigner != token.Issuer() {
```
roman-khimov, line 206: Better just inline this check, it's always needed for both types of tokens.

pkg/services/object/acl/acl.go
```diff
@@ -221,41 +224,35 @@ func (c *Checker) CheckEACL(msg any, reqInfo v2.RequestInfo) error {
...
 		return errBearerExpired
 	}
 
 	// 2. Then check if bearer token is signed correctly.
-	_, err := icrypto.VerifyTokenSignature(token)
+	signer, err := icrypto.VerifyTokenSignature(&token)
 	if err != nil {
 		return fmt.Errorf("verify bearer token signature: %w", err)
 	}
+	if issuer := token.Issuer(); !issuer.IsZero() && signer != issuer {
```
carpawell, line 238: if `issuer.IsZero`, we continue? could not find info if missing issuer is ok there: https://github.com/nspcc-dev/neofs-api/commit/f49089fc59d0af2e2765c869e4438a44a47e29b6

roman-khimov: At this stage non-zero issuer can be enforced, I think. Bearer tokens are not saved anywhere long-term and I'd expect all of the old ones to expire already.

pkg/innerring/processors/container/common.go
cthulhu-rider: done

pkg/services/container/server.go
cthulhu-rider: done

pkg/core/object/fmt.go
cthulhu-rider: done

internal/crypto/errors.go
```diff
@@ -0,0 +1,11 @@
+package crypto
+
+import (
+	"fmt"
+
+	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
+)
+
+func schemeError(s neofscrypto.Scheme, cause error) error {
+	return fmt.Errorf("scheme %v: %w", s, cause)
```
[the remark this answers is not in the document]
cthulhu-rider: `%v` and `%s` are the same for `fmt.Stringer`s

pkg/services/object/acl/acl.go
cthulhu-rider: initially was done for compatibility with tokens issued before https://github.com/nspcc-dev/neofs-api/commit/f49089fc59d0af2e2765c869e4438a44a47e29b6. Changed to the required issuer field

roman-khimov (approved):

PR #3221, opened 2025-03-11
writecache: retry flush after delay if an error is received

### LINKED ISSUE #3099: `ignoreErrors` in WC, dump shard routine, etc
I'm always frustrated when we are not sure if it is OK to ignore every error or not: https://github.com/nspcc-dev/neofs-node/pull/3091#discussion_r1930884103.

Ignore only "read" errors in processes that are about making as many saving things as possible. It makes no sense to continue flushing or dumping objects if the receiving side (disk) is not ready to take any more objects (no space left, global disk errors).

In fact, it may be already implemented but it needs to be rechecked.

The worst thing about skipping "write" errors is wasting time on useless operations. At least, it looks like. Maybe it is not that bad to really _ignore every error_ if it was asked so.

## Additional context
https://github.com/nspcc-dev/neofs-node/pull/3091

End-rey:
Closes #3099.

Do I have to do this for every error, or just like now, if we can't write an object to the main storage?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Changelog for NeoFS Node
 - "new transaction callback finished with error" logs for "already exists in mempool" error (#3218)
 
 ### Changed
+- Retry flush from write-cache after 10s delay if an error is received (#3221)
 
 ### Removed
 
--- a/pkg/local_object_storage/writecache/flush.go
+++ b/pkg/local_object_storage/writecache/flush.go
@@ -18,6 +18,8 @@ import (
 const (
 	// defaultFlushInterval is default time interval between successive flushes.
 	defaultFlushInterval = 10 * time.Second
+	// defaultErrorDelay is default time delay for retrying a flush after receiving an error.
+	defaultErrorDelay = 10 * time.Second
 	// defaultWorkerCount is a default number of workers that flush objects.
 	defaultWorkerCount = 20
 )
@@ -39,6 +41,15 @@ func (c *cache) flushScheduler() {
 
 	for {
 		select {
+		case <-c.flushErrCh:
+			c.log.Warn("flush scheduler paused due to error",
+				zap.Duration("delay", defaultErrorDelay))
+
+			time.Sleep(defaultErrorDelay)
+
+			for len(c.flushErrCh) > 0 {
+				<-c.flushErrCh
+			}
 		case <-ticker.C:
 			c.modeMtx.RLock()
 			if c.readOnly() {
@@ -48,6 +59,12 @@ func (c *cache) flushScheduler() {
 			c.modeMtx.RUnlock()
 			err := c.fsTree.IterateAddresses(func(addr oid.Address) error {
 				select {
+				case <-c.flushErrCh:
+					select {
+					case c.flushErrCh <- struct{}{}:
+					default:
+					}
+					return errors.New("stopping iteration due to error")
 				case c.flushCh <- addr:
 					return nil
 				case <-c.closeCh:
@@ -76,10 +93,19 @@ func (c *cache) flushWorker(id int) {
 				c.modeMtx.RUnlock()
 				continue
 			}
-			if err := c.flushSingle(addr, true); err != nil {
-				c.log.Warn("flush failed", zap.Int("worker", id), zap.Error(err))
-			}
+			err := c.flushSingle(addr, true)
 			c.modeMtx.RUnlock()
+
+			if err != nil {
+				select {
+				case c.flushErrCh <- struct{}{}:
+				default:
+				}
+				c.log.Error("worker can't flush an object due to error",
+					zap.Int("worker", id),
+					zap.Stringer("addr", addr),
+					zap.Error(err))
+			}
 		case <-c.closeCh:
 			return
 		}
--- a/pkg/local_object_storage/writecache/writecache.go
+++ b/pkg/local_object_storage/writecache/writecache.go
@@ -59,6 +59,9 @@ type cache struct {
 	mode    mode.Mode
 	modeMtx sync.RWMutex
 
+	// flushErrCh is a channel for error handling while flushing.
+	flushErrCh chan struct{}
+
 	// compressFlags maps address of a big object to boolean value indicating
 	// whether object should be compressed.
 	compressFlags map[string]struct{}
@@ -94,8 +97,9 @@ var (
 // New creates new writecache instance.
 func New(opts ...Option) Cache {
 	c := &cache{
-		flushCh: make(chan oid.Address),
-		mode:    mode.ReadWrite,
+		flushCh:    make(chan oid.Address),
+		flushErrCh: make(chan struct{}, 1),
+		mode:       mode.ReadWrite,
 
 		compressFlags: make(map[string]struct{}),
 		options: options{
--- a/pkg/local_object_storage/writecache/flush_test.go
+++ b/pkg/local_object_storage/writecache/flush_test.go
@@ -1,15 +1,19 @@
 package writecache
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"path/filepath"
+	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
 
 	objectCore "github.com/nspcc-dev/neofs-node/pkg/core/object"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor"
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor/common"
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor/fstree"
 	meta "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/metabase"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/shard/mode"
 	checksumtest "github.com/nspcc-dev/neofs-sdk-go/checksum/test"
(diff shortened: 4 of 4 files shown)


carpawell:
Tests say we have a race now in WC.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Changelog for NeoFS Node
 - "new transaction callback finished with error" logs for "already exists in mempool" error (#3218)
 
 ### Changed
+- Retry flush from write-cache after delay if an error is received (#3221)
```
cthulhu-rider, line 13: i'd clarify
```suggestion
- Retry flush from write-cache after 10s delay if an error is received (#3221)
```
since it's known and static

pkg/local_object_storage/writecache/flush.go
```diff
@@ -39,6 +41,19 @@ func (c *cache) flushScheduler() {
 
 	for {
 		select {
+		case err := <-c.flushErrCh:
+			c.retryMtx.Lock()
+			c.lastError = err
+			c.log.Warn("flush scheduler paused due to error",
```
cthulhu-rider, line 47: u dont use concurrent data for this, no need to keep the lock

pkg/local_object_storage/writecache/flush.go
```diff
@@ -47,6 +62,13 @@ func (c *cache) flushScheduler() {
 			}
 			c.modeMtx.RUnlock()
 			err := c.fsTree.IterateAddresses(func(addr oid.Address) error {
+				c.retryMtx.Lock()
+				if c.lastError != nil {
+					c.retryMtx.Unlock()
+					return fmt.Errorf("stopping iteration due error: %w", c.lastError)
```
cthulhu-rider, line 68: concurrent data is read w/o a lock

pkg/local_object_storage/writecache/flush.go
```diff
@@ -47,6 +62,13 @@ func (c *cache) flushScheduler() {
 			}
 			c.modeMtx.RUnlock()
 			err := c.fsTree.IterateAddresses(func(addr oid.Address) error {
+				c.retryMtx.Lock()
```
cthulhu-rider, line 65: `RLock()` is enough

pkg/local_object_storage/writecache/flush.go
```diff
@@ -76,10 +98,23 @@ func (c *cache) flushWorker(id int) {
...
-				c.log.Warn("flush failed", zap.Int("worker", id), zap.Error(err))
-			}
+			err := c.flushSingle(addr, true)
 			c.modeMtx.RUnlock()
+
+			if err != nil {
+				c.retryMtx.Lock()
+				if c.lastError == nil {
+					c.lastError = err
+					c.log.Error("flush worker stopped due to error",
```
cthulhu-rider, line 108: again no concurrency

pkg/local_object_storage/writecache/flush.go
```diff
@@ -76,10 +98,23 @@ func (c *cache) flushWorker(id int) {
...
+						zap.Int("worker", id),
+						zap.Error(err))
+					c.retryMtx.Unlock()
+					c.flushErrCh <- err
```
cthulhu-rider, line 112: i think we have an excessive mixture of error and notification about it here. If scheduler already sleeps due to ealier error, this will block for up to 10s. After waking up the scheduler will probably catch this error and go back to sleep. Although this error could have occurred almost simultaneously with the previous one

i'd try to simplify. When some worker encounters an error, it notifies the scheduler about this (if not yet, meaning no other worker has done this earlier). When scheduler receives the notfication, it just makes `time.Sleep()`, frees notification channel and keeps goin. While the scheduler sleeps, the workers dont get work, so they sleep too

what do u think?

pkg/local_object_storage/writecache/writecache.go
```diff
@@ -59,6 +59,11 @@ type cache struct {
 	mode    mode.Mode
 	modeMtx sync.RWMutex
 
+	// error handling while flushing.
+	flushErrCh chan error
+	retryMtx   sync.RWMutex
+	lastError  error
```
cthulhu-rider, line 65:
```suggestion
	lastFlushError  error
```
(if we'll decide to keep it)

pkg/local_object_storage/writecache/flush.go
End-rey: > When some worker encounters an error, it notifies the scheduler about this (if not yet, meaning no other worker has done this earlier). When scheduler receives the notfication, it just makes time.Sleep(), frees notification channel and keeps goin.

That's how it's supposed to be now, isn't it? I anticipated that when checking `c.lastFlushError == nil`, cases of the same errors in a row would be removed, and in theory, while the scheduler sleeps, the workers also sleep.

cthulhu-rider: we need to bubble one single signal from any worker to scheduler to trigger its sleep. I dont say ur proposal works incorrectly, my point is that it implements needed behavior to complicated. Mb im wrong

End-rey: I understood about the notification and made it easier. However, I could not remove the memorization of the error, since it is needed to interrupt iteration by addresses in scheduler. I've tried different ways, and I think it's the best option right now.

cthulhu-rider: > since it is needed to interrupt iteration by addresses in scheduler

did u try simple channel probe?
```
Iterate(...) error {
  select {
  case <-c.flushSleepCh: // chan struct{}
 	return return errors.New("stopping iteration due error")
  case <-c.closeCh:
	return errors.New("closed during iteration")
  case c.flushCh <- addr:
	return nil
  }
}

```

End-rey: I tried this option and it was not easy for me to implement it, especially in the place where the signal was sent from the workers.
The best thing I've come up with in `flushWorkers`:
```go
if err != nil {
  select {
  case <-c.flushSleepCh:
  default:
    close(c.flushSleepCh)
    select {
    case c.flushErrCh <- struct{}{}:
    default:
    }
    c.log.Error("flush worker stopped due to error",
      zap.Int("worker", id),
      zap.Error(err))
  }
}
```
However, with a large number of workers, sometimes occurs `panic: close of closed channel`.

cthulhu-rider: ofc u must never try to close channel concurrently, and u dont need to close `flushSleepCh`. Instead workers do:
```
if err != nil {
  select {
  case c.flushSleepCh <- struct{}{}:
  default: // nothing, other worker sent the signal
  }
}
```
and scheduler only reads the `flushSleepCh`

End-rey:
@cthulhu-rider check, please, is that what you meant? It really looks better.

cthulhu-rider:
> @cthulhu-rider check, please, is that what you meant? It really looks better.

one channel is enough for you. In the current version it is `flushErrCh`

End-rey:
> one channel is enough for you. In the current version it is `flushErrCh`

So, is it ok if `IterateAddresses` continues to work when the error appears? Because of this, the workers will still accept addresses.

cthulhu-rider:
> > one channel is enough for you. In the current version it is `flushErrCh`
> 
> 
> 
> So, is it ok if `IterateAddresses` continues to work when the error appears? Because of this, the workers will still accept addresses.
> 
> 

instead, breaking iteration is rly needed. And u can do it with the same channel

cthulhu-rider (approved):

pkg/local_object_storage/writecache/flush.go
```diff
@@ -39,6 +41,15 @@ func (c *cache) flushScheduler() {
 
 	for {
 		select {
+		case <-c.flushErrCh:
+			c.log.Warn("flush scheduler paused due to error",
+				zap.Duration("delay", defaultErrorDelay))
+
+			time.Sleep(defaultErrorDelay)
+
+			for len(c.flushErrCh) > 0 {
```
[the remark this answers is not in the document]
End-rey: The point of this cycle is to drain the channel from errors that could still have entered the channel before sleep.

pkg/local_object_storage/writecache/flush.go
```diff
@@ -76,10 +93,18 @@ func (c *cache) flushWorker(id int) {
...
 			}
-			if err := c.flushSingle(addr, true); err != nil {
-				c.log.Warn("flush failed", zap.Int("worker", id), zap.Error(err))
-			}
+			err := c.flushSingle(addr, true)
 			c.modeMtx.RUnlock()
+
+			if err != nil {
+				select {
+				case c.flushErrCh <- struct{}{}:
```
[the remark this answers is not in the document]
roman-khimov: You have multiple flusher routines (`flushWorker`), _all_ of them will fail to flush some object when out of space. They _all_ will try pushing an event into `flushErrCh`. One will succeed, others will stuck, then the `flushScheduler` routine will extract one event, another routine will put its event into buffer, others will still be stuck. Then it will drain the channel (a single event that is present in the queue), but some other channel will just unblock and push its event that will be picked up again by `flushScheduler`. This will eventually be resolved, but recovery can take more time than `defaultErrorDelay`.

pkg/local_object_storage/writecache/flush.go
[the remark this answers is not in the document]
End-rey: Yes, and therefore this 1 error, if there is one, needs to be cleaned up.

pkg/local_object_storage/writecache/flush.go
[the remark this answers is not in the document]
roman-khimov: Correct! OK.

pkg/local_object_storage/writecache/flush.go
```diff
@@ -76,10 +93,18 @@ func (c *cache) flushWorker(id int) {
...
+				default:
+				}
+				c.log.Error("flush worker stopped due to error",
```
roman-khimov, line 104: But why "stopped" here if it doesn't really stop?

End-rey: `worker can't flush an object due to error` ok?

roman-khimov: Yep. And you can add an address as well, just in case.

roman-khimov (approved):

PR #3226, opened 2025-03-12
Fix redundant epoch ticks by IR

cthulhu-rider:


cthulhu-rider:
macos tests PASS now. Action fails on publishing (#3228)

pkg/innerring/innerring.go
```diff
@@ -72,8 +72,8 @@ type (
 		// event producers
 		morphListener   event.Listener
 		mainnetListener event.Listener
-		blockTimers     []*timer.BlockTimer
 		epochTimer      *timer.BlockTimer
+		initEpochTimer  atomic.Pointer[timer.BlockTimer]
```
roman-khimov, line 76: Is it still `init`? Looks like it's the only one left.

cthulhu-rider: didnt get what u mean. This is the `initialEpochTicker` revealed, previously it was in `blockTimers` slice

roman-khimov: OK, looks like I've not noticed the other one.

roman-khimov (approved):

PR #3231, opened 2025-03-15
Feat/meta/search

carpawell:


carpawell:
https://rest.fs.neo.org/HXSaMJXk2g8C14ht8HSi7BBaiYZ1HeWh2xnWPGQCg4H6/2768-1742054444/index.html

carpawell:
Suggest this PR. Some functions are repeated now and there could be a better sharing but I went through the code many-many times trying to duplicate things and think that this is the most stable way for search routine to keep working. After additional external tests enabling and some feedback from performance tests, I will deduplicate some more things in a smaller PR(s).

pkg/core/object/metadata.go
```diff
@@ -197,3 +200,641 @@ func VerifyHeaderForMetadata(hdr object.Object) error {
...
+	errWrongPrimaryAttribute = errors.New("wrong primary attribute")
+	errWrongKeyValDelim      = errors.New("wrong key-value delimiter")
+
+	// AttributeDelimiter is attribute key and value separator used in metadata DB.
+	AttributeDelimiter = []byte{0x00}
+	metaOIDPrefix      = []byte{metaPrefixID}
+
+	maxUint256 = new(big.Int).SetBytes([]byte{255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255,
+		255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255})
+	maxUint256Neg = new(big.Int).Neg(maxUint256)
```
roman-khimov, line 225: maxNeg sounds a lot like `min` to me.

pkg/core/object/metadata.go
```diff
@@ -197,3 +200,641 @@ func VerifyHeaderForMetadata(hdr object.Object) error {
...
+type ParsedIntFilter struct {
+	// AutoMatch means every existing value if acceptable for filters.
+	AutoMatch bool
+	// IntValue is parsed integer value from attribute.
+	IntValue *big.Int
+	// RawValue is raw attribute value in the original form.
+	RawValue []byte
+}
+
+// AttributeGetter provides access to attributes based on data base's indexes.
```
roman-khimov, line 247: data base?

pkg/core/object/metadata.go
```diff
@@ -197,3 +200,641 @@ func VerifyHeaderForMetadata(hdr object.Object) error {
...
+// the provided arguments. resHolder must not be nil. additionalCheck may or may
+// not be nil, it will be called once on every object that matches all the previous
+// (filters-based) checks. fInt must have preparsed (from string attribute form)
+// integer values from the fs.
+// Returned handler must be called on a sorted list of key-value pairs without
+// any pair omission and/or duplication. Pairs must be common NeoFS indexes for
+// object header's fields. See [meta] (storage engine's package) for details.
+func MetaDataKVHandler(resHolder *SearchResult, attrGetter AttributeGetter, additionalCheck AdditionalObjectChecker, fs object.SearchFilters, fInt map[int]ParsedIntFilter, attrs []string, cursor *SearchCursor, count uint16) func(k, v []byte) bool {
+	if wantCap := int(count); cap(resHolder.Objects) != wantCap {
+		resHolder.Objects = slices.Grow(resHolder.Objects, wantCap)
```
roman-khimov, line 278: Luckily it's just 32K.

pkg/core/object/metadata.go
```diff
@@ -197,3 +200,641 @@ func VerifyHeaderForMetadata(hdr object.Object) error {
...
+// MetaDataKVHandler returns a handler that filters out and writes search results
+// for Search NeoFS operation (see [client.SearchObjects] for details) based on
+// the provided arguments. resHolder must not be nil. additionalCheck may or may
+// not be nil, it will be called once on every object that matches all the previous
+// (filters-based) checks. fInt must have preparsed (from string attribute form)
+// integer values from the fs.
+// Returned handler must be called on a sorted list of key-value pairs without
+// any pair omission and/or duplication. Pairs must be common NeoFS indexes for
+// object header's fields. See [meta] (storage engine's package) for details.
+func MetaDataKVHandler(resHolder *SearchResult, attrGetter AttributeGetter, additionalCheck AdditionalObjectChecker, fs object.SearchFilters, fInt map[int]ParsedIntFilter, attrs []string, cursor *SearchCursor, count uint16) func(k, v []byte) bool {
```
roman-khimov, line 276: Scary as hell.

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -394,180 +209,41 @@ func (db *DB) search(cnr cid.ID, fs object.SearchFilters, fInt map[int]ParsedInt
...
-	var n uint16
-	var more bool
-	var id, dbVal, primDBVal []byte
 	var keyBuf keyBuffer
-	var wasPrimMatch bool
 	attrSkr := &metaAttributeSeeker{keyBuf: &keyBuf, bkt: metaBkt}
 	curEpoch := db.epochState.CurrentEpoch()
-	dbValInt := new(big.Int)
-nextPrimKey:
-	for ; bytes.HasPrefix(primKey, cursor.primKeysPrefix); primKey, _ = primCursor.Next() {
```
roman-khimov, line 426: What if this is an `iter.Seq2` (or `iter.Seq) providing these values? Can this lead to a more humane `MetaDataKVHandler`? Without result placeholders and functions returned?

pkg/core/object/metadata.go
```diff
@@ -197,3 +200,641 @@ func VerifyHeaderForMetadata(hdr object.Object) error {
...
+	case object.MatchNumLE:
+		return dbVal.Cmp(fltVal) <= 0
+	}
+}
+
+// PreprocessSearchQuery accepts verified search filters, requested attributes
+// along and continuation cursor, verifies the cursor and returns additional
+// arguments to pass into [DB.Search]. If the query is valid but unreachable,
+// [ErrUnreachableQuery] is returned.
+func PreprocessSearchQuery(fs object.SearchFilters, attrs []string, cursor string) (*SearchCursor, map[int]ParsedIntFilter, error) {
```
roman-khimov, line 627: This `map[int]ParsedIntFilter` is also not very helpful, it can be like `object.SearchFilters -> InternalSearchFilters` with this data embedded into appropriate fields.

pkg/core/object/metadata.go
carpawell: https://github.com/nspcc-dev/neofs-node/blob/092ff01ee240397eda24d76105bb9c89760742f6/pkg/local_object_storage/metabase/metadata.go#L44

pkg/core/object/metadata.go
carpawell: https://github.com/nspcc-dev/neofs-node/blob/092ff01ee240397eda24d76105bb9c89760742f6/pkg/local_object_storage/metabase/metadata.go#L415

pkg/core/object/metadata.go
carpawell: what exactly?

pkg/local_object_storage/metabase/metadata.go
carpawell: it can help a little, i really tried to share this KVs handling logic as simply as possible and thought about `iter` pkg, but nothing really simple for now from my side. i can implement your suggestion, let me repeat your suggestion: `MetaDataKVHandler` (naming will differ) may accept some `iter.Seq` that provides KVs from different DBs implementations, right?
key problem for me is trying to write a doc for such func (IMO doc problem eloquently explains our current problem in this situation and why i did not do this): "iterate all KVs (_somehow_), put an iterator to `MetaDataKVHandler` from a different package and it will return results in your package". currently, it just returns a criterion for KV handler that could be called on every KV pair and it will include filtered KVs to a result; even though its signature scares, it is simple as it is

i don't mind, just want to hear your thoughts

pkg/core/object/metadata.go
carpawell: https://github.com/nspcc-dev/neofs-node/blob/092ff01ee240397eda24d76105bb9c89760742f6/pkg/local_object_storage/metabase/metadata.go#L217

carpawell:
@roman-khimov, I did many tests, many runs, many fixes, and many sharings but in fact, only `pkg/services/meta` package has smth functionally new (and its usage in the `main` ofc), so I answered with links to some of your threads that point to the current master, I haven't even changed naming. Also look at the main `Search` method that is the same for metaservice and metabase.
If I thought that it would be helpful, I would have done it in this PR, but I still think that for a review and a faster first test run, it will be better to accept this PR in this way and then improve code sharing/naming. So up to you.

pkg/core/object/metadata.go
roman-khimov: SearchV2, first of all. We're risking a lot here, I can't accept it to 0.45.2. The interface itself is quite complex.

pkg/core/object/metadata.go
roman-khimov: I know, I know, but you're touching it and it's a part of `MetaDataKVHandler` complexity problem as well.

pkg/core/object/metadata.go
carpawell: > The interface itself is quite complex.

will `iter.Seq` do the trick for you? (thread below)

roman-khimov: Not yet known. `iter.Seq` will have some other problems (DB transaction management).

roman-khimov:
Conflicts.

pkg/core/object/metadata.go
```diff
@@ -197,3 +200,641 @@ func VerifyHeaderForMetadata(hdr object.Object) error {
...
+	var lastMatchedPrimKey []byte
+	var n uint16
+	var more bool
+	var id, dbVal, primDBVal []byte
+	var wasPrimMatch bool
+	dbValInt := new(big.Int)
+
+	return func(k, v []byte) bool {
+		defer func() {
+			if resHolder.Err == nil && more {
```
cthulhu-rider, line 293: isnt `resHolder.Err == nil` always when `more`?

pkg/services/meta/meta.go
```diff
@@ -25,7 +25,7 @@ const (
 
 	// rootKey is the key for the last known state root in KV data base
 	// associated with MPT.
-	rootKey = 0x00
+	rootKey = 0xff
```
cthulhu-rider, line 28: back compatible change?

cmd/neofs-node/object.go
```diff
@@ -621,14 +624,27 @@ func (x *fsChainForObjects) IsOwnPublicKey(pubKey []byte) bool {
 func (x *fsChainForObjects) LocalNodeUnderMaintenance() bool { return x.isMaintenance.Load() }
 
 type storageForObjectService struct {
-	local  *engine.StorageEngine
-	putSvc *putsvc.Service
-	keys   *util.KeyStorage
+	cnrs        containercore.Source
+	localEngine *engine.StorageEngine
```
cthulhu-rider, line 628: `local` was clear enough

pkg/core/object/metadata.go
```diff
@@ -568,7 +568,9 @@ func matchValues(dbVal []byte, matcher object.SearchMatchType, fltVal []byte) bo
 	}
 }
 
-func restoreIntAttribute(b []byte) (string, error) {
+// RestoreIntAttribute restores from raw signed uint256 format its string
+// representation.
+func RestoreIntAttribute(b []byte) (string, error) {
```
cthulhu-rider, line 573: exporting should never be done for tests imo

pkg/services/meta/search.go
```diff
@@ -0,0 +1,128 @@
...
+
+type attrGetter struct {
+	keyBuff []byte
+
+	db storage.Store
+}
+
+func (a *attrGetter) Get(oID []byte, attributeKey string) (attributeValue []byte, err error) {
+	defer func() {
+		a.keyBuff = a.keyBuff[:0]
```
cthulhu-rider, line 103: redundant since it is unknown whether the `keyBuf` will still be used

pkg/services/meta/search.go
```diff
@@ -0,0 +1,128 @@
...
+	rng.Start = a.keyBuff
+	a.db.Seek(rng, func(k, v []byte) bool {
+		if !bytes.HasPrefix(k, a.keyBuff) {
+			return false
+		}
+		if len(k[len(a.keyBuff):]) == 0 {
+			err = fmt.Errorf("invalid meta bucket key (prefix 0x%X): missing attribute value", k[0])
+			return false
+		}
+		attributeValue = k[len(a.keyBuff):]
```
cthulhu-rider, line 122: shouldnt be a clone to avoid mutation?

pkg/services/object/server.go
```diff
@@ -2071,7 +2070,7 @@ func (s *server) processSearchRequest(ctx context.Context, req *protoobject.Sear
 			return nil, fmt.Errorf("merge results from container nodes: %w", err)
 		}
 		if more {
-			if newCursor, err = meta.CalculateCursor(fs, res[len(res)-1]); err != nil {
+			if newCursor, err = objectcore.CalculateCursor(fs, res[len(res)-1]); err != nil {
```
cthulhu-rider, line 2073: shouldnt be triggered by the commit changing tests

cthulhu-rider (commented):
right direction overall

go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/multiformats/go-multiaddr v0.12.2
 	github.com/nspcc-dev/hrw/v2 v2.0.3
 	github.com/nspcc-dev/locode-db v0.6.0
-	github.com/nspcc-dev/neo-go v0.108.1
+	github.com/nspcc-dev/neo-go v0.108.2-0.20250227114656-68d7e8e01cd8
```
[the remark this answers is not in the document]
roman-khimov: Not yet, there is no release with the appropriate fix. Not a big deal either in this case.

pkg/services/meta/meta.go
roman-khimov: Shouldn't be a problem, it's an experimental feature, not available in production anywhere.

pkg/core/object/metadata.go
```diff
@@ -197,3 +200,641 @@ func VerifyHeaderForMetadata(hdr object.Object) error {
...
+	// ErrInvalidCursor is returned when [SearchCursor] has invalid seek
+	// values.
+	ErrInvalidCursor = errors.New("invalid cursor")
+
+	errInvalidPrimaryFilter  = errors.New("invalid primary filter")
+	errWrongPrimaryAttribute = errors.New("wrong primary attribute")
+	errWrongKeyValDelim      = errors.New("wrong key-value delimiter")
+
+	// AttributeDelimiter is attribute key and value separator used in metadata DB.
+	AttributeDelimiter = []byte{0x00}
```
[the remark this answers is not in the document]
carpawell: tbh, i would move it to another package with a better naming, this PR moves it closer to `MergeSearchResults` and `VerifyHeaderForMetadata`. renamed

pkg/core/object/metadata.go
```diff
@@ -197,3 +200,641 @@ func VerifyHeaderForMetadata(hdr object.Object) error {
...
+	// Get must return attribute value for a pair of object ID and attribute
+	// value. (nil, nil) must be returned if value is missing. err should mean
+	// only data base errors.
+	Get(oID []byte, attributeKey string) (attributeValue []byte, err error)
+}
+
+// AdditionalObjectChecker is an additional check that may be provided to
+// [MetaDataKVHandler]. `false` return value stops object from being recorded
+// to [SearchResult], `true` keeps searching without additional filtering.
+type AdditionalObjectChecker func(oid.ID) (match bool)
```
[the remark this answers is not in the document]
carpawell: i do not feel like alliases are needed to be used. all I heard or read is that they are mainly for temporary fixes, here I wanted to highlight that they are smth different (not every search routing requires them)

> it's not used as a separate type

can you explain? when are types used as types?

pkg/core/object/metadata.go
```diff
@@ -197,3 +200,641 @@ func VerifyHeaderForMetadata(hdr object.Object) error {
...
+
+// SearchResult is a [MetaDataKVHandler] operation's results holder.
+type SearchResult struct {
+	Objects             []client.SearchResultItem
+	UpdatedSearchCursor []byte
+	Err                 error
```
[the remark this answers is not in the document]
carpawell: because it is created, filled and handled outside of `MetaDataKVHandler`? it is a closure and there is nothing we can do about it

pkg/core/object/metadata.go
```diff
@@ -197,3 +200,641 @@ func VerifyHeaderForMetadata(hdr object.Object) error {
...
+}
+
+// MetaDataKVHandler returns a handler that filters out and writes search results
+// for Search NeoFS operation (see [client.SearchObjects] for details) based on
```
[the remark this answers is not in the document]
carpawell: > ref to nowhere

fixed

> I'd just state why it exists for developers and dont burden with a detailed docs

since it is already done, I would keep it if you do not see any mistakes in this doc

pkg/core/object/metadata.go
```diff
@@ -197,3 +200,641 @@ func VerifyHeaderForMetadata(hdr object.Object) error {
...
+// the provided arguments. resHolder must not be nil. additionalCheck may or may
+// not be nil, it will be called once on every object that matches all the previous
+// (filters-based) checks. fInt must have preparsed (from string attribute form)
+// integer values from the fs.
+// Returned handler must be called on a sorted list of key-value pairs without
+// any pair omission and/or duplication. Pairs must be common NeoFS indexes for
+// object header's fields. See [meta] (storage engine's package) for details.
+func MetaDataKVHandler(resHolder *SearchResult, attrGetter AttributeGetter, additionalCheck AdditionalObjectChecker, fs object.SearchFilters, fInt map[int]ParsedIntFilter, attrs []string, cursor *SearchCursor, count uint16) func(k, v []byte) bool {
+	if wantCap := int(count); cap(resHolder.Objects) != wantCap {
```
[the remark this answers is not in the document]
carpawell: you mean `cap` is always enough? it is an external thing to this package, with two completely different places for usage, I would keep it

dropped

cmd/neofs-node/object.go
carpawell: there was `meta` and it is always local, so IMO it was not

another change cancelled this change anyway

pkg/core/object/metadata.go
carpawell: this code is used 3 times, do you mean making 3 different copy of it?

pkg/services/object/server.go
carpawell: why? commit message has some words about it. without moving it, we have cyclic deps, it must be moved. all I can suggest is moving in a separate commit (why? commit is for sharing only, it changes nothing, and `CalculateCursor` is shared in it)

pkg/core/object/metadata.go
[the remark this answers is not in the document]
cthulhu-rider: > when are types used as types?

for example, when they expose interface. Aliases cant do that. To me, this type is just to shorten the code, no more, so i'd make it an alias

pkg/services/meta/search.go
cthulhu-rider: i mean iiuc following code doesnt need this instruction to work properly. If so, it's no-op completely

pkg/services/object/server.go
cthulhu-rider: ok lets have it. It confuses very much when code is changed for test purposes in most of the cases

roman-khimov (approved):

PR #3232, opened 2025-03-16
chore: make function comment match function name

growfrow:
make function comment match function name


roman-khimov (approved):

growfrow:
Thanks for your review.

Issue #3229, opened 2025-03-13
Node fails to process SEARCH requests when the other node is down


roman-khimov:
## Expected Behavior
An attempt to make a connection, timeout, returning whatever we can, no reconnection attempts for some time.

## Current Behavior
Looks like requests get stuck.

## Possible Solution
Not yet known.

## Steps to Reproduce (for bugs)
4 SNs in VM in a single container, shut one VM down, make some SEARCHv2 (via S3 for example).

## Regression
Maybe.

## Your Environment
* Version used: 0.45.1

roman-khimov:
https://pkg.go.dev/google.golang.org/grpc#DialContext
> DialContext calls NewClient and then exits idle mode.

https://pkg.go.dev/google.golang.org/grpc#NewClient
> No I/O is performed.

https://github.com/nspcc-dev/neofs-sdk-go/blob/441b37a4bc03673f3786360787ead1dfc61ebfe8/client/client.go#L178-L194

Notice `parentCtx` being used for `EndpointInfo` probe. This makes us violate the dial timeout when the remote node doesn't answer in any way.

roman-khimov:
```
func main() {
        // Create client instance
        var prm client.PrmInit
        c, err := client.New(prm)
        if err != nil {
                fmt.Println("New:", err)
                os.Exit(1)
        }

        start := time.Now()

        // Connect to the NeoFS server
        var prmDial client.PrmDial
        prmDial.SetServerURI("grpc://192.168.1.22:8080") // endpoint address
        prmDial.SetTimeout(5 * time.Second)
        prmDial.SetStreamTimeout(15 * time.Second)

        err = c.Dial(prmDial)
        fmt.Println(err, time.Since(start))
}
```
I can get
```
gRPC dial: context deadline exceeded 5.004549752s
```
with SDK RC13 for black hole addresses (that exist on the network (ARP/ping), but do not respond to connection attempts in any way), so https://github.com/nspcc-dev/neofs-sdk-go/pull/696 not fixing the problem is not that surprising (even though it's still a nice patch to me).

cthulhu-rider:
if remote server shuts down, SN detects this only through active ops, e.g. on next SearchV2 request. At the same time, node fully relies to the connection context in terms of interruption. With this, if client does not set its deadline, the 1st request will hang. If the client ends up doing Ctrl-C, the server will catch `context cancelled` status and the connection will not be marked as dead

[...]

PR #3204, opened 2025-03-05
node/config: use list instead of maps for options

End-rey:
Closes #2993.

PR #3236, opened 2025-03-18
Stop supporting SN `ExternalAddr` attribute

cthulhu-rider:


roman-khimov:
It's all good, but the intention was to have 0.45.2 with connection fixes while this one breaks configuration compatibility and requires 0.46.0 (which raises the question of #3204 as well). Is this required for #3229?

cthulhu-rider:
this is not required for #3229

i think conn fixes deserve to be postponed. Too deep rework, and we aint dealing with regression

roman-khimov:
Conflicts.

PR #3241, opened 2025-03-20
Reuse generic meta index for some metabase tasks

roman-khimov:
This drops some useless code and make metabase reuse a bit more of generic meta index. In general it's more efficient, other code can be changed in a similar fashion.


pkg/local_object_storage/metabase/iterators.go
```diff
@@ -56,61 +58,100 @@ func (db *DB) IterateExpired(epoch uint64, h ExpiredObjectHandler) error {
...
+		return math.MaxUint64, oid.ID{}
+	}
+
+	return binary.BigEndian.Uint64(f256[32-8:]), id
+}
+
+func (db *DB) iterateExpired(tx *bbolt.Tx, curEpoch uint64, h ExpiredObjectHandler) error {
+	var (
+		expStart  = make([]byte, 1+len(object.AttributeExpirationEpoch)+1+1)
+		typPrefix = make([]byte, 1+objectKeySize+len(object.FilterType)+1)
```
carpawell, line 86:
```suggestion
		typPrefix = make([]byte, 1+oid.Size+len(object.FilterType)+1)
```

also one of the `1` also can be reused as `len` expression or as a const

pkg/local_object_storage/metabase/iterators.go
```diff
@@ -56,61 +58,100 @@ func (db *DB) IterateExpired(epoch uint64, h ExpiredObjectHandler) error {
 	})
 }
 
-func (db *DB) iterateExpired(tx *bbolt.Tx, epoch uint64, h ExpiredObjectHandler) error {
+func keyToEpochOID(k []byte, expStart []byte) (uint64, oid.ID) {
+	if !bytes.HasPrefix(k, expStart) || len(k) < len(expStart)+32+objectKeySize {
+		return math.MaxUint64, oid.ID{}
+	}
+	var (
+		f256 = k[len(expStart) : len(expStart)+32]
```
carpawell, line 66: "f"?

carpawell (approved):

pkg/local_object_storage/metabase/iterators.go
roman-khimov: fixed, can be renamed

carpawell: > fixed

make sense but initially it broke me

PR #3246, opened 2025-03-24
Fix attribute collecting in SearchV2

cthulhu-rider:


pkg/local_object_storage/metabase/metadata.go
```diff
@@ -534,8 +534,8 @@ nextPrimKey:
 		// collect attributes
 		collected := make([]string, len(attrs))
 		if len(attrs) > 0 {
+			var err error
 			if intPrimMatcher {
-				var err error
```
roman-khimov, line 538: Looks like belonging to the second commit, not the first one.

roman-khimov (approved):

pkg/local_object_storage/metabase/metadata.go
cthulhu-rider: yeah fixed

Issue #2994, opened 2024-10-31
Make validation more simple and reuse config in components


End-rey:
## Is your feature request related to a problem? Please describe.

I'm always frustrated when additional proprietary logic is used for validation, when existing solutions can be used. It is also unpleasant when the config structure is used only for validation, although it can be used in various components.

## Describe the solution you'd like

Make validation through [viper.UnmarshalExact](https://pkg.go.dev/github.com/spf13/viper#UnmarshalExact) and reuse the config struct in the components.

## Describe alternatives you've considered

Keeping as is, but with a lot of code and duplicate structures.

## Additional context

Refs #2981.
First, you need to change the config so that you can perform validation simply: #2993.

Issue #736, opened 2021-08-02
Smart contract script hash auto negotiation


alexvanin:
NeoFS Adm utility deployes side chain network with NNS contract that stores data about all side chain smart contract script hashes. Inner Ring app and Storage app can use data from NNS contract. However we should not change the configuration. If smart contracts are present in local config, then use values from there.

fyrchik:
The code to retrieve hashes from NNS can be found in `cmd/neofs-adm`. Currently they are stored as hex-encoded little-endian string (this is the form we use in every CLI). It can be more convenient to store them as raw bytes if we use them mostly for automatic processing vs user-level display.

alexvanin:
What about main chain contract addresses? They are not presented in NNS contract. I would like to keep them in Inner Ring node config.

realloc:
Mainnet contract hash is well-known for N3 mainnet/testnet and may be kept as a config option. But we should keep in mind that it still needs update for devenv.

PR #3247, opened 2025-03-24
ir: reuse config struct in components

End-rey:
Refs #2994.


roman-khimov (commented):
I'm surprised you've got more added lines overall. This suggests we're doing something wrong here. But I need more time to check what exactly is the problem.

pkg/util/config/config.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	cfg.Normalize()
+
+	return nil
+}
+
+// BindEnvForStruct goes through all fields of the structure,
+// searches for the corresponding env variables by tags, and calls viper.Set.
+func BindEnvForStruct(v *viper.Viper, s any, envPrefix string) {
+	bindFields(v, reflect.ValueOf(s), "", envPrefix)
```
cthulhu-rider, line 55: reflection is most likely not the best way. Viper handles ENVs under the hood, so what exact problem ur solving here?

pkg/util/config/config.go
```diff
@@ -0,0 +1,182 @@
...
+func PublicKeyHook() mapstructure.DecodeHookFunc {
+	return func(
+		f reflect.Type,
+		t reflect.Type,
+		data any,
+	) (any, error) {
+		if t == reflect.TypeOf(keys.PublicKey{}) {
+			str, ok := data.(string)
+			if !ok {
+				return nil, fmt.Errorf("invalid public key %v", data)
```
cthulhu-rider, line 119: when type is wrong, it doesnt matter what value it is
```suggestion
				return nil, fmt.Errorf("public key field type %T instead of %T", data, str)
```

pkg/util/config/config.go
```diff
@@ -0,0 +1,182 @@
...
+// that converts int and int64 to uint32, ensuring they are within the valid range.
+func Uint32StrictHook() mapstructure.DecodeHookFunc {
+	return func(
+		f reflect.Type,
+		t reflect.Type,
+		data any,
+	) (any, error) {
+		if t.Kind() == reflect.Uint32 {
+			switch v := data.(type) {
+			case int, int64:
```
cthulhu-rider, line 142: can `data` be any of these `case`s it is `reflect.Uint32`?

pkg/util/config/config.go
```diff
@@ -0,0 +1,182 @@
...
+		}
+
+		if fieldVal.Kind() == reflect.Struct || (fieldVal.Kind() == reflect.Ptr && fieldVal.Elem().Kind() == reflect.Struct) {
+			bindFields(v, fieldVal, viperKey, envPrefix)
+		}
+	}
+}
+
+// PublicKeyHook returns a mapstructure decode hook func that converts a string to a keys.PublicKey.
+func PublicKeyHook() mapstructure.DecodeHookFunc {
```
cthulhu-rider, line 110:
```suggestion
func PublicKeyHook() mapstructure.DecodeHookFuncType {
```

pkg/util/config/config.go
```diff
@@ -0,0 +1,182 @@
...
+	return func(
+		f reflect.Type,
```
cthulhu-rider, line 112: `f` is never used, wouldnt `mapstructure.DecodeHookFuncValue

cmd/neofs-ir/internal/validate/config.go
```diff
@@ -1,284 +0,0 @@
-package validate
-
-import "time"
-
-type validConfig struct {
```
cthulhu-rider, line 5: isn't it enough to just export it? Then move to another package in a separate change

pkg/util/config/config.go
```diff
@@ -0,0 +1,182 @@
...
+			switch v := data.(type) {
+			case int, int64:
+				intVal := reflect.ValueOf(v).Int()
+				if intVal < 0 || intVal > math.MaxUint32 {
+					return nil, fmt.Errorf("%d out of allowable range [%d:%d]", intVal, 0, math.MaxUint32)
+				}
+				return uint32(intVal), nil
+
+			case float32, float64:
+				return nil, fmt.Errorf("unexpected format (%.2f) for uint32", v)
```
cthulhu-rider, line 150: https://github.com/nspcc-dev/neofs-node/pull/3247/files#r2013577820

pkg/util/config/config.go
```diff
@@ -0,0 +1,182 @@
...
+func Uint160Hook() mapstructure.DecodeHookFunc {
+	return func(
+		f reflect.Type,
+		t reflect.Type,
+		data any,
+	) (any, error) {
+		if t == reflect.TypeOf(util.Uint160{}) {
+			str, ok := data.(string)
+			if !ok {
+				return nil, fmt.Errorf("invalid hash %v", data)
```
cthulhu-rider, line 167: https://github.com/nspcc-dev/neofs-node/pull/3247/files#r2013577820

pkg/util/config/config.go
```diff
@@ -0,0 +1,182 @@
...
+			}
+
+			if str == "" {
+				return util.Uint160{}, nil // if address is not set, then NNS resolver should be used
```
cthulhu-rider, line 171: these hashes aint always related to `NNS`, drop this comment

pkg/util/config/config.go
```diff
@@ -0,0 +1,182 @@
...
+			}
+
+			addr, err := util.Uint160DecodeStringLE(str)
+			if err != nil {
+				return nil, fmt.Errorf("can't parse contract address %s: %w", str, err)
```
cthulhu-rider, line 176: do not include `str` to the error, user knows what he set

pkg/util/config/config.go
```diff
@@ -0,0 +1,182 @@
...
+	"reflect"
+	"strings"
+
+	"github.com/mitchellh/mapstructure"
+	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
+	"github.com/nspcc-dev/neo-go/pkg/util"
+	"github.com/spf13/viper"
+)
+
+type config interface {
```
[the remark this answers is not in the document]
End-rey: It will also be used for the implementation of the node config.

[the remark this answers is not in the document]
cthulhu-rider: those changes can bring this unification. Without them, i see this as an overhead

pkg/util/config/config.go
End-rey: https://github.com/spf13/viper/blob/4ddd7c091cc0153bda143ddb90e29b2160869ce9/viper.go#L442

`viper`, when accessing the key, directly accesses the ENV. With `UnmarshalExact`, ENV variables are not involved, so you need to search for them yourself and assign them.

cmd/neofs-ir/internal/validate/config.go
End-rey: Inside the previous config, there were anonymous structures that need to be accessed in tests, and some in the code, so I broke them into separate substructures.

cthulhu-rider: > and some in the code

acceptable then

pkg/innerring/config.go
```diff
@@ -33,258 +24,189 @@ const (
...
+		return fmt.Errorf("unsupported storage type '%s': '%s'", storageTypeKey, typ)
 	}
 
 	var committeeKey = cfgPathFSChainLocalConsensus + ".committee"
-	c.Committee, err = parseConfigPublicKeys(v, committeeKey, "committee members")
-	if err != nil {
-		return c, err
-	} else if len(c.Committee) == 0 {
-		return c, fmt.Errorf("empty committee members '%s'", committeeKey)
+	if !cfg.IsSet(committeeKey) {
```
roman-khimov, line 77: I think this duplicates the check below, if it's not set the length is going to be zero anyway.

pkg/innerring/config/config.go
```diff
@@ -0,0 +1,162 @@
...
+
+type Experimental struct {
+	ChainMetaData bool `mapstructure:"chain_meta_data"`
+}
+
+// HandleDeprecated normalizes values if outdated names were used.
+// Use it after setting all the keys.
+func (c *Config) HandleDeprecated() {
+	const (
+		deprecatedMorphPrefix = "morph"
```
roman-khimov, line 124: Wow, we can drop this now. It's 0.46.0, the migration should've happened earlier.

pkg/innerring/config.go
```diff
@@ -33,258 +24,189 @@ const (
...
-				return fmt.Errorf("parse unsigned integer: %w", err)
+		for k, v := range cfgFSChainLocalConsensus.ValidatorsHistory.Height {
+			if k%committeeSize != 0 {
+				return fmt.Errorf("height %d is not a multiple of the %q size", k, committeeKey)
 			}
-
-			if height%committeeSize != 0 {
-				return fmt.Errorf("height %d is not a multiple of the %q size", height, committeeKey)
+			if v <= 0 {
+				return fmt.Errorf("value %d is out of allowable range", v)
```
roman-khimov, line 112: "is not a positive number" and k is even more helpful here for the user (as well as pointing to `ValidatorsHistory`).

internal/configutil/unmarshal.go
```diff
@@ -0,0 +1,177 @@
...
+	HandleDeprecated()
+}
+
+// Unmarshal loads config from Viper into the structure pointed to by cfg.
+// Also, assigns all the necessary env variables by `envPrefix`.
+// It decodes with custom support for keys.PublicKeys, uint32, and util.Uint160,
+// applies keys, and handle deprecated fields.
+// Returns an error if decoding fails.
+func Unmarshal(v *viper.Viper, cfg config, envPrefix string) error {
+	bindEnvForStruct(v, cfg, envPrefix)
```
roman-khimov, line 29: But why doing it _before_ unmarshal? IIRC environment has a priority over config.

pkg/innerring/config/audit.go
```diff
@@ -0,0 +1,31 @@
+package config
+
+import "time"
+
+type Audit struct {
```
roman-khimov, line 5: Comments everywhere, these are exported types.

pkg/innerring/config/contracts.go
```diff
@@ -0,0 +1,23 @@
...
+	Balance    string   `mapstructure:"balance"`
+	Container  string   `mapstructure:"container"`
+	Netmap     string   `mapstructure:"netmap"`
+	Proxy      string   `mapstructure:"proxy"`
+	Reputation string   `mapstructure:"reputation"`
+	Alphabet   Alphabet `mapstructure:"alphabet"`
+}
+
+type Alphabet struct {
+	Az      string `mapstructure:"az"`
```
roman-khimov, line 16: Wow. How many times have we used these settings? Theoretically, this list should support all of `pkg/util/glagolitsa`, but my proposal is to drop it completely (@cthulhu-rider, @carpawell).

pkg/innerring/config/contracts.go
```diff
@@ -0,0 +1,23 @@
+package config
+
+type Contracts struct {
```
roman-khimov, line 3: This one is also useful only for main net contracts (neofs/processing). The only that could be relevant for the FS chain is unfortunately completely missing. Suggesting simplification here as well, these settings predate NNS-based configs (#736) and I don't think they were ever used after the switch.

pkg/innerring/config.go
```diff
@@ -33,258 +24,189 @@ const (
...
+
+	if !cfg.IsSet(cfgPathFSChainLocalConsensus) {
+		return fmt.Errorf("missing root section '%s'", cfgPathFSChainLocalConsensus)
 	}
 
-	_uint, err := parseConfigUint64Range(v, cfgPathFSChainLocalConsensus+".magic", "network magic", 1, math.MaxUint32)
-	if err != nil {
-		return c, err
+	magicKey := cfgPathFSChainLocalConsensus + ".magic"
+	if !cfg.IsSet(magicKey) {
```
roman-khimov, line 49: One things that bugs me is that we still have a mix of prefixes and structures in many places. Can this be avoided? Like here, unset magic is not much different from 0 magic ("not set or zero" is a good enough error message).

roman-khimov (commented):
I see it's a bit cleaner now and there is some generic code laid down here that's going to be reused by node configs as well as additional tests.

internal/configutil/unmarshal.go
End-rey: `bindEnvForStruct` sets new values for the `viper` variable that already reads the config, so the priority of ENV over the config is maintained.

pkg/innerring/config/audit.go
End-rey: Should I leave a comment for all fields of structures as well?

pkg/innerring/config.go
End-rey: Couldn't figure out how to check this for structures, for example for `fschain.consensus` key, or for `fschain.consensus.p2p.peers.min` key which can be 0, but if it is not specified, it is calculated.

pkg/innerring/config/audit.go
roman-khimov: Not required.

pkg/innerring/config.go
roman-khimov: At least we should try to minimize that, then we'll see what's left.

End-rey: I've already minimized it as much as I could, leaving only the section checks: `.consensus`, `.rpc`, `.p2p`, `.ping` and  1 check for field `fschain.consensus.p2p.peers.min`.

roman-khimov: OK, good.

roman-khimov:
Conflicts.

PR #3251, opened 2025-03-26
metabase: GC objects using new searchV2 index

### LINKED ISSUE #3245: Migrate metabase deletion code to searchV2 indexes
Implement the scheme from https://github.com/nspcc-dev/neofs-node/issues/3239#issuecomment-2735543688.

carpawell:
1. Cleanup new index based only on current new index's state
2. Keep actual stored objects' size based only on new index's state
3. Delete old indexes based on their old state, but do not involve them in new routines

Tests for counters and parent objects are based on PUT + DELETE operations, so they are actual for the new code too. Closes #3245.

### DIFF
--- a/pkg/local_object_storage/metabase/delete.go
+++ b/pkg/local_object_storage/metabase/delete.go
@@ -2,14 +2,10 @@ package meta
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
-	"slices"
 
-	"github.com/nspcc-dev/neofs-node/pkg/core/object"
 	storagelog "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/internal/log"
-	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/util/logicerr"
-	apistatus "github.com/nspcc-dev/neofs-sdk-go/client/status"
+	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 	objectSDK "github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"go.etcd.io/bbolt"
@@ -149,70 +145,12 @@ func (db *DB) delete(tx *bbolt.Tx, addr oid.Address, currEpoch uint64) (bool, bo
 		}
 	}
 
-	// unmarshal object, work only with physically stored (raw == true) objects
-	obj, err := db.get(tx, addr, key, false, true, currEpoch)
+	payloadSize, err := deleteMetadata(tx, db.log, addr.Container(), addr.Object(), false)
 	if err != nil {
-		var siErr *objectSDK.SplitInfoError
-		var notFoundErr apistatus.ObjectNotFound
-
-		if errors.As(err, &notFoundErr) || errors.As(err, &siErr) {
-			return false, false, 0, nil
-		}
-
-		if !errors.Is(err, logicerr.Error) {
-			// data corruption, it should be removed anyway
-			err = deleteAllDataByAddress(tx, addr)
-			if err != nil {
-				db.log.Warn("cleaning corrupted object failed", zap.Stringer("addr", addr), zap.Error(err))
-				return false, false, 0, nil
-			}
-		}
-
-		return false, false, 0, err
+		return false, false, 0, fmt.Errorf("can't remove metadata indexes: %w", err)
 	}
 
-	// if object is an only link to a parent, then remove parent
-	if parent := obj.Parent(); parent != nil && !parent.GetID().IsZero() {
-		err = db.deleteObject(tx, parent, true)
-		if err != nil {
-			return false, false, 0, fmt.Errorf("could not remove parent object: %w", err)
-		}
-	}
-
-	// remove object
-	err = db.deleteObject(tx, obj, false)
-	if err != nil {
-		return false, false, 0, fmt.Errorf("could not remove object: %w", err)
-	}
-
-	return true, removeAvailableObject, obj.PayloadSize(), nil
-}
-
-func (db *DB) deleteObject(
-	tx *bbolt.Tx,
-	obj *objectSDK.Object,
-	isParent bool,
-) error {
-	err := delUniqueIndexes(tx, obj, isParent)
-	if err != nil {
-		return fmt.Errorf("can't remove unique indexes")
-	}
-
-	err = updateListIndexes(tx, obj, delListIndexItem)
-	if err != nil {
-		return fmt.Errorf("can't remove list indexes: %w", err)
-	}
-
-	err = updateFKBTIndexes(tx, obj, delFKBTIndexItem)
-	if err != nil {
-		return fmt.Errorf("can't remove fake bucket tree indexes: %w", err)
-	}
-
-	if err = deleteMetadata(tx, obj.GetContainerID(), obj.GetID()); err != nil {
-		return fmt.Errorf("can't remove metadata indexes: %w", err)
-	}
-
-	return nil
+	return true, removeAvailableObject, payloadSize, nil
 }
 
 func delUniqueIndexItem(tx *bbolt.Tx, item namedBucketItem) {
@@ -222,30 +160,29 @@ func delUniqueIndexItem(tx *bbolt.Tx, item namedBucketItem) {
 	}
 }
 
-func delFKBTIndexItem(tx *bbolt.Tx, item namedBucketItem) error {
+func delFKBTIndexItem(tx *bbolt.Tx, item namedBucketItem) {
 	bkt := tx.Bucket(item.name)
 	if bkt == nil {
-		return nil
+		return
 	}
 
 	fkbtRoot := bkt.Bucket(item.key)
 	if fkbtRoot == nil {
-		return nil
+		return
 	}
 
 	_ = fkbtRoot.Delete(item.val) // ignore error, best effort there
-	return nil
 }
 
-func delListIndexItem(tx *bbolt.Tx, item namedBucketItem) error {
+func delListIndexItem(tx *bbolt.Tx, item namedBucketItem) {
 	bkt := tx.Bucket(item.name)
 	if bkt == nil {
-		return nil
+		return
 	}
 
 	lst, err := decodeList(bkt.Get(item.key))
 	if err != nil || len(lst) == 0 {
-		return nil
+		return
 	}
 
 	// remove element from the list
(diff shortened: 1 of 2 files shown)


pkg/local_object_storage/metabase/delete.go
```diff
@@ -149,43 +151,107 @@ func (db *DB) delete(tx *bbolt.Tx, addr oid.Address, currEpoch uint64) (bool, bo
 		}
 	}
 
-	// unmarshal object, work only with physically stored (raw == true) objects
-	obj, err := db.get(tx, addr, key, false, true, currEpoch)
-	if err != nil {
-		var siErr *objectSDK.SplitInfoError
-		var notFoundErr apistatus.ObjectNotFound
+	{
```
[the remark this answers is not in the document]
roman-khimov: Yeah, this `db.get()` is exactly what we're trying to avoid.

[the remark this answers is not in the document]
carpawell: to me, only two possibilities are correct:
1. old indexes are used to delete old indexes
2. new indexes are used to delete new indexes

if drop this block, nothing can delete old indexes. only if we use new ones to build an almost full header again to be able to delete every old thing we have. i do not see much difference with "Unmarshaling is needed to get object parent" then. i wanted (and started) to delete old indexes completely but it takes some time and touches many things, so I opened this one first
even if i build full header based on new metadata, that is a code that should be dropped right before 0.46.0 release anyway (as i understand, new meta is what we use since now for everything)

pkg/local_object_storage/metabase/delete.go
```diff
@@ -149,43 +151,107 @@ func (db *DB) delete(tx *bbolt.Tx, addr oid.Address, currEpoch uint64) (bool, bo
...
+	k, _ := c.Seek(startPref)
+	if !bytes.Equal(k, startPref) {
+		return 0, oid.ID{}, logicerr.Wrap(apistatus.ObjectNotFound{})
 	}
 
-	return true, removeAvailableObject, obj.PayloadSize(), nil
+	var size uint64
+	var parent oid.ID
+
+	for k, _ = c.Next(); bytes.HasPrefix(k[1:], oID[:]); k, _ = c.Next() {
```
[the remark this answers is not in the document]
carpawell: depends on whether you believe two GETs are faster than a single SEEK for values that are close for sure

pkg/local_object_storage/metabase/delete.go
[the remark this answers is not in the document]
cthulhu-rider: meta bucket contain all header fields, so it's enough to remove any related object data incl. old indexes. So, no need to decode the full header

pkg/local_object_storage/metabase/delete.go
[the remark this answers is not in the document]
cthulhu-rider: i believe in benchmarks. Without them, better to use as simple approach as possible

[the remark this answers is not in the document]
roman-khimov: It depends on how many attributes there are in an object, get is seek internally, but moving the cursor isn't free either.

pkg/local_object_storage/metabase/delete.go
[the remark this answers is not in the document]
roman-khimov: You don't need all of the header here, only specific parts of it in specific cases. But that's the essence of the switch --- you start using new indexes where appropriate (like size/type things, it doesn't matter where you get the data from, you'll need it anyway) _replacing_ old ones. And yes, currently you'll be dealing with both old and new indexes during delete (but not in some other cases, like #3241). Once we have no users for old indexes --- they can be deleted.

pkg/local_object_storage/metabase/delete.go
```diff
@@ -149,43 +151,107 @@ func (db *DB) delete(tx *bbolt.Tx, addr oid.Address, currEpoch uint64) (bool, bo
...
-	// remove object
-	err = db.deleteObject(tx, obj, false)
-	if err != nil {
-		return false, false, 0, fmt.Errorf("could not remove object: %w", err)
+	return true, removeAvailableObject, size, nil
+}
+
+func getSizeAndParent(tx *bbolt.Tx, cID cid.ID, oID oid.ID) (uint64, oid.ID, error) {
+	bkt := tx.Bucket(metaBucketKey(cID))
+	c := bkt.Cursor()
```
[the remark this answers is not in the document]
carpawell: there was a pattern that ensured every important bucket was created at the init step, i supported it. but meta bkt is not ensured, ok, fixed

pkg/local_object_storage/metabase/delete.go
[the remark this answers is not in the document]
carpawell: > i believe in benchmarks

there is no single bench for 10 objs, 1m objs, cached db, db-on-disk. i followed the idea that get is seek and also there should not be many attrs _in a common case_

anyway, currently it requires user attributes, so it must be SEEK

pkg/local_object_storage/metabase/delete.go
```diff
@@ -149,43 +169,222 @@ func (db *DB) delete(tx *bbolt.Tx, addr oid.Address, currEpoch uint64) (bool, bo
...
+	}
+	err = deleteMetadata(tx, cID, oID)
+	if err != nil {
+		return false, false, 0, fmt.Errorf("can't remove metadata indexes: %w", err)
+	}
+
+	return true, removeAvailableObject, h.payloadSize, nil
+}
+
+func getGCHeader(tx *bbolt.Tx, cID cid.ID, oID oid.ID) (gcHeader, error) {
```
cthulhu-rider, line 257: we dont read old index anymore which is good. The proposed approach seems generally correct if we limit ourselves to existing `deleteObject` and `deleteMetadata` funcs

however, the desired result can be achieved easier imo. I suppose we dont need an intermediate pseudo-header struct at all. I tend to think `cID`, `oID` and metabucket is all we need to remove everything possible in the whole DB:
- this func scans object in the metabucket, so it may delete new index itself
- having the new index, the old one can be removed on the spot
- `deleteMetadata` repeats iterating over metabucket almost the same, hinting that they are worth merging

there are several corner cases like broken values of the new index (this func fails on them now) or its divergence with the old one. They are abnormal, but solvable in general. Practice showed that ignoring only results in garbage getting stuck in the DB, although we can still drop more

roman-khimov: @carpawell, we're discussing this longer than the time needed to fix it. You don't need to reimplement them, you just need to give them enough data to do their job.

cthulhu-rider: i guessed it's enough to extend `deleteMetadata` iterator which fits well semantically under the removal of any indexes. It does not require a new function and structure. Mb im missing smth, but i don't see the point in reimplementing it

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -147,7 +158,89 @@ func deleteMetadata(tx *bbolt.Tx, cnr cid.ID, id oid.ID) error {
...
+				val:  id[:],
+			})
+			if err != nil {
+				return 0, err
+			}
+		case object.FilterPayloadSize:
+			size, _ = strconv.ParseUint(string(attrV), 10, 64)
+		default:
+			const systemFieldPrefix = "$Object:"
+			if !bytes.HasPrefix(attrK, []byte(systemFieldPrefix)) {
```
cthulhu-rider, line 234: else worth log, normally we dont expect sys attrs here

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -167,10 +260,15 @@ func deleteMetadata(tx *bbolt.Tx, cnr cid.ID, id oid.ID) error {
...
 	for i := range ks {
 		if err := metaBkt.Delete(ks[i]); err != nil {
-			return err
+			return 0, err
 		}
 	}
-	return nil
+
+	if !parent.IsZero() {
+		_, _ = deleteMetadata(tx, cnr, parent, true)
```
cthulhu-rider, line 268: why error is ignored?

pkg/local_object_storage/metabase/delete.go
cthulhu-rider: ten times better now

pkg/local_object_storage/metabase/metadata.go
carpawell: not all cases are provided in this `switch`, there will always be some other fields. it is not an error check, it is just a not-trigger-when-not-needed check

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -131,14 +132,24 @@ func PutMetadataForObject(tx *bbolt.Tx, hdr object.Object, hasParent, phy bool)
...
-		return nil
+		return 0, nil
 	}
+	var err error
+	var parent oid.ID
+	var size uint64
+	if !isParent {
+		v := metaBkt.Get(slices.Concat([]byte{metaPrefixIDAttr}, id[:], []byte(object.FilterPhysical), objectcore.MetaAttributeDelimiter, []byte(binPropMarker)))
+		if v == nil {
+			return 0, nil
```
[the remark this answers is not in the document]
carpawell: to keep prev behavior. tests want parent object removal to be no-op. it was more evident in the pref version of PR

pkg/local_object_storage/metabase/metadata.go
cthulhu-rider: log is always better than ignoring. But caller already logs child's errors, I'd keep similar records together in the code. Error can be from `bbolt/Bucket.Delete` for both child and parent objects. And we return child's error while ignore parent's ones, although they can be similar. Sounds like unnecessary duality

> err will lead to incorrect counters maintenance

nothing stops us from returning both child size and parent error in this case. The main thing is to process the return correctly

pkg/local_object_storage/metabase/metadata.go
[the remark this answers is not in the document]
cthulhu-rider: i'd skip this check, it is meaningless to me. Which test fails without it and on which assert?

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -147,7 +159,87 @@ func deleteMetadata(tx *bbolt.Tx, cnr cid.ID, id oid.ID) error {
...
+				owner := user.ID(attrV)
+				delFKBTIndexItem(tx, namedBucketItem{
+					name: ownerBucketName(cnr, bktKeyBuff),
+					key:  []byte(owner.EncodeToString()),
+					val:  id[:],
+				})
+			} else {
+				l.Warn("unexpected object's owner index",
+					zap.Stringer("addr", oid.NewAddress(cnr, id)),
+					zap.Binary("value", attrV),
```
cthulhu-rider, line 186: can be a big pile of trash, len is enough

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -131,23 +133,113 @@ func PutMetadataForObject(tx *bbolt.Tx, hdr object.Object, hasParent, phy bool)
...
+				)
+			}
+		case object.FilterPayloadChecksum:
+			err = delListIndexItem(tx, namedBucketItem{
```
cthulhu-rider, line 190: https://github.com/nspcc-dev/neofs-node/pull/3251#discussion_r2028493754

pkg/local_object_storage/metabase/metadata.go
cthulhu-rider: u mean not all possible old indexes are cased in the switch now? They all should be covered

pkg/local_object_storage/metabase/metadata.go
carpawell: > nothing stops us from returning both child size and parent error in this case

golang basis error to me: handling return value when `err != nil`. a child removal was requested, it is done, some side indexes were not removed, warn errors could be found in logs. key goal is achieved: results returned, strange side errors are logged, I do not understand you

pkg/local_object_storage/metabase/metadata.go
carpawell: not all newly indexed info was indexed in the old indexes. i carefully looked at the code and deleted only things that were deleted before; no functional changes

pkg/local_object_storage/metabase/delete.go
carpawell: to me: 1-hour intermediate refactor PR is taking 2nd week. structs/allocations/fast/slow/old funcs refactoring... i cannot remember the last time i did not understand what i was doing so much

also, the added code will not be removed in a single block/func/routine, it will have to be refactored carefully too

pkg/local_object_storage/metabase/metadata.go
cthulhu-rider: > a child removal was requested, it is done, some side indexes were not removed, warn errors could be found in logs.

log records appear for both child and parent, but in different places of code. That's what I don't see the point in. As well as in returning an error for child and ignoring for parent. They can behave the same way, either with returned error or internal log

cthulhu-rider (approved):

roman-khimov (approved):

PR #3257, opened 2025-03-26
Fix/meta/big object handling

carpawell:
SearchV2 PR reveals this problem.

### DIFF
--- a/pkg/services/meta/containers.go
+++ b/pkg/services/meta/containers.go
@@ -29,6 +29,7 @@ type containerStorage struct {
 	m           sync.RWMutex
 	mptOpsBatch map[string][]byte
 
+	l    *zap.Logger
 	path string
 	mpt  *mpt.Trie
 	db   storage.Store
@@ -162,7 +163,7 @@ func (s *containerStorage) putRawIndexes(ctx context.Context, l *zap.Logger, ee 
 		batch[string(append([]byte{oidIndex}, commsuffix...))] = []byte{}
 		if len(e.deletedObjects) > 0 {
 			batch[string(append([]byte{deletedIndex}, commsuffix...))] = e.deletedObjects
-			evWithMpt.additionalKVs, err = deleteObjectsOps(batch, s.db, e.deletedObjects, false)
+			evWithMpt.additionalKVs, err = s.deleteObjectsOps(batch, e.deletedObjects, false)
 			if err != nil {
 				l.Error("cleaning deleted object", zap.Stringer("oid", e.oID), zap.Error(err))
 				continue
@@ -296,9 +297,14 @@ func fillObjectIndex(batch map[string][]byte, h objectsdk.Object, isParent bool)
 	}
 }
 
-func deleteObjectsOps(dbKV map[string][]byte, s storage.Store, objects []byte, canDeleteLockObjects bool) (map[string][]byte, error) {
+func (s *containerStorage) deleteObjectsOps(dbKV map[string][]byte, objects []byte, canDeleteLockObjects bool) (map[string][]byte, error) {
 	rng := storage.SeekRange{}
 	mptKV := make(map[string][]byte)
+
+	if len(objects) == 0 {
+		return mptKV, nil
+	}
+	objects = s.expandChildren(objects)
 
 	// nil value means "delete" operation
 
@@ -311,7 +317,7 @@ func deleteObjectsOps(dbKV map[string][]byte, s storage.Store, objects []byte, c
 		var objFound bool
 		var err error
 
-		s.Seek(rng, func(k, v []byte) bool {
+		s.db.Seek(rng, func(k, v []byte) bool {
 			if bytes.Compare(k, stopKey) > 0 {
 				return false
 			}
@@ -399,7 +405,7 @@ func lastObjectKey(rawOID []byte) []byte {
 	return append(res, rawOID...)
 }
 
-func storageForContainer(rootPath string, cID cid.ID) (*containerStorage, error) {
+func storageForContainer(l *zap.Logger, rootPath string, cID cid.ID) (*containerStorage, error) {
 	p := path.Join(rootPath, cID.EncodeToString())
 
 	st, err := storage.NewBoltDBStore(dbconfig.BoltDBOptions{FilePath: p, ReadOnly: false})
@@ -422,6 +428,7 @@ func storageForContainer(rootPath string, cID cid.ID) (*containerStorage, error)
 	}
 
 	return &containerStorage{
+		l:           l.With(zap.Stringer("cid", cID)),
 		path:        p,
 		mpt:         mpt.NewTrie(prevRootNode, mpt.ModeLatest, storage.NewMemCachedStore(st)),
 		db:          st,
@@ -562,7 +569,7 @@ func (s *containerStorage) handleNewEpoch(e uint64) error {
 	}
 
 	dbBatch := make(map[string][]byte)
-	mptBatch, err := deleteObjectsOps(dbBatch, s.db, rawOIDs, true)
+	mptBatch, err := s.deleteObjectsOps(dbBatch, rawOIDs, true)
 	if err != nil {
 		return fmt.Errorf("making delete operations batch: %w", err)
 	}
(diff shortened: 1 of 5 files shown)


pkg/services/meta/containers.go
```diff
@@ -289,6 +289,10 @@ func fillObjectIndex(batch map[string][]byte, h objectsdk.Object) {
 			putPlainAttribute(batch, id, ak, av)
 		}
 	}
+
+	if hasParent {
+		fillObjectIndex(batch, *h.Parent(), true)
```
cthulhu-rider, line 294: `h.Parent()` can be deduplicated

pkg/services/meta/containers.go
```diff
@@ -591,3 +593,121 @@ func (s *containerStorage) handleNewEpoch(e uint64) error {
 
 	return errors.Join(mptErr, dbErr)
 }
+
+func expandChildren(st storage.Store, rootOIDs []byte) []byte {
+	if len(rootOIDs) == 0 {
```
cthulhu-rider, line 598: i'd do this check in `deleteObjectsOps` which also checks for emptiness

pkg/services/meta/containers.go
```diff
@@ -591,3 +593,121 @@ func (s *containerStorage) handleNewEpoch(e uint64) error {
...
+	//  3. first part ID -> all children search
+
+	var childrenWithParentID [][]byte
+	var rng storage.SeekRange
+	rng.Prefix = slices.Concat([]byte{attrPlainToOIDIndex}, []byte(objectsdk.FilterParentID), object.AttributeDelimiter)
+	rng.Start = make([]byte, oid.Size)
+
+	rootsSorted := slices.SortedFunc(slices.Chunk(rootOIDs, oid.Size), bytes.Compare)
+	lastRoot := rootsSorted[len(rootsSorted)-1]
+	copy(rng.Start, rootsSorted[0])
```
cthulhu-rider, line 615: wouldnt shallow copy be enough?

pkg/services/meta/containers.go
```diff
@@ -591,3 +593,121 @@ func (s *containerStorage) handleNewEpoch(e uint64) error {
...
+
+	st.Seek(rng, func(k, _ []byte) bool {
+		currRoot := k[len(rng.Prefix) : len(rng.Prefix)+oid.Size]
```
cthulhu-rider, line 618: len should be checked to prevent `index out of range` panic

pkg/services/meta/containers.go
```diff
@@ -591,3 +593,121 @@ func (s *containerStorage) handleNewEpoch(e uint64) error {
...
+		if len(rootsSorted) == 0 || bytes.Compare(currRoot, lastRoot) > 0 {
```
cthulhu-rider, line 619: we can catch `rootsSorted` emptiness in the cut place below and avoid extra iteration

pkg/services/meta/containers.go
```diff
@@ -591,3 +593,121 @@ func (s *containerStorage) handleNewEpoch(e uint64) error {
...
+		for i := 0; i < len(rootsSorted); i++ {
+			switch bytes.Compare(currRoot, rootsSorted[i]) {
+			case -1:
+				return true
+			case +1:
+				rootsSorted = rootsSorted[i+1:]
+				i = -1
+				continue
+			case 0:
+				childrenWithParentID = append(childrenWithParentID, slices.Clone(k[len(rng.Prefix)+oid.Size+1:]))
```
cthulhu-rider, line 631: len check will also ensure appended el's len is `oid.Size`

pkg/services/meta/containers.go
cthulhu-rider, line 619: if we pre-cmp against `lastRoot` anyway, we can also catch equality case

pkg/services/meta/containers.go
```diff
@@ -591,3 +593,121 @@ func (s *containerStorage) handleNewEpoch(e uint64) error {
...
+				rootsSorted = rootsSorted[i+1:]
+				return true
+			}
+		}
+
+		return true
```
cthulhu-rider, line 637: seems like here we are done

cthulhu-rider (commented):
finished partially, will back a bit later

pkg/services/meta/containers.go
```diff
@@ -587,3 +593,121 @@ func (s *containerStorage) handleNewEpoch(e uint64) error {
...
+		currRoot := k[len(rng.Prefix) : len(rng.Prefix)+oid.Size]
+		if len(rootsSorted) == 0 || bytes.Compare(currRoot, lastRoot) > 0 {
+			return false
+		}
+		for i := 0; i < len(rootsSorted); i++ {
+			switch bytes.Compare(currRoot, rootsSorted[i]) {
+			case -1:
+				return true
+			case +1:
+				rootsSorted = rootsSorted[i+1:]
```
cthulhu-rider, line 627: `i` is always `0` here? Seems like `for ; len(rootsSorted) > 0; rootsSorted = rootsSorted[1:] {}` loop would be more simple

pkg/services/meta/containers.go
```diff
@@ -587,3 +593,121 @@ func (s *containerStorage) handleNewEpoch(e uint64) error {
...
+	})
+
+	if len(childrenWithParentID) == 0 {
+		// all objects are roots, no additional children
+		return rootOIDs
+	}
+
+	firstPartOIDs := make([][]byte, 0, len(childrenWithParentID))
+	rng.Prefix = []byte{oidToAttrIndex}
+	copy(rng.Start, childrenWithParentID[0])
```
cthulhu-rider, line 647: `+ objectsdk.FilterFirstSplitObject + AttributeDelimiter`?

cthulhu-rider, line 647: sort before?

pkg/services/meta/containers.go
```diff
@@ -587,3 +593,121 @@ func (s *containerStorage) handleNewEpoch(e uint64) error {
...
+		currChild := k[1 : 1+oid.Size]
+		for i := 0; i < len(childrenWithParentID); i++ {
+			switch bytes.Compare(currChild, childrenWithParentID[i]) {
+			case -1:
+				return true
+			case +1:
+				// should never happen, it means we have info
+				// about child object with parent ID but child
+				// object does not have first part ID, nothing
+				// can be done, just skip
```
cthulhu-rider, line 666: log?

carpawell: what this log should say? like we have some garbage or unknown db state?

pkg/services/meta/containers.go
carpawell: changed but to even more simpler way

pkg/services/meta/containers.go
carpawell: if move empty check, yes, changed

pkg/services/meta/containers.go
carpawell: dropped this check at all

pkg/services/meta/containers.go
carpawell: again do not agree that db scheme missmatch does not deserve panic but fixed

pkg/services/meta/containers.go
cthulhu-rider: the same as the comment. It's worth leaving at least one way to detect this case. Log costs nothing for this

pkg/services/meta/containers.go
cthulhu-rider: ... and drop here as never happens

pkg/services/meta/containers.go
```diff
@@ -587,3 +597,115 @@ func (s *containerStorage) handleNewEpoch(e uint64) error {
...
+	}
+
+	slices.SortFunc(firstPartOIDs, bytes.Compare)
+	children := slices.Concat(firstPartOIDs...) // first objects are children too
+	rng.Prefix = slices.Concat([]byte{attrPlainToOIDIndex}, []byte(objectsdk.FilterFirstSplitObject), object.AttributeDelimiter)
+	rng.Start = firstPartOIDs[0]
+
+	st.Seek(rng, func(k, _ []byte) bool {
+		currChild := k[len(rng.Prefix) : len(rng.Prefix)+oid.Size]
+		if len(firstPartOIDs) == 0 {
```
cthulhu-rider, line 691: https://github.com/nspcc-dev/neofs-node/pull/3257#discussion_r2016324082

pkg/services/meta/containers.go
```diff
@@ -587,3 +597,115 @@ func (s *containerStorage) handleNewEpoch(e uint64) error {
...
+
+	var childrenWithParentID [][]byte
+	var rng storage.SeekRange
+	rng.Prefix = slices.Concat([]byte{attrPlainToOIDIndex}, []byte(objectsdk.FilterParentID), object.AttributeDelimiter)
+
+	rootsSorted := slices.SortedFunc(slices.Chunk(rootOIDs, oid.Size), bytes.Compare)
+	rng.Start = rootsSorted[0]
+
+	st.Seek(rng, func(k, _ []byte) bool {
+		if len(k) < len(rng.Prefix)+oid.Size+attributeDelimiterLen+oid.Size {
```
cthulhu-rider, line 620:
```suggestion
		if len(k) != len(rng.Prefix)+oid.Size+attributeDelimiterLen+oid.Size {
```
?

pkg/services/meta/containers.go
```diff
@@ -587,3 +597,115 @@ func (s *containerStorage) handleNewEpoch(e uint64) error {
...
+		// all objects are roots, no additional children
+		return rootOIDs
+	}
+
+	firstPartOIDs := make([][]byte, 0, len(childrenWithParentID))
+	slices.SortFunc(childrenWithParentID, bytes.Compare)
+	rng.Prefix = []byte{oidToAttrIndex}
+	rng.Start = slices.Concat(childrenWithParentID[0], []byte(objectsdk.FilterFirstSplitObject), object.AttributeDelimiter)
+	st.Seek(rng, func(k, _ []byte) bool {
+		currChild := k[1 : 1+oid.Size]
```
cthulhu-rider, line 655: https://github.com/nspcc-dev/neofs-node/pull/3257#discussion_r2016321087

pkg/services/meta/containers.go
```diff
@@ -587,3 +597,115 @@ func (s *containerStorage) handleNewEpoch(e uint64) error {
...
+	var childrenWithParentID [][]byte
+	var rng storage.SeekRange
+	rng.Prefix = slices.Concat([]byte{attrPlainToOIDIndex}, []byte(objectsdk.FilterParentID), object.AttributeDelimiter)
+
+	rootsSorted := slices.SortedFunc(slices.Chunk(rootOIDs, oid.Size), bytes.Compare)
+	rng.Start = rootsSorted[0]
+
+	st.Seek(rng, func(k, _ []byte) bool {
+		if len(k) < len(rng.Prefix)+oid.Size+attributeDelimiterLen+oid.Size {
+			// unexpected, just skip
```
cthulhu-rider, line 621: ignoring is usually bad, i'd fail or at least log

pkg/services/meta/containers.go
```diff
@@ -587,3 +597,115 @@ func (s *containerStorage) handleNewEpoch(e uint64) error {
...
+		return rootOIDs
+	}
+
+	slices.SortFunc(firstPartOIDs, bytes.Compare)
+	children := slices.Concat(firstPartOIDs...) // first objects are children too
+	rng.Prefix = slices.Concat([]byte{attrPlainToOIDIndex}, []byte(objectsdk.FilterFirstSplitObject), object.AttributeDelimiter)
+	rng.Start = firstPartOIDs[0]
+
+	st.Seek(rng, func(k, _ []byte) bool {
+		currChild := k[len(rng.Prefix) : len(rng.Prefix)+oid.Size]
```
cthulhu-rider, line 690: https://github.com/nspcc-dev/neofs-node/pull/3257#discussion_r2016321087

pkg/services/meta/containers.go
cthulhu-rider: don't we always exit the loop with an empty slice? I think we can just `return false` after `for`

pkg/services/meta/containers.go
cthulhu-rider: it wasn't bad actually, but had micro effect, so be it

pkg/services/meta/containers.go
carpawell: can't promise i wouldn't suggest such code again; i do not feel these checks are useful. fixed

pkg/services/meta/containers.go
carpawell: ok

pkg/services/meta/containers.go
carpawell: useless check after `return false` change

pkg/services/meta/containers.go
carpawell: > Log costs nothing for this

it is a log record inside an opened db tx that is asked to be idempotent (ok, it is a `Storage` interface but we all know what it means in practice)

fixed

roman-khimov (approved):
This has some potential for optimization and it's a bit hard to follow `[]byte` and `[][]byte` juggling for oid lists, but should be sufficient for now.

carpawell:
> it's a bit hard to follow []byte and [][]byte juggling for oid lists

I did not change what was done before so that is why this went this way. Previously, it was just take `[]byte` from contract and put to storage. Now it is a little bit trickier, will be improved eventually. I did not want to stop merging this because that is just a fast fix for searchv2 PR.

cthulhu-rider (approved):

PR #3258, opened 2025-03-27
ir: drop session key check

roman-khimov:
We've got test_static_session_token_container_delete failing after 4b7eee5c02938b9c2e84fe1e78a61f40581af4a4. This happens because deletion event has no real key and we're substituting it by the key used when container was created. It's an issuer key effectively, not a session key. And IMO this check is not very useful, we're checking signature against session-bound key anyway adn we trust this key because the token is issued by correct owner. The key from event is not very useful from the security perspective then.

The code extracting the key is moved as well because we can avoid decoding for the tokenized case.

### DIFF
--- a/pkg/innerring/processors/container/common.go
+++ b/pkg/innerring/processors/container/common.go
@@ -79,10 +79,6 @@ func (cp *Processor) verifySignature(v signatureVerificationData) error {
 			return errors.New("session token is not signed by the container owner")
 		}
 
-		if !tok.AssertAuthKey(&key) {
-			return errors.New("signed with a non-session key")
-		}
-
 		if !tok.AssertVerb(v.verb) {
 			return errWrongSessionVerb
 		}


roman-khimov:
Full test run: https://rest.fs.neo.org/HXSaMJXk2g8C14ht8HSi7BBaiYZ1HeWh2xnWPGQCg4H6/2872-1743096316/index.html

pkg/innerring/processors/container/common.go
```diff
@@ -79,10 +79,6 @@ func (cp *Processor) verifySignature(v signatureVerificationData) error {
 			return errors.New("session token is not signed by the container owner")
 		}
 
-		if !tok.AssertAuthKey(&key) {
```
cthulhu-rider, line 82: Iinm Chuck could attach the token Alice issued to Bob along with his own signature to mutate Alice's container. This condition prevents such a malicious behavior. Dropping the condition leaves [`session_key`](https://github.com/nspcc-dev/neofs-api/blob/33e3a77997674e576070db452f84583576a86368/session/types.proto#L112-L113) field unhandled

roman-khimov: https://github.com/nspcc-dev/neofs-sdk-go/blob/ac5d07bb6540e4f1f097a28e21ee20e1381f31ba/session/container.go#L246 uses `authKey` (which is a `session_key`) to verify provided signature. So the signature has to be created using the key specified in the token.

cthulhu-rider: alright. Then Chuck cannot change the request payload, so he can only fake the fact that Bob _sent_ the transaction when Bob wasn't going to. I see it as a special case of a replay attack. This check does not protect anyway, so lets drop

cthulhu-rider (approved):

carpawell (approved):
> against session-bound key anyway ADN we trust this key because

PR #3260, opened 2025-03-28
Fix/meta/subscriptions

carpawell:


pkg/services/meta/notifications.go
```diff
@@ -32,14 +32,11 @@ func (m *Meta) subscribeForBlocks(ch chan<- *block.Header) (string, error) {
...
+// unsubscribeFromBlocks reqauires [Meta.cliM] to be taken.
 func (m *Meta) unsubscribeFromBlocks() {
-	var err error
-	m.cliM.Lock()
-	defer m.cliM.Unlock()
-
 	m.l.Debug("unsubscribing from blocks")
 
-	err = m.ws.Unsubscribe(m.blockSubID)
+	err := m.ws.Unsubscribe(m.blockSubID)
```
cthulhu-rider, line 39: unrelated to deadlock

cthulhu-rider: u refactored var declaration in the fix

carpawell: hm, ok. it looked strange without mutexes, so i changed. reverted then

roman-khimov (approved):

PR #3261, opened 2025-04-01
writecache: distribute load during flushing

### LINKED ISSUE #1150: Distribute load in writecache during flushing
https://github.com/nspcc-dev/neofs-node/blob/master/pkg/local_object_storage/writecache/flush.go#L191
Currently the algorithm is pretty simple. We have 3 channels:
1) for persisted objects where only meta needs to be written
2) for objects which are written directly to the underlying store (e.g. because the cache is full)
3) for objects which are stored in writecache and are being flushed to the main memory in the background

Each flush worker receives from a random channel, but is guaranteed to check his priority channel first (all 3 have equal probability of becoming a priority channel for some worker).

This can be suboptimal. In this task:
1. Build a model and come up with some algorithms optimize `Put`/`Get` throughput and/or latency.
2. Implement what we think is the best, possibly with some configuration options.

As a simple example, we could have channels (1) and (2) to have more priority than channel (3).

End-rey:
Closes #1150.

### DIFF
--- a/cmd/neofs-lens/internal/storage/root.go
+++ b/cmd/neofs-lens/internal/storage/root.go
@@ -145,20 +145,18 @@ func openEngine() (*engine.StorageEngine, error) {
 
 	var shardsWithMeta []shardOptsWithID
 	for _, shCfg := range shards {
-		var writeCacheOpts []writecache.Option
-		if wcRead := shCfg.WritecacheCfg; wcRead.Enabled {
-			writeCacheOpts = append(writeCacheOpts,
-				writecache.WithPath(wcRead.Path),
-				writecache.WithMaxObjectSize(wcRead.MaxObjSize),
-				writecache.WithMaxCacheSize(wcRead.SizeLimit),
-				writecache.WithNoSync(wcRead.NoSync),
-			)
-		}
-
+		var (
+			wcMaxBatchSize      uint64
+			wcMaxBatchCount     int
+			wcMaxBatchThreshold uint64
+		)
 		var ss []blobstor.SubStorage
 		for _, sRead := range shCfg.SubStorages {
 			switch sRead.Typ {
 			case fstree.Type:
+				wcMaxBatchSize = uint64(sRead.CombinedSizeLimit)
+				wcMaxBatchCount = sRead.CombinedCountLimit
+				wcMaxBatchThreshold = uint64(sRead.CombinedSizeThreshold)
 				ss = append(ss, blobstor.SubStorage{
 					Storage: fstree.New(
 						fstree.WithPath(sRead.Path),
@@ -184,6 +182,19 @@ func openEngine() (*engine.StorageEngine, error) {
 				// should never happen, that has already
 				// been handled: when the config was read
 			}
+		}
+
+		var writeCacheOpts []writecache.Option
+		if wcRead := shCfg.WritecacheCfg; wcRead.Enabled {
+			writeCacheOpts = append(writeCacheOpts,
+				writecache.WithPath(wcRead.Path),
+				writecache.WithMaxObjectSize(wcRead.MaxObjSize),
+				writecache.WithMaxCacheSize(wcRead.SizeLimit),
+				writecache.WithNoSync(wcRead.NoSync),
+				writecache.WithMaxFlushBatchSize(wcMaxBatchSize),
+				writecache.WithMaxFlushBatchCount(wcMaxBatchCount),
+				writecache.WithMaxFlushBatchThreshold(wcMaxBatchThreshold),
+			)
 		}
 
 		var sh shardOptsWithID
--- a/pkg/local_object_storage/blobstor/fstree/fstree.go
+++ b/pkg/local_object_storage/blobstor/fstree/fstree.go
@@ -56,6 +56,13 @@ type Info struct {
 type writer interface {
 	writeData(oid.ID, string, []byte) error
 	finalize() error
+	writeBatch([]writeDataUnit) error
+}
+
+type writeDataUnit struct {
+	id   oid.ID
+	path string
+	data []byte
 }
 
 const (
@@ -341,6 +348,33 @@ func (t *FSTree) Put(addr oid.Address, data []byte) error {
 	if err != nil {
 		return fmt.Errorf("write object data into file %q: %w", p, err)
 	}
+	return nil
+}
+
+// PutBatch puts a batch of objects in the storage.
+func (t *FSTree) PutBatch(objs map[oid.Address][]byte) error {
+	if t.readOnly {
+		return common.ErrReadOnly
+	}
+
+	writeDataUnits := make([]writeDataUnit, 0, len(objs))
+	for addr, data := range objs {
+		p := t.treePath(addr)
+		if err := util.MkdirAllX(filepath.Dir(p), t.Permissions); err != nil {
+			return fmt.Errorf("mkdirall for %q: %w", p, err)
+		}
+		writeDataUnits = append(writeDataUnits, writeDataUnit{
+			id:   addr.Object(),
+			path: p,
+			data: t.Compress(data),
+		})
+	}
+
+	err := t.writer.writeBatch(writeDataUnits)
+	if err != nil {
+		return fmt.Errorf("cannot write batch: %w", err)
+	}
+
 	return nil
 }
 
--- a/pkg/local_object_storage/blobstor/fstree/fstree_write_linux.go
+++ b/pkg/local_object_storage/blobstor/fstree/fstree_write_linux.go
@@ -72,6 +72,17 @@ func newSpecificWriter(t *FSTree) writer {
 }
 
 func (w *linuxWriter) newSyncBatch() (*syncBatch, error) {
+	sb, err := w.createBatch()
+	if err != nil {
+		return nil, err
+	}
+	sb.lock.Lock()
+	sb.timer = time.AfterFunc(w.combinedWriteInterval, sb.sync)
+	sb.ready = make(chan struct{})
+	return sb, nil
+}
+
+func (w *linuxWriter) createBatch() (*syncBatch, error) {
 	fd, err := unix.Open(w.root, w.bFlags, w.perm)
 	if err != nil {
 		return nil, err
@@ -79,11 +90,8 @@ func (w *linuxWriter) newSyncBatch() (*syncBatch, error) {
 	sb := &syncBatch{
 		fd:       fd,
 		procname: "/proc/self/fd/" + strconv.FormatUint(uint64(fd), 10),
-		ready:    make(chan struct{}),
 		noSync:   w.noSync,
 	}
-	sb.lock.Lock()
-	sb.timer = time.AfterFunc(w.combinedWriteInterval, sb.sync)
 	return sb, nil
 }
 
@@ -113,8 +121,12 @@ func (b *syncBatch) intSync() {
 	if b.err == nil && err != nil {
 		b.err = err
 	}
-	close(b.ready)
-	_ = b.timer.Stop() // True is stopped, but false is "AfterFunc already running".
+	if b.ready != nil {
+		close(b.ready)
+	}
+	if b.timer != nil {
+		_ = b.timer.Stop() // True is stopped, but false is "AfterFunc already running".
+	}
 }
 
 func (b *syncBatch) wait() error {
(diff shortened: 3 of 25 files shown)


pkg/local_object_storage/writecache/flush.go
```diff
@@ -171,10 +171,10 @@ func (c *cache) flushSingle(addr oid.Address, ignoreErrors bool) error {
 }
 
 // flushObject is used to write object directly to the main storage.
-func (c *cache) flushObject(obj *object.Object, data []byte) error {
+func (c *cache) flushObject(obj *object.Object) error {
```
roman-khimov, line 257: But it's exactly to avoid marshaling, marshaling isn't free.

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -354,6 +356,44 @@ func (t *FSTree) Put(addr oid.Address, data []byte) error {
...
+		putPrms = append(putPrms, putPrm{
+			id:   obj.GetID(),
+			path: p,
+			data: t.Compress(obj.Marshal()),
+		})
+	}
+
+	_, ok := t.writer.(*genericWriter)
+	if ok {
+		t.log.Debug("writing batch for generic writer is not supported, use a common put")
```
roman-khimov, line 380: Not sure it's useful.

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -354,6 +356,44 @@ func (t *FSTree) Put(addr oid.Address, data []byte) error {
...
+// PutBatch puts a batch of objects in the storage.
+func (t *FSTree) PutBatch(objs []*objectSDK.Object) error {
+	if t.readOnly {
+		return common.ErrReadOnly
+	}
+
+	putPrms := make([]putPrm, 0, len(objs))
+	for _, obj := range objs {
+		p := t.treePath(object.AddressOf(obj))
+		if err := util.MkdirAllX(filepath.Dir(p), t.Permissions); err != nil {
```
roman-khimov, line 368: It's not needed for the optimized writer.

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -354,6 +356,44 @@ func (t *FSTree) Put(addr oid.Address, data []byte) error {
...
+			return fmt.Errorf("mkdirall for %q: %w", p, err)
+		}
+		putPrms = append(putPrms, putPrm{
+			id:   obj.GetID(),
+			path: p,
+			data: t.Compress(obj.Marshal()),
+		})
+	}
+
+	_, ok := t.writer.(*genericWriter)
```
roman-khimov, line 378: It can be handled in `func (w *genericWriter) writeBatch(objs []putPrm) error`, you don't need this is the common code.

pkg/local_object_storage/blobstor/fstree/fstree_write_linux.go
```diff
@@ -243,3 +243,42 @@ func (w *linuxWriter) writeFile(p string, data []byte) error {
 	}
 	return nil
 }
+
+func (w *linuxWriter) writeBatch(objs []putPrm) error {
+	w.batchLock.Lock()
+	defer w.batchLock.Unlock()
```
roman-khimov, line 249: No need to even look at this batch, just forget about it. Create a batch of your own and work with it here.

pkg/local_object_storage/blobstor/fstree/fstree_write_linux.go
```diff
@@ -243,3 +243,42 @@ func (w *linuxWriter) writeFile(p string, data []byte) error {
...
+		w.batch = nil
+	}
+
+	sb, err := w.newSyncBatch()
+	if err != nil {
+		return fmt.Errorf("create sync batch: %w", err)
+	}
+	defer sb.lock.Unlock()
+	w.batch = sb
+	_ = sb.timer.Stop()
```
roman-khimov, line 266: You can split `newSyncBatch()` to have some time-less batch without creating/stopping it.

pkg/local_object_storage/blobstor/fstree/fstree_write_linux.go
```diff
@@ -243,3 +243,42 @@ func (w *linuxWriter) writeFile(p string, data []byte) error {
...
+		}
+		w.batch = nil
+	}
+
+	sb, err := w.newSyncBatch()
+	if err != nil {
+		return fmt.Errorf("create sync batch: %w", err)
+	}
+	defer sb.lock.Unlock()
+	w.batch = sb
```
roman-khimov, line 265: No, this batch is independent.

pkg/local_object_storage/blobstor/fstree/fstree_write_linux.go
```diff
@@ -243,3 +243,42 @@ func (w *linuxWriter) writeFile(p string, data []byte) error {
...
+	for _, obj := range objs {
+		err = sb.write(obj)
+		if err != nil {
+			w.batch = nil
+			return fmt.Errorf("write object: %w", err)
+		}
+	}
+
+	sb.intSync()
+	err = sb.wait()
```
roman-khimov, line 277: Can be simplified as well, you own this thread, you don't need to close/wait on channel, the error is available too.

pkg/local_object_storage/blobstor/put.go
```diff
@@ -75,3 +71,81 @@ func (b *BlobStor) Put(addr oid.Address, obj *objectSDK.Object, raw []byte) ([]b
...
+		return batchSID, nil
+	} else if !errors.Is(err, common.ErrNoSpace) && err != ErrNoPlaceFound {
+		return nil, err
+	}
+
+	var (
+		sid []byte
+		err error
+	)
+	// if it didn't work out to put it in a batch, then we put it separately
```
roman-khimov, line 92: Do you think it makes any difference?

pkg/local_object_storage/blobstor/put.go
```diff
@@ -81,7 +81,7 @@ func (b *BlobStor) PutBatch(objs []*objectSDK.Object) ([]byte, error) {
 
 	if batchSID, err := b.tryPutBatch(objs); err == nil {
 		return batchSID, nil
-	} else if !errors.Is(err, common.ErrNoSpace) && err != ErrNoPlaceFound {
+	} else if !errors.Is(err, common.ErrNoSpace) && !errors.Is(err, ErrNoPlaceFound) {
```
roman-khimov, line 84: Doesn't belong to ad9ab71d88569031fd02f08e13ce9a5632276328

pkg/local_object_storage/writecache/flush.go
```diff
@@ -16,12 +17,14 @@ import (
...
-	// defaultFlushInterval is default time interval between successive flushes.
-	defaultFlushInterval = 10 * time.Second
 	// defaultErrorDelay is default time delay for retrying a flush after receiving an error.
 	defaultErrorDelay = 10 * time.Second
 	// defaultWorkerCount is a default number of workers that flush objects.
 	defaultWorkerCount = 20
+	// defaultMaxBatchDelay is a delay before pushing small objects.
+	defaultMaxBatchDelay = 1 * time.Second
+	// defaultMaxBatchSize is the maximum size of the sum of the sizes of small objects that flush together.
+	defaultMaxBatchSize = 1 << 16
```
roman-khimov, line 27: You need to pass the same parameters as are used for fstree (count/size).

pkg/local_object_storage/writecache/state.go
```diff
@@ -30,11 +30,22 @@ func (x *counters) Delete(addr oid.Address) {
...
+	x.mu.RLock()
+	defer x.mu.RUnlock()
 	return x.size
 }
 
+func (x *counters) Map() map[oid.Address]uint64 {
+	x.mu.RLock()
+	defer x.mu.RUnlock()
+
+	result := make(map[oid.Address]uint64, len(x.objMap))
```
roman-khimov, line 42: https://pkg.go.dev/maps#Clone

pkg/local_object_storage/writecache/flush.go
```diff
@@ -37,45 +40,141 @@ func (c *cache) runFlushLoop() {
...
-					return errors.New("closed during iteration")
+					return
 				}
-			}, true)
+			}
+
+			batch = append(batch, as.addr)
+			batchSize += uint64(as.size)
+
+			if batchTimer == nil && len(batch) == 1 {
```
roman-khimov, line 170: Why `len()` check?

pkg/local_object_storage/writecache/flush.go
```diff
@@ -204,6 +287,100 @@ func (c *cache) flushObject(obj *object.Object) error {
...
+			!errors.Is(err, blobstor.ErrNoPlaceFound) {
+			for _, obj := range objs {
+				addr := objectCore.AddressOf(obj)
+				c.reportFlushError("can't flush an object to blobstor",
+					addr.EncodeToString(), err)
+			}
+		}
+		return err
+	}
+
```
roman-khimov, line 326: I don't see the main map (obj->size) being updated for the batch.

cthulhu-rider (commented):
finished 1st two small commits. Mb return to the others a bit later

pkg/local_object_storage/blobstor/fstree/fstree.go
End-rey: Without this, I get an error `no such file or directory`.

pkg/local_object_storage/blobstor/fstree/fstree_write_linux.go
End-rey: I tried to simplify it, is that what you meant?

pkg/local_object_storage/blobstor/put.go
End-rey: Deleted the following code.

pkg/local_object_storage/writecache/flush.go
End-rey: It was difficult, but I passed the same parameters. Is that what it takes?

pkg/local_object_storage/writecache/flush.go
End-rey: I start the timer as soon as 1 object is in the batch.

pkg/local_object_storage/writecache/flush.go
End-rey: This is done below:
https://github.com/nspcc-dev/neofs-node/blob/dd3cf83843335a0e8fee5b05f427c3a1317a6a1d/pkg/local_object_storage/writecache/flush.go#L353

pkg/local_object_storage/writecache/flush.go
```diff
@@ -150,7 +150,7 @@ func (c *cache) flushSingle(addr oid.Address, ignoreErrors bool) error {
 		return err
 	}
 
-	err = c.flushObject(&obj, data)
+	err = c.flushObject(obj)
```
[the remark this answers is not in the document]
End-rey: I made a separate function `getObject` because the code was repeated and it returns a pointer.

cmd/neofs-lens/internal/storage/root.go
```diff
@@ -210,6 +206,18 @@ func openEngine() (*engine.StorageEngine, error) {
...
 		}
 
+		var writeCacheOpts []writecache.Option
+		if wcRead := shCfg.WritecacheCfg; wcRead.Enabled {
+			writeCacheOpts = append(writeCacheOpts,
+				writecache.WithPath(wcRead.Path),
+				writecache.WithMaxObjectSize(wcRead.MaxObjSize),
+				writecache.WithMaxCacheSize(wcRead.SizeLimit),
+				writecache.WithNoSync(wcRead.NoSync),
+				writecache.WithMaxFlushBatchDelay(wcMaxBatchDelay),
```
roman-khimov, line 216: 1. Delay can be different for WC.
2. Count/size/thresh are all important for proper batches, batch size is not the only parameter.

pkg/local_object_storage/writecache/flush.go
```diff
@@ -16,12 +17,14 @@ import (
...
 
 const (
-	// defaultFlushInterval is default time interval between successive flushes.
-	defaultFlushInterval = 10 * time.Second
 	// defaultErrorDelay is default time delay for retrying a flush after receiving an error.
 	defaultErrorDelay = 10 * time.Second
 	// defaultWorkerCount is a default number of workers that flush objects.
 	defaultWorkerCount = 20
+	// defaultMaxBatchDelay is a default delay before pushing small objects.
+	defaultMaxBatchDelay = 10 * time.Millisecond
```
roman-khimov, line 25: WC can have this delay be substantially bigger. Remember that FSTree has to balance PUT delay (batching obviously adds some time to every object) and throughput. WC can have this balance shifted significantly to throughput because it already holds the object and PUT was finished some time ago, now it's an internal node thing. A second is perfectly fine here.

roman-khimov (approved):
Good one! Rebase, please.

PR #3264, opened 2025-04-02
Refactor and improve verification of objects' signatures

cthulhu-rider:
continues #3216


pkg/core/object/fmt.go
```diff
@@ -158,17 +158,17 @@ func (v *FormatValidator) Validate(obj *object.Object, unprepared bool) error {
...
+
 		if err := validateSignatureKey(obj); err != nil {
 			return fmt.Errorf("could not validate signature key: %w", err)
 		}
 
 		if err := v.checkExpiration(*obj); err != nil {
 			return fmt.Errorf("object did not pass expiration check: %w", err)
 		}
-
-		if err := obj.CheckHeaderVerificationFields(); err != nil {
```
roman-khimov, line 169: I agree that it has precedence over the expiration check, however its relation to `validateSignatureKey` is far less obvious. The chain of trust is "container owner -> session key -> object signature" and I'd expect following this chain for any checks.

At the same time _format_ checks have precedence over signature checks, currently they're somewhat mixed (related to following commits).

pkg/core/object/fmt.go
```diff
@@ -161,14 +161,15 @@ func (v *FormatValidator) Validate(obj *object.Object, unprepared bool) error {
...
 		}
-		if err := obj.VerifyID(); err != nil {
-			return fmt.Errorf("could not validate header fields: invalid identifier: %w", err)
-		}
 
 		if err := validateSignatureKey(obj); err != nil {
 			return fmt.Errorf("could not validate signature key: %w", err)
 		}
 
+		if err := obj.VerifyID(); err != nil {
```
roman-khimov, line 165: That's exactly what should be done first. It's cheap and it's a format violation. If it's broken signatures don't matter.

internal/crypto/tokens.go
```diff
@@ -1,35 +1,29 @@
...
-	"fmt"
 
 	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
 	"github.com/nspcc-dev/neofs-sdk-go/user"
 )
 
 // TODO: https://github.com/nspcc-dev/neofs-node/issues/2795 after API stabilization, move some components to SDK
 
-// SignedToken is a common interface of signed NeoFS tokens.
-type SignedToken interface {
```
roman-khimov, line 15: `SignedEntity` then? Not sure why it's removed completely.

pkg/core/object/fmt.go
```diff
@@ -188,7 +188,7 @@ func validateSignature(obj *object.Object) error {
 	}
 
 	if err := icrypto.AuthenticateObject(*obj); err != nil {
-		return errors.New("invalid signature")
+		return fmt.Errorf("authenticate: %w", err)
```
roman-khimov: Not sure it's easily understandable, usually "authentication" is applied to users.

internal/crypto/tokens.go
cthulhu-rider: used once now

pkg/core/object/fmt.go
cthulhu-rider: we check the authenticity of object creation by its owner. Signature may be valid, but  Any suggestions?

pkg/core/object/fmt.go
cthulhu-rider: kept the order

pkg/core/object/fmt.go
cthulhu-rider: raised

pkg/core/object/fmt.go
roman-khimov: I'd say it's verification. But the difference is not so obvious here.

cthulhu-rider: i can leave this `verify signature:`

roman-khimov (approved):

pkg/core/object/fmt.go
roman-khimov: I'd be somewhat ambiguous this way, we have a number of signatures and even some other checks in `AuthenticateObject`.

cthulhu-rider: then im leaving `authenticate` until a better option

PR #3270, opened 2025-04-04
Drop tree service

roman-khimov:
Bye-bye, you won't be missed.

With S3 0.36.0 we can finally do this.


CHANGELOG.md
```diff
@@ -46,6 +49,9 @@ along with the alphabet, were removed from the config.
 
 In SN configuration section `contracts` is gone and no longer needed.
 
+Experimental tree service is completely removed in this release, please drop
```
cthulhu-rider, line 52: i'd also suggest to rm corresponding DB files (iiuc they wont dissapear automatically)

cthulhu-rider (approved):

CHANGELOG.md
roman-khimov: Makes sense.

PR #3271, opened 2025-04-04
Refactor and improve verification of requests' signatures

cthulhu-rider:
continues #3216

### DIFF
--- a/internal/crypto/requests.go
+++ b/internal/crypto/requests.go
@@ -1,11 +1,18 @@
 package crypto
 
 import (
+	apistatus "github.com/nspcc-dev/neofs-sdk-go/client/status"
 	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
 )
 
 // VerifyRequestSignatures checks whether all request signatures are set and
-// valid.
+// valid. Returns [apistatus.SignatureVerification] otherwise.
 func VerifyRequestSignatures[B neofscrypto.ProtoMessage](req neofscrypto.SignedRequest[B]) error {
-	return neofscrypto.VerifyRequestWithBuffer(req, nil)
+	err := neofscrypto.VerifyRequestWithBuffer(req, nil)
+	if err != nil {
+		var st apistatus.SignatureVerification
+		st.SetMessage(err.Error())
+		return st
+	}
+	return nil
 }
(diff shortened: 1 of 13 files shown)


internal/crypto/requests.go
```diff
@@ -14,5 +18,34 @@ func VerifyRequestSignatures[B neofscrypto.ProtoMessage](req neofscrypto.SignedR
 		st.SetMessage(err.Error())
 		return st
 	}
+
```
roman-khimov, line 21: Not needed.

pkg/services/object/server.go
```diff
@@ -495,15 +497,16 @@ func (s *server) Delete(ctx context.Context, req *protoobject.DeleteRequest) (*p
...
+	if err != nil {
 		return s.makeStatusDeleteResponse(err), nil
 	}
 
 	if s.fsChain.LocalNodeUnderMaintenance() {
 		return s.makeStatusDeleteResponse(apistatus.ErrNodeUnderMaintenance), nil
 	}
 
-	reqInfo, err := s.reqInfoProc.DeleteRequestToInfo(req)
+	reqInfo, err := s.reqInfoProc.DeleteRequestToInfo(req, author, authorPub)
```
roman-khimov, line 509: This passing of data around doesn't seem elegant.

internal/crypto/requests.go
```diff
@@ -14,5 +18,34 @@ func VerifyRequestSignatures[B neofscrypto.ProtoMessage](req neofscrypto.SignedR
...
+// AuthenticateRequest checks whether all request signatures are set and valid.
+// Returns request author ID along with public key if so, or
+// [apistatus.SignatureVerification] error otherwise.
+func AuthenticateRequest[B neofscrypto.ProtoMessage](req neofscrypto.SignedRequest[B]) (user.ID, []byte, error) {
+	if err := VerifyRequestSignatures(req); err != nil {
+		return user.ID{}, nil, err
+	}
+
+	var sig *refs.Signature
```
roman-khimov, line 33: This part can just be a separate method called from `acl/v2` for a request. It doesn't reuse much of existing code anyway, the logic is pretty separate from `VerifyRequestSignatures`.

pkg/services/object/server.go
cthulhu-rider: redone

roman-khimov (approved):

PR #3274, opened 2025-04-05
object: do not allow split of split

### LINKED ISSUE #2451: Root object is inaccessible from metabase if it contains split fields
* see #2448

currently, root objects has no split fields. In theory, they may have (multi split chain?). Metabase accepts such objects carried by the last child or the linking one. Then metabase fails `Exists` op with `no split info on parent object` error (object couldn't be read from the engine/node). This behavior is incorrect. Possible corrections:
* deny storage of such objects
* accept objects but don't fail like now

carpawell:
Closes #2451.

### DIFF
--- a/pkg/core/object/fmt.go
+++ b/pkg/core/object/fmt.go
@@ -128,6 +128,10 @@ func (v *FormatValidator) Validate(obj *object.Object, unprepared bool) error {
 	par := obj.Parent()
 
 	if obj.HasParent() {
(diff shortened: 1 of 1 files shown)


roman-khimov (approved):

cthulhu-rider (approved):

cthulhu-rider:
@carpawell lets resolve and merge

Issue #202, opened 2020-11-24
Object service does not verify eACL signature


cthulhu-rider:
`acl` sub-service of `object` service receives eACL table directly from morph storage:

`table, _, err := s.w.GetEACL(cid)`

Is it correct to ignore the table signature or service must verify it?

alexvanin:
As for me, I don't think node should check signatures when getting eACL directly from NEO RPC node for internal purposes (like checking eACL rules).

I think node _can_ do that if smart contract returns both signature and public key, but in general it looks like overhead for me. NeoFS node relies on its NEO RPC node as it is source of truth.

/cc @realloc

alexvanin:
We've decided to check signatures if they are available. For object service it is possible to do after https://github.com/nspcc-dev/neofs-node/pull/317 

However `GetEACL` may return signature without public key if this public key was revoked by container owner.

PR #3276, opened 2025-04-07
node: Do not verify signature of eACLs from the Container contract

cthulhu-rider:
for #3194, smth definite should be done for this. I decided to trust the contract and drop verification. If we'll decide to keep the check, it should be globalized/deduplicated and refactored same as #3265


roman-khimov (approved):
Refs. ad348afcd6c85c1c3111e3665043dfea4386a465, #202. This is just weird, we get a whole network map and a lot of other things from chain without signatures and it's never a problem. We trust chain data, there is no way around it.

PR #3278, opened 2025-04-08
node/object: Sync response processing of forwarded and new Head requests

cthulhu-rider:
relates #3276 a bit, but less obvious. We may discuss necessity of checking integrity of the headers came from remote SNs. But i think checks should be the same as those done on Put (`FormatValidator`, #3264). For now, i propose to drop this check just to sync with SDK client behavior


roman-khimov (approved):
Header is a self-descriptive structure, it maps clearly to OID and no other header maps to the same one. So at the `GET` level it needs no signatures at all.

PR #3280, opened 2025-04-09
Speed up peapod-to-fstree

roman-khimov:


carpawell (approved):

pkg/local_object_storage/blobstor/common/storage.go
```diff
@@ -69,15 +78,32 @@ func Copy(dst, src Storage) error {
...
-		if err != nil {
-			return fmt.Errorf("put object %s into destination sub-storage: %w", addr, err)
+		if batchSize <= 1 {
+			err = dst.Put(addr, data)
+			if err != nil {
+				return fmt.Errorf("put object %s into destination sub-storage: %w", addr, err)
+			}
+			return nil
+		}
+		objBatch[addr] = data
```
cthulhu-rider, line 88: iinm `objBatch` can be non-empty after `src.Iterate` and we have to finish recording

End-rey (approved):
Linter says that `epochState` in migrator is not used.

PR #3286, opened 2025-04-11
Feat/improve meta debug via lens

carpawell:


pkg/local_object_storage/metabase/status.go
```diff
@@ -146,13 +156,31 @@ func readBuckets(tx *bbolt.Tx, cID cid.ID, objKey []byte) []BucketValue {
...
+	if mBucket == nil {
+		return oldIndexes, nil
+	}
+
+	c := mBucket.Cursor()
+	pref := slices.Concat([]byte{metaPrefixIDAttr}, objKey)
+	k, _ := c.Seek(pref)
+	for ; bytes.HasPrefix(k, pref); k, _ = c.Next() {
+		kCut := k[len(pref):]
+		i := bytes.Index(kCut, object.MetaAttributeDelimiter)
```
cthulhu-rider, line 177: `bytes.Cut` simplifies delimiter cut

pkg/local_object_storage/metabase/status.go
```diff
@@ -146,13 +156,31 @@ func readBuckets(tx *bbolt.Tx, cID cid.ID, objKey []byte) []BucketValue {
...
+		if i == -1 {
+			continue
+		}
+
+		newIndexes = append(newIndexes, HeaderField{K: string(kCut[:i]), V: kCut[i+len(object.MetaAttributeDelimiter):]})
```
cthulhu-rider, line 182: `V` goes out the `tx`, it must be cloned

we may also use const `attributeDelimiterLen`

cmd/neofs-lens/internal/meta/status.go
```diff
@@ -60,6 +60,12 @@ func statusFunc(cmd *cobra.Command, _ []string) error {
 		cmd.Printf("\tBucket: %d\n"+
 			"\tValue (HEX): %s\n", bucket.BucketIndex, valStr)
 	}
+	if len(res.HeaderIndex) > 0 {
+		cmd.Println("Header field indexes:")
+		for _, field := range res.HeaderIndex {
+			cmd.Printf("\t%s: %x\n", field.K, field.V)
```
cthulhu-rider, line 66: `field.K` is a UTF-8 w/o zeros, it may be not human-readable

carpawell: what do you suggest here? it will always have human-readable system fields, and will likely have readable user attrs (why uft8 but not readable?). if we are afraid that it may not be readable, then it will not be readable for sure if we encode it somehow and reading it (first of all it is a debug util that is supposed to help with smth) will be harder

carpawell: ok, print only system attrs

cthulhu-rider: i just noticed that. In general, not all attributes can be read, copy-pasted, etc. But i checked that, for example, `neofs-cli object get`, also prints them as is (`%s`)

so it makes no sense to behave differently in this case only. Lets go back to the original version of ur code

cthulhu-rider (approved):

cmd/neofs-lens/internal/meta/status.go
roman-khimov: There is `%q`.

PR #3288, opened 2025-04-11
Reuse node config in components

### LINKED ISSUE #2994: Make validation more simple and reuse config in components

End-rey:
I have expanded the work with env variables. But in the future, I see that viper will be able to solve this problem on its own.
This feature is experimental, but already implemented https://github.com/spf13/viper/pull/1429.
There is a problem with the slices that has not been resolved yet https://github.com/spf13/viper/issues/1732.

Closes #2994.


roman-khimov:
For experimental and other viper things we need an issue, we'll of course use new APIs once they're available.

cmd/neofs-node/config/fschain/chain.go
```diff
@@ -0,0 +1,44 @@
+package fschainconfig
+
+import (
+	"time"
+)
+
+const (
+	//DialTimeoutDefault is a default dial timeout of FS chain client connection.
```
roman-khimov, line 8: Missing space after `//`

cmd/neofs-node/config/fschain/chain.go
```diff
@@ -0,0 +1,44 @@
...
+	DialTimeoutDefault = time.Minute
+	// CacheTTLDefault is a default value for cached values TTL.
+	// It is 0, because actual default depends on block time.
+	CacheTTLDefault = time.Duration(0)
+	// ReconnectionRetriesNumberDefault is a default value for reconnection retries.
```
roman-khimov, line 13: btw, it's `the default` everywhere, it's the one and only value here, not some abstract default.

cmd/neofs-node/config/internal/size.go
```diff
@@ -0,0 +1,101 @@
...
+
+// The following code is taken from https://github.com/spf13/viper/blob/master/util.go
+// with minor corrections (allow to use both `k` and `kb` forms).
+// Seems like viper allows to convert sizes but corresponding parser in `cast` package
+// is missing.
+
+// safeMul returns size*multiplier, rounding down to the
+// multiplier/1024 number of bytes.
+// Returns 0 if overflow is detected.
+func safeMul(size float64, multiplier uint64) uint64 {
```
roman-khimov, line 48: I'd avoid `float64` here, btw. We never officially supported "2.5G", I think.

cmd/neofs-node/config/profiler/profiler.go
```diff
@@ -0,0 +1,27 @@
...
+
+const (
+	// ShutdownTimeoutDefault is a default value for profiler HTTP service timeout.
+	ShutdownTimeoutDefault = 30 * time.Second
+	// AddressDefault is a default value for profiler HTTP service endpoint.
+	AddressDefault = "localhost:6060"
+)
+
+// Pprof contains configuration for the pprof profiler.
+type Pprof struct {
```
roman-khimov, line 13: It's the same structure as for prometheus, can be generalized.

cmd/neofs-node/config/validate_test.go
```diff
@@ -250,80 +248,31 @@ storage:
...
+			_, err = New(WithConfigFile(configFilePath))
 			fmt.Println(err)
 			if (err != nil) != tt.wantErr {
 				t.Errorf("CheckForUnknownFields() error = %v, wantErr %v", err, tt.wantErr)
 			}
 		})
 	}
 }
-
-func TestCheckForUnknownFieldsExamples(t *testing.T) {
```
roman-khimov, line 272: No more tests for `config/`?

cmd/neofs-node/validate.go
```diff
@@ -1,71 +1,103 @@
...
+
 	// shard configuration validation
 
 	shardNum := 0
 	paths := make(map[string]pathDescription)
-	return engineconfig.IterateShards(c, false, func(sc *shardconfig.Config) error {
-		if sc.WriteCache().Enabled() {
-			err := addPath(paths, "writecache", shardNum, sc.WriteCache().Path())
+	return engineconfig.IterateShards(&c.Storage, false, func(sc *shardconfig.ShardDetails) error {
+		if sc.Mode.String() == "UNDEFINED" {
```
roman-khimov, line 66: I'd expect some non-string comparison here. Maybe even `!sc.Mode.IsValid()` that would do the magic internally.

config/example/node.env
```diff
@@ -14,7 +14,7 @@ NEOFS_PROMETHEUS_SHUTDOWN_TIMEOUT=15s
 NEOFS_NODE_WALLET_PATH=./wallet.json
 NEOFS_NODE_WALLET_ADDRESS=NcpJzXcSDrh5CCizf4K9Ro6w4t59J5LKzz
 NEOFS_NODE_WALLET_PASSWORD=password
-NEOFS_NODE_ADDRESSES="s01.neofs.devenv:8080 /dns4/s02.neofs.devenv/tcp/8081 grpc://127.0.0.1:8082 grpcs://localhost:8083"
+NEOFS_NODE_ADDRESSES="s01.neofs.devenv:8080,/dns4/s02.neofs.devenv/tcp/8081,grpc://127.0.0.1:8082,grpcs://localhost:8083"
```
roman-khimov, line 17: Does it work with spaces now? Did it work previously?

cmd/neofs-node/config/validate_test.go
End-rey: These tests are now included here: https://github.com/nspcc-dev/neofs-node/blob/e329da3cca9af69847a1b718daa7f8f35a36e734/cmd/neofs-node/validate_test.go

cmd/neofs-lens/internal/storage/root.go
```diff
@@ -162,10 +81,10 @@ func openEngine() (*engine.StorageEngine, error) {
 						fstree.WithPath(sRead.Path),
 						fstree.WithPerm(sRead.Perm),
 						fstree.WithDepth(sRead.Depth),
-						fstree.WithNoSync(sRead.NoSync),
+						fstree.WithNoSync(*sRead.NoSync),
```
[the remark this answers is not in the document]
End-rey: I checked it here:
https://github.com/nspcc-dev/neofs-node/blob/e329da3cca9af69847a1b718daa7f8f35a36e734/cmd/neofs-node/config/engine/shard/blobstor/blobstor.go#L60
This was done in order to check whether the value was set or not.

cmd/neofs-node/config/control/control.go
```diff
@@ -0,0 +1,22 @@
...
+
+	GRPC struct {
+		Endpoint string `mapstructure:"endpoint"`
+	} `mapstructure:"grpc"`
+}
+
+// Normalize sets default values for Control configuration.
+func (c *Control) Normalize() {
+	if c.GRPC.Endpoint == "" {
+		c.GRPC.Endpoint = GRPCEndpointDefault
```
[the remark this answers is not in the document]
End-rey: The same logic was there before that. If it's not needed, I can delete it.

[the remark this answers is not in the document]
roman-khimov: It'd be useful if there was any real default. But we don't have any officially, so this can be omitted. If we're to ever have any it's not hard to add a check.
https://github.com/nspcc-dev/neofs-node/blob/master/docs/storage-node-configuration.md#grpc-section

config/example/node.env
roman-khimov: We should always think of compatibility. If this worked with spaces it should continue working with spaces.

cmd/neofs-node/config/config.go
```diff
@@ -1,82 +1,101 @@
...
+		if strings.HasPrefix(k, key) {
+			delete(c.isSet, k)
 		}
 	}
+}
 
-	return nil
+// UnsetAll unsets all keys from config.
+func (c *Config) UnsetAll() {
+	c.isSet = nil
```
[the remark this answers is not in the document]
End-rey: If `c.isSet` is nil, then here `_, ok := c.isSet[key]`, ok will be false.

roman-khimov (approved):

internal/configutil/unmarshal.go
```diff
@@ -24,16 +22,19 @@ type config interface {
...
 		mapstructure.ComposeDecodeHookFunc(
-			mapstructure.StringToTimeDurationHookFunc(),
-			mapstructure.StringToSliceHookFunc(","),
-			publicKeyHook(),
-			uint32StrictHook(),
-			uint160Hook(),
+			append([]mapstructure.DecodeHookFunc{
+				mapstructure.StringToTimeDurationHookFunc(),
+				mapstructure.StringToSliceHookFunc(" "),
+				//mapstructure.StringToSliceHookFunc(","),
```
carpawell, line 33: should be commented?

carpawell (approved):
I would keep file names in `noefs-node/config/*/config.go` or rename test files accordingly.

internal/configutil/unmarshal.go
End-rey: No...

PR #3293, opened 2025-04-14
writecache: expose metrics

### LINKED ISSUE #3201: Check and improve WC metrics
Related to #1390, let's check and expose more things about WC (size?) via metrics.

End-rey:
Closes #3201.


pkg/local_object_storage/writecache/put.go
```diff
@@ -23,6 +23,10 @@ func (c *cache) Put(addr oid.Address, obj *objectSDK.Object, data []byte) error
 		return ErrReadOnly
 	}
 
+	if c.metrics != nil {
```
roman-khimov, line 26: Better be done after the size check since in that case WC isn't doing anything.

pkg/local_object_storage/writecache/delete.go
```diff
@@ -15,17 +15,22 @@ func (c *cache) Delete(addr oid.Address) error {
...
 		storagelog.Write(c.log,
-			storagelog.AddressField(saddr),
+			storagelog.AddressField(addr),
 			storagelog.StorageTypeField(wcStorageType),
 			storagelog.OpField("DELETE"),
 		)
 		c.objCounters.Delete(addr)
+		if c.metrics != nil {
+			c.metrics.DecWCObjectCount()
+			c.metrics.SetWCSize(c.objCounters.Size())
```
roman-khimov, line 32: You can technically use `AddWCSize` here as well, just negate the size.

pkg/metrics/writecache.go
```diff
@@ -0,0 +1,99 @@
...
+
+	objectCount prometheus.Gauge
+	size        prometheus.Gauge
+}
+
+func newWritecacheMetrics() writecacheMetrics {
+	var (
+		putDuration = prometheus.NewHistogram(prometheus.HistogramOpts{
+			Namespace: storageNodeNameSpace,
+			Subsystem: writecahceSubsystem,
```
roman-khimov, line 24: Does it work correctly with multiple shards?

pkg/local_object_storage/writecache/delete.go
End-rey: To do this, you need to know the size, but it is not possible to get it optimally.

pkg/metrics/writecache.go
End-rey: Added a shard label for all metrics.

End-rey:
I noticed that for 5 puts with big and small objects:
```
neofs_node_writecache_put_time_count{shard="CQ6vCQ1B9ec1Gm6eKB1yT4"} 3
neofs_node_writecache_put_time_count{shard="TisnMndhZimyiDL6McXTcU"} 2
```
There is a correct number of small objects that flushed:
```
neofs_node_writecache_flush_batch_time_count{shard="CQ6vCQ1B9ec1Gm6eKB1yT4"} 1
neofs_node_writecache_flush_batch_time_count{shard="TisnMndhZimyiDL6McXTcU"} 1
```
But very strange number for big objects:
```
neofs_node_writecache_flush_single_time_count{shard="CQ6vCQ1B9ec1Gm6eKB1yT4"} 76
neofs_node_writecache_flush_single_time_count{shard="TisnMndhZimyiDL6McXTcU"} 29
```
This is most likely due to the fact that the object is still not flushed and it is still in the size map, and each time it is sent to the worker.

roman-khimov (approved):

PR #3297, opened 2025-04-14
Rm/object header limitation

carpawell:


carpawell:
@roman-khimov, @cthulhu-rider, we have a problem with binary replication, we wanted not to unmarshal anything when replicating but now if we want to limit replication with too big header, we need to either unmarshal it, or add some custom  light bytes reader that says from the binary object how big a header is. Not that hard but some magic code for a single object, do we need it?

roman-khimov:
No.

carpawell:
Ready then.

pkg/services/object/acl/v2/service.go
```diff
@@ -491,14 +491,9 @@ func (b Service) PutRequestToInfo(request *protoobject.PutRequest) (RequestInfo,
...
-	if !replication {
-		// header length is unchecked for replication because introducing a restriction
-		// should not prevent the replication of objects created before.
-		// See also https://github.com/nspcc-dev/neofs-api/issues/293
-		var hdrLen = header.MarshaledSize()
-		if hdrLen > objectsdk.MaxHeaderLen {
-			return RequestInfo{}, user.ID{}, fmt.Errorf("object header length exceeds the limit: %d>%d", hdrLen, objectsdk.MaxHeaderLen)
-		}
+	var hdrLen = header.MarshaledSize()
+	if hdrLen > objectsdk.MaxHeaderLen {
```
roman-khimov, line 495: Not sure why we're dealing with it here at all, it's supposed to be about permissions, not format correctness.

carpawell: idk, but it cannot be put to the "validator" that we have now, since it is reused for both cases: prepared and unprepared objects, so a header without checksums, etc is not a full header

pkg/services/replicator/process.go
```diff
@@ -21,6 +22,14 @@ type TaskResult interface {
 // HandleTask executes replication task inside invoking goroutine.
 // Passes all the nodes that accepted the replication to the TaskResult.
 func (p *Replicator) HandleTask(ctx context.Context, task Task, res TaskResult) {
+	if task.obj != nil && task.obj.HeaderLen() > object.MaxHeaderLen {
+		p.log.Warn("replication task with too big header",
+			zap.Int("header len", task.obj.HeaderLen()),
```
cthulhu-rider, line 27: no need to recalc the len

pkg/services/object/acl/v2/service.go
cthulhu-rider: agree, we no longer depend on replicate-ness, so better to move to fmt space

> unprepared objects, so a header without checksums, etc is not a full header

unprepared header can only grow

carpawell: > unprepared header can only grow

if an unprepared header is 16k-1 len, a prepared one overflows for sure

cthulhu-rider: This fact is unrelated to the place for this condition

roman-khimov (approved):

roman-khimov:
Changelog

pkg/services/object/server.go
```diff
@@ -1676,6 +1676,12 @@ func (s *server) Replicate(_ context.Context, req *protoobject.ReplicateRequest)
 			Message: "missing header field in the object field",
 		}}, nil
 	}
+	if hdrLen := hdr.MarshaledSize(); hdrLen > object.MaxHeaderLen {
```
cthulhu-rider, line 1679: `VerifyAndStoreObjectLocally` is doing this now

cthulhu-rider (approved):

PR #3302, opened 2025-04-15
cli: fix timeout can expire before RPC

### LINKED ISSUE #3279: neofs-cli starts keeping track of time before RPC is done
neofs-cli has a timeout for almost every operation, it is supposed to be a remote execution timeout. However timer starts ticking as the first operation in the cli: https://github.com/nspcc-dev/neofs-node/blob/69ee0fec4f94402198fd8ff408a18634ba2ab8d5/cmd/neofs-cli/modules/object/get.go#L46 makes reading password, files from FS, etc part of the execution. As a result, RPC may start with already expired context, exiting cli with a confusing message (it has not even tried to make RPC): `can't create API client: can't init SDK client: write request: open gRPC connection: gRPC dial: context deadline exceeded`

## Expected Behavior
Timeout is for operation being transferred to remote server.

## Current Behavior
Timeout may expire even if a user is only preparing cli for making RPC.

## Possible Solution
Read state first, init context with deadline then.

Do not provide timeout flag, do not enter password for 15+ seconds, press enter, see message (and do not see if enter is pressed earlier that 15s).

End-rey:
Closes #3279.

### DIFF
--- a/cmd/neofs-cli/modules/accounting/balance.go
+++ b/cmd/neofs-cli/modules/accounting/balance.go
@@ -26,8 +26,6 @@ var accountingBalanceCmd = &cobra.Command{
 	Long:  `Get internal balance of NeoFS account`,
 	Args:  cobra.NoArgs,
 	RunE: func(cmd *cobra.Command, _ []string) error {
-		ctx := context.Background()
-
 		var idUser user.ID
 
 		pk, err := key.GetOrGenerate(cmd)
@@ -42,6 +40,7 @@ var accountingBalanceCmd = &cobra.Command{
 			return fmt.Errorf("can't decode owner ID wallet address: %w", idUser.DecodeString(balanceOwner))
 		}
 
+		ctx := context.Background()
 		cli, err := internalclient.GetSDKClientByFlag(ctx, commonflags.RPC)
 		if err != nil {
 			return err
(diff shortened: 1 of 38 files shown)


cmd/neofs-cli/modules/accounting/balance.go
```diff
@@ -42,6 +40,7 @@ var accountingBalanceCmd = &cobra.Command{
 			return fmt.Errorf("can't decode owner ID wallet address: %w", idUser.DecodeString(balanceOwner))
 		}
 
+		ctx := context.Background()
```
roman-khimov, line 43: No timeout for this command?

End-rey: It wasn't there initially. Add it?

roman-khimov: Yeah, it's an additional fix in its essence, but it'd be nice to do while we're here.

cmd/neofs-cli/modules/accounting/balance.go
```diff
@@ -40,7 +39,9 @@ var accountingBalanceCmd = &cobra.Command{
 			return fmt.Errorf("can't decode owner ID wallet address: %w", idUser.DecodeString(balanceOwner))
 		}
 
-		ctx := context.Background()
+		ctx, cancel := commonflags.GetCommandContext(cmd)
```
roman-khimov, line 42: Needs to be mentioned in changelog as well.

roman-khimov (commented):
OK otherwise

PR #3306, opened 2025-04-16
peapod-to-fstree: ignore corrupt objects without container

End-rey:
I don't know how to check it, but in theory it should work.

### DIFF
--- a/cmd/peapod-to-fstree/main.go
+++ b/cmd/peapod-to-fstree/main.go
@@ -46,7 +46,14 @@ func (s *onlyObjectsWithContainersStorage) Iterate(objHandler func(oid.Address, 
 		}
 
 		return objHandler(addr, data, id)
-	}, nil)
+	}, func(addr oid.Address, err error) error {
+		_, found := s.containers[addr.Container()]
+		if !found {
+			log.Printf("skipping corrupted object without container '%s': %v\n", addr, err)
+			return nil
+		}
+		return err
+	})
 }
 
 func main() {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ Changelog for NeoFS Node
 - neofs-adm uses alphabet0/1/2 names for wallet files by default now (#3268)
 - Distribute load in writecache during flushing (#3261)
 - Faster migration in peapod-to-fstree (#3280)
-- `peapod-to-fstree` migrates only objects with existing containers (#3283)
+- `peapod-to-fstree` migrates only objects with existing containers (#3283, #3306)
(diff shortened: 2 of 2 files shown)


cmd/peapod-to-fstree/main.go
```diff
@@ -46,7 +46,13 @@ func (s *onlyObjectsWithContainersStorage) Iterate(objHandler func(oid.Address,
 		}
 
 		return objHandler(addr, data, id)
-	}, nil)
+	}, func(addr oid.Address, err error) error {
+		_, found := s.containers[addr.Container()]
+		if !found {
```
roman-khimov, line 51: Still log the event, please.

cmd/peapod-to-fstree/main.go
```diff
@@ -46,7 +46,14 @@ func (s *onlyObjectsWithContainersStorage) Iterate(objHandler func(oid.Address,
...
+			log.Printf("skipping object without container '%s': %v", addr, err)
```
carpawell, line 52:
```suggestion
			log.Printf("skipping corrupted object without container '%s': %v", addr, err)
```
we skip many-many objects, not all of them are logged

cmd/peapod-to-fstree/main.go
carpawell: it will be the object we will stop at, if there is `return err` below, it will be pronted anyway

carpawell: it's another matter if you meant that you shouldn't stop...

carpawell: `errorHandler` that return non-nil `err` (current PR's state) stop iteration

roman-khimov (approved):

cmd/peapod-to-fstree/main.go
```diff
@@ -46,7 +46,14 @@ func (s *onlyObjectsWithContainersStorage) Iterate(objHandler func(oid.Address,
...
 
 		return objHandler(addr, data, id)
-	}, nil)
+	}, func(addr oid.Address, err error) error {
+		_, found := s.containers[addr.Container()]
+		if !found {
+			log.Printf("skipping corrupted object without container '%s': %v", addr, err)
+			return nil
+		}
+		return err
```
cthulhu-rider, line 55: shouldnt we return nil to continue storage traverse?

cthulhu-rider: seems like no https://github.com/nspcc-dev/neofs-node/pull/3306#discussion_r2046881011

cthulhu-rider (approved):

PR #3309, opened 2025-04-17
fix/meta GC test

carpawell:


carpawell:
It really seems like the only problem that cannot be checked on my machine, but that reproduces often on CI.

roman-khimov (approved):

PR #3310, opened 2025-04-17
Drop storage ID

roman-khimov:


pkg/local_object_storage/blobstor/get.go
```diff
@@ -12,47 +12,35 @@ import (
...
 	b.modeMtx.RLock()
 	defer b.modeMtx.RUnlock()
 
-	if storageID == nil {
-		for i := range b.storage {
-			obj, err := b.storage[i].Storage.Get(addr)
-			if err == nil || !errors.As(err, new(apistatus.ObjectNotFound)) {
-				return obj, err
-			}
+	for i := range b.storage {
```
carpawell, line 19: why do you want to merge it now? it does some functional changes. once peapod is dropped, this PR would do more correct changes, and there is no magic `for` over every storage we have. this PR moves master to some strange state, i do not agree with it

roman-khimov: What are the functional changes here? Yes, substorages are iterated over, but it's the same if we don't pass any storage ID. And practically we have one substorage per shard everywhere except for three machines (and those to be converted soon).

This cuts one of the problematic things from the codebase, then we can move on to the next one, not trying to solve everything in a single PR.

carpawell (approved):

cthulhu-rider (approved):

PR #3312, opened 2025-04-18
services/object: use SearchV2 for search objects

### LINKED ISSUE #3143: Use SearchV2 for internal SN needs
afaik the only place for crossnode search is a tombstone verification https://github.com/nspcc-dev/neofs-node/blob/2c1f717dab7d9dda1cbadc2160d3c4c74fdd4ee6/pkg/services/object/tombstone/verify.go#L26

End-rey:
Closes #3143.

I hope I understood correctly what exactly needs to be changed.

### DIFF
--- a/cmd/neofs-node/object.go
+++ b/cmd/neofs-node/object.go
@@ -40,8 +40,10 @@ import (
 	objectSDK "github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	protoobject "github.com/nspcc-dev/neofs-sdk-go/proto/object"
+	protosession "github.com/nspcc-dev/neofs-sdk-go/proto/session"
 	apireputation "github.com/nspcc-dev/neofs-sdk-go/reputation"
 	"github.com/nspcc-dev/neofs-sdk-go/user"
+	"github.com/nspcc-dev/neofs-sdk-go/version"
 	"go.uber.org/zap"
 )
 
@@ -244,6 +246,7 @@ func initObjectService(c *cfg) {
 	mNumber, err := c.shared.basics.cli.MagicNumber()
 	fatalOnErr(err)
 
+	os := &objectSource{get: sGet}
 	sPut := putsvc.NewService(&transport{clients: putConstructor}, c,
 		putsvc.WithNetworkMagic(mNumber),
 		putsvc.WithKeyStorage(keyStorage),
@@ -257,7 +260,7 @@ func initObjectService(c *cfg) {
 		putsvc.WithWorkerPools(c.cfgObject.pool.putRemote, c.cfgObject.pool.putLocal),
 		putsvc.WithLogger(c.log),
 		putsvc.WithSplitChainVerifier(split.NewVerifier(sGet)),
-		putsvc.WithTombstoneVerifier(tombstone.NewVerifier(objectSource{sGet, sSearch})),
+		putsvc.WithTombstoneVerifier(tombstone.NewVerifier(os)),
 	)
 
 	sDelete := deletesvc.New(
@@ -314,6 +317,7 @@ func initObjectService(c *cfg) {
 		keys:    keyStorage,
 	}
 	server := objectService.New(objSvc, mNumber, fsChain, storage, c.metaService, c.shared.basics.key.PrivateKey, c.metricsCollector, aclChecker, aclSvc, coreConstructor)
+	os.server = server
 
 	for _, srv := range c.cfgGRPC.servers {
 		protoobject.RegisterObjectServiceServer(srv, server)
@@ -639,16 +643,9 @@ func (x storageForObjectService) GetSessionPrivateKey(usr user.ID, uid uuid.UUID
 
 type objectSource struct {
 	get    *getsvc.Service
-	search *searchsvc.Service
-}
-
-type searchWriter struct {
-	ids []oid.ID
-}
-
-func (w *searchWriter) WriteIDs(ids []oid.ID) error {
-	w.ids = append(w.ids, ids...)
-	return nil
+	server interface {
+		ProcessSearch(ctx context.Context, req *protoobject.SearchV2Request) ([]client.SearchResultItem, []byte, error)
+	}
 }
 
 func (o objectSource) Head(ctx context.Context, addr oid.Address) (*objectSDK.Object, error) {
@@ -664,20 +661,30 @@ func (o objectSource) Head(ctx context.Context, addr oid.Address) (*objectSDK.Ob
 	return hw.h, err
 }
 
-func (o objectSource) Search(ctx context.Context, cnr cid.ID, filters objectSDK.SearchFilters) ([]oid.ID, error) {
-	var sw searchWriter
-
-	var sPrm searchsvc.Prm
-	sPrm.SetWriter(&sw)
-	sPrm.WithSearchFilters(filters)
-	sPrm.WithContainerID(cnr)
-
-	err := o.search.Search(ctx, sPrm)
+func (o objectSource) SearchOne(ctx context.Context, cnr cid.ID, filters objectSDK.SearchFilters) (oid.ID, error) {
+	var id oid.ID
+	res, _, err := o.server.ProcessSearch(ctx, &protoobject.SearchV2Request{
+		Body: &protoobject.SearchV2Request_Body{
+			ContainerId: cnr.ProtoMessage(),
+			Version:     1,
+			Filters:     filters.ProtoMessage(),
+			Cursor:      "",
+			Count:       1,
+			Attributes:  nil,
+		},
+		MetaHeader: &protosession.RequestMetaHeader{
+			Version: version.Current().ProtoMessage(),
+			Ttl:     2,
+		},
+	})
 	if err != nil {
-		return nil, err
+		return id, err
 	}
 
-	return sw.ids, nil
+	if len(res) == 1 {
+		return res[0].ID, nil
+	}
+	return id, nil
 }
 
 // IsLocalNodePublicKey checks whether given binary-encoded public key is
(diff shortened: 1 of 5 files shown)


roman-khimov (commented):
It still does too much to me. This affects Search RPC operation, while the intention was to convert internal SN users to proper SearchV2. We'll need to do this anyway once we're to drop the old Search, so better isolate SN calls (not related to Search processing) and deal only with them.

End-rey:
@roman-khimov Is that what you need? Am I in the right direction?

cmd/neofs-node/object.go
```diff
@@ -665,19 +681,73 @@ func (o objectSource) Head(ctx context.Context, addr oid.Address) (*objectSDK.Ob
 }
 
 func (o objectSource) Search(ctx context.Context, cnr cid.ID, filters objectSDK.SearchFilters) ([]oid.ID, error) {
-	var sw searchWriter
+	mProcessedNodes := make(map[string]struct{})
+	var wg sync.WaitGroup
+	var mu sync.Mutex
+
+	result := make([]oid.ID, 0)
+	err := o.containers.ForEachRemoteContainerNode(cnr, func(node netmapsdk.NodeInfo) {
```
roman-khimov, line 689: Not sure why you have to do it this way, it should be a regular SearchV2 call just like it was with the old Search call.

cmd/neofs-node/object.go
```diff
@@ -665,19 +681,73 @@ func (o objectSource) Head(ctx context.Context, addr oid.Address) (*objectSDK.Ob
 }
 
 func (o objectSource) Search(ctx context.Context, cnr cid.ID, filters objectSDK.SearchFilters) ([]oid.ID, error) {
```
roman-khimov, line 683: This API can be reworked to be sv2-compatible as well. It is important to have it being used properly. For example, tombstone verifier for v2 doesn't need all oids, it can set lower limit. For v1 it can avoid subsequent link object HEAD by requesting payload size (see #3324, it's not as reliable as children check, but it's good enough for old v1 objects).

pkg/services/object/tombstone/verify.go
```diff
@@ -130,28 +132,26 @@ func (v *Verifier) verifyMember(ctx context.Context, cnr cid.ID, member oid.ID)
 
 func (v *Verifier) verifyV1Child(ctx context.Context, cnr cid.ID, sID object.SplitID) error {
 	filters := object.SearchFilters{}
+	filters.AddPayloadSizeFilter(object.MatchNumGT, 0)
```
roman-khimov, line 135: 1. The first filter should be the most restrictive one, in this case it'd be split ID filter.
2. `> 0` is wrong, we need `== 0` and in fact we can limit the count to 1 as well then, v1 link objects have no payload and we want link objects here.

pkg/services/object/tombstone/verify.go
```diff
@@ -174,7 +174,7 @@ func (v *Verifier) verifyV2Child(ctx context.Context, cnr cid.ID, firstObject oi
 		// no link object, child can be deleted
 		return nil
 	case 1:
-		return fmt.Errorf("found link object %s", ids[0])
+		return fmt.Errorf("found link object %s", ids[0].ID)
 	default:
```
roman-khimov, line 178: You can only have 0 or 1 result now.

pkg/services/object/server.go
```diff
@@ -192,7 +192,7 @@ type server struct {
 }
 
 // New provides protoobject.ObjectServiceServer for the given parameters.
-func New(hs Handlers, magicNumber uint32, fsChain FSChain, st Storage, metaSvc *metasvc.Meta, signer ecdsa.PrivateKey, m MetricCollector, ac aclsvc.ACLChecker, rp ACLInfoExtractor, cs searchsvc.ClientConstructor) protoobject.ObjectServiceServer {
+func New(hs Handlers, magicNumber uint32, fsChain FSChain, st Storage, metaSvc *metasvc.Meta, signer ecdsa.PrivateKey, m MetricCollector, ac aclsvc.ACLChecker, rp ACLInfoExtractor, cs searchsvc.ClientConstructor) *server {
```
roman-khimov, line 195: Public API can't return private types.

pkg/services/object/server.go
```diff
@@ -1928,18 +1935,34 @@ func (s *server) processSearchRequest(ctx context.Context, req *protoobject.Sear
 	default:
 	}
 
+	ttl := req.MetaHeader.GetTtl()
+	if ttl == 0 {
```
roman-khimov, line 1939: This (and filter management) should be handled before any network interactions (like container getter above).

cmd/neofs-node/object.go
```diff
@@ -664,20 +661,34 @@ func (o objectSource) Head(ctx context.Context, addr oid.Address) (*objectSDK.Ob
...
 }
 
-func (o objectSource) Search(ctx context.Context, cnr cid.ID, filters objectSDK.SearchFilters) ([]oid.ID, error) {
-	var sw searchWriter
-
-	var sPrm searchsvc.Prm
-	sPrm.SetWriter(&sw)
-	sPrm.WithSearchFilters(filters)
-	sPrm.WithContainerID(cnr)
+func (o objectSource) Search(ctx context.Context, cnr cid.ID, filters objectSDK.SearchFilters, childV2 bool) ([]client.SearchResultItem, error) {
```
roman-khimov, line 664: This is not a generic search. It fits the purpose, but it should be named a bit differently then.

cmd/neofs-node/object.go
```diff
@@ -664,20 +661,34 @@ func (o objectSource) Head(ctx context.Context, addr oid.Address) (*objectSDK.Ob
...
+	var attrs []string
+	count := uint32(1)
+	if !childV2 {
+		count = 1000
+		attrs = []string{objectSDK.FilterPayloadSize}
+	}
 
-	err := o.search.Search(ctx, sPrm)
+	req := &protoobject.SearchV2Request{
```
roman-khimov, line 672: Can this wrapping/unwrapping be avoided?

cmd/neofs-node/object.go
```diff
@@ -314,6 +317,7 @@ func initObjectService(c *cfg) {
 		keys:    keyStorage,
 	}
 	server := objectService.New(objSvc, mNumber, fsChain, storage, c.metaService, c.shared.basics.key.PrivateKey, c.metricsCollector, aclChecker, aclSvc, coreConstructor)
+	os.server = server
```
roman-khimov, line 320: Seems racy, `os` is already given to the `tombstone.NewVerifier()`. Not sure there are any threads involved though, maybe not a problem.

pkg/services/object/tombstone/verify.go
```diff
@@ -131,31 +131,17 @@ func (v *Verifier) verifyMember(ctx context.Context, cnr cid.ID, member oid.ID)
 func (v *Verifier) verifyV1Child(ctx context.Context, cnr cid.ID, sID object.SplitID) error {
 	filters := object.SearchFilters{}
 	filters.AddSplitIDFilter(object.MatchStringEqual, sID)
+	filters.AddPayloadSizeFilter(object.MatchNumGE, 0)
+	filters.AddPayloadSizeFilter(object.MatchNumLE, 0)
```
roman-khimov, line 135: Why not `EQ 0`?

pkg/services/object/tombstone/verify.go
```diff
@@ -131,31 +131,17 @@ func (v *Verifier) verifyMember(ctx context.Context, cnr cid.ID, member oid.ID)
...
-			return fmt.Errorf("heading %s object that was searched: %w", addr, err)
-		}
-
-		if len(header.Children()) != 0 {
-			return fmt.Errorf("found link object %s", addr)
-		}
+	if len(ids) == 1 {
+		return fmt.Errorf("found link object %s", ids[0])
 	}
-
```
roman-khimov, line 158: Useless change.

pkg/services/object/server.go
```diff
@@ -177,7 +177,7 @@ const (
 	maxObjAddrRespAmount = maxRespDataChunkSize / addrMsgSize // each address is about 72 bytes
 )
 
-type server struct {
+type Server struct {
```
roman-khimov, line 180: Needs a comment if exported.

cmd/neofs-node/object.go
```diff
@@ -664,20 +661,30 @@ func (o objectSource) Head(ctx context.Context, addr oid.Address) (*objectSDK.Ob
...
-func (o objectSource) Search(ctx context.Context, cnr cid.ID, filters objectSDK.SearchFilters) ([]oid.ID, error) {
-	var sw searchWriter
-
-	var sPrm searchsvc.Prm
-	sPrm.SetWriter(&sw)
-	sPrm.WithSearchFilters(filters)
-	sPrm.WithContainerID(cnr)
-
-	err := o.search.Search(ctx, sPrm)
+func (o objectSource) SearchOne(ctx context.Context, cnr cid.ID, filters objectSDK.SearchFilters) ([]oid.ID, error) {
```
roman-khimov, line 664: If it's `SearchOne` it can just return `oid.ID` and caller can check for `.IsZero()`.

pkg/services/object/tombstone/verify.go
End-rey: https://github.com/nspcc-dev/neofs-sdk-go/blob/1af995ff59815aa59aba913d8481523b0e71de64/object/search.go#L27-L37

There is no such comparison.

roman-khimov: https://github.com/nspcc-dev/neofs-sdk-go/blob/1af995ff59815aa59aba913d8481523b0e71de64/object/search.go#L29

It's the same effectively.

roman-khimov (approved):
Changelog, please.

cthulhu-rider (approved):

PR #3168, opened 2025-02-21
Zero byte in object attributes

cthulhu-rider:
w8 4 @roman-khimov to bring changes into API

PR #3316, opened 2025-04-21
Select via search

roman-khimov:
Migrations to be added.


pkg/local_object_storage/metabase/version.go
```diff
@@ -278,3 +294,57 @@ func migrateFrom4Version(db *DB) error {
...
+
+		if hdr.HasParent() {
+			queue = append(queue, id)
+			if len(queue) == int(limit) {
+				break
+			}
+		}
+	}
+	for _, id := range queue {
+		_ = b.Delete(slices.Concat(attrIDPrefix, id[:]))
```
cthulhu-rider, line 346: why u decided to ignore error?

pkg/local_object_storage/metabase/exists.go
```diff
@@ -91,21 +92,32 @@ func objectStatus(tx *bbolt.Tx, addr oid.Address, currEpoch uint64) uint8 {
...
-		if prevEpochBkt != nil {
-			rawOID := objectKey(oID, make([]byte, objectKeySize))
-			if prevEpochBkt.Get(rawOID) != nil {
-				expired = true
-			}
+	var (
+		expired        bool
+		oID            = addr.Object()
+		cID            = addr.Container()
+		metaBucketName = append([]byte{metadataPrefix}, cID[:]...)
```
cthulhu-rider, line 99: `metaBucketKey(cID)`. And can be inlined

pkg/local_object_storage/metabase/exists.go
```diff
@@ -91,21 +92,32 @@ func objectStatus(tx *bbolt.Tx, addr oid.Address, currEpoch uint64) uint8 {
...
+			cur       = metaBucket.Cursor()
+			expPrefix = make([]byte, attrIDFixedLen+len(objectSDK.AttributeExpirationEpoch))
+		)
+		expPrefix[0] = metaPrefixIDAttr
+		copy(expPrefix[1:], oID[:])
+		copy(expPrefix[1+len(oID):], objectSDK.AttributeExpirationEpoch)
+
+		expKey, _ := cur.Seek(expPrefix)
+		if bytes.HasPrefix(expKey, expPrefix) {
+			var val = expKey[len(expPrefix):]
```
cthulhu-rider, line 114: `+attributeDelimiterLen`?

pkg/local_object_storage/metabase/exists.go
```diff
@@ -91,21 +92,32 @@ func objectStatus(tx *bbolt.Tx, addr oid.Address, currEpoch uint64) uint8 {
...
+
+			objExpiration, err := strconv.ParseUint(string(val), 10, 64)
+			// Technically it should be currEpoch > objExpiration
```
cthulhu-rider, line 117: mentioned condition is rly expected. Do we plan to change it?

pkg/local_object_storage/metabase/version.go
```diff
@@ -275,9 +275,14 @@ func migrateFrom4Version(db *DB) error {
...
-		var buckets [][]byte
+		var (
+			buckets          [][]byte
+			obsoletePrefixes = []byte{unusedSmallPrefix, unusedOwnerPrefix,
+				unusedUserAttributePrefix, unusedPayloadHashPrefix,
+				unusedSplitPrefix, unusedFirstObjectIDPrefix}
+		)
 		err := tx.ForEach(func(name []byte, _ *bbolt.Bucket) error {
-			if len(name) > 0 && name[0] == unusedSmallPrefix {
+			if bytes.IndexByte(obsoletePrefixes, name[0]) != -1 {
```
cthulhu-rider, line 285: `slices.Contains` is more natural here to me, although this is correct

pkg/local_object_storage/metabase/select.go
```diff
@@ -51,636 +33,35 @@ func (db *DB) Select(cnr cid.ID, filters object.SearchFilters) ([]oid.Address, e
...
-		default:
-			continue // ignore unknown search attributes
-		}
-
-		if !matchFunc.matchSlow(f[i].Header(), data, f[i].Value()) {
-			return false
+		if len(newCursor) == 0 {
+			break
 		}
+		cursor = string(newCursor)
```
cthulhu-rider, line 61: shouldnt it be Base64-encoded? Afaik `PreprocessSearchQuery` expects it in API format

pkg/local_object_storage/metabase/exists.go
roman-khimov: It's already included (`attrIDFixedLen`), just not copied since it's zero.

pkg/local_object_storage/metabase/exists.go
roman-khimov: Can be an additional issue, but it really fails this way.

pkg/local_object_storage/metabase/version.go
cthulhu-rider: it does not state this is the only possible error. Althought it is unexpected. Func returns an error, nothing breaks us to handle the case

pkg/local_object_storage/metabase/exists.go
cthulhu-rider: tricky for me, it's worth leaving a comment at least

pkg/local_object_storage/metabase/exists.go
roman-khimov: #3321.

pkg/core/object/metadata.go
```diff
@@ -619,16 +618,16 @@ func intMatches(dbVal *big.Int, matcher object.SearchMatchType, fltVal *big.Int)
 // along and continuation cursor, verifies the cursor and returns additional
 // arguments to pass into [DB.Search]. If the query is valid but unreachable,
 // [ErrUnreachableQuery] is returned.
-func PreprocessSearchQuery(fs object.SearchFilters, attrs []string, cursor string) (*SearchCursor, map[int]ParsedIntFilter, error) {
+func PreprocessSearchQuery(fs object.SearchFilters, attrs []string, cursor string) ([]SearchFilter, *SearchCursor, error) {
```
carpawell, line 621: reason for changing the order?

pkg/local_object_storage/metabase/version.go
carpawell: still do not log? even though added logger as an arg

pkg/local_object_storage/metabase/exists.go
carpawell: excuse me?.. we have a separate var for it, use it everywhere, and you decided not to use it only here because it is the same as the default value for a byte? just imagine we decide to change it to smth else once in the future and this code is still the same

cthulhu-rider: see comment now, it's enough

carpawell: > It can't realistically be changed to me

https://github.com/nspcc-dev/neofs-node/pull/3168? Am I right that the argument is like "I think we will never change code/make a mistake"?

carpawell: > see comment now, it's enough

when we change the delimiter, we change const's value, we do not read every comment in every metabase's file

roman-khimov: It _can not_ be changed. Any other value breaks the basic properties of these indexes. That's exactly why #3168 fixed it from the value that just doesn't really work (but was brought in because 0 is a valid UTF-8).

carpawell (approved):

PR #3319, opened 2025-04-22
cli: Add command accepting container creation request in JSON format

cthulhu-rider:


cmd/neofs-cli/modules/request/container.go
```diff
@@ -0,0 +1,239 @@
...
+		} else if !errors.Is(err, apistatus.ErrContainerNotFound) {
+			return fmt.Errorf("status failure: %w", err)
+		}
+
+		const maxAttempts = 3
+		if attempt == maxAttempts {
+			break
+		}
+
+		const pollInterval = 5 * time.Second
```
carpawell, line 233: i think this is mainly  test command, maybe more often that 5 sec?

cmd/neofs-cli/modules/request/container.go
```diff
@@ -0,0 +1,239 @@
...
+		cmd.Println("Signatures are valid. Checking status...")
+
+		if err := apistatus.ToError(resp.GetMetaHeader().GetStatus()); err == nil {
+			cmd.Println("Status OK. Operation succeeded.")
+			return nil
+		} else if !errors.Is(err, apistatus.ErrContainerNotFound) {
+			return fmt.Errorf("status failure: %w", err)
+		}
+
+		const maxAttempts = 3
```
carpawell, line 228: try based on user inputs? like timeout in our every command?

cthulhu-rider: it's 10s now. In practice, container is either created fast, or never

i tend to think that good tools work without explicitly specified timeout in 99% cases cuz timeout values they are often too opaque

cmd/neofs-cli/modules/request/container.go
```diff
@@ -0,0 +1,239 @@
...
+	bodyDec.DisallowUnknownFields()
+	var body createContainerBody
+	if err := bodyDec.Decode(&body); err != nil {
+		return fmt.Errorf("failed to decode request body JSON: %w", err)
+	}
+	if body.Container == nil {
+		return fmt.Errorf("invalid request body JSON: missing container")
+	}
+	if body.Container.Version == nil {
+		return fmt.Errorf("invalid request body JSON: invalid container: missing nonce")
```
roman-khimov, line 82: s/nonce/version/

cthulhu-rider: fixed

roman-khimov:
Conflicts

PR #3324, opened 2025-04-24
Drop a bit more metabase buckets

roman-khimov:


pkg/local_object_storage/writecache/flush_test.go
```diff
@@ -210,6 +210,7 @@ func TestFlush(t *testing.T) {
 }
 
 func TestFlushPerformance(t *testing.T) {
+	t.Skip()
```
carpawell, line 213: issue/fix here?

pkg/local_object_storage/metabase/get.go
```diff
@@ -183,93 +182,19 @@ func getSplitInfoError(tx *bbolt.Tx, cnr cid.ID, key []byte) error {
...
-	}
-
-	var oID oid.ID
-	var err error
-
-	err = bkt.ForEach(func(k, _ []byte) error {
-		err = oID.Decode(k)
+	var cur = metaBkt.Cursor()
+	k, _ := cur.Seek([]byte{metaPrefixID})
+	for ; len(k) > 0 && len(unique) < limit && k[0] == metaPrefixID; k, _ = cur.Next() {
```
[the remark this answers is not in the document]
roman-khimov: Yeah, I had this thought as well, but decided to keep it compatible initially. Can be improved.

pkg/local_object_storage/metabase/get.go
```diff
@@ -108,32 +109,48 @@ func getFromBucket(tx *bbolt.Tx, name, key []byte) []byte {
...
+	)
+	if metaBucket == nil {
+		return childOID
 	}
 
-	relativeLst, err := decodeList(parentBucket.Get(key))
-	if err != nil {
-		return nil, err
+	var (
+		parentPrefix = make([]byte, 1+len(objectSDK.FilterParentID)+1+len(parentID)+1)
```
cthulhu-rider: i'd use `attributeDelimiterLen` instead of some `+1`

pkg/local_object_storage/metabase/get.go
```diff
@@ -108,32 +109,48 @@ func getFromBucket(tx *bbolt.Tx, name, key []byte) []byte {
...
+		cur          = metaBucket.Cursor()
+	)
+	parentPrefix[0] = metaPrefixAttrIDPlain
+	off := 1 + copy(parentPrefix[1:], objectSDK.FilterParentID)
+	off += copy(parentPrefix[off:], objectcore.MetaAttributeDelimiter)
+	copy(parentPrefix[off:], parentID[:])
```
cthulhu-rider, line 124: one more delim? Although resulting value is correct. `slices.Concat` as an option

pkg/local_object_storage/metabase/status.go
```diff
@@ -64,7 +64,8 @@ func (db *DB) ObjectStatus(address oid.Address) (ObjectStatus, error) {
 
 		removedStatus := inGraveyardWithKey(addrKey, graveyardBkt, garbageObjectsBkt, garbageContainersBkt)
 
-		if removedStatus != 0 && objectLocked(tx, cID, oID) || inBucket(tx, primaryBucketName(cID, key), objKey) || inBucket(tx, parentBucketName(cID, key), objKey) {
+		childForParent := getChildForParent(tx, cID, oID, key)
```
cthulhu-rider, line 67: i'd inline, it may be not needed

pkg/local_object_storage/metabase/get.go
```diff
@@ -183,93 +182,19 @@ func getSplitInfoError(tx *bbolt.Tx, cnr cid.ID, key []byte) error {
...
+	for ; len(k) > 0 && len(unique) < limit && k[0] == metaPrefixID; k, _ = cur.Next() {
+		obj, err := oid.DecodeBytes(k[1:])
 		if err != nil {
-			return fmt.Errorf("object ID parsing: %w", err)
-		}
-
-		resMap[oID] = struct{}{}
-		if len(resMap) == limit {
-			return errBreakIter
+			return fmt.Errorf("garbage prefixID key for container %s: %v", cID, k)
```
cthulhu-rider: garbage can be enormous, I would leave len only, i.e. incl. the `err`

pkg/local_object_storage/metabase/exists.go
```diff
@@ -191,19 +192,71 @@ func inBucket(tx *bbolt.Tx, name, key []byte) bool {
...
+				isV1 = true
+				splitInfo.SetSplitID(objectSDK.NewSplitIDFromV2(attrVal))
+			}
+			if string(attrKey) == objectSDK.FilterPayloadSize && string(attrVal) == "0" {
+				isEmpty = true
+			}
+			if string(attrKey) == objectSDK.FilterFirstSplitObject {
+				firstID, err := oid.DecodeBytes(attrVal)
+				if err != nil {
+					return nil, fmt.Errorf("invalid first ID attribute in %s/%s", cnr, objID)
```
cthulhu-rider, line 240: i'd include `err`

pkg/local_object_storage/metabase/exists.go
```diff
@@ -191,19 +192,71 @@ func inBucket(tx *bbolt.Tx, name, key []byte) bool {
...
+		// This is not perfect since the last part can have zero
+		// payload technically, but we don't have proper child info
+		// and it's practically good enough for v1 compatibility
+		// (v2 doesn't have this problem).
+		if isLink || (isV1 && isEmpty) {
+			splitInfo.SetLink(objID)
+		}
+		// We should have one or two IDs here, if one is link the
+		// other is not.
+		if (isV1 && !isEmpty) || (!isV1 && !isLink) {
```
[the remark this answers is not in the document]
roman-khimov: Middles have no parent header with id, only the right one and link do. Practically I have not found any case of empty right child on m1/m3 and m4 (partially, was interrupted), so it works and we don't expect any new v1 objects.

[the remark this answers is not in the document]
cthulhu-rider: > Middles have no parent header with id

no parent ID, right. But afair they all have split ID and some payload. Lookin at `isV1 && !isEmpty` condition

pkg/local_object_storage/metabase/get.go
roman-khimov: Some components are strings, some are bytes and some are not-really-slices, `slices.Concat` is not very helpful for the mix.

pkg/local_object_storage/metabase/status.go
roman-khimov: Currently it's used twice.

cthulhu-rider: i mean var, not func

roman-khimov: That condition (the next line) is complex enough already.

pkg/local_object_storage/metabase/exists.go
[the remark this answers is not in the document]
roman-khimov: Yes, but we're iterating for objects that do have parent ID here. Only two valid cases here: last one or link.

cthulhu-rider (approved):

pkg/local_object_storage/metabase/status.go
roman-khimov: This method is far from being optimal DB-wise (doube `objectLocked`, a lot of the same accesses in `readBuckets`), but it's only used for control which is not on the hot path.

PR #3331, opened 2025-04-28
node/object: Optimize filtered attribute-less SearchV2 queries

cthulhu-rider:


roman-khimov (commented):
Why not doing this for local queries only? It won't break anything and you could keep the same request with all signatures and things.

cthulhu-rider:
then I doubt SN merging results can build VAL_OID continuation cursor in general. It will be just OID. Continuation of such query is questionable to me. Mb we can emulate the cursor too, will see

cthulhu-rider:
> Although it can be problematic for objects in the middle of the list (we don't know the attribute value for them), hmm...

yep, that's the general challenge to me

in these terms, `EQ` filters are nice cuz all results have the same expected value

cthulhu-rider:
solvin conflicts...

roman-khimov (commented):
Looks legit, comment/SDK/API need to be updated. Check tests also.

roman-khimov (approved):

PR #3336, opened 2025-05-07
client/notary: check notary on a chain by roles

End-rey:


pkg/morph/client/notary.go
```diff
@@ -127,8 +126,11 @@ func (c *Client) ProbeNotary() (res bool) {
 		return false
 	}
 
-	_, err := conn.client.GetContractStateByAddressOrName(nativenames.Notary)
-	return err == nil
+	keys, err := conn.roleList(noderoles.P2PNotary)
+	if err != nil {
+		return false
+	}
+	return len(keys) > 0
```
roman-khimov, line 133: `return err == nil && len(keys) > 0`

roman-khimov (approved):

PR #3337, opened 2025-05-12
object/get: fix that object can't be got without a link object

### LINKED ISSUE #3275: Object can't be constructed without a corresponding link object
https://github.com/nspcc-dev/neofs-testcases/pull/1015 (test_object_can_be_get_without_link_object)

1. Upload a complex object
2. Get a corresponding link object
3. Remove the link object via `neofs-cli control drop-objects` from all SNs
4. Try to get the original complex object - expect success, got failure.

This is the flipping failure, locally always passes, in github - 50/50.

End-rey:
Closes #3275.

Is it a proper decision?

### DIFF
--- a/pkg/services/object/get/assembly_v2.go
+++ b/pkg/services/object/get/assembly_v2.go
@@ -9,6 +9,8 @@ import (
 	"go.uber.org/zap"
 )
 
+var errNoLinkNoLastPart = errors.New("no link and no last part in split info")
+
 func (exec *execCtx) processV2Split(si *objectSDK.SplitInfo) {
 	if si.GetFirstPart().IsZero() {
 		exec.log.Debug("no first ID found in V2 split")
@@ -26,6 +28,7 @@ func (exec *execCtx) processV2Split(si *objectSDK.SplitInfo) {
 	prev := si.GetLastPart()
 	if prev.IsZero() {
 		exec.log.Debug("neither link, not last part is set in the v2 split information")
+		exec.err = errNoLinkNoLastPart
 		return
 	}
 
--- a/pkg/services/object/get/get.go
+++ b/pkg/services/object/get/get.go
@@ -2,6 +2,7 @@ package getsvc
 
 import (
 	"context"
+	"errors"
 
 	"github.com/nspcc-dev/neofs-node/pkg/util"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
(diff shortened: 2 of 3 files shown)


pkg/services/object/get/get.go
```diff
@@ -110,6 +112,9 @@ func (exec *execCtx) analyzeStatus(execCnr bool) {
 	case statusVIRTUAL:
 		exec.log.Debug("requested object is virtual")
 		exec.assemble()
+		if errors.Is(exec.err, apistatus.ErrObjectNotFound) {
+			exec.analyzeStatus(execCnr)
```
roman-khimov, line 116: This can lead to infinite recursion, suppose you can't get some object's part -- you'd get the same "not found" error and will retry indefinitely.

pkg/services/object/get/assembly_v2.go
```diff
@@ -57,6 +57,8 @@ func (exec *execCtx) processV2Link(linkID oid.ID) bool {
 	linkObj, ok := exec.getChild(linkID, nil, true)
 	if !ok {
 		exec.log.Debug("failed to read link object")
+		exec.status = statusNotFound
+		exec.err = apistatus.ErrObjectNotFound
 		return false
```
roman-khimov, line 62: But why `processV2Split()` doesn't handle this `false` correctly already? It's supposed to handle link object problems, `true` is returned when the status from `processV2Link` is final and `false` is returned when we should fallback to chain processing.

End-rey:
I didn't find any other good way to get the last part, except through the `executeOnContainer`, because the part lies on one of the nodes. Is there any other way?

pkg/services/object/get/get.go
```diff
@@ -110,6 +111,11 @@ func (exec *execCtx) analyzeStatus(execCnr bool) {
 	case statusVIRTUAL:
 		exec.log.Debug("requested object is virtual")
 		exec.assemble()
+		if errors.Is(exec.err, errNoLinkNoLastPart) && execCnr {
+			exec.err = nil
```
cthulhu-rider, line 115: does this instruction affect anything?

roman-khimov: I think it can be relevant for the next execute/analyze cycle. In successful case `errNoLinkNoLastPart` can remain there.

carpawell: tbh, i would expect to keep this err as a local result and be overwritten by remote results if any

End-rey: Initially, it was supposed to save from endless execution, but now it's not so important.

carpawell: > In successful case errNoLinkNoLastPart can remain there.

every successful case should overwrite status and nils the error: https://github.com/nspcc-dev/neofs-node/blob/4ab3a9a5c8183d65c7a0086748b889f3ca32957b/pkg/services/object/get/remote.go#L37-L38

cthulhu-rider: i think we should not do this. `execCnr` operand prevent endless looping, while

> In successful case errNoLinkNoLastPart can remain there.

must never happen cuz this would be a problem to me, and this line must not hide it

cthulhu-rider (approved):

carpawell (approved):

PR #3341, opened 2025-05-13
configutil: fix ENV variable rewrite entire structure

End-rey:
Fix that new variable from the ENV that is part of an array can rewrite the entire structure, not just this variable.

https://github.com/nspcc-dev/neofs-testcases/issues/1051#issuecomment-2873675554

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@ Changelog for NeoFS Node
 - IR contract approval no longer supports neofsid accounts (#3256)
 - IR no longer processes bind/unbind neofsid requests (#3256)
 - SN `ExternalAddr` attribute is ignored now (#3235)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ Changelog for NeoFS Node
 - Unclosed network connections in CLI (#3326)
 - SN no longer fails tombstone verification on `ALREADY_REMOVED` member's status (#3327)
 - Panic in search (#3333)
+- ENV variable rewrite entire structure of config (#3341)
```
roman-khimov, line 33: I think this can be added into
> - Use list instead of maps for options in node config (#3204)

Because this problem didn't exist in 0.45.2.

roman-khimov (approved):

PR #3343, opened 2025-05-14
Fix/meta/uncontrollable number of subs

carpawell:


pkg/services/object/server.go
```diff
@@ -245,9 +245,10 @@ func (s *Server) sendStatusPutResponse(stream protoobject.ObjectService_PutServe
 }
 
 type putStream struct {
-	ctx    context.Context
-	signer ecdsa.PrivateKey
-	base   *putsvc.Streamer
+	ctx     context.Context
+	signer  ecdsa.PrivateKey
+	base    *putsvc.Streamer
+	metaSvc *metasvc.Meta
```
cthulhu-rider, line 251: seems unused

pkg/services/object/put/distributed.go
```diff
@@ -187,12 +190,34 @@ func (t *distributedTarget) Close() (oid.ID, error) {
...
+			if await {
+				t.metaSvc.UnsubscribeFromObject(addr)
+			}
+			return oid.ID{}, fmt.Errorf("failed to submit %s object meta information: %w", addr, err)
+		}
+
+		if await {
+			select {
+			case <-time.After(15 * time.Second):
+				t.metaSvc.UnsubscribeFromObject(addr)
```
cthulhu-rider, line 211: this is done in 2 cases, i think it'd be easier to just `defer` right after `NotifyObjectSuccess`. Correct though

pkg/services/object/put/distributed.go
```diff
@@ -187,12 +190,34 @@ func (t *distributedTarget) Close() (oid.ID, error) {
...
-			return oid.ID{}, fmt.Errorf("failed to submit %s object meta information: %w", id, err)
+			if await {
+				t.metaSvc.UnsubscribeFromObject(addr)
+			}
+			return oid.ID{}, fmt.Errorf("failed to submit %s object meta information: %w", addr, err)
+		}
+
+		if await {
+			select {
+			case <-time.After(15 * time.Second):
```
cthulhu-rider, line 210: also worth checking original request's context expiration

pkg/services/meta/meta.go
```diff
@@ -61,6 +61,48 @@ type wsClient interface {
...
+	delete(on.subs, addr)
+}
+
+func (on *objectNotifier) notifyRecieved(oid oid.Address) {
+	on.m.Lock()
+	defer on.m.Unlock()
+
+	ch, ok := on.subs[oid]
+	if ok {
+		select {
```
cthulhu-rider, line 98: can be just `close(ch)`

pkg/services/meta/notifications.go
```diff
@@ -626,3 +626,16 @@ func (m *Meta) handleEpochNotification(e uint64) error {
 
 	return nil
 }
+
+// NotifyObjectSuccess subscribes channel for object notification chain inclusion.
+// Channel must be read before subscription is made and writing to it must be
+// non-blocking.
+func (m *Meta) NotifyObjectSuccess(ch chan<- error, addr oid.Address) {
```
cthulhu-rider, line 633: error is never actually written to `ch`, i'd simplify to `chan<- struct{}`

pkg/services/meta/meta.go
```diff
@@ -61,6 +61,48 @@ type wsClient interface {
...
+}
+
+func (on *objectNotifier) unsubscribe(addr oid.Address) {
+	on.m.Lock()
+	defer on.m.Unlock()
+
+	delete(on.subs, addr)
+}
+
+func (on *objectNotifier) notifyRecieved(oid oid.Address) {
```
cthulhu-rider, line 92: `addr` is closer and avoids pkg name collision

carpawell: yeah, changed to addr in the last moment and have not fixed every place

pkg/services/meta/notifications.go
carpawell: i do expect it to be more expressive in the future (like be possible to use VUB and to timeout inside meta service too, if no tx will be added to the chain) but not for now

pkg/services/meta/meta.go
carpawell: ok for now

pkg/services/object/put/distributed.go
carpawell: well, i wanted to do so, but in a success case (most of the expected cases) it will be an additional sync lock for nothing

pkg/services/meta/notifications.go
cthulhu-rider: i'd extend when rly needed

cthulhu-rider (approved):
changelog?

pkg/services/meta/blocks.go
```diff
@@ -49,6 +49,8 @@ func (m *Meta) handleBlock(ctx context.Context, b *block.Header) error {
 				continue
 			}
 
+			m.notifier.notifyRecieved(oid.NewAddress(ev.cID, ev.oID))
```
roman-khimov, line 52: received

roman-khimov (approved):

PR #3348, opened 2025-05-14
go.mod: upgrade to the latest and greatest NeoGo

roman-khimov:
0.109.1 is out and we need it.


carpawell (approved):
`out and we need it` and no changes except `go.mod` is strange for commit message, imo

roman-khimov:
No API changes compared to 0.109.0, but an updated WSClient should be helpful. State also matters a lot and we have a correct one now for Echidna here.

carpawell:
I know-know, but for git's history, it says nothing useful.

PR #3355, opened 2025-05-16
blockchain: decrease default proto tick interval

AnnaShaleva:
Context: slow P2P sync on NeoFS IR nodes with internal consensus. This commit decreases default P2P proto tick interval which directly affects the frequency of ping/pong message handlers. A side effect is more frequent requests for missing blocks, and hence, faster P2P sync. Also, I'd say that previous value (2s) is too slow for 1-second networks.

Despite the fact that P2P synchronisation benefits from this commit, the real effect of this commit on the NeoFS performance must be evaluated, because this commit significantly increases the number of P2P messages in the network.


AnnaShaleva:
I'll make all appropriate changes (CHANGELOG update, etc.) if we decide to merge this PR.

AnnaShaleva:
This problem is not urgent, since there's no need to introduce this kludge. We'll wait for BlockFetcher-enabled IR version.

PR #3360, opened 2025-05-20
Drop `MaintenanceModeAllowed` network setting

### LINKED ISSUE #3328: Drop MaintenanceModeAllowed
I'm always frustrated when we have excessive configuration parameters.

Drop `MaintenanceModeAllowed`. Make it always allowed. Account for that during basic income distribution, nodes in maintenance mode receive nothing, but otherwise they're valid nodes.

End-rey:
Closes #3328.


pkg/innerring/processors/netmap/nodevalidation/state/validator.go
```diff
@@ -35,34 +21,30 @@ type NetworkSettings interface {
...
 
-// SetNetworkSettings specifies provider of the NetworkSettings interface.
-// MUST be called before any VerifyAndUpdate call. Parameter MUST NOT be nil.
-func (x *NetMapCandidateValidator) SetNetworkSettings(netSettings NetworkSettings) {
-	x.netSettings = netSettings
+// New creates a new instance of the NetMapCandidateValidator.
+//
+// The created NetMapCandidateValidator does not require additional
+// initialization and is completely ready for work.
+func New() *NetMapCandidateValidator {
```
roman-khimov, line 30: Why do we need this `New()`?

roman-khimov (commented):
Surprised to see a test failure here, but it seems to be unrelated.

pkg/innerring/processors/netmap/nodevalidation/state/validator.go
End-rey: I did this because other validators were also created in this way:
https://github.com/nspcc-dev/neofs-node/blob/56099822da4a8903deaf53aa26df334e08443f99/pkg/innerring/processors/netmap/nodevalidation/structure/validator.go#L15-L17

https://github.com/nspcc-dev/neofs-node/blob/d0d38e70b4ab30ccab0fda598adf6d89d9b48f10/pkg/innerring/processors/netmap/nodevalidation/locode/validator.go#L22-L24

roman-khimov:
Still a test failure here that is not observed on the master branch (or #3365 for a quick run), this needs to be checked.

End-rey:
@roman-khimov, the tests are not working because I removed the setting `MaintenanceModeAllowed` output from the command `neofs-cli netmap netinfo`. Should I return the output or fix the test?

roman-khimov:
OK, this means we need to adjust tests to ignore it. Let's wait for https://github.com/nspcc-dev/neofs-testcases/issues/1067.

roman-khimov:
BTW, one thing is missing here:
> Account for that during basic income distribution, nodes in maintenance mode receive nothing, but otherwise they're valid nodes

We need to check for state somewhere around `pkg/innerring/settlement.go` or `pkg/innerring/processors/settlement/basic/collect.go` where node lists are created.

End-rey:
@roman-khimov, is that what you meant?

carpawell (approved):

roman-khimov (approved):

roman-khimov:
Waiting for tests.

PR #3371, opened 2025-05-26
Drop `blobstor` package

### LINKED ISSUE #3358: Drop blobstor package
I'm always frustrated when I see the code we don't need in node.

`blobstor` package was supposed to handle substorages, but we don't have substorages since #3330/#2497. Which means that effectively it's all dead code, `shard` can use backing storage directly via some interface _it_ has (and not `blobstor/common`). We still need to keep extensibility here (other types of storages can be developed in future), so `shard` shouldn't depend on `fstree` directly.

Any common code like copying or tests should be kept and adapted for this as well (minimal interfaces that work for `fstree`).

Doubt we have many, code that does nothing complicates the node for no reason.

End-rey:
Closes #3358.

Do I need to change the section name in the SN config?

### DIFF
--- a/pkg/local_object_storage/writecache/flush.go
+++ b/pkg/local_object_storage/writecache/flush.go
@@ -6,10 +6,8 @@ import (
 	"sort"
 	"time"
 
-	objectCore "github.com/nspcc-dev/neofs-node/pkg/core/object"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor/common"
 	apistatus "github.com/nspcc-dev/neofs-sdk-go/client/status"
-	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"go.uber.org/zap"
 )
@@ -233,18 +231,15 @@ func (c *cache) flushSingle(addr oid.Address, ignoreErrors bool) error {
 	if c.metrics.mr != nil {
 		defer elapsed(c.metrics.AddWCFlushSingleDuration)()
 	}
-	obj, data, err := c.getObject(addr)
+	data, err := c.getObject(addr)
 	if err != nil {
 		if ignoreErrors {
 			return nil
 		}
 		return err
 	}
-	if obj == nil {
-		return nil
-	}
 
-	err = c.flushObject(obj, data)
+	err = c.flushObject(addr, data)
 	if err != nil {
 		return err
 	}
@@ -258,9 +253,7 @@ func (c *cache) flushSingle(addr oid.Address, ignoreErrors bool) error {
 }
 
 // flushObject is used to write object directly to the main storage.
-func (c *cache) flushObject(obj *object.Object, data []byte) error {
-	addr := objectCore.AddressOf(obj)
-
+func (c *cache) flushObject(addr oid.Address, data []byte) error {
 	err := c.storage.Put(addr, data)
 	if err != nil {
 		if !errors.Is(err, common.ErrNoSpace) && !errors.Is(err, common.ErrReadOnly) {
@@ -287,15 +280,12 @@ func (c *cache) flushBatch(addrs []oid.Address, ignoreErrors bool) error {
 
 	objs := make(map[oid.Address][]byte, len(addrs))
 	for _, addr := range addrs {
-		obj, data, err := c.getObject(addr)
+		data, err := c.getObject(addr)
 		if err != nil {
 			if ignoreErrors {
 				continue
 			}
 			return err
-		}
-		if obj == nil {
-			continue
 		}
 
 		objs[addr] = data
@@ -321,7 +311,7 @@ func (c *cache) flushBatch(addrs []oid.Address, ignoreErrors bool) error {
 	return err
 }
 
-func (c *cache) getObject(addr oid.Address) (*object.Object, []byte, error) {
+func (c *cache) getObject(addr oid.Address) ([]byte, error) {
 	sAddr := addr.EncodeToString()
 
 	data, err := c.fsTree.GetBytes(addr)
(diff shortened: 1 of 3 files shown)


roman-khimov:
> Do I need to change the section name in the SN config?

No.

pkg/local_object_storage/writecache/flush_test.go
```diff
@@ -424,6 +429,16 @@ func newObject(t *testing.T, size int) (*object.Object, []byte) {
 	return obj, obj.Marshal()
 }
 
+func storageSetMode(ss *storage.SubStorage, m mode.Mode) error {
```
roman-khimov, line 432: Should be no-op if mode doesn't change.

pkg/local_object_storage/shard/storage/storage.go
```diff
@@ -0,0 +1,110 @@
...
+	Policy  func(*objectSDK.Object, []byte) bool
+
+	Log *zap.Logger
+}
+
+// Exists checks if the object is presented in storage.
+//
+// Returns any error encountered that did not allow
+// to completely check object existence.
+func (ss *SubStorage) Exists(addr oid.Address) (bool, error) {
```
roman-khimov, line 25: What's the value? It's just `ss.Storage.Exists()`.

pkg/local_object_storage/shard/storage/storage.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// Put stores an object in a storage.
+func (ss *SubStorage) Put(addr oid.Address, obj *objectSDK.Object, raw []byte) error {
+	if raw == nil {
+		// marshal object
+		raw = obj.Marshal()
+	}
+
+	if ss.Policy == nil || ss.Policy(obj, raw) {
```
roman-khimov, line 41: No policies --- no problem.

pkg/local_object_storage/shard/storage/storage.go
```diff
@@ -0,0 +1,110 @@
...
+			storagelog.Write(ss.Log,
+				storagelog.AddressField(obj.Addr),
+				storagelog.OpField("PUT"),
+			)
+		}
+	}
+	return nil
+}
+
+func (ss *SubStorage) canStoreBatch(objs []PutBatchPrm) bool {
```
roman-khimov, line 100: Same here.

pkg/local_object_storage/shard/status.go
```diff
@@ -12,19 +11,33 @@ import (
...
+	Blob       StorageObjectStatus
 	Metabase   meta.ObjectStatus
 	Writecache writecache.ObjectStatus
 	Errors     []error
 }
 
+// StorageObjectStatus represents the status of the object in the storage,
+// containing the type and path of the storage and an error if it
+// occurred.
+type StorageObjectStatus struct {
```
roman-khimov, line 23: Can be `ObjectStatus` as well, `shard.ObjectStatus` doesn't seem alien.

pkg/local_object_storage/writecache/flush.go
```diff
@@ -287,7 +285,7 @@ func (c *cache) flushBatch(addrs []oid.Address, ignoreErrors bool) error {
 		defer elapsed(c.metrics.AddWCFlushBatchDuration)()
 	}
 
-	objs := make([]blobstor.PutBatchPrm, 0, len(addrs))
+	objs := make(map[oid.Address][]byte, len(addrs))
 	for _, addr := range addrs {
 		obj, data, err := c.getObject(addr)
```
roman-khimov, line 290: If we don't need `obj`, we can avoid decoding it as well. Just `GetBytes()` and go.

pkg/local_object_storage/blobstor/exists_test.go
```diff
@@ -1,84 +0,0 @@
...
-	"testing"
-
-	objectCore "github.com/nspcc-dev/neofs-node/pkg/core/object"
-	cidtest "github.com/nspcc-dev/neofs-sdk-go/container/id/test"
-	objectSDK "github.com/nspcc-dev/neofs-sdk-go/object"
-	oidtest "github.com/nspcc-dev/neofs-sdk-go/object/id/test"
-	"github.com/stretchr/testify/require"
-)
-
-func TestExists(t *testing.T) {
```
roman-khimov, line 15: Is it just deleted or there is a new test added somewhere?

pkg/local_object_storage/blobstor/generic_test.go
```diff
@@ -1,24 +0,0 @@
...
-import (
-	"os"
-	"path/filepath"
-	"strconv"
-	"testing"
-
-	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/internal/storagetest"
-)
-
-func TestGeneric(t *testing.T) {
```
roman-khimov, line 12: Not sure what happened to this one as well.

pkg/local_object_storage/shard/status.go
roman-khimov: OK then.

cmd/neofs-lens/internal/fstree/root.go
```diff
@@ -3,8 +3,8 @@ package fstree
 import (
 	"fmt"
 
-	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor/compression"
-	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor/fstree"
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/shard/compression"
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/shard/fstree"
```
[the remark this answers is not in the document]
carpawell: I don't mind dropping, but I wouldn't change for an easier way. also depends on our common naming convention: is "blobstor" as disgusting as "morph", "side chain", "blobovnicza", etc, for us?

PR #3373, opened 2025-05-30
Update SDK to the latest master

### LINKED ISSUE #3359: Object exists but inaccessible in Mainnet
in full-public Mainnet container 
```
$ neofs-cli container get --cid CeeroywT8ppGE4HGjhpzocJkdb2yu3wD5qCGFTjkw1Cc -r st1.storage.fs.neo.org:8080

container ID: CeeroywT8ppGE4HGjhpzocJkdb2yu3wD5qCGFTjkw1Cc
owner ID: NMecMz4YurGg2nVaFuu7hqP3MbXCdRon1c
basic ACL: fffffff
       RangeHASH    Range      Search     Delete     Put        Head       Get
0 0    1 1 1 1      1 1 1 1    1 1 1 1    1 1 1 1    1 1 1 1    1 1 1 1    1 1 1 1
X F    U S O B      U S O B    U S O B    U S O B    U S O B    U S O B    U S O B
  X-Sticky F-Final U-User S-System O-Others B-Bearer
created: 2021-08-30 13:33:16 +0300 MSK
attributes:
	Timestamp=1630319596
placement policy:
REP 2 IN X
CBF 2
SELECT 2 FROM F AS X
FILTER Deployed EQ NSPCC AS F
```

there is an object which is stored, but:
```
 neofs-cli object get --cid CeeroywT8ppGE4HGjhpzocJkdb2yu3wD5qCGFTjkw1Cc --oid Cu6Vyun5WQeEnKuwsGFgPAfqyzuxcVN5aBrV83fbUVTA -r st2.storage.fs.neo.org:8080 -v
Missing wallet in the configuration.
Generating random private key for command processing...
Private key generated successfully. Public key: fd0d740abe058271a316e8212986012bc039f3d12c70c59bf584186c90c783259d5e489cc4142d88d8fd341baef65bdf4fa32318bf712fe9adeba008e3e8db41

TTL: 2
Trying to read session from the file...
File with session token is not provided.
rpc error: init object reading on client: read header: status: code = 2048 message = access to object operation denied: access to operation OBJECT_GET is denied by extended ACL check: can't parse headers: invalid header: invalid session token: missing token lifetime
```

seems like object was created with session token w/o lifetime field. Currently, such objects are invalid. Seems like this object was stored before this check was added to the PUT handler

NOTE that we should ensure this is the case

## Expected Behavior
op goes OK

## Current Behavior
no data access

## Possible Solution
ignore session token problems in read ops. If SN responds with an object, it was valid at the time of writing. Something may have been tightened, but it should not be retroactive

## Regression
guess it was retreivable some time ago, but now it's not. From the other side, it's residue in the system is questionable

## Your Environment
* Version used:
```
$ neofs-cli --version
NeoFS CLI
Version: 0.45.2-262-g97bc4ed3-dirty 
GoVersion: go1.23.0
```

End-rey:
In this version of the SDK, old objects that do not yet have a lifetime in the session token are valid.

Closes #3359.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Changelog for NeoFS Node
 - Split object with a missing link object can't be retrieved in some cases (#3337)
 - Invalid URI error text in `request create-container` CLI command (#3366)
 - Invalid object ID error text in `object delete` CLI command (#3366)
+- Old objects without a session token lifetime are not valid (#3373)
 
 ### Changed
 - Make `MaintenanceModeAllowed` network setting always allowed (#3360)
@@ -18,6 +19,7 @@ Changelog for NeoFS Node
 ### Removed
 
 ### Updated
+- `github.com/nspcc-dev/neofs-sdk-go` dependency to `v1.0.0-rc.13.0.20250530123548-f8dbe53f3996` (#3373)
 
 ### Updating from v0.46.1
 
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,7 @@ require (
 	github.com/nspcc-dev/neo-go v0.109.1
 	github.com/nspcc-dev/neofs-api-go/v2 v2.14.1-0.20240827150555-5ce597aa14ea
 	github.com/nspcc-dev/neofs-contract v0.22.0
-	github.com/nspcc-dev/neofs-sdk-go v1.0.0-rc.13.0.20250516062127-1f3abc2c67ca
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ Changelog for NeoFS Node
 ### Removed
 
 ### Updated
+- `github.com/nspcc-dev/neofs-sdk-go` dependency to `v1.0.0-rc.13.0.20250530123548-f8dbe53f3996` (#3373)
```
roman-khimov, line 21: A `Fixed` note for the bug would be nice to have as well.

PR #3375, opened 2025-06-02
Feat/non notary meta on chain

carpawell:


pkg/morph/client/constructor.go
```diff
@@ -250,6 +259,28 @@ func newActor(ws *rpcclient.WSClient, acc *wallet.Account, cfg cfg) (*actor.Acto
...
+	return actor.New(ws, []actor.SignerAccount{
+		{
+			Signer: transaction.Signer{
+				Account: proxy,
+				Scopes:  transaction.None,
+			},
+			Account: wallet.NewContractAccount(proxy),
+		},
+		{
+			Signer: transaction.Signer{
```
roman-khimov, line 272: Do we need this signer at all?

carpawell: i thought as a chain guy you would support signing an operation by a real sender. in practice, no, we do not need it

roman-khimov: I don't support useless signatures. We have enough of them already here.

carpawell: ok

PR #3381, opened 2025-06-05
Drop WC max object size

roman-khimov:


pkg/local_object_storage/writecache/put.go
```diff
@@ -23,11 +23,6 @@ func (c *cache) Put(addr oid.Address, obj *objectSDK.Object, data []byte) error
 		return ErrReadOnly
 	}
 
-	sz := uint64(len(data))
-	if sz > c.maxObjectSize {
-		return ErrBigObject
```
End-rey, line 28: Unused variable 'ErrBigObject'.

roman-khimov: Fixed!

Issue #1803, opened 2022-09-20
Share functionality of placement building between components


cthulhu-rider:
Object placement is build using `netmap.Source` and `container.Source` in most scenarios. Also there are some similar cases when we use same placement approach, e.g. shard evacuation introduced [here](https://github.com/nspcc-dev/neofs-node/pull/1800/) and remote placement [builder](https://github.com/nspcc-dev/neofs-node/blob/c7f85994e550d4c990c08d1ea6529ec99c954770/pkg/services/object/util/placement.go#L72).

I propose to concentrate all placement functionality in single shared code space in order to easily reuse it in new tasks.

fyrchik:
With a proper interface, tree server can use this component too.

cthulhu-rider:
* #2674 brings foundation for this
* #2722 continues (they need to be sync)

caching is very meaningful for app performance and efficiency. Some [attempts](https://github.com/nspcc-dev/neofs-node/blob/master/pkg/services/object_manager/placement/netmap.go) were made, but they are buggy, so lets also fix them

suggest to w8 4 prelim work finish

roman-khimov:
Do we have anything left here?

cthulhu-rider:
for SN:
1. `ControlService.Evacuate` https://github.com/nspcc-dev/neofs-node/blob/f33949e1958717a33c821376d1cb5b9d5bfcb69f/pkg/services/control/server/evacuate.go#L65-L78
2. object ACL https://github.com/nspcc-dev/neofs-node/blob/f33949e1958717a33c821376d1cb5b9d5bfcb69f/pkg/services/object/acl/v2/classifier.go#L130
3. routing for container volume estimations https://github.com/nspcc-dev/neofs-node/blob/f33949e1958717a33c821376d1cb5b9d5bfcb69f/cmd/neofs-node/container.go#L410-L427
4. Tree service (multi-op) https://github.com/nspcc-dev/neofs-node/blob/f33949e1958717a33c821376d1cb5b9d5bfcb69f/pkg/services/tree/container.go#L35-L60

object ACL should definitely use shared policy component. Evacuator is rarely used, but better to use it too. Tree and estimations... i wouldnt bother them tbh

finally, whole [package](https://github.com/nspcc-dev/neofs-node/tree/master/pkg/services/object_manager/placement) should be purged

IR and CLI also apply storage policies to network maps for various needs. Caching doesn't help them, so I would leave everything as is

[...]

PR #3385, opened 2025-06-10
sn/object/acl: Use cached container SN lists for ACL processing

cthulhu-rider:


cthulhu-rider:
definitely #1803, tagged

roman-khimov (approved):

Issue #3397, opened 2025-06-11
Resync new epoch's next tick correctly after connection is lost


carpawell:
If connection was lost, IR [resets the epoch timer](https://github.com/nspcc-dev/neofs-node/blob/aa4ee681911d1820dc15bf617676c27ab0c79de1/pkg/innerring/innerring.go#L1154-L1156), which means it will tick next time only after EpochDuration time. But it is wrong, the next epoch should tick depending on the current height and the number of blocks after the last tick. 

## Expected Behavior
No connection loss can stop epoch flow.

## Current Behavior
If a connection is lost by more than 1/3 - 1 nodes, a new epoch will not occur.

## Possible Solution
Extend timers to be possible to reset at a custom height.
OR
Just run a one-time timer after reconnection, we have such too.

## Steps to Reproduce (for bugs)
Stop RPCs for 3 out of 7 IRs for like... minutes, retunrn them back and see that there is no new epoch. 

## Context
Mainnet update.

## Regression
I dont think so.

## Your Environment
0.47.0

PR #3404, opened 2025-06-18
Fix/ir timers

carpawell:


pkg/innerring/blocktimer.go
```diff
@@ -1,92 +0,0 @@
...
-		epoch      epochState        // to specify which epoch to stop, and epoch duration
-
-		stopEstimationDMul uint32 // X: X/Y of epoch in blocks
-		stopEstimationDDiv uint32 // Y: X/Y of epoch in blocks
-
-		basicIncome subEpochEventHandler
-	}
-)
-
-func newEpochTimer(args *epochTimerArgs) *timer.BlockTimer {
```
roman-khimov, line 40: Not sure it was worth renaming it (at least not in a separate commit), makes diff larger on GH.

pkg/innerring/innerring.go
```diff
@@ -256,18 +239,56 @@ func (s *Server) Start(ctx context.Context, intError chan<- error) (err error) {
...
+		if err != nil {
+			return fmt.Errorf("can't read last current block header (#%d): %w", blockHeight, err)
+		}
+
+		nextTickAfter = epochTickAfter(lastTickH.Timestamp, currentH.Timestamp, epochDuration)
+	}
+
+	if nextTickAfter == 0 {
+		for _, h := range s.newEpochTickHandlers() {
+			h()
```
roman-khimov, line 283: Seems to be wrong, consider delayed epoch tick, it's not correct to call handlers in this case (even though it shouldn't hurt at `Start()`).

pkg/innerring/processors/netmap/process_epoch.go
```diff
@@ -30,14 +30,7 @@ func (np *Processor) processNewEpoch(ev netmapEvent.NewEpoch) {
...
 
-	h, err := np.netmapClient.Morph().TxHeight(ev.TxHash())
-	if err != nil {
-		l.Warn("can't get transaction height",
-			zap.String("hash", ev.TxHash().StringLE()),
-			zap.Error(err))
-	}
-
-	if err := np.epochTimer.ResetEpochTimer(h); err != nil {
+	if err := np.epochTimer.ResetEpochTimer(); err != nil {
```
roman-khimov, line 33: Not using block time here?

pkg/innerring/config.go
```diff
@@ -194,7 +194,7 @@ func validateBlockchainConfig(cfg *config.Config) error {
 // auto-deployment.
 func setNetworkSettingsDefaults(netCfg *deploy.NetworkConfiguration) {
 	netCfg.MaxObjectSize = 64 << 20 // in bytes of object payload
-	netCfg.EpochDuration = 240      // in FS chain blocks (e.g. ~1h for 15s block interval)
+	netCfg.EpochDuration = 3600     // in seconds
```
roman-khimov, line 197: Please don't do this now. This will delay deployments in some cases, it'll be fixed with initial forced ticks, but this requires a refactor. Also tests are using 1s blocks with this default epoch config and can experience delays.

pkg/innerring/innerring.go
```diff
@@ -1181,7 +1181,7 @@ func (s *Server) restartFSChain() error {
 		return fmt.Errorf("FS chain config reinitialization: %w", err)
 	}
 
-	if err = s.resetEpochTimer(true); err != nil {
+	if err = s.resetEpochTimer(false); err != nil {
```
roman-khimov, line 1184: Seems like this flag is always `false` now.

pkg/innerring/blocktimer.go
carpawell: > Not sure it was worth renaming it

renaming "block" timer file that does not count blocks is wrong? why? i do like renaming usually, but there is literally no "block timers" anymore in the repo

pkg/innerring/innerring.go
carpawell: we can discuss it. in fact, these were the last lines i added; it does not affect timers, it is more about #3397 situation. it is likely the best thing an ir that has not received an epoch event but that truly believes that it should have ticked it already can do

pkg/innerring/innerring.go
carpawell: this flag is `afterNewEpochEvent`, so it is obviously `true` for resetting after new epoch event: https://github.com/nspcc-dev/neofs-node/blob/903c9d95018c8687820f4ab23f6cfdccdc74fd45/pkg/innerring/state.go#L201

pkg/innerring/processors/netmap/process_epoch.go
carpawell: oh, you mean that after an epoch is done, we should use the last block's time to know when to tick next time?

roman-khimov: Exactly. Block time != time when we get the event. And we're monitoring next block for block time again, not local clock.

pkg/innerring/config.go
roman-khimov: We will change it when we're ready everywhere, that's it.

pkg/innerring/blocktimer.go
```diff
@@ -30,31 +30,68 @@ type (
...
+	timers []*time.Timer
+
+	eHandlers     []newEpochHandler
+	deltaHandlers []deltaHandler
+}
+
+func (et *epochTimer) reset(tickAfter, epochDuration time.Duration) {
+	et.stop()
+
+	et.timers = append(et.timers, time.AfterFunc(tickAfter, func() {
```
roman-khimov: The fact that it's no longer tied to blocks is something I'm still thinking about.

pkg/innerring/innerring.go
roman-khimov: You can differentiate missing tick from node missing a tick.

pkg/innerring/blocktimer.go
carpawell: ... and what do you think? it is related but not on the timer layer, but somewhere above inside IR

pkg/innerring/blocktimer.go
```diff
@@ -30,31 +30,82 @@ type (
...
-			return uint32(args.epoch.EpochDuration()), nil
-		},
-		func() {
-			for _, handler := range args.newEpochHandlers {
-				handler()
-			}
-		},
-	)
+type deltaHandler struct {
+	m          sync.Mutex
```
roman-khimov: I think locking can be simplified here, you can have a queue of handlers with the only thread managing it (the same that receives block events).

carpawell: ok, as long as it is done [block by block](https://github.com/nspcc-dev/neofs-node/blob/fce19b776491b3489a79fc74dd5505ebe75add0b/pkg/morph/event/listener.go#L197-L215), it can be dropped. but the main internal mutex still must be kept, since netmap handling and ws reconnection are separate routines, and these changes may appear async at any time

roman-khimov (approved):

PR #3406, opened 2025-06-19
Support unsigned object GET/HEAD responses

cthulhu-rider:
* based on and blocked by https://github.com/nspcc-dev/neofs-sdk-go/pull/719

tested S3 and REST GWs require signatures to be presented. Passed tests assert proposed behavior is backward compatible. Note that clients which do _not_ set API version are also served with signatures

### DIFF
--- a/pkg/services/object/server.go
+++ b/pkg/services/object/server.go
@@ -709,29 +709,22 @@ func convertHeadPrm(signer ecdsa.PrivateKey, req *protoobject.HeadRequest, resp 
 			return nil, err
 		}
 
-		nodePub := node.PublicKey()
 		var hdr *object.Object
 		return hdr, c.ForEachGRPCConn(ctx, func(ctx context.Context, conn *grpc.ClientConn) error {
 			var err error
-			hdr, err = getHeaderFromRemoteNode(ctx, conn, nodePub, req)
+			hdr, err = getHeaderFromRemoteNode(ctx, conn, req)
 			return err // TODO: log error
 		})
 	})
 	return p, nil
 }
 
-func getHeaderFromRemoteNode(ctx context.Context, conn *grpc.ClientConn, nodePub []byte, req *protoobject.HeadRequest) (*object.Object, error) {
+func getHeaderFromRemoteNode(ctx context.Context, conn *grpc.ClientConn, req *protoobject.HeadRequest) (*object.Object, error) {
 	resp, err := protoobject.NewObjectServiceClient(conn).Head(ctx, req)
 	if err != nil {
 		return nil, fmt.Errorf("sending the request failed: %w", err)
 	}
 
-	if err := internal.VerifyResponseKeyV2(nodePub, resp); err != nil {
-		return nil, err
-	}
-	if err := neofscrypto.VerifyResponseWithBuffer(resp, nil); err != nil {
-		return nil, fmt.Errorf("response verification failed: %w", err)
-	}
 	if err := checkStatus(resp.GetMetaHeader().GetStatus()); err != nil {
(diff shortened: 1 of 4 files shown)


pkg/services/object/server.go
```diff
@@ -1147,6 +1168,9 @@ func continueGetFromRemoteNode(ctx context.Context, conn *grpc.ClientConn, req *
 				if !headWas {
 					return io.ErrUnexpectedEOF
 				}
+				if !bytes.Equal(x.payloadHashGot.Sum(nil), x.payloadHashCheck) {
+					return errors.New("received payload mismatches checksum from header")
```
roman-khimov, line 1172: IIUC it's too late at this stage. And not sure we can improve anything. Maybe this check can be left for client to perform.

pkg/services/object/server.go
```diff
@@ -1177,27 +1212,31 @@ func continueGetFromRemoteNode(ctx context.Context, conn *grpc.ClientConn, req *
...
 				continue
 			}
-			if err := stream.WriteChunk(respChunk); err != nil {
+			if err := x.respStream.WriteChunk(respChunk); err != nil {
 				return fmt.Errorf("could not write object chunk in Get forwarder: %w", err)
 			}
 			readPayload += len(fullChunk)
-			*respondedPayload += len(respChunk)
+			x.respondedPayload += len(respChunk)
+			x.payloadHashGot.Write(respChunk) // never returns an error according to docs
```
roman-khimov, line 1239: Maybe it can be checked here? We always know the proper length.

cmd/neofs-cli/internal/client/client.go
```diff
@@ -547,7 +547,6 @@ func GetObject(ctx context.Context, prm GetObjectPrm) (*GetObjectRes, error) {
 	}
 
 	getPrm.WithXHeaders(prm.xHeaders...)
-	getPrm.SkipChecksumVerification()
```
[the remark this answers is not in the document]
cthulhu-rider: not to me. There were no checks before, now they are enabled by active change

pkg/services/object/server.go
cthulhu-rider: to me, it's pretty expected to verify checksum when the data is over. Normally, we'll continue and get EOF insta

we dont check for truncated/excessive payload here, regardless of hash. Seems like it's worth doing

cmd/neofs-cli/internal/client/client.go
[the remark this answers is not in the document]
roman-khimov: Well, new SDK adds them, so it's kinda expected and requires no addition/removal of lines.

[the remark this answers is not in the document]
roman-khimov: @carpawell, what do you think?

pkg/services/object/internal/client/client.go
```diff
@@ -148,6 +148,7 @@ func GetObject(prm GetObjectPrm) (*GetObjectRes, error) {
 	}
 
 	prm.cliPrm.WithXHeaders(prm.xHeaders...)
+	prm.cliPrm.SkipChecksumVerification()
```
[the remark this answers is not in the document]
cthulhu-rider: `GetObject` func is used to get requested root object or its children from remote container SNs

pkg/services/object/server.go

pkg/services/object/server.go
cthulhu-rider: considering that client from SDK does checks by default now, I suggest not adding them to server for now

roman-khimov (approved):

PR #3413, opened 2025-06-23
sn/object: Forward original GET/HEAD/RANGE request to container members

cthulhu-rider:
i stick to the proposed change in cost/benefit ratio

at the same time, the explicit context of request forwarding is lost. I dont see a problem with this for now. If the need arises, I'd consider more lightweight approach than recursive headers


roman-khimov:
Rebase?

cthulhu-rider:
this cannot be applied that simple. In particular, missing object leads to leads to an infinite loop (deadline exceeded) instead of 404. #3978 is needed

roman-khimov:
Replaced by #4021?

Issue #2563, opened 2023-09-09
Local object store behavior for Put of already existing objects


roman-khimov:
## Is your feature request related to a problem? Please describe.
`common.Storage` interface exposes a `Put()` function that's supposed to be pushing objects into the store. One can `Put` an object and then `Put` it again. What happens next? Currently:
 * `Put` succeeds, returning `nil`
 * data is overwritten (FSTree/Peapod/bbcz)

In general we suppose that objects are immutable and we have the same data for the same object. Which means that the first point is likely correct. If that's the case then the second one doesn't matter. However if you're to change the logic of FSTree to preserve an already existing file `TestRefillMetabaseCorrupted` suddenly starts to fail because it relies on this overwriting semantics.

## Describe the solution you'd like
We need to specify what the proper behavior is for all stores.

## Describe alternatives you've considered
And it's not that easy, because:
 * we can have concurrent writers pushing the same data, these prefer to see `nil` and don't care about overwriting, that's the most common case
 * old data is supposed to be good, so overwriting is not necessary
 * we can also have errors leading to different data being pushed for the same address, in this case keeping old data is also more safe
 * but in this case returning `nil` looks bad since writer would think that he has successfully pushed something into the store, but any `Get` would reveal that it's not exactly true (the data is different)
 * if we're to overwrite then `nil` makes total sense
 * old data _can_ be corrupted as well, so we may want to do a forced write (although it can be done with Delete/Put)
 * checking data consistency and returning different status based on that sounds nice, but can lead to substantial overhead that is not justified for the most common case
 * always returning EEXIST is an option (let the upper layer handle), but can be daunting for callers

So, what do we really want from our stores?

PR #2068, opened 2022-11-17
Speed up writecache initialization

fyrchik:

PR #1462, opened 2022-06-01
Some writecache fixes

fyrchik:
1. Allow to fetch data on demand in `FSTree`.
2. Initialize flush mark LRU cache on startup.

PR #3417, opened 2025-06-24
writecache: drop "init flush marks" code

roman-khimov:
The problem is that it can take a significant amount of time to perform while the value is questionable.

It was initially added in 8717820038b6462ec61c9edc7fd20a23ab84ec1d (#1462) when write cache had completely different structure and function. 733a792afa4cf7b3b0cf748aed5257afa6f8f03d tried to optimize it in #2068, but most of the code was subsequently deleted and as write cache is no longer a read cache this lead to the remaining code doing only the following:
 * checking objects against the metabase
 * checking objects against the underlying long-term storage

If the first one is ever triggered (we detect non-existent object) this just means WC is out of sync with metabase which can happen for a number of other reasons. Lower level blobstor can also get out of sync with metabase in some cases, so it's not like this is something unique. If we're to drop this check we'd just flush some dead object, but then GC should still be able to remove it if it's deleted normally. If it's some other type of race --- like WC accepted an object and metabase not yet then we shouldn't delete it at all for safety reasons.

Then if some valid object was previously flushed already we'll retry flushing it again and it's not a problem, EEXIST is not an error (but #2563), so we'd waste some time, but it's not very likely we're going to since this case is rare.

So it should be safe to drop this useless code.


carpawell (approved):
Have you considered "SN was gracefully shutdown" == "WC is clear"? This also looks like a solution to me (continuation of this PR, in fact), but depends on how long it may last.

roman-khimov:
Flushing several GBs of 1K objects can take a lot of time, that'd be a problem of "I can't stop the node gracefully" then.

cthulhu-rider (approved):

PR #3429, opened 2025-06-27
Functionality for Reed-Solomon object encoding

cthulhu-rider:
currently, i'm trying to implement single-level encoding. It's simpler and less resource intensive. Plus it's within current multi-level slicing restrictions like https://github.com/nspcc-dev/neofs-node/commit/60faa73a3829a410ecbae4c7e00260913f51ae4a (smth else?)

an obvious alternative would be a two-level format, where the resulting objects of slicing by length (up to MaxObjectSize) are wrapped into new objects due to the recursive [structure](https://github.com/nspcc-dev/neofs-api/blob/a0a8cb7b23a1fe651ec6eeddb77263fb05f1cfb5/object/types.proto#L270-L271). I haven't found any advantages in it yet


cthulhu-rider:
works overall. I tested this against SNs v0.47.1. Data shards are accepted well. Linker is denied:
1. being sharded as other split members, its RS shards have `LINK` type, so SNs deny them on payload verification ([code](https://github.com/nspcc-dev/neofs-node/blob/d651e2d3b0a2315d5e8e0f6d8c0fbe899c443b75/pkg/core/object/fmt.go#L255))
2. then i tried to make its RS shards to have `REGULAR` type with recovery `__NEOFS__EC_RS_SRC_TYPE: LINK` attribute. This leads to other [failure](https://github.com/nspcc-dev/neofs-node/blob/d651e2d3b0a2315d5e8e0f6d8c0fbe899c443b75/pkg/core/object/fmt.go#L185)

Issue #3391, opened 2025-06-10
"signatures were not found" in some tests


roman-khimov:
## Expected Behavior
No errors.

## Current Behavior
```
output: rpc error: finish object stream: status: code = 1024 message = incomplete object PUT by placement: number of replicas cannot be met for list #0: 2 required, 0 nodes remaining (last node error: signatures were not found in object's metadata)
```
https://github.com/nspcc-dev/neofs-node/actions/runs/15564064452/job/43823539301

## Possible Solution
Fix something.

## Steps to Reproduce (for bugs)
It happens from time to time in tests.

## Regression
Looks like related to meta.

## Your Environment
* Version used: master & 0.46.1

roman-khimov:
Fresh one (0.47.1): https://rest.fs.neo.org/HXSaMJXk2g8C14ht8HSi7BBaiYZ1HeWh2xnWPGQCg4H6/1199-1751307236/index.html#suites/44e11ced39071e8d0cfc11c5b94622ba/5733e482a8c55836/

roman-khimov:
https://rest.fs.neo.org/HXSaMJXk2g8C14ht8HSi7BBaiYZ1HeWh2xnWPGQCg4H6/1203-1751429468/index.html#suites/44e11ced39071e8d0cfc11c5b94622ba/dd499407d068cf55/

roman-khimov:
https://rest.fs.neo.org/HXSaMJXk2g8C14ht8HSi7BBaiYZ1HeWh2xnWPGQCg4H6/3514-1751552688/index.html#suites/44e11ced39071e8d0cfc11c5b94622ba/a174fb6ff57b8ba5/

PR #3443, opened 2025-07-03
Fix replication signature collecting on object PUT

cthulhu-rider:


carpawell (approved):
While this fixes big objects, I am a little surprised that `distributedTarget` is reused for many objects and not reinitialized (it stores so many internals related to a specific object, epoch, block state, etc).

carpawell:
> #3391, @carpawell?

I would like to think this way, but not sure yet, will think a little bit more.

carpawell:
>> I am a little surprised that distributedTarget is reused for many objects and not reinitialized
>
>did not think it'd surprise u 3665c3b1b6bae2f24033db42d05f5850767777ff

Yeah, i think that is the reason, it was not always like that, and i have got used to it. Currently it still [initializes](https://github.com/nspcc-dev/neofs-node/blob/5b1b437597e987bcab83c61a5d22e936be81a350/pkg/services/object/put/streamer.go#L226) smth inside, mb that was the reason for my confusion. I am not sure i understand what unexported initializers of unexported targets initialize, and i am for a more straightforward object flow.

roman-khimov (approved):

PR #3444, opened 2025-07-03
adm: drop fschain init, fix #2905

### LINKED ISSUE #2905: Reconsider init code in `neofs-adm`
I'm always frustrated when we need to support code that is not used. TBW, I am unsure when it worked last time and if it feels good at all. IR deploys a network on its own even in our stands, devenv, and testcases. 

I would just drop it cause I do not see any use cases when it can work out.

Refactor it a little and make it easier to support. Think about when this code can be used and change it accordingly. As I understand @roman-khimov has something to add to this paragraph.

## Additional context
It uses neo-go a lot and every time we need to update neo-go, this is a _huge_ code part that we need to take care of. I prefer to see and support neo-go in the contracts repo (deploy code) and in the `morph` pkgs only.

roman-khimov:
Obsolete, unused, hard to maintain.


carpawell (commented):
No release shouting "do not use me"?

carpawell (approved):

roman-khimov:
Wait a minute, a bit more can be deleted.

carpawell (approved):

PR #3445, opened 2025-07-03
More fixes for replication signature collecting on object PUT

cthulhu-rider:


pkg/services/object/put/distributed.go
```diff
@@ -275,6 +279,10 @@ func (t *distributedTarget) sendObject(node nodeDesc) error {
 				continue
 			}
 
+			if t.metainfoConsistencyAttr == "" {
```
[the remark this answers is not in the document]
cthulhu-rider: do we even need these signatures in this case? They are requested when server is in container regardless of consistency settings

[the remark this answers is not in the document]
cthulhu-rider: what problem do u see? We dont change server behavior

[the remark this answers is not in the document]
cthulhu-rider: seems so https://github.com/nspcc-dev/neofs-node/blob/de0d6c01f53c080425f308be0309728787b17445/pkg/services/object/server.go#L1793-L1794

pkg/services/object/put/distributed.go
```diff
@@ -96,7 +96,7 @@ func (t *distributedTarget) WriteHeader(hdr *objectSDK.Object) error {
 		if t.placementIterator.localOnly {
 			t.encodedObject, err = encodeObjectWithoutPayload(*hdr, int(payloadLen))
 		} else {
-			t.encodedObject, err = encodeReplicateRequestWithoutPayload(t.localNodeSigner, *hdr, int(payloadLen), true)
+			t.encodedObject, err = encodeReplicateRequestWithoutPayload(t.localNodeSigner, *hdr, int(payloadLen), t.metainfoConsistencyAttr != "")
```
carpawell, line 99: while i agree this is not finished (like the whole meta service), i am not sure it's true that replication approval is not needed in replication

pkg/services/object/put/distributed.go
```diff
@@ -224,6 +203,36 @@ func (t *distributedTarget) Close() (oid.ID, error) {
 	return id, nil
 }
 
+func (t *distributedTarget) getShardObjectMetadata() []byte {
```
carpawell, line 206: i do not understand what "shard" means here. either we keep in sync with all the previous code, or we describe what "shard" means

pkg/services/object/put/distributed.go
```diff
@@ -224,6 +203,36 @@ func (t *distributedTarget) Close() (oid.ID, error) {
...
+	if t.objSharedMeta != nil {
```
carpawell, line 207: when this can happen?

pkg/services/object/put/distributed.go
```diff
@@ -201,7 +180,7 @@ func (t *distributedTarget) Close() (oid.ID, error) {
 			t.metaSvc.NotifyObjectSuccess(objAccepted, addr)
 		}
 
-		err = t.cnrClient.SubmitObjectPut(t.objSharedMeta, t.collectedSignatures)
+		err = t.cnrClient.SubmitObjectPut(t.getShardObjectMetadata(), t.collectedSignatures)
```
carpawell, line 183: this was done this way cause i expected meta-on-chain to be the main option in our consistency problem. @roman-khimov, has smth changed?

pkg/services/object/put/distributed.go
```diff
@@ -215,19 +215,10 @@ func (t *distributedTarget) getShardObjectMetadata() []byte {
...
-	typ := t.objMeta.Type()
-	switch typ {
-	case objectSDK.TypeTombstone:
-		deletedObjs = t.objMeta.Objects()
-	case objectSDK.TypeLock:
-		lockedObjs = t.objMeta.Objects()
-	default:
-	}
-
+	deletedObjs := t.objMeta.Objects()
```
carpawell, line 218: have you checked that a TS cannot lock and a LOCK cannon delete smth?

pkg/services/object/put/distributed.go
cthulhu-rider: it was an attempt to do encoding only when necessary. Separate topic, besides, it requires reset and synchronization, which I forgot about, so reverted for now

pkg/services/object/put/distributed.go
cthulhu-rider: renamed to "current"

carpawell (approved):

roman-khimov (approved):

PR #3451, opened 2025-07-07
fstree: make a new object storage structure

End-rey:


End-rey:
The new structure significantly improves `Head` performance for compressed objects (up to 93.28% faster, 21.54% geomean reduction) and reduces memory usage (36.60% geomean reduction). However, it degrades `Get` performance (13.94% geomean slower, up to 70.89% worse).

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -90,6 +91,26 @@ const (
 	// combinedDataOff is the offset from the start of the combined prefix to object data.
 	// It's also the length of the prefix in total.
 	combinedDataOff = combinedLengthOff + combinedLenSize
+
+	// streamPrefix is the prefix for streamed objects. It is used to distinguish
+	// streamed objects from regular ones.
+	streamPrefix = 0x7e
```
roman-khimov, line 97: Not really needed, there is versioning in `combined` format.

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -375,6 +396,28 @@ func (t *FSTree) PutBatch(objs map[oid.Address][]byte) error {
...
+func (t *FSTree) joinHeaderAndPayload(header, payload []byte) []byte {
+	if header == nil {
+		return payload
+	}
+
+	hLen := len(header)
+	payload = t.Compress(payload)
+	pLen := len(payload)
+
+	data := make([]byte, hLen+pLen+streamDataOff)
```
roman-khimov, line 409: Allocation. Which is not really needed.

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -329,7 +350,7 @@ func (t *FSTree) getPath(addr oid.Address) (string, error) {
 }
 
 // Put puts an object in the storage.
-func (t *FSTree) Put(addr oid.Address, data []byte) error {
+func (t *FSTree) Put(addr oid.Address, data []byte, header []byte) error {
```
roman-khimov, line 353: You better keep the old `data` semantics and find header/payload chunks of it.

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -485,6 +538,23 @@ func (t *FSTree) readFullObject(f io.Reader, initial []byte, size int64) ([]byte
...
+	if hLen > 0 {
+		data = data[streamDataOff:]
+		payload, err := t.Decompress(data[hLen:])
+		if err != nil {
+			return nil, fmt.Errorf("decompress payload: %w", err)
+		}
+		pLen := len(payload)
+		payloadNum := protowire.Number(4)
+		n := protowire.SizeTag(payloadNum) + protowire.SizeVarint(uint64(pLen))
+		buf := make([]byte, int(hLen)+pLen+n)
```
roman-khimov, line 551: This explains `Get()` problems. What happens to `GetStream()`?

pkg/local_object_storage/blobstor/fstree/fstree.go
End-rey: That is, to parse `data` during the `Put`?

roman-khimov: That's acceptable to me because the cost of Put is dominated by disk operation time, walking through protobuf is easy here.

pkg/local_object_storage/blobstor/fstree/fstree.go
End-rey: Added to commit message `GetStream` bench test.

roman-khimov:
Can you make a separate PR with tests? They're relevant regradless of outcome of this PR. And for it specifically the results don't look positive, either something can be improved in the implementation and it'd be more efficient or it's just not worth the trouble.

PR #3457, opened 2025-07-08
EC partitioning (placement)

cthulhu-rider:
will be merged with #3429 eventually

will PR some independent changes separately


roman-khimov (commented):
What happens if we're to PUT a prepared object into EC-enabled container?

cthulhu-rider:
lookin at failing units

internal/ec/ec.go
```diff
@@ -0,0 +1,81 @@
...
+	if err := enc.Encode(parts); err != nil {
+		return nil, fmt.Errorf("calculate Reed-Solomon parity: %w", err)
+	}
+
+	return parts, nil
+}
+
+// Decode decodes source data of known len from EC parts obtained by applying
+// specified rule.
+func Decode[LT constraints.Unsigned](rule Rule, dataLen LT, parts [][]byte) ([]byte, error) {
```
roman-khimov, line 63: `dataLen` is just `int` to me. Yeah, with the regular Go int/uint caveat, but it's used for slicing only and it's just `int` in Go.

internal/ec/ec.go
```diff
@@ -0,0 +1,81 @@
...
+	required := make([]bool, rule.DataPartNum+rule.ParityPartNum)
+	for i := range rule.DataPartNum {
+		required[i] = true
+	}
+
+	if err := dec.ReconstructSome(parts, required); err != nil {
+		return nil, fmt.Errorf("restore Reed-Solomon: %w", err)
+	}
+
+	// TODO: last part may be shorter, do not overallocate buffer.
```
roman-khimov, line 79: I doubt it's a problem practically, just a tiny rounding error.

internal/object/attributes.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	return i, nil
+}
+
+// GetIntAttribute looks up for specified int attribute in the given object
+// header. Returns [ErrAttributeNotFound] if the attribute is missing.
+//
+// GetIntAttribute ignores all attribute values except the first.
+func GetIntAttribute(hdr object.Object, attr string) (int, error) {
```
roman-khimov, line 40: This wants to be in SDK, isn't it?

pkg/services/object/put/distributed.go
```diff
@@ -62,6 +65,10 @@ type distributedTarget struct {
 	keyStorage        *svcutil.KeyStorage
 
 	localOnly bool
+
+	// When object from request is an EC part, ecPart.RuleIndex is >= 0.
+	// Undefined when policy have no EC rules.
```
roman-khimov, line 70: has

pkg/services/object/put/ec.go
```diff
@@ -0,0 +1,135 @@
...
+func (t *distributedTarget) saveECPart(part object.Object, objMeta objectcore.ContentMeta, encObj encodedObject, idx, total int, nodeList []netmap.NodeInfo) error {
+	return t.distributeObject(part, objMeta, encObj, func(obj object.Object, objMeta objectcore.ContentMeta, encObj encodedObject) error {
+		return t.distributeECPart(obj, objMeta, encObj, idx, total, nodeList)
+	})
+}
+
+func (t *distributedTarget) distributeECPart(part object.Object, objMeta objectcore.ContentMeta, enc encodedObject, idx, total int, nodeList []netmap.NodeInfo) error {
+	var firstErr error
+	for {
+		err := t.saveECPartOnNode(part, objMeta, enc, nodeList[idx])
```
roman-khimov, line 106: How about saving on backup nodes (from `nodeList[DATA+PARITY:]` range) if this fails?

pkg/services/object/put/distributed.go
```diff
@@ -165,7 +161,8 @@ func (t *distributedTarget) Close() (oid.ID, error) {
 			err = errIncompletePut{singleErr: fmt.Errorf("%w (last node error: %w)", errNotEnoughNodes{required: 1}, err)}
 		}
 	} else {
-		err = t.placementIterator.iterateNodesForObject(id, t.sendObject)
+		broadcast := tombOrLink || typ == objectSDK.TypeLock || len(t.obj.Children()) > 0
```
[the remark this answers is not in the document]
cthulhu-rider: this is definitely should not happen, will review

internal/ec/ec.go
cthulhu-rider: thinkin about parts of several MB len with the tiny last one (1B in particular)

pkg/services/object/put/ec.go
cthulhu-rider: worthwile addition

internal/ec/ec.go
roman-khimov: But aren't EC part all ~same size? The difference I expect here is on the order of NUM_OF_PARTS bytes.

pkg/services/object/put/distributed.go
```diff
@@ -165,7 +161,8 @@ func (t *distributedTarget) Close() (oid.ID, error) {
 			err = errIncompletePut{singleErr: fmt.Errorf("%w (last node error: %w)", errNotEnoughNodes{required: 1}, err)}
 		}
 	} else {
-		err = t.placementIterator.iterateNodesForObject(id, t.sendObject)
+		broadcast := tombOrLink || (!t.localOnly && typ == objectSDK.TypeLock) || len(t.obj.Children()) > 0
```
roman-khimov: Makes me wonder about `localOnly && tombOrLink`, but at least that's the way it was done previously.

PR #3461, opened 2025-07-11
Object PUT unit testing

cthulhu-rider:
wanna this before #3457 to preserve REP behavior and prevent regressions

### DIFF
--- a/pkg/services/object/put/service.go
+++ b/pkg/services/object/put/service.go
@@ -123,7 +123,12 @@ func NewService(transport Transport, neoFSNet NeoFSNetwork, m *meta.Meta, opts .
 		opts[i](c)
 	}
 
-	c.fmtValidator = object.NewFormatValidator(c.cnrClient.Morph(), neoFSNet, c.fmtValidatorOpts...)
+	var fmtValidatorChain object.FSChain
+	if c.cnrClient != nil {
+		fmtValidatorChain = c.cnrClient.Morph()
+	}
+
+	c.fmtValidator = object.NewFormatValidator(fmtValidatorChain, neoFSNet, c.fmtValidatorOpts...)
 	c.metaSvc = m
 
 	return &Service{
--- a/pkg/services/object/put/service.go
+++ b/pkg/services/object/put/service.go
@@ -90,8 +90,6 @@ type cfg struct {
 
 	cnrSrc container.Source
 
-	netMapSrc netmap.Source
-
 	remotePool util.WorkerPool
 
 	fmtValidator *object.FormatValidator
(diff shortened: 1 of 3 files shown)


pkg/services/object/put/service.go
```diff
@@ -123,7 +123,12 @@ func NewService(transport Transport, neoFSNet NeoFSNetwork, m *meta.Meta, opts .
 		opts[i](c)
 	}
 
-	c.fmtValidator = object.NewFormatValidator(c.cnrClient.Morph(), neoFSNet, c.fmtValidatorOpts...)
+	var fmtValidatorChain object.FSChain
+	if c.cnrClient != nil {
```
roman-khimov, line 127: Can it ever be `nil` for real?

pkg/services/object/put/service_test.go
```diff
@@ -0,0 +1,642 @@
...
+	return int(res)
+}
+
+func assertObjectIntegrity(t *testing.T, obj object.Object) {
+	require.NoError(t, obj.CheckVerificationFields())
+
+	require.NotNil(t, obj.Version())
+	require.Equal(t, version.Current(), *obj.Version())
+
+	require.NotZero(t, obj.GetContainerID())
```
roman-khimov, line 618: Likely we know specific values we expect.

pkg/services/object/put/service_test.go
```diff
@@ -0,0 +1,642 @@
...
+	restored.SetPayload(gotPayload)
+
+	return restored
+}
+
+func splitMembersCount(limit, ln uint64) int {
+	if ln <= limit {
+		return 1
+	}
+	res := ln/limit + 1 // + LINK
```
roman-khimov, line 605: `(ln+limit-1)/limit + 1` and no `if` below? Although it's a test...

pkg/services/object/put/service_test.go
```diff
@@ -0,0 +1,642 @@
...
+}
+
+type testCluster struct {
+	nodeServices      nodeServices
+	nodeNetworks      []mockNetwork
+	nodeSessions      []mockNodeSession
+	nodeLocalStorages []inMemLocalStorage
+}
+
+func (x testCluster) allStoredObjects() iter.Seq[[]object.Object] {
```
roman-khimov, line 474: Looks like you're never iterating it partially, so `iter.Seq` doesn't buy us anything here.

pkg/services/object/put/service.go
cthulhu-rider: only in test

pkg/services/object/put/service_test.go
cthulhu-rider: yeah, this is a test, i've kept it w/o smart math intentionally

pkg/services/object/put/service_test.go
cthulhu-rider: this func checks general correctness, input-dependent checks are done where input is init

pkg/services/object/put/service_test.go
cthulhu-rider: changed to slice

roman-khimov (approved):

PR #3466, opened 2025-07-14
services/object: use stream API for `Get` operation

### LINKED ISSUE #3439: Stream GET replies
I'm always frustrated when we're wasting time and memory for nothing. Current GET (and GETRANGE) implementation does complete buffering for both local and networked requests.

Network API always had a concept of payload stream, FSTree can also provide it now after #3431. The only thing needed is to make it available from shard/local engine API and use it in object service. Play with buffers a bit (64K should be OK after #3406 and SDK version update).

Keep pulling megabytes of data into RAM, wait for this data to be read completely before replying to user.

End-rey:
Closes #3439.

### DIFF
--- a/pkg/services/object/get/assemble.go
+++ b/pkg/services/object/get/assemble.go
@@ -66,7 +66,7 @@ func (exec *execCtx) assemble() {
 
 			if ok := exec.overtakePayloadInReverse(children[len(children)-1]); ok {
 				// payload of all children except the last are written, write last payload
-				exec.writeObjectPayload(exec.collectedObject)
+				exec.copyChild(exec.lastChildID, &exec.lastChildRange, false)
 			}
 		}
 	} else if prev != nil {
@@ -76,8 +76,12 @@ func (exec *execCtx) assemble() {
 			//  * else go right-to-left with GET and compose in single object before writing
 
 			if ok := exec.overtakePayloadInReverse(*prev); ok {
-				// payload of all children except the last are written, write last payloa
-				exec.writeObjectPayload(exec.collectedObject)
+				var rng *objectSDK.Range
+				if exec.ctxRange() != nil {
+					rng = &exec.lastChildRange
+				}
+				// payload of all children except the last are written, write last payload
+				exec.copyChild(exec.lastChildID, rng, false)
 			}
 		}
 	} else {
@@ -85,12 +89,12 @@ func (exec *execCtx) assemble() {
 	}
 }
 
-func (exec *execCtx) initFromChild(obj oid.ID) (prev *oid.ID, children []oid.ID) {
+func (exec *execCtx) initFromChild(obj oid.ID) (*oid.ID, []oid.ID) {
 	exec.log.Debug("starting assembling from child", zap.Stringer("child ID", obj))
 
-	child, ok := exec.getChild(obj, nil, true)
+	child, ok := exec.headChild(obj)
 	if !ok {
-		return
+		return nil, nil
 	}
 
 	par := child.Parent()
@@ -99,12 +103,10 @@ func (exec *execCtx) initFromChild(obj oid.ID) (prev *oid.ID, children []oid.ID)
 
 		exec.log.Debug("received child with empty parent", zap.Stringer("child ID", obj))
 
-		return
+		return nil, nil
 	}
 
-	exec.collectedObject = par
-
-	var payload []byte
+	exec.collectedHeader = par
 
 	if rng := exec.ctxRange(); rng != nil {
 		seekOff := rng.GetOffset()
@@ -121,30 +123,42 @@ func (exec *execCtx) initFromChild(obj oid.ID) (prev *oid.ID, children []oid.ID)
 			exec.err = &errOutOfRange
 			exec.status = statusAPIResponse
 
-			return
+			return nil, nil
 		}
 
 		childSize := child.PayloadSize()
 
-		exec.curOff = parSize - childSize
+		startRight := parSize - childSize
+		exec.curOff = startRight
 
 		from := uint64(0)
-		if exec.curOff < seekOff {
-			from = seekOff - exec.curOff
+		if startRight < seekOff {
+			from = seekOff - startRight
 		}
 
 		to := uint64(0)
-		if seekOff+seekLen > exec.curOff+from {
-			to = seekOff + seekLen - exec.curOff
+		if seekOff+seekLen > startRight+from {
+			to = seekOff + seekLen - startRight
+			if to > childSize {
+				to = childSize
+			}
 		}
 
-		payload = child.Payload()[from:to]
-		rng.SetLength(seekLen - to + from)
-	} else {
-		payload = child.Payload()
+		segLen := uint64(0)
+		if to > from {
+			segLen = to - from
+		}
+		rng.SetLength(seekLen - segLen)
+
+		if segLen > 0 {
+			exec.lastChildRange.SetOffset(from)
+			exec.lastChildRange.SetLength(segLen)
+		} else {
+			exec.lastChildRange.SetLength(0)
+		}
 	}
 
-	exec.collectedObject.SetPayload(payload)
+	exec.lastChildID = child.GetID()
 
 	idPrev := child.GetPreviousID()
 	if !idPrev.IsZero() {
@@ -163,12 +177,8 @@ func (exec *execCtx) overtakePayloadDirectly(children []oid.ID, rngs []objectSDK
 			r = &rngs[i]
 		}
 
-		child, ok := exec.getChild(children[i], r, !withRng && checkRight)
-		if !ok {
-			return
-		}
-
-		if ok := exec.writeObjectPayload(child); !ok {
+		retrieved, wrote := exec.copyChild(children[i], r, !withRng && checkRight)
+		if !retrieved && !wrote {
 			return
 		}
 	}
--- a/pkg/services/object/get/assemble.go
+++ b/pkg/services/object/get/assemble.go
@@ -118,9 +118,7 @@ func (exec *execCtx) initFromChild(obj oid.ID) (*oid.ID, []oid.ID) {
 		seekTo := seekOff + seekLen
 
 		if seekTo < seekOff || parSize < seekOff || parSize < seekTo {
-			var errOutOfRange apistatus.ObjectOutOfRange
-
-			exec.err = &errOutOfRange
+			exec.err = apistatus.ErrObjectOutOfRange
 			exec.status = statusAPIResponse
 
 			return nil, nil
--- a/pkg/services/object/get/exec.go
+++ b/pkg/services/object/get/exec.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/ecdsa"
 	"errors"
+	"io"
 
 	clientcore "github.com/nspcc-dev/neofs-node/pkg/core/client"
 	"github.com/nspcc-dev/neofs-node/pkg/core/object"
@@ -12,6 +13,12 @@ import (
 	objectSDK "github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"go.uber.org/zap"
+)
+
+const (
+	// streamChunkSize is the size of the chunk that is used to read/write
+	// object payload from/to the stream.
+	streamChunkSize = 64 * 1024 // 64 KiB
 )
 
 type statusError struct {
@@ -34,11 +41,16 @@ type execCtx struct {
 	// If debug level is enabled, all messages also include info about processing request.
 	log *zap.Logger
 
-	collectedObject *objectSDK.Object
+	collectedHeader *objectSDK.Object
+	collectedReader io.ReadCloser
 
 	curOff uint64
 
 	head bool
+
+	// range assembly (V1 split) helpers
+	lastChildID    oid.ID
+	lastChildRange objectSDK.Range
 }
 
 type execOption func(*execCtx)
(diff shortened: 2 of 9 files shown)


pkg/local_object_storage/shard/get.go
```diff
@@ -140,3 +142,53 @@ func (s *Shard) getBytesWithMetadataLookup(addr oid.Address, skipMeta bool) ([]b
...
+// GetStream reads an object from shard as a stream. skipMeta flag allows to fetch object from
+// the blobstor directly.
+//
+// Returns the object header and a reader for the payload.
+// The caller is responsible for closing the returned io.ReadCloser if it is not nil.
+//
+// Returns any error encountered that did not allow to completely read the object part.
+// Returns an error of type apistatus.ObjectNotFound if the requested object is missing in shard.
+// Returns an error of type apistatus.ObjectAlreadyRemoved if the requested object has been marked as removed in shard.
+// Returns the object.ErrObjectIsExpired if the object is presented but already expired.
```
roman-khimov, line 155: `is present` (I know it's copypasted, but let's not reintroduce this error again)

pkg/local_object_storage/shard/get.go
```diff
@@ -140,3 +142,53 @@ func (s *Shard) getBytesWithMetadataLookup(addr oid.Address, skipMeta bool) ([]b
...
+		if err != nil {
+			return err
+		}
+		res = obj
+		reader = r
+		return nil
+	}
+
+	wc := func(c writecache.Cache) error {
+		// For writecache, we fall back to regular Get since it's typically small objects
```
roman-khimov, line 176: But not necessarily. Why don't you let write cache's FSTree handle this?

pkg/services/object/get/assemble.go
```diff
@@ -168,7 +168,7 @@ func (exec *execCtx) overtakePayloadDirectly(children []oid.ID, rngs []objectSDK
 			return
 		}
 
-		if ok := exec.writeObjectPayload(child); !ok {
+		if ok := exec.writeObjectPayload(child, nil); !ok {
```
roman-khimov, line 171: Why not streaming here? `getChild()` should be adapted for this as well. This optimization is very much relevant for big objects that stream GBs of data.

pkg/services/object/get/util.go
```diff
@@ -145,12 +149,19 @@ func (c *clientWrapper) getObject(exec *execCtx, info coreclient.NodeInfo) (*obj
...
 				// fallback to GET here.
-				obj, err := c.get(exec, key)
+				obj, reader, err := c.get(exec, key)
 				if err != nil {
-					return nil, err
+					return nil, nil, err
+				}
+				defer reader.Close()
+
+				// Read the whole object and extract range
```
roman-khimov, line 158: Why? Range requests return only the data requested, locally there is `GetRange()`.

End-rey: I didn't really understand what this was about, but now I've done the payload processing as well as the `GetRange` from `FSTree`.

pkg/local_object_storage/blobstor/common/storage.go
```diff
@@ -26,6 +27,7 @@ type Storage interface {
 	GetBytes(oid.Address) ([]byte, error)
 	Get(oid.Address) (*objectSDK.Object, error)
```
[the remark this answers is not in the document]
End-rey: It is used in these operations: `Evacuate`, `ObjectStatus`, `inhume`.

pkg/local_object_storage/shard/get.go
```diff
@@ -140,3 +141,51 @@ func (s *Shard) getBytesWithMetadataLookup(addr oid.Address, skipMeta bool) ([]b
...
+// the blobstor directly.
+//
+// Returns the object header and a reader for the payload.
+// The caller is responsible for closing the returned io.ReadCloser if it is not nil.
+//
+// Returns any error encountered that did not allow to completely read the object part.
+// Returns an error of type apistatus.ObjectNotFound if the requested object is missing in shard.
+// Returns an error of type apistatus.ObjectAlreadyRemoved if the requested object has been marked as removed in shard.
+// Returns the object.ErrObjectIsExpired if the object is present but already expired.
+func (s *Shard) GetStream(addr oid.Address, skipMeta bool) (*objectSDK.Object, io.ReadCloser, error) {
```
cthulhu-rider, line 155: u add same extension for writecache, and then for shard and engine merged w/ the service changes. I'd make extensions for storage components 1st. This will trigger simple refactoring of the service. Test all `GetStream` methods properly. Then, when storage is rdy, change the service

pkg/services/object/get/util.go
```diff
@@ -158,50 +157,48 @@ func (c *clientWrapper) getObject(exec *execCtx, info coreclient.NodeInfo) (*obj
...
+				}{
+					Reader: io.LimitReader(reader, int64(to-from)),
+					Closer: reader,
+				}, nil
+			}
+			return nil, nil, fmt.Errorf("init payload reading: %w", err)
 		}
-
-		return payloadOnlyObject(w.Bytes()), nil
+		return object.New(), rdr, nil
```
cthulhu-rider, line 195: since we have separate return for payload now, suggest to return `rdr` only. But make sure callers won't NPE

pkg/services/object/get/util.go
```diff
@@ -222,72 +219,50 @@ func (c *clientWrapper) get(exec *execCtx, key *ecdsa.PrivateKey) (*object.Objec
...
 		if err != nil {
-			return nil, err
+			return nil, nil, err
 		}
 
 		o := object.New()
 		o.SetPayload(r)
 
-		return o, nil
+		return o, nil, nil
```
cthulhu-rider, line 246: we return stream only in remote case. Suggest to do the same here. Although `engine.GetRange` is not streamed yet, we can prepare for this in advance

pkg/services/object/get/util.go
```diff
@@ -222,72 +219,50 @@ func (c *clientWrapper) get(exec *execCtx, key *ecdsa.PrivateKey) (*object.Objec
...
 
 func (w *partWriter) WriteChunk(p []byte) error {
 	return w.chunkWriter.WriteChunk(p)
 }
 
 func (w *partWriter) WriteHeader(o *object.Object) error {
+	// Range responses use only ChunkWriter and didn't emit headers.
+	// Treat header writes as a no-op when no header writer is set so callers can
+	// still "trigger" header-dependent flows without affecting the client stream.
+	if w.headWriter == nil {
```
cthulhu-rider, line 260: seems like previously this method was not called for RANGE, otherwise NPE would occur. May u pls explain why we need this now?

pkg/services/object/get/exec.go
```diff
@@ -198,6 +206,61 @@ func (exec *execCtx) getChild(id oid.ID, rng *objectSDK.Range, withHdr bool) (*o
 	return child, ok
 }
 
+func (exec *execCtx) getChildStream(id oid.ID, rng *objectSDK.Range, withHdr bool) (*objectSDK.Object, io.ReadCloser, bool) {
```
cthulhu-rider, line 209: i was surprised to see a pipe. But it seems reasonable if we're limited to the declared interface. But according to the existing use cases, it can be avoided: the result is only needed to copy to the original stream. Therefore, this could be
```suggestion
func (exec *execCtx) copyChild(dst ObjectWriter, id oid.ID, rng *objectSDK.Range, withHdr bool) bool {
```

cthulhu-rider, line 209: do we actually need to return header? Seems like we copy payload bytes only

pkg/services/object/get/assemble.go
```diff
@@ -163,12 +163,15 @@ func (exec *execCtx) overtakePayloadDirectly(children []oid.ID, rngs []objectSDK
 			r = &rngs[i]
 		}
 
-		child, ok := exec.getChild(children[i], r, !withRng && checkRight)
+		child, reader, ok := exec.getChildStream(children[i], r, !withRng && checkRight)
 		if !ok {
+			if reader != nil {
```
cthulhu-rider, line 168: i would not expect this possible. Better to have `reader` and `ok` mutually exclusive. Otherwise all callers will have to do this. Method can close opened reader on failure itself

cthulhu-rider (commented):
viewed partially, lets discuss these ones first

pkg/services/object/get/exec.go
End-rey: I tried to do it. Is that what you meant?

pkg/services/object/get/exec.go
```diff
@@ -322,9 +374,29 @@ func (exec *execCtx) writeObjectPayload(obj *objectSDK.Object) bool {
 	return err == nil
 }
 
+// writeFromStream writes payload from stream.
+func (exec *execCtx) writeFromStream(reader io.ReadCloser) error {
```
cthulhu-rider, line 378:
```suggestion
func copyPayloadStream(w ChunkWriter, r io.Reader) error
```
would be more reusable

pkg/services/object/get/exec.go
```diff
@@ -299,12 +345,18 @@ func (exec *execCtx) writeCollectedHeader() bool {
...
-func (exec *execCtx) writeObjectPayload(obj *objectSDK.Object) bool {
+func (exec *execCtx) writeObjectPayload(obj *objectSDK.Object, reader io.ReadCloser) bool {
 	if exec.headOnly() {
 		return true
 	}
 
-	err := exec.prm.objWriter.WriteChunk(obj.Payload())
+	var err error
+	if reader != nil {
+		defer func() { _ = reader.Close() }()
```
cthulhu-rider, line 355: lets log this error, it can be interesting if occurs

pkg/services/object/get/util.go
```diff
@@ -296,3 +262,31 @@ func (h *hasherWrapper) WriteChunk(p []byte) error {
 func prettyRange(rng *object.Range) string {
 	return fmt.Sprintf("[%d:%d]", rng.GetOffset(), rng.GetLength())
 }
+
+// DirectChildWriter streams child object payload directly into destination ChunkWriter
+// while capturing the header.
+type DirectChildWriter struct {
```
cthulhu-rider, line 268:
```suggestion
type directChildWriter struct {
```
same for `NewDirectChildWriter`

pkg/services/object/get/util.go
```diff
@@ -296,3 +262,31 @@ func (h *hasherWrapper) WriteChunk(p []byte) error {
...
+	w.obj = obj
+	w.headerSet = true
+	return nil
+}
+
+func (w *DirectChildWriter) WriteChunk(p []byte) error {
+	return w.dest.WriteChunk(p)
+}
+
+func (w *DirectChildWriter) Object() *object.Object { return w.obj }
```
cthulhu-rider, line 291: no need, code has access to the field in all use cases

pkg/services/object/get/util.go
```diff
@@ -296,3 +262,31 @@ func (h *hasherWrapper) WriteChunk(p []byte) error {
 func prettyRange(rng *object.Range) string {
 	return fmt.Sprintf("[%d:%d]", rng.GetOffset(), rng.GetLength())
 }
+
+// DirectChildWriter streams child object payload directly into destination ChunkWriter
+// while capturing the header.
+type DirectChildWriter struct {
+	obj       *object.Object
```
cthulhu-rider, line 269:
```suggestion
	hdr       *object.Object
```

pkg/services/object/get/util.go
```diff
@@ -296,3 +262,31 @@ func (h *hasherWrapper) WriteChunk(p []byte) error {
...
+	dest      ChunkWriter
+	headerSet bool
+}
+
+func NewDirectChildWriter(dest ChunkWriter) *DirectChildWriter {
+	return &DirectChildWriter{
+		obj:  object.New(),
```
cthulhu-rider, line 276: why we init it? Seems like we can drop this, and then drop `HeaderSet` cuz it will be the same as `w.hdr != nil`

pkg/services/object/get/get_test.go
```diff
@@ -476,7 +497,8 @@ func generateChain(ln int, cnr cid.ID) ([]*objectSDK.Object, []oid.ID, []byte) {
 }
 
 func TestGetRemoteSmall(t *testing.T) {
-	ctx := context.Background()
+	ctx, cancel := context.WithTimeout(context.Background(), 150*time.Second)
```
cthulhu-rider, line 500: rly needed?

pkg/services/object/get/exec.go
```diff
@@ -198,6 +206,44 @@ func (exec *execCtx) getChild(id oid.ID, rng *objectSDK.Range, withHdr bool) (*o
 	return child, ok
 }
 
+// copyChild fetches child object payload and streams it directly into current exec writer.
+// Returns if child header was received and if full payload was successfully written.
+func (exec *execCtx) copyChild(id oid.ID, rng *objectSDK.Range, withHdr bool) (bool, bool) {
```
cthulhu-rider, line 211: how about `getChild` method above? Is there anything stopping u from converting it into streams?

pkg/services/object/get/get_test.go
End-rey: No, it happened during development. I thought it would be useful to add it to the tests. OK, I'll move it to a separate commit.

cthulhu-rider: > How many seconds of context do you suggest?

just leave `context.Background()`

pkg/services/object/get/exec.go
```diff
@@ -280,7 +299,7 @@ func (exec *execCtx) writeCollectedHeader() bool {
 	}
 
 	err := exec.prm.objWriter.WriteHeader(
-		exec.collectedObject.CutPayload(),
+		exec.collectedHeader.CutPayload(),
```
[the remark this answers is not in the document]
End-rey: From `forwarder` it can be.

pkg/services/object/get/exec.go
```diff
@@ -299,12 +318,23 @@ func (exec *execCtx) writeCollectedHeader() bool {
...
 	if exec.headOnly() {
 		return true
 	}
 
-	err := exec.prm.objWriter.WriteChunk(obj.Payload())
+	var err error
+	if reader != nil {
+		defer func() {
+			err := reader.Close()
+			if err == nil {
```
cthulhu-rider, line 330:
```suggestion
			if err != nil {
```

pkg/services/object/get/assemble.go
```diff
@@ -121,37 +109,76 @@ func (exec *execCtx) initFromChild(obj oid.ID) (prev *oid.ID, children []oid.ID)
...
+	for !prevID.IsZero() {
+		ph, ok := exec.headChild(prevID)
+		if !ok {
+			return nil
+		}
+		chain = append(chain, ph.GetID())
+		prevID = ph.GetPreviousID()
+	}
+	// reverse to chronological order
+	for l, r := 0, len(chain)-1; l < r; l, r = l+1, r-1 {
```
cthulhu-rider, line 175: https://pkg.go.dev/slices#Reverse

cthulhu-rider (approved):
@End-rey good!

lets cmp performance of 64K and 256K buffers. If there is no growth or it is insignificant, it makes sense to leave 64

and conflicts ofc

pkg/services/object/get/exec.go
```diff
@@ -167,35 +179,42 @@ func (exec *execCtx) headOnly() bool {
...
 
-	return child, ok
+	if !ok {
+		if hdr != nil {
+			return true, false
+		}
+		return false, false
+	}
+
+	return true, true
```
roman-khimov, line 217: `return (hdr != nil), ok`?

pkg/services/object/get/util.go
```diff
@@ -221,55 +216,28 @@ func (c *clientWrapper) get(exec *execCtx, key *ecdsa.PrivateKey) (*object.Objec
...
 	if err != nil {
-		return nil, fmt.Errorf("init object reading:: %w", err)
-	}
-
-	buf := make([]byte, hdr.PayloadSize())
-
-	_, err = rdr.Read(buf)
-	if err != nil && !errors.Is(err, io.EOF) {
-		return nil, fmt.Errorf("read payload: %w", err)
+		return nil, nil, fmt.Errorf("init object reading:: %w", err)
```
roman-khimov, line 219: `::`, btw. and it better be `reader`.

roman-khimov (approved):
Please add some high-level statistics to 256K commit message (maybe to the first one as well), we need this to be kept in git log.

carpawell (approved):

PR #3478, opened 2025-07-22
Fix/system objects without payload

carpawell:


pkg/local_object_storage/metabase/control.go
```diff
@@ -102,6 +102,8 @@ func (db *DB) init(reset bool) error {
 		}
 	}
 
+	curreEpoch := db.epochState.CurrentEpoch()
```
roman-khimov, line 105: current?

pkg/local_object_storage/metabase/exists.go
```diff
@@ -143,18 +143,18 @@ func objectStatus(tx *bbolt.Tx, metaCursor *bbolt.Cursor, addr oid.Address, curr
...
 		graveyardBkt         = tx.Bucket(graveyardBucketName)
 	)
-	return inGraveyardWithKey(metaCursor, addrKey, graveyardBkt, garbageObjectsBkt, garbageContainersBkt)
+	return inGraveyardWithKey(metaCursor, currEpoch, addrKey, graveyardBkt, garbageObjectsBkt, garbageContainersBkt)
 }
 
-func inGraveyardWithKey(metaCursor *bbolt.Cursor, addrKey []byte, graveyard, garbageObjectsBCK, garbageContainersBCK *bbolt.Bucket) uint8 {
-	if associatedWithTypedObject(metaCursor, oid.ID(addrKey[cid.Size:]), objectSDK.TypeTombstone) {
+func inGraveyardWithKey(metaCursor *bbolt.Cursor, currEpoch uint64, addrKey []byte, graveyard, garbageObjectsBCK, garbageContainersBCK *bbolt.Bucket) uint8 {
+	if associatedWithTypedObject(currEpoch, metaCursor, oid.ID(addrKey[cid.Size:]), objectSDK.TypeTombstone) {
```
roman-khimov, line 157: But tombstones are set forever. Even if they expire, removed object is removed object. I don't think we need it for tombstones.

pkg/local_object_storage/metabase/db_test.go
```diff
@@ -44,11 +45,11 @@ func testSelect(t *testing.T, db *meta.DB, cnr cid.ID, fs object.SearchFilters,
 }
 
 func newDB(t testing.TB, opts ...meta.Option) *meta.DB {
-	path := t.Name()
+	p := t.TempDir()
 
 	bdb := meta.New(
 		append([]meta.Option{
-			meta.WithPath(path),
+			meta.WithPath(path.Join(p, "meta.db")),
```
roman-khimov, line 52: Something different.

cmd/neofs-ir/main.go
```diff
@@ -106,6 +106,7 @@ func main() {
 
 	select {
 	case <-ctx.Done():
+		log.Info("application is shutting down...")
```
roman-khimov, line 109: Maybe add a cause? Somewhat generic log.

pkg/local_object_storage/metabase/db_test.go
carpawell: you mean pick another name or what?

roman-khimov: Unrelated to the patch it's included into.

cmd/neofs-ir/main.go
carpawell: ok, added

roman-khimov: You can cancel with cause then. Currently it's just `signal.NotifyContext()`, but it can be something different.

PR #3483, opened 2025-07-25
Fix/incorrect TS metabase resync

carpawell:


pkg/local_object_storage/metabase/put.go
```diff
@@ -156,7 +157,7 @@ func handleNonRegularObject(tx *bbolt.Tx, currEpoch uint64, obj objectSDK.Object
 				}
 
 				garbageObjectsBKT := tx.Bucket(garbageObjectsBucketName)
-				err = garbageObjectsBKT.Put(target[:], zeroValue)
+				err = garbageObjectsBKT.Put(addressKey(oid.NewAddress(cID, target), make([]byte, addressKeySize)), zeroValue)
```
roman-khimov, line 160: So, it's not just resync?

roman-khimov (commented):
Changelog?

pkg/local_object_storage/metabase/put.go
carpawell: current state: lock object cannot be removed by gc because of incorrect gc bucket internals -> expired lock object was not removed from blobstor -> it was resynced from time to time -> _correct_ TS cannot resync because of stuck LOCK

roman-khimov:
Replaced by #3484.

PR #3484, opened 2025-07-28
Fix incorrect metabase garbage bucket setting on tombstone writing

cthulhu-rider:
from #3483

### DIFF
--- a/pkg/local_object_storage/metabase/metadata.go
+++ b/pkg/local_object_storage/metabase/metadata.go
@@ -422,3 +422,26 @@ func (x *metaAttributeSeeker) Get(id []byte, attr string) (attributeValue []byte
 	}
 	return key[len(pref):], nil
 }
+
+// returns both zero if object is missing.
+func resolveContainerByOID(tx *bbolt.Tx, metaOIDKey []byte) (cid.ID, error) {
+	var res cid.ID
+
+	err := tx.ForEach(func(name []byte, bkt *bbolt.Bucket) error {
+		if len(name) != 1+cid.Size || name[0] != metadataPrefix {
+			return nil
+		}
+
+		if bkt.Get(metaOIDKey) != nil {
+			res = cid.ID(name[1:]) // len is checked above
+			return errBreakBucketForEach
+		}
+
+		return nil
+	})
+	if errors.Is(err, errBreakBucketForEach) {
+		err = nil
+	}
+
+	return res, err
+}
--- a/pkg/local_object_storage/metabase/version.go
+++ b/pkg/local_object_storage/metabase/version.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"slices"
 
+	"github.com/mr-tron/base58"
 	objectconfig "github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/object"
 	objectcore "github.com/nspcc-dev/neofs-node/pkg/core/object"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/util/logicerr"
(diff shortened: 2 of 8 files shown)


cthulhu-rider:
> We shouldn't have many of these, luckily. Just testnet.

yep, i also relied on this, otherwise the code'd have become more complicated with iterations

pkg/local_object_storage/metabase/garbage.go
```diff
@@ -0,0 +1,53 @@
...
+
+		if cnr.IsZero() {
+			continue // TODO: consider cleaning
+		}
+
+		if err := garbageCursor.Delete(); err != nil {
+			return fmt.Errorf("remove broken item: %w", err)
+		}
+
+		newItems[cnr] = append(newItems[cnr], k) // put after so as not to break the cursor
```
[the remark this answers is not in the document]
cthulhu-rider: could not reproduce the problem, and took into account that https://pkg.go.dev/go.etcd.io/bbolt#Cursor.Delete do not state this, like https://pkg.go.dev/go.etcd.io/bbolt#Bucket.ForEach does (although it is a cursor essentially)

well, we'll have to use more mem for the sake of safety

pkg/local_object_storage/metabase/control.go
```diff
@@ -115,10 +115,16 @@ func (db *DB) init(reset bool) error {
...
 			}
 
-			_, err := tx.CreateBucketIfNotExists(name)
+			bkt, err := tx.CreateBucketIfNotExists(name)
 			if err != nil {
 				return fmt.Errorf("could not create static bucket %s: %w", k, err)
 			}
+
+			if k == string(garbageObjectsBucketName) {
+				if err := fixGarbageBucketKeys(tx, bkt); err != nil {
```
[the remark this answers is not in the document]
cthulhu-rider: moved to migration routine

pkg/local_object_storage/metabase/garbage.go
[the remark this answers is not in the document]
cthulhu-rider: done

pkg/local_object_storage/metabase/metadata.go
```diff
@@ -422,3 +422,26 @@ func (x *metaAttributeSeeker) Get(id []byte, attr string) (attributeValue []byte
 	}
 	return key[len(pref):], nil
 }
+
+// returns both zero if object is missing.
+func resolveContainerByOID(tx *bbolt.Tx, metaOIDKey []byte) (cid.ID, error) {
```
roman-khimov, line 427: Looks like it never returns non-nil error.

pkg/local_object_storage/metabase/version.go
```diff
@@ -566,3 +568,65 @@ func migrateFrom5Version(db *DB) error {
...
+	garbageCursor := garbageBkt.Cursor()
+	for k, _ := garbageCursor.First(); k != nil; k, _ = garbageCursor.Next() {
+		if len(k) != oid.Size {
+			continue
+		}
+
+		copy(metaOIDKey[1:], k)
+
+		cnr, err := resolveContainerByOID(tx, metaOIDKey[:])
+		if err != nil {
```
roman-khimov, line 597: So this is a dead code.

roman-khimov (approved):

pkg/local_object_storage/metabase/metadata.go
cthulhu-rider: only if blindly believe that `tx.ForEach(f)` returns `f` errors only. This is not stated in docs

roman-khimov: OK, it'd be a DB error in which case failing completely makes sense.

PR #3487, opened 2025-07-29
Fix metabase resync for object + tombstone + lock case

cthulhu-rider:


pkg/local_object_storage/metabase/put.go
```diff
@@ -31,6 +31,16 @@ var (
 // associated with the locked one, i.e. there DB contains object of [objectSDK.TypeLock] type
 // associated with the tombstone target. If target is not locked, Put marks is as garbage.
 func (db *DB) Put(obj *objectSDK.Object) error {
+	return db._Put(obj, false)
+}
+
+// PutResync is similar [DB.Put] but handles objects of [objectSDK.TypeTombstone] type specifically:
+// PutResync does not check for locks and does not mark the target as garbage.
```
roman-khimov, line 38: But this will break GC, isn't it? We would also get both TS and L, so what's the status of the object? The best way to resolve this conflict is to try leveraging timestamps. Because normally it shouldn't happen at all (#3492).

cthulhu-rider: > But this will break GC, isn't it?

i need to check code, but i think GC should not break. If there is tombstone in metabase, it should work and result in the object being deleted

> We would also get both TS and L, so what's the status of the object?

Locked

> The best way to resolve this conflict is to try leveraging timestamps.

we can try, but they can be the same in T and L. I dont think there is a stable way to choose b/w L and T. Taking into account #3492, we better keep undefined behavior (depending on order) in this case. In other words, to reject this change

cthulhu-rider:
replaced with #3495

PR #3527, opened 2025-08-13
Fix bearer token cache

cthulhu-rider:


pkg/services/object/acl/v2/service.go
```diff
@@ -218,21 +218,21 @@ func (b Service) getVerifiedBearerToken(mh *protosession.RequestMetaHeader, reqC
 	res, ok := b.bearerTokenCommonCheckCache.Get(cacheKey)
 	if !ok {
 		// TODO: Signed data is used twice - for cache key and to check the signature. Coding can be deduplicated.
-		res.token, res.err = b.decodeAndVerifyBearerTokenCommon(m, ownerCnr)
+		res.token, res.err = b.decodeAndVerifyBearerTokenCommon(m)
```
carpawell, line 221: ist bearer->eacl table->container->owner a const chain for any token?

cthulhu-rider: it is, and that's what tricked me. `ownerCnr` is an owner of the container from request body

PR #3532, opened 2025-08-15
Storage `GetStream` API

End-rey:
Part of #3466.

### DIFF
--- a/pkg/services/object/get/util.go
+++ b/pkg/services/object/get/util.go
@@ -219,7 +219,18 @@ func (e *storageEngineWrapper) get(exec *execCtx) (*object.Object, error) {
 		return o, nil
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/metrics/engine.go
```diff
@@ -72,6 +73,13 @@ func newEngineMetrics() engineMetrics {
 			Help:      "Engine 'head' operations handling time",
 		})
 
+		streamDuration = prometheus.NewHistogram(prometheus.HistogramOpts{
```
carpawell, line 76: hm, i look at it and think if it can be useful as a metric that just answers how long we were acquiring an object stream, but not a streaming duration itself. mb that is ok

carpawell (approved):

PR #3536, opened 2025-08-19
test: Utilize checking `zap.Logger` entries

cthulhu-rider:


cthulhu-rider:
one more place came with #3531, adding...

internal/testutil/log.go
```diff
@@ -0,0 +1,111 @@
...
+		require.NoError(x.t, err, i)
+
+		v, ok = m[logMessageKey]
+		require.True(x.t, ok, i)
+		require.IsType(x.t, "", v)
+		res[i].Message = v.(string)
+
+		delete(m, logTimeKey)
+		delete(m, logLevelKey)
+		delete(m, logMessageKey)
```
carpawell, line 106: i would add doc info that messages are dropped after every `Assert*`, this requires testing asserts (if many) without repetitions, and i think it is important

internal/testutil/log.go
```diff
@@ -0,0 +1,111 @@
...
+func (x *LogBuffer) AssertEmpty() {
+	x.AssertEqual(nil)
+}
+
+// AssertSingle asserts that log has given entry only.
+func (x *LogBuffer) AssertSingle(e LogEntry) {
+	x.AssertEqual([]LogEntry{e})
+}
+
+// AssertEqual asserts that log has exact entries.
```
carpawell, line 66: can you add a few words that the order matters? i guess we may have smth like `Match` in the future for parallel testing logs when the order is not defined

internal/testutil/log.go
cthulhu-rider: they are never dropped now

internal/testutil/log.go
cthulhu-rider: sure, done

carpawell (approved):

End-rey (approved):

PR #3553, opened 2025-08-29
Refactor policy processing in Policer

cthulhu-rider:


cthulhu-rider:
i want to add unit tests to prevent regressions and later extend for EC rules

pkg/local_object_storage/metabase/iterators.go
```diff
@@ -177,68 +176,6 @@ func (db *DB) iterateExpired(tx *bbolt.Tx, curEpoch uint64, h ExpiredObjectHandl
...
 
-// IterateCoveredByTombstones iterates over all objects in DB which are covered
-// by tombstone with string address from tss. Locked objects are not included
-// (do not confuse with objects of type LOCK).
-//
-// If h returns ErrInterruptIterator, nil returns immediately.
-// Returns other errors of h directly.
-//
-// Does not modify tss.
-func (db *DB) IterateCoveredByTombstones(tss map[string]oid.Address, h func(oid.Address) error) error {
```
roman-khimov, line 188: @End-rey, is it of any use for GC?

pkg/services/policer/policer.go
```diff
@@ -140,8 +140,6 @@ func New(opts ...Option) *Policer {
 		opts[i](c)
 	}
 
-	c.log = c.log.With(zap.String("component", "Object Policer"))
```
roman-khimov, line 142: I think this breaks common pattern used by other services.

pkg/services/policer/check.go
```diff
@@ -79,12 +79,13 @@ func (p *Policer) processObject(ctx context.Context, addrWithType objectcore.Add
...
 
-	cnr, err := p.cnrSrc.Get(idCnr)
+	nn, repRules, err := p.network.GetNodesForObject(addr)
 	if err != nil {
-		p.log.Error("could not get container",
+		p.log.Error("could not build placement vector for object",
 			zap.Stringer("cid", idCnr),
 			zap.Error(err),
 		)
+
```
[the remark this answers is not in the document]
cthulhu-rider: occasion, reverted

pkg/local_object_storage/metabase/iterators.go
End-rey: As I understand it, we are mainly looking at the lifetime of an object, and therefore there is no point in iterating over tombstones.

PR #3570, opened 2025-09-09
Support EC in container storage policies

cthulhu-rider:
* based on https://github.com/nspcc-dev/neofs-sdk-go/pull/736


roman-khimov:
Obviously I asked before looking into the code. OK then.

PR #3582, opened 2025-09-17
Decouple expired object collector from epoch ticker

### LINKED ISSUE #3522: Decouple expired object collector from epoch ticker
I'm always frustrated when I'm thinking of how expired objects are handled (and even more frustrated when I'm fixing thing as in #3511):
 * expired objects are just another sort of garbage
 * there can be 1M or more objects expired in an epoch
 * marking them as garbage separately from deletion just adds useless pressure to metabase

1. Move their handling from epoch event handler to regular GC cycle. Epoch event just sets some known epoch and that's it.
2. Grab a number of them (same limit as in GC) when doing regular GC cycle.
3. Handle them as usual, if we find that some object shouldn't be deleted (a lock is stored on the other shard), copy this lock to current shard.

No alternatives.

End-rey:
Closes #3522.

### DIFF
--- a/pkg/local_object_storage/engine/gc_test.go
+++ b/pkg/local_object_storage/engine/gc_test.go
@@ -11,14 +11,12 @@ import (
 	"github.com/nspcc-dev/neofs-node/pkg/core/object"
 	meta "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/metabase"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/shard"
-	"github.com/nspcc-dev/neofs-node/pkg/util"
 	statusSDK "github.com/nspcc-dev/neofs-sdk-go/client/status"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 	cidtest "github.com/nspcc-dev/neofs-sdk-go/container/id/test"
 	objectSDK "github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	usertest "github.com/nspcc-dev/neofs-sdk-go/user/test"
-	"github.com/panjf2000/ants/v2"
 	"github.com/stretchr/testify/require"
 )
 
@@ -37,14 +35,6 @@ func TestChildrenExpiration(t *testing.T) {
 				meta.WithEpochState(es),
 			),
 			shard.WithExpiredObjectsCallback(e.processExpiredObjects),
-			shard.WithGCWorkerPoolInitializer(func(sz int) util.WorkerPool {
-				pool, err := ants.NewPool(sz)
-				if err != nil {
-					panic(err)
-				}
-
-				return pool
-			}),
 			shard.WithGCRemoverSleepInterval(100*time.Millisecond),
 		)
 		require.NoError(t, err)
--- a/pkg/local_object_storage/shard/gc.go
+++ b/pkg/local_object_storage/shard/gc.go
@@ -7,7 +7,6 @@ import (
 
 	meta "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/metabase"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/shard/mode"
-	"github.com/nspcc-dev/neofs-node/pkg/util"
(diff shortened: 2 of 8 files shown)


pkg/local_object_storage/shard/gc.go
```diff
@@ -206,108 +276,9 @@ func (s *Shard) removeGarbage() {
...
-	}
-	return expired, nil
-}
-
-// HandleExpiredLocks unlocks all objects which were locked by lockers.
-// If successful, marks lockers themselves as garbage.
+// FreeLockedBy unlocks all objects which were locked by lockers.
 // Returns every object that is unlocked.
-func (s *Shard) HandleExpiredLocks(lockers []oid.Address) []oid.Address {
+func (s *Shard) FreeLockedBy(lockers []oid.Address) []oid.Address {
```
roman-khimov, line 243: Why do we still have this one, why deleting a lock is not sufficient?

roman-khimov (commented):
@cthulhu-rider, there can be some intersections with #3549.

pkg/local_object_storage/engine/gc_test.go
```diff
@@ -165,3 +167,114 @@ func checkObjectsAsyncRemoval(t *testing.T, e *StorageEngine, cnr cid.ID, objs .
...
+
+		_, err := e.Get(objAddr)
+		require.Error(t, err, statusSDK.ErrObjectAlreadyRemoved)
+
+		e.HandleNewEpoch(currEpoch + 1)
+
+		require.Eventually(t, func() bool {
+			_, err := e.Get(objAddr)
+			return errors.Is(err, statusSDK.ErrObjectNotFound)
+		}, 3*time.Second, 100*time.Millisecond)
```
cthulhu-rider, line 268: why wait time and interval varies? We have `1s` (`checkObjectsAsyncRemoval`), `2s` and `3s`; `100ms` and `200ms`

pkg/local_object_storage/shard/gc.go
```diff
@@ -119,24 +98,11 @@ func (gc *gc) listenEvents() {
 			}
 
 			v.prevGroup.Wait()
-
-			v.prevGroup.Add(len(v.handlers))
-
-			for i := range v.handlers {
-				h := v.handlers[i]
-
-				err := gc.workerPool.Submit(func() {
```
cthulhu-rider, line 120: `drop unused parameters` commit message is not entirely true. We used worker pool, now we don't

pkg/local_object_storage/shard/gc.go
```diff
@@ -119,24 +98,11 @@ func (gc *gc) listenEvents() {
...
-					gc.log.Warn("could not submit GC job to worker pool",
-						zap.Error(err),
-					)
-
-					v.prevGroup.Done()
-				}
-			}
+			v.prevGroup.Add(1)
+			go func() {
+				v.handler(event)
```
cthulhu-rider, line 103: moving from `v.handlers` slice to single `v.handler` is not actually related to the pool drop patch, it's an autonomous refactoring

pkg/local_object_storage/shard/gc.go
```diff
@@ -206,108 +214,83 @@ func (s *Shard) removeGarbage() {
...
+func (s *Shard) collectExpiredObjects() {
+	epoch := s.gc.currentEpoch.Load()
+	doneUpTo := s.gc.processedEpoch.Load()
+	if s.info.Mode.NoMetabase() || doneUpTo == epoch {
 		return
 	}
-	if len(expired) == 0 {
-		log.Debug("no expired objects")
+	if doneUpTo > epoch {
+		s.log.Warn("current epoch is less than the last processed epoch",
```
roman-khimov, line 186: It'd be strange to see this, but at least GC should be mentioned then, otherwise this message doesn't have a lot of meaning.

pkg/local_object_storage/shard/delete.go
```diff
@@ -59,3 +59,10 @@ func (s *Shard) deleteObjs(addrs []oid.Address) error {
 
 	return nil
 }
+
+// DeleteNow deletes provided objects immediately without prior garbage marking.
+func (s *Shard) DeleteNow(addrs ...oid.Address) error {
```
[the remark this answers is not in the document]
roman-khimov: Still here, btw. Used in tests only.

PR #3584, opened 2025-09-17
ir: fix not using the fee parameter

### LINKED ISSUE #3579: IR `fee.main_chain` config is unused
seems like it is unused since 8150a2e2ea85e849aeba23913a8e2b12028639ea. Better to drop it then

End-rey:
Closes #3579.

### DIFF
--- a/pkg/morph/client/neofs/client.go
+++ b/pkg/morph/client/neofs/client.go
@@ -34,6 +34,7 @@ func NewFromMorph(cli *client.Client, contract util.Uint160, fee fixedn.Fixed8, 
 	for i := range opts {
 		opts[i](o)
 	}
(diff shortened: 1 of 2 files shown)


pkg/morph/client/neofs/client.go
```diff
@@ -28,7 +28,7 @@ const (
 )
 
 // NewFromMorph wraps client to work with NeoFS contract.
-func NewFromMorph(cli *client.Client, contract util.Uint160, fee fixedn.Fixed8, opts ...Option) (*Client, error) {
+func NewFromMorph(cli *client.Client, contract util.Uint160, opts ...Option) (*Client, error) {
```
roman-khimov, line 31: I think this can be kept as is, just handle client.WithStaticFeeIncrement internally.

PR #3598, opened 2025-09-24
refactor: use slices.Contains to simplify code

russcoss:
There is a [new function](https://pkg.go.dev/slices@go1.21.0#Contains) added in the go1.21 standard library, which can make the code more concise and easy to read.

### DIFF
--- a/pkg/services/object/acl/v2/util_test.go
+++ b/pkg/services/object/acl/v2/util_test.go
@@ -1,6 +1,7 @@
 package v2
 
 import (
+	"slices"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


pkg/services/object/acl/v2/util_test.go
```diff
@@ -43,11 +44,8 @@ func TestIsVerbCompatible(t *testing.T) {
 	for op, list := range table {
 		for _, verb := range verbs {
 			var contains bool
```
roman-khimov, line 46: `var contains = slices.Contains(list, verb)` then, no need for `if`.

russcoss: Good idea. Modified.

PR #3600, opened 2025-09-25
EC RANGE

cthulhu-rider:


pkg/services/object/get/ec.go
```diff
@@ -355,6 +361,506 @@ func (s *Service) getECPartFromNode(ctx context.Context, cnr cid.ID, parent oid.
...
+			return written, fmt.Errorf("%w: write [%d:%d] range of EC part #%d: %w", errStreamFailure, from, to, partIdx, err)
+		}
+
+		written += to - from
+	}
+
+	return written, nil
+}
+
+func (s *Service) getECParentPayloadLen(ctx context.Context, localNodeKey ecdsa.PrivateKey, cnr cid.ID, id oid.ID, sTok *session.Object,
```
cthulhu-rider, line 551: it's worth noting that method of obtaining parent's payload len will change to GET first part in 17ca59f4af6015df6ad859bf3eca65e99dd62f2a

i'd change it right here, but for now I'll leave the original version

roman-khimov (approved):

PR #3602, opened 2025-09-26
adm/fschain: do not mess with flags requirements

carpawell:


carpawell:
Full run will be there: https://github.com/nspcc-dev/neofs-node/actions/runs/18035755908

roman-khimov (approved):

PR #3603, opened 2025-09-26
node/objectsvc: clean up signatures correctly

carpawell:
Placement vectors must be the same length, but signatures should be cleaned up correctly. Problem was added in a9202b1c6b8776beb2ff8276144d7bb947934e00.


carpawell:
No meta problems: https://rest.fs.neo.org/HXSaMJXk2g8C14ht8HSi7BBaiYZ1HeWh2xnWPGQCg4H6/4085-1758910925/index.html#

PR #3606, opened 2025-09-30
SDK 15

roman-khimov:


CHANGELOG.md
```diff
@@ -33,9 +33,10 @@ Changelog for NeoFS Node
 - `neofs-cli object head --main-only` no-op flag (#3509)
 - IR's estimations timer management (#3539)
 - `neofs-adm fschain estimations` command (#3572)
+- `--oid` flag from CLI object search commands (#3603)
```
carpawell, line 36: what does this refer to? same below

pkg/services/object/server.go
```diff
@@ -2074,8 +2082,12 @@ func (s *Server) ProcessSearch(ctx context.Context, req *protoobject.SearchV2Req
 			res[i].Attributes = nil
 		}
 	}
+	var incValue = incomplete.Load()
+	if incValue == nil {
```
carpawell, line 2164:
```
if incValue != nil {
	return res, newCursor, incValue.(error)
}
```
?

pkg/services/object/put/distributed.go
```diff
@@ -155,6 +155,9 @@ func (t *distributedTarget) Close() (oid.ID, error) {
 
 	err := t.saveObject(*t.obj, objMeta, t.encodedObject)
 	if err != nil {
+		if errors.Is(err, apistatus.ErrIncomplete) {
+			return t.obj.GetID(), nil
```
carpawell, line 159: so we do not return an incomplete error in that case? or i got smth wrong?

carpawell (commented):
Do we need additional support in `neofs-cli`? https://github.com/nspcc-dev/neofs-node/blob/582380b995ff920c1cd4fe1384abb84e538dd4e8/cmd/neofs-cli/internal/common/exit.go#L27-L50

CHANGELOG.md
roman-khimov: Ah, this was sitting for a while in the queue, to be fixed.

pkg/services/object/put/distributed.go
roman-khimov: Good catch, fixed.

roman-khimov:
We have a lot of other codes as well, but we only check for some specific ones in CLI. I don't have any strict preferences here for now.

roman-khimov:
Added incomplete/busy codes to CLI.

PR #3607, opened 2025-09-30
EC+Split

cthulhu-rider:
* based on #3600

### DIFF
--- a/pkg/services/object/get/ec.go
+++ b/pkg/services/object/get/ec.go
@@ -293,6 +293,8 @@ func (s *Service) getECPart(ctx context.Context, cnr cid.ID, parent oid.ID, sTok
 	if _, err := io.ReadFull(rc, buf); err != nil {
 		if errors.Is(err, io.EOF) { // empty payload is caught above
 			err = io.ErrUnexpectedEOF
+		} else {
+			err = convertContextStatus(err)
 		}
 		return object.Object{}, nil, fmt.Errorf("read full payload: %w", err)
 	}
--- a/pkg/services/object/get/ec.go
+++ b/pkg/services/object/get/ec.go
@@ -50,20 +50,6 @@ func (s *Service) copyLocalECPart(dst ObjectWriter, cnr cid.ID, parent oid.ID, p
 	return nil
 }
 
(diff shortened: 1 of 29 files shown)


pkg/services/object/put/distributed.go
```diff
@@ -178,7 +178,7 @@ func (t *distributedTarget) saveObject(obj objectSDK.Object, objMeta object.Cont
 
 	repRules := t.containerNodes.PrimaryCounts()
 	ecRules := t.containerNodes.ECRules()
-	if typ := obj.Type(); len(repRules) > 0 || typ == objectSDK.TypeTombstone || typ == objectSDK.TypeLock {
+	if typ := obj.Type(); len(repRules) > 0 || typ == objectSDK.TypeTombstone || typ == objectSDK.TypeLock || typ == objectSDK.TypeLink {
```
roman-khimov: Links can have quite some payload. Although typically I'd expect <10GB objects and that are just ~6K. Still, consider 60M link, can it be split?

pkg/services/object/get/ec.go
```diff
@@ -192,7 +184,77 @@ func (s *Service) restoreFromECParts(ctx context.Context, cnr cid.ID, parent oid
...
+	}
+
+	if err := dst.WriteHeader(parentHdr); err != nil {
+		return fmt.Errorf("%w: write parent header: %w", errStreamFailure, err)
+	}
+
+	var partTimeout time.Duration
+	if deadline, ok := ctx.Deadline(); ok {
+		// don't take into account that last part may be tiny
+		if partTimeout = time.Until(deadline) / time.Duration(len(sizeSplitParts)); partTimeout <= 0 {
```
roman-khimov: When do you expect this condition to be met? Why do we care at all? What if one object is fetching slowly, but others are OK?

cthulhu-rider: > When do you expect this condition to be met?

if context is almost done. Too small timeout is set by the client, or network took most of it

> Why do we care at all? What if one object is fetching slowly, but others are OK?

The delay for each part is most likely the same, so it's unlikely OK to download everything before the deadline. Although, in rare cases, it might be possible. In general, we don't have to do this

roman-khimov (approved):

PR #3609, opened 2025-10-01
Feat/multiple reports per epoch

carpawell:


pkg/timers/timer.go
```diff
@@ -0,0 +1,114 @@
+package timers
+
+import (
+	"fmt"
+	"sync"
+)
+
+type (
+	// Tick is a hander that must be executed every time any of [EpochTimers]
```
cthulhu-rider, line 9:
```suggestion
	// Tick is a handler that must be executed every time any of [EpochTimers]
```

cmd/neofs-node/container.go
```diff
@@ -79,11 +83,135 @@ func initContainerService(c *cfg) {
...
+	var (
+		m     sync.RWMutex
+		cache = make(map[cid.ID]report)
+	)
+
+	return func() {
+		epoch := c.CurrentEpoch()
 		st := c.cfgObject.cfgLocalStorage.localStorage
-		l := c.log.With(zap.String("component", "containerReports"), zap.Uint64("epoch", ev.EpochNumber()))
+		l = l.With(zap.Uint64("epoch", epoch))
```
cthulhu-rider, line 214:
```suggestion
		l := l.With(zap.Uint64("epoch", epoch))
```

cmd/neofs-node/container.go
```diff
@@ -79,11 +83,135 @@ func initContainerService(c *cfg) {
...
+			return
+		}
+		height, err := c.cli.GetBlockCount()
+		if err != nil {
+			l.Warn("failed to get the current height to reset epoch timer", zap.Error(err))
+			return
+		}
+		currH, err := c.cli.GetBlockHeader(height - 1)
+		if err != nil {
+			l.Warn("failed to get the current block", zap.Error(err))
```
cthulhu-rider, line 193:
```suggestion
			l.Warn("failed to get the current block to reset epoch timer", zap.Error(err))
```

cmd/neofs-node/container.go
```diff
@@ -79,11 +83,135 @@ func initContainerService(c *cfg) {
...
+	l := c.log.With(zap.String("component", "containerReports"))
+
+	// hardcoded in neofs-contracts as of 957152c7da6eb0a20745cfa1c4513b3c3512db6e
+	const maxReportsPerEpoch = 3
+
+	nm, err := c.nCli.NetMap()
+	if err != nil {
+		fatalOnErr(fmt.Errorf("failed to get netmap to initialize container service: %w", err))
+	}
+	durrSec, err := c.nCli.EpochDuration()
```
cthulhu-rider, line 108:
```suggestion
	durSec, err := c.nCli.EpochDuration()
```

cmd/neofs-node/container.go
```diff
@@ -79,11 +83,135 @@ func initContainerService(c *cfg) {
...
+	if indexInNM == -1 {
+		indexInNM = nmLen
+	}
+
+	var (
+		reportStep   time.Duration
+		stepsInEpoch uint32
+	)
+	if nmLen != 0 {
+		reportStep = min(time.Second, durr/maxReportsPerEpoch/time.Duration(nmLen))
```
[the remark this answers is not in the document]
roman-khimov: OK, let's try it this way.

config/example/ir.yaml
```diff
@@ -146,9 +146,6 @@ timers:
   collect_basic_income:
     mul: 1 # Multiplier in x/y relation of when to start basic income asset collection within the epoch
     div: 2 # Divider in x/y relation of when to start basic income asset collecting within the epoch
-  distribute_basic_income:
```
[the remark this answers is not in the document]
carpawell: btw, so much pain with changing it, mb it is worth a separate PR

cmd/neofs-node/container.go
carpawell: does it help with smth?

cmd/neofs-node/container.go
```diff
@@ -79,11 +85,129 @@ func initContainerService(c *cfg) {
...
+	indexInNM := slices.IndexFunc(nm.Nodes(), func(node netmapsdk.NodeInfo) bool {
+		return bytes.Equal(node.PublicKey(), c.binPublicKey)
+	})
+	if indexInNM == -1 {
+		indexInNM = nmLen
+	}
+
+	var stepsInEpoch uint32
+	if nmLen != 0 {
+		// likely there is no need to report less often than once per second
```
carpawell, line 126: @roman-khimov, do you agree with it? another suggestion is block time as the first arg in `min`

roman-khimov: It's ok.

cmd/neofs-node/container.go
```diff
@@ -79,11 +83,135 @@ func initContainerService(c *cfg) {
...
+	}
+	durr := time.Duration(durrSec) * time.Second
+
+	var (
+		indexInNM = -1
+		nmLen     = len(nm.Nodes())
+	)
+	for i, node := range nm.Nodes() {
+		if bytes.Equal(node.PublicKey(), c.binPublicKey) {
+			indexInNM = i
```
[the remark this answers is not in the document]
roman-khimov: What we need here is to have nodes reporting at different times (not all 1K nodes pushing data for 100K containers at the same time). In that sense we don't care about particular order, netmap or even specific time, we just need different nodes to use some different time offsets. If this scheme provides that, it's ok.

cmd/neofs-node/container.go
cthulhu-rider: not clone the clone

https://go.dev/play/p/8_am7FVgVsW

cmd/neofs-node/container.go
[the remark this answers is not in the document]
carpawell: i guess anything that returns iterator.Iterator is always sorted, that is how contracts and storage.Find work in VM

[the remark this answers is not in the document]
cthulhu-rider: anyway, as u noticed, netmap can change. So nvm

cthulhu-rider (approved):

PR #3611, opened 2025-10-02
morph: fix parsing container load reports

carpawell:
It has an additional field now after https://github.com/nspcc-dev/neofs-contract/pull/517.


carpawell:
full will be there: https://github.com/nspcc-dev/neofs-node/actions/runs/18205391751

PR #3612, opened 2025-10-03
Fix EC PUT broadcast

cthulhu-rider:


pkg/services/object/put/distributed.go
```diff
@@ -177,16 +177,27 @@ func (t *distributedTarget) saveObject(obj objectSDK.Object, objMeta object.Cont
...
+	ecRules := t.containerNodes.ECRules()
 	if typ := obj.Type(); len(repRules) > 0 || typ == objectSDK.TypeTombstone || typ == objectSDK.TypeLock {
 		broadcast := typ == objectSDK.TypeTombstone || typ == objectSDK.TypeLink || (!t.localOnly && typ == objectSDK.TypeLock) || len(obj.Children()) > 0
+
+		useRepRules := repRules
+		if broadcast && len(ecRules) > 0 {
+			useRepRules = make([]uint, len(repRules)+len(ecRules))
+			copy(useRepRules, repRules)
+			for i := range ecRules {
+				useRepRules[len(repRules)+i] = 1
```
roman-khimov, line 189: Technically it's even more important to push ts/lock to all nodes in EC group, how about data+parity here instead of 1?

cthulhu-rider: pretty reasonable. Done

PR #3614, opened 2025-10-03
EC HEAD

cthulhu-rider:


pkg/services/object/get/util.go
```diff
@@ -368,6 +357,47 @@ func (c *clientCacheWrapper) InitGetObjectStream(ctx context.Context, node netma
 	return hdr, rc, nil
 }
 
+func (c *clientCacheWrapper) Head(ctx context.Context, node netmap.NodeInfo, pk ecdsa.PrivateKey, cnr cid.ID, id oid.ID,
+	sTok *session.Object) (object.Object, error) {
```
roman-khimov, line 361: No bearer? It seems to be a generic Head, not EC-specific one. It should support bearer, but then EC can never pass it.

pkg/services/object/get/ec.go
```diff
@@ -60,6 +62,129 @@ func (s *Service) copyECObject(ctx context.Context, cnr cid.ID, parent oid.ID, s
...
+}
+
+func (s *Service) getECObjectHeader(ctx context.Context, cnr cid.ID, id oid.ID, sTok *session.Object,
+	ecRules []iec.Rule, sortedNodeLists [][]netmap.NodeInfo) (object.Object, error) {
+	localNodeKey, err := s.keyStore.GetKey(nil)
+	if err != nil {
+		return object.Object{}, fmt.Errorf("get local SN private key: %w", err)
+	}
+
+	// TODO: sort EC rules by complexity and try simpler ones first. Note that rule idxs passed as arguments must be kept.
```
roman-khimov, line 86: Doubt it changes much for HEAD, especially if we're to consider https://github.com/nspcc-dev/neofs-api/issues/348.

pkg/services/object/get/ec.go
```diff
@@ -60,6 +62,129 @@ func (s *Service) copyECObject(ctx context.Context, cnr cid.ID, parent oid.ID, s
...
+		hdr, err := s.getECObjectHeaderByRule(ctx, *localNodeKey, cnr, id, sTok, ecRules[i], sortedNodeLists[i])
+		if err == nil || errors.Is(err, apistatus.ErrObjectAlreadyRemoved) || errors.Is(err, ctx.Err()) {
+			return hdr, err
+		}
+
+		if i == 0 {
+			firstErr = err
+		}
+
+		s.log.Info("failed to fetch object header by EC rule",
```
roman-khimov, line 100: I doubt we do the same for HEADs, some random OID can flood the logs quickly this way.

pkg/services/object/get/ec.go
```diff
@@ -60,6 +62,129 @@ func (s *Service) copyECObject(ctx context.Context, cnr cid.ID, parent oid.ID, s
...
+		node := sortedNodes[i]
+		eg.Go(func() error {
+			if err := ctx.Err(); err != nil {
+				// errgroup call the function even when context is done. Routines for backup
+				// nodes should catch this asap to do nothing.
+				return err
+			}
+
+			if s.neoFSNet.IsLocalNodePublicKey(node.PublicKey()) {
+				// TODO: it'd be more efficient to get payload len only, not a full header
```
roman-khimov, line 129: Isn't header is what we need here?

pkg/services/object/get/ec.go
```diff
@@ -60,6 +62,129 @@ func (s *Service) copyECObject(ctx context.Context, cnr cid.ID, parent oid.ID, s
...
+
+	var hdrTarget atomic.Value
+	var firstErrTarget atomic.Value
+
+	eg, ctx := errgroup.WithContext(ctx)
+	eg.SetLimit(min(totalParts, 3))
+
+	for i := range sortedNodes {
+		node := sortedNodes[i]
+		eg.Go(func() error {
```
roman-khimov, line 121: Concurrent execution is an overkill here. Every part has the header, so chances of the first try to succeed are very high.

cthulhu-rider: reasonable, done

pkg/services/object/get/ec.go
cthulhu-rider: copy-paste, dropped

pkg/services/object/get/ec.go
cthulhu-rider: ok, debug

pkg/services/object/get/util.go
roman-khimov: I was sure it's there in `InitGetObjectStream`, but seems like it's not. OK, let's not touch this for now. Wrapped wrappers of wrappers in another issue.

roman-khimov (approved):

PR #3615, opened 2025-10-03
Fix/object counters with new GC

carpawell:


pkg/local_object_storage/metabase/delete.go
```diff
@@ -154,6 +154,16 @@ func (db *DB) delete(tx *bbolt.Tx, addr oid.Address, currEpoch uint64) (bool, bo
 		return false, false, 0, fmt.Errorf("can't remove metadata indexes: %w", err)
 	}
 
+	// if object is not available, counters have already been handled in
+	// `Inhume` call, but if we are removing an available object, this is
+	// either a force removal or GC work, and counters must be kept up-to-date
+	if removeAvailableObject {
+		err = changeContainerInfo(tx, cID, -int(payloadSize), -1)
+		if err != nil {
+			return true, removeAvailableObject, payloadSize, fmt.Errorf("can't update container info: %w", err)
```
cthulhu-rider, line 161: no point in non-zero results with error

carpawell: for some reason, this time i decided to highlight that this is not critical and returned values are ok, but container info changing routine failed. dont mind dropping non-zeor values

roman-khimov (approved):

PR #3619, opened 2025-10-03
ir: move `fschain_autodeploy` out of top level

### LINKED ISSUE #3361: Move `fschain_autodeploy` out of top level
I'm always frustrated when:
 * we have a config that is `true` for all networks yet it has to be configured explicitly
 * we have a lot of top level configs without logical grouping

1. Create an `fschain.disable_autodeploy` with an inverse meaning, we use autodeploy by default, but don't if `disable_autodeploy` is true. This config shouldn't be used normally, only in case of emergency/node failing to work with autodeploy for whatever reason (although practically it's super-stable now).
2. Make `fschain_autodeploy` default to `true`, but support it for one more release (simplifying transition).
3. In subsequent release drop `fschain_autodeploy` completely.

Keeping things as is is not very convenient.

## Additional context
#2423.

End-rey:
Closes #3361.


CHANGELOG.md
```diff
@@ -64,6 +65,9 @@ Drop `--main-only` flag from all `neofs-cli object head` commands.
 Drop IR's `timers.stop_estimation.*` and `timers.distribute_basic_income` configuration values, they are not used anymore.
 `neofs-adm fschain estimations` was removed.
 
+Use new `fschain.disable_autodeploy` IR configuration option instead of deprecated `fschain_autodeploy`,
+that was removed in next release.
```
carpawell, line 69: mention that default value for the old config has changed?

carpawell: but in fact, i think it is useless work, we will just drop it in the next release, nobody gonna change this exact value in their configs, they will just use the new one

pkg/innerring/innerring.go
```diff
@@ -418,6 +418,18 @@ func New(ctx context.Context, log *zap.Logger, cfg *config.Config, errChan chan<
 
 	serveMetrics(server, cfg)
 
+	// TODO: remove deprecated option in future releases
```
carpawell, line 421: issue?

pkg/innerring/innerring.go
```diff
@@ -418,6 +418,18 @@ func New(ctx context.Context, log *zap.Logger, cfg *config.Config, errChan chan<
...
+	if cfg.IsSet("fschain_autodeploy") && cfg.IsSet("fschain.disable_autodeploy") &&
```
carpawell, line 422: i think any case when both values are set can be reported as an error, this is smth abnormal to me

pkg/innerring/innerring.go
```diff
@@ -418,6 +418,15 @@ func New(ctx context.Context, log *zap.Logger, cfg *config.Config, errChan chan<
 
 	serveMetrics(server, cfg)
 
+	// TODO: remove deprecated 'fschain_autodeploy' config option in future release
+	if cfg.IsSet("fschain_autodeploy") && cfg.IsSet("fschain.disable_autodeploy") {
+		return nil, fmt.Errorf("'fschain_autodeploy' and 'fschain.disable_autodeploy' set simultaneously")
+	}
+	if cfg.IsSet("fschain_autodeploy") {
```
cthulhu-rider, line 425: may be merged with the previous `if`. Not a big deal taking into account further removal of both blocks

PR #3649, opened 2025-10-23
adm: add storage load inspection command

carpawell:


cmd/neofs-adm/internal/modules/fschain/load.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	if len(items) == 0 {
+		return errNoMoreVectors
+	}
+
+	var r cnrrpc.ContainerNodeReport
+	for i, item := range items {
+		err := r.FromStackItem(item)
+		if err != nil {
+			return fmt.Errorf("reading %d stackitem of as raw node report: %w", i, err)
```
End-rey: `of as` confuses me

PR #3650, opened 2025-10-24
Feat/unpaid containers handling

carpawell:


carpawell:
Still not sure about dropping container immediately after it is unpaid. How long to wait?

cmd/neofs-node/container.go
```diff
@@ -351,6 +354,82 @@ func setContainerNotificationParser(c *cfg, sTyp string, p event.NotificationPar
...
+
+		cID := cid.ID(ev.ContainerID)
+		if ev.Unpaid {
+			l.Info("container status has changed to unpaid",
+				zap.Stringer("cID", cID),
+				zap.Uint64("epoch", ev.Epoch))
+
+			p.statuses[cID] = int64(ev.Epoch)
+		} else {
+			l.Debug("container status has changed to paid",
```
roman-khimov: Info?

Consider the following scenario as well:
 * container gets "unpaid" status
 * all data is deleted from nodes
 * its size is zero, so there are no settlements to make
 * but its status is still "unpaid" even if owner has some balance

Can this be a problem?

cmd/neofs-node/container.go
```diff
@@ -351,6 +354,82 @@ func setContainerNotificationParser(c *cfg, sTyp string, p event.NotificationPar
...
+			m:          sync.RWMutex{},
+			statuses:   make(map[cid.ID]int64),
+			balanceCli: c.bCli,
+		}
+	)
+
+	const changeUnpaidStatusEventName = "ChangePaymentStatus"
+	typ := event.TypeFromString(changeUnpaidStatusEventName)
+	c.cfgBalance.parsers[typ] = balanceEvent.ParseChangePaymentStatus
+	c.cfgBalance.subscribers[typ] = append(c.cfgBalance.subscribers[typ], event.WorkerPoolHandler(c.cfgBalance.workerPool, func(e event.Event) {
```
roman-khimov: Frankly, a pool of workers is too much for the thing they do (change some status in a map).

pkg/local_object_storage/shard/gc.go
```diff
@@ -97,9 +97,11 @@ func (gc *gc) listenEvents() {
 
 			v.prevGroup.Wait()
 			v.prevGroup.Add(1)
+			gc.wg.Add(1)
```
roman-khimov: Seems wrong. Rather it should be some delayed `v.prevGroup.Wait()` since `gc.wg` has exactly two things to care about. Duplicating an item in different work groups doesn't seem to be a good practice.

roman-khimov (approved):

PR #3671, opened 2025-11-07
Session token V2

End-rey:
Refs https://github.com/nspcc-dev/neofs-api/pull/350, https://github.com/nspcc-dev/neofs-sdk-go/pull/750

### DIFF
--- a/cmd/neofs-node/object.go
+++ b/cmd/neofs-node/object.go
@@ -17,6 +17,7 @@ import (
 	coreclient "github.com/nspcc-dev/neofs-node/pkg/core/client"
 	containercore "github.com/nspcc-dev/neofs-node/pkg/core/container"
 	"github.com/nspcc-dev/neofs-node/pkg/core/netmap"
+	"github.com/nspcc-dev/neofs-node/pkg/core/nns"
 	objectcore "github.com/nspcc-dev/neofs-node/pkg/core/object"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/engine"
 	morphClient "github.com/nspcc-dev/neofs-node/pkg/morph/client"
@@ -49,6 +50,7 @@ import (
 	protoobject "github.com/nspcc-dev/neofs-sdk-go/proto/object"
 	protosession "github.com/nspcc-dev/neofs-sdk-go/proto/session"
 	apireputation "github.com/nspcc-dev/neofs-sdk-go/reputation"
+	sessionv2 "github.com/nspcc-dev/neofs-sdk-go/session/v2"
 	"github.com/nspcc-dev/neofs-sdk-go/user"
 	"github.com/nspcc-dev/neofs-sdk-go/version"
 	"go.uber.org/zap"
@@ -217,11 +219,14 @@ func initObjectService(c *cfg) {
 
 	c.workers = append(c.workers, c.policer)
 
+	nnsResolver := nns.NewResolver(c.cli)
+
 	sGet := getsvc.New(c,
 		getsvc.WithLogger(c.log),
 		getsvc.WithLocalStorageEngine(ls),
 		getsvc.WithClientConstructor(coreConstructor),
 		getsvc.WithKeyStorage(keyStorage),
+		getsvc.WithNNSResolver(nnsResolver),
 	)
 
 	*c.cfgObject.getSvc = *sGet // need smth better
@@ -235,6 +240,7 @@ func initObjectService(c *cfg) {
 		searchsvc.WithLocalStorageEngine(ls),
 		searchsvc.WithClientConstructor(coreConstructor),
 		searchsvc.WithKeyStorage(keyStorage),
+		searchsvc.WithNNSResolver(nnsResolver),
 	)
 
 	mNumber, err := c.cli.MagicNumber()
@@ -256,6 +262,7 @@ func initObjectService(c *cfg) {
 		putsvc.WithLogger(c.log),
 		putsvc.WithSplitChainVerifier(split.NewVerifier(sGet)),
 		putsvc.WithTombstoneVerifier(tombstone.NewVerifier(os)),
+		putsvc.WithNNSResolver(nnsResolver),
 	)
 
 	sDelete := deletesvc.New(
@@ -268,6 +275,7 @@ func initObjectService(c *cfg) {
 			cfg: c,
 		}),
 		deletesvc.WithKeyStorage(keyStorage),
+		deletesvc.WithNNSResolver(nnsResolver),
 	)
 
 	objSvc := &objectSvc{
@@ -294,6 +302,7 @@ func initObjectService(c *cfg) {
 			netmapContract: c.nCli,
 		}),
 		v2.WithContainerSource(c.cnrSrc),
+		v2.WithTimeProvider(&c.chainTime),
 	)
 	addNewEpochAsyncNotificationHandler(c, func(event.Event) {
 		aclSvc.ResetTokenCheckCache()
@@ -619,6 +628,14 @@ func (x storageForObjectService) VerifyAndStoreObjectLocally(obj object.Object) 
 
 func (x storageForObjectService) GetSessionPrivateKey(usr user.ID, uid uuid.UUID) (ecdsa.PrivateKey, error) {
 	k, err := x.keys.GetKey(&util.SessionInfo{ID: uid, Owner: usr})
+	if err != nil {
+		return ecdsa.PrivateKey{}, err
+	}
+	return *k, nil
+}
+
+func (x storageForObjectService) GetSessionV2PrivateKey(issuer user.ID, subjects []sessionv2.Target) (ecdsa.PrivateKey, error) {
+	k, err := x.keys.GetKeyBySubjects(issuer, subjects)
 	if err != nil {
 		return ecdsa.PrivateKey{}, err
 	}
--- a/pkg/core/object/fmt.go
+++ b/pkg/core/object/fmt.go
@@ -10,13 +10,16 @@ import (
 	"github.com/nspcc-dev/neo-go/pkg/core/transaction"
 	"github.com/nspcc-dev/neo-go/pkg/neorpc/result"
 	"github.com/nspcc-dev/neo-go/pkg/smartcontract/trigger"
+	"github.com/nspcc-dev/neo-go/pkg/util"
 	icrypto "github.com/nspcc-dev/neofs-node/internal/crypto"
 	"github.com/nspcc-dev/neofs-node/pkg/core/container"
 	"github.com/nspcc-dev/neofs-node/pkg/core/netmap"
+	"github.com/nspcc-dev/neofs-node/pkg/core/nns"
 	"github.com/nspcc-dev/neofs-node/pkg/core/version"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	"github.com/nspcc-dev/neofs-sdk-go/session/v2"
 )
 
 // FormatValidator represents an object format validator.
@@ -25,6 +28,7 @@ type FormatValidator struct {
 	fsChain        FSChain
 	netmapContract NetmapContract
 	containers     container.Source
+	resolver       session.NNSResolver
 }
 
 // FormatValidatorOption represents a FormatValidator constructor option.
@@ -85,6 +89,7 @@ type TombVerifier interface {
 // [FormatValidator] to work.
 type FSChain interface {
 	InvokeContainedScript(tx *transaction.Transaction, header *block.Header, _ *trigger.Type, _ *bool) (*result.Invoke, error)
+	HasUserInNNS(name string, addr util.Uint160) (bool, error)
 }
 
 // NetmapContract represents Netmap contract deployed in the FS chain required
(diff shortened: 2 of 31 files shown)


pkg/services/container/server.go
```diff
@@ -100,15 +106,42 @@ type sessionTokenCommonCheckResult struct {
...
 
+type sessionTokenV2CommonCheckResult struct {
+	token sessionv2.Token
+	err   error
+}
+
+type nnsResolver struct {
+	fsChain FSChain
+
+	nnsCache *lru.Cache[string, bool]
```
carpawell: nns checks never expire (if size is big enough)?

pkg/services/object/put/service_test.go
```diff
@@ -405,9 +530,15 @@ func testSlicingREP3(t *testing.T, cluster *testCluster, ln uint64, repNodes, cn
 
 		assertObjectIntegrity(t, restoredObj)
 		require.True(t, bytes.Equal(srcObj.Payload(), restoredObj.Payload()))
-		require.Equal(t, sessionToken, *restoredObj.SessionToken())
 		require.Equal(t, srcObj.GetContainerID(), restoredObj.GetContainerID())
-		require.Equal(t, sessionToken.Issuer(), restoredObj.Owner())
+		if isSessionTokenV2 {
+			// TODO: fix Lifetime fields different
```
carpawell: issue?

pkg/services/object/acl/v2/service.go
```diff
@@ -122,21 +192,35 @@ func New(fsChain FSChain, opts ...Option) Service {
...
 }
 
-func (b Service) getVerifiedSessionToken(mh *protosession.RequestMetaHeader, reqVerb sessionSDK.ObjectVerb, reqCnr cid.ID, reqObj oid.ID) (*sessionSDK.Object, error) {
+func (b Service) getVerifiedSessionToken(mh *protosession.RequestMetaHeader, reqVerb sessionSDK.ObjectVerb,
+	reqVerbV2 sessionSDKv2.Verb, reqCnr cid.ID, reqObj oid.ID) (verifiedSession, error) {
 	for omh := mh.GetOrigin(); omh != nil; omh = mh.GetOrigin() {
 		mh = omh
 	}
+
+	mV2 := mh.GetSessionTokenV2()
```
carpawell: can you, please, tell me or point me to the place where there are no tokens and storage nodes are still trying to PUT object? i think with sessionV2 support, we have changed the rules. i think almost every PUT now should be sessionV2-ed; otherwise, our new feature is not fully supported, imo. in practice, i think we should force the use of some session or accept only fully prepared objects. placing objects in an open container just because it is allowed to be owned by anyone (and without sessions, SNs own every object they have handled) is weird, imo
@roman-khimov, @cthulhu-rider

roman-khimov: Keep as is.

pkg/services/container/server.go
```diff
@@ -242,6 +287,81 @@ func (s *Server) checkSessionIssuer(id cid.ID, issuer user.ID) error {
 	return nil
 }
 
+func (s *Server) getVerifiedSessionTokenV2FromMetaHeader(mh *protosession.RequestMetaHeader, reqVerb sessionv2.Verb, reqCnr cid.ID) (*sessionv2.Token, []byte, error) {
+	for omh := mh.GetOrigin(); omh != nil; omh = mh.GetOrigin() {
```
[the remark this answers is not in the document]
roman-khimov: SNs don't need to sign tokens for each other normally.

pkg/services/container/server_test.go
```diff
@@ -128,6 +133,10 @@ func (x testFSChain) InvokeContainedScript(*transaction.Transaction, *block.Head
 	panic("unimplemented")
 }
 
+func (testFSChain) HasUserInNNS(string, neoutil.Uint160) (bool, error) {
```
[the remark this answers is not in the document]
End-rey: Added.

cmd/neofs-cli/modules/object/util.go
```diff
@@ -406,26 +408,43 @@ func CreateSessionV2(cmd *cobra.Command, dst SessionPrm, key *ecdsa.PrivateKey,
...
+			if err != nil {
+				return fmt.Errorf("set V2 session objects: %w", err)
+			}
+		}
+		contexts = []sessionv2.Context{ctx2}
 	case *client.PrmObjectDelete:
-		verb = sessionv2.VerbObjectDelete
+		// For DELETE, create context with two verbs:
+		// 1. DELETE for the deleted object
+		// 2. PUT for the tombstone
```
roman-khimov: Shouldn't be required, tombstone PUT is DELETE ACL-wise.

pkg/services/container/server.go
```diff
@@ -242,6 +293,81 @@ func (s *Server) checkSessionIssuer(id cid.ID, issuer user.ID) error {
...
+
+	ok, err := token.AssertAuthority(user.NewFromECDSAPublicKey(s.signer.PublicKey), s.resolver)
+	if err != nil {
+		return token, fmt.Errorf("check v2 token authority: %w", err)
+	}
+	if !ok {
+		return token, errors.New("v2 token doesn't authorize this node")
+	}
+
+	currentTime := time.Now()
```
roman-khimov: Technically the latest blockchain time should be used here. Local time can be manipulated, blockchain time not really. But this can be left in a separate issue if we don't have a good source of this data currently.

pkg/services/object/acl/v2/service.go
```diff
@@ -202,6 +290,103 @@ func (b Service) verifySessionTokenAgainstRequest(token sessionSDK.Object, reqVe
...
+	for origin != nil {
+		sig, ok := origin.Signature()
+		if !ok {
+			return nil, errors.New("missing signature in V2 session token")
+		}
+		key = sig.PublicKeyBytes()
+		origin = origin.Origin()
+	}
+
+	neoPub, err := keys.NewPublicKeyFromBytes(key, elliptic.P256())
```
roman-khimov: Can be an N3 signature at any step of v2 chain.

pkg/services/object/acl/v2/service.go
```diff
@@ -202,6 +290,103 @@ func (b Service) verifySessionTokenAgainstRequest(token sessionSDK.Object, reqVe
...
+	if !token.ValidAt(currentTime) {
+		return token, fmt.Errorf("%s: V2 token is invalid at %s Unix time", invalidRequestMessage, currentTime)
+	}
+
+	return token, nil
+}
+
+func (b Service) verifySessionTokenV2AgainstRequest(token sessionSDKv2.Token, reqVerb sessionSDKv2.Verb, reqCnr cid.ID, reqObj oid.ID) error {
+	// When we PUT a tombstone, the requested object is not zero,
+	// but the session is not bound to it because the user cannot predict its ID.
```
roman-khimov: Something has to substitute it for associated ID above. IIRC this was the case once upon a time for v1 tokens.

cmd/neofs-cli/modules/object/util.go
End-rey: It seems strange and incorrect to me when the request changes from `PUT` for the tombstone to `DELETE` with the OID of the tombstone itself. In that case, I suggest specifying zero OID for `Put` with a tombstone here:
https://github.com/nspcc-dev/neofs-node/blob/3f6a367ade41019a7a9fce405372b33d57dc3fc7/pkg/services/object/acl/v2/service.go#L638-L646
And replace `PUT` with `DELETE`, as for token v1.

End-rey: Passed the associated ID to the function.

pkg/innerring/processors/container/common.go
```diff
@@ -112,3 +113,39 @@ func (cp *Processor) checkTokenLifetime(token session.Container) error {
...
+
+	if tok.OriginalIssuer() != v.ownerContainer {
+		return errors.New("owner differs with original token issuer")
+	}
+
+	idx, err := cp.cnrClient.Morph().BlockCount()
+	if err != nil {
+		return fmt.Errorf("get current block count: %w", err)
+	}
+	bh, err := cp.cnrClient.Morph().GetBlockHeader(idx - 1)
```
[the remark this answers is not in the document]
End-rey: There is a problem that arises in tests. The new block may not yet have reached the current timestamp, so I decided to add the `MsPerBlock` network setting. However, I don't know if this is a problem with the tests or if it can really happen. In the test settings, the parameter was 50ms, but it seemed as if the blocks appeared after 500ms, and then it didn't pass. For now, I rounded this time to seconds and everything works, but I don't know how correct this is.

pkg/services/object/acl/v2/service.go
End-rey: Signature verification occurs much earlier in `ValidateWithNNS`, so this is not required.

pkg/services/container/server.go
```diff
@@ -100,15 +106,46 @@ type sessionTokenCommonCheckResult struct {
...
+}
+
+type nnsResolver struct {
+	fsChain FSChain
+
+	nnsCache *lru.Cache[string, bool]
+}
+
+func (r nnsResolver) HasUser(name string, userID user.ID) (bool, error) {
+	cacheKey := fmt.Sprintf("%s|%s", name, userID.String())
```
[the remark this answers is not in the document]
roman-khimov: Don't `fmt.Sprintf()`, you can concatenate just about anything to make a string.

pkg/core/object/fmt.go
```diff
@@ -25,6 +28,26 @@ type FormatValidator struct {
...
+	resolver       nnsResolver
+}
+
+type nnsResolver struct {
+	fsChain  FSChain
+	nnsCache *lru.Cache[string, bool]
+}
+
+func (r nnsResolver) HasUser(name string, userID user.ID) (bool, error) {
+	cacheKey := name + userID.String()
```
roman-khimov: You can avoid `.String()` call at all.

pkg/innerring/processors/container/common.go
```diff
@@ -112,3 +113,62 @@ func (cp *Processor) checkTokenLifetime(token session.Container) error {
...
+	if err != nil {
+		return time.Time{}, fmt.Errorf("get FS chain block interval: %w", err)
+	}
+	durBlockTime := time.Duration(blockTimeMs) * time.Millisecond
+	if cp.chainTime != nil {
+		ct := cp.chainTime.Now()
+		if !ct.IsZero() {
+			return ct.Add(durBlockTime), nil
+		}
+		// If provider returns zero, fall back to chain.
```
roman-khimov: I think you can safely use `time.Now()`. Asking the chain is additional complexity that it not really required.

pkg/innerring/innerring.go
```diff
@@ -202,6 +204,9 @@ func (s *Server) Start(ctx context.Context, intError chan<- error) (err error) {
 		)
 
 		s.epochTimers.UpdateTime(b.Timestamp)
+		if s.chainTime != nil {
```
roman-khimov: When do you expect it to be `nil`?

pkg/innerring/processors/container/common.go
```diff
@@ -112,3 +134,31 @@ func (cp *Processor) checkTokenLifetime(token session.Container) error {
...
+		if !tok.AssertContainer(v.verbV2, v.idContainer) {
+			return errWrongCID
+		}
+	}
+
+	if tok.OriginalIssuer() != v.ownerContainer {
+		return errors.New("owner differs with original token issuer")
+	}
+
+	msPerBlock, err := cp.cnrClient.Morph().MsPerBlock()
```
roman-khimov: Still questionable to me. The time is not yet come and the token should still be valid. Test synchronization is a test problem.

cmd/neofs-cli/modules/object/util.go
```diff
@@ -501,3 +453,47 @@ func openFileForPayload(name string) (io.WriteCloser, error) {
 	}
 	return f, nil
 }
+
+func parseSessionSubjects(cmd *cobra.Command, ctx context.Context, cli *client.Client) ([]sessionv2.Target, error) {
+	sessionSubjects, _ := cmd.Flags().GetStringSlice("session-subjects")
```
carpawell: why not const? a complex string for setting and getting flags in different places

pkg/morph/client/netmap/epoch.go
```diff
@@ -91,3 +91,32 @@ func (c *Client) GetEpochBlock(epoch uint64) (uint32, error) {
 
 	return uint32(bn.Uint64()), nil
 }
+
+func (c *Client) GetEpochBlockByTime(t uint32) (uint32, error) {
```
carpawell: no doc?

cmd/neofs-node/object.go
```diff
@@ -921,3 +937,29 @@ func convertECRules(from []netmapsdk.ECRule) []iec.Rule {
 	}
 	return to
 }
+
+type nnsResolverAdapter struct {
```
carpawell: the same thing used 3 times in 3 commits, right?

cmd/neofs-cli/modules/object/util.go
```diff
@@ -405,6 +383,29 @@ func CreateSessionV2(cmd *cobra.Command, dst SessionPrm, key *ecdsa.PrivateKey,
 		return fmt.Errorf("set subjects: %w", err)
 	}
 
+	common.PrintVerbose(cmd, "Creating server-side session key via RPC...")
+
+	var sessionPrm client.PrmSessionCreate
+	sessionPrm.SetExp(10) // 10 epochs
```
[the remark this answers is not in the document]
End-rey: Honestly, not yet. I took the base value. `KeyStorage` is cleared by epochs, and I haven't figured out how to correlate this with time yet. 

@roman-khimov What can be done? Leave the issue for the future?

carpawell (approved):

PR #3672, opened 2025-11-10
Drop lock bucket

roman-khimov:


pkg/local_object_storage/metabase/lock.go
```diff
@@ -201,17 +202,20 @@ func associatedWithTypedObject(currEpoch uint64, metaCursor *bbolt.Cursor, idObj
...
 
-	return false
+	return false, oid.ID{}
 }
 
 // checks if specified object is locked in the specified container.
 func objectLocked(tx *bbolt.Tx, currEpoch uint64, metaCursor *bbolt.Cursor, idCnr cid.ID, idObj oid.ID) bool {
-	if associatedWithTypedObject(currEpoch, metaCursor, idObj, object.TypeLock) {
-		return true
+	locked, lockID := associatedWithTypedObject(currEpoch, metaCursor, idObj, object.TypeLock)
```
carpawell: > Lock can be deleted manually

why? https://github.com/nspcc-dev/neofs-api/blob/fc8584b766d6b544337fc055dca34c29b0155d31/object/service.proto#L92-L93

pkg/local_object_storage/engine/put.go
```diff
@@ -141,14 +141,14 @@ func (e *StorageEngine) Put(obj *objectSDK.Object, objBin []byte) error {
 // First return value is true iff put has been successfully done.
 // Second return value is true iff object already exists.
```
carpawell: you have broken this desc

carpawell (commented):
Once you drop `Lock(..., locked []oid.ID)` API from the engine, you must prohibit deprecated LOCK objects with payload on node's API layer. Either i do not see it, or it is not done.

pkg/local_object_storage/metabase/lock.go
roman-khimov: Control service.

roman-khimov: In fact we have a test that does exactly that, that's how it was discovered initially.

pkg/local_object_storage/engine/put.go
```diff
@@ -213,7 +211,7 @@ func (e *StorageEngine) putToShard(sh shardWrapper, ind int, pool util.WorkerPoo
 
 	<-exitCh
 
-	return putSuccess, alreadyExists, overloaded
+	return alreadyExists, overloaded, putError
```
cthulhu-rider: `putError == nil` on `overload`. Seems like some callers can go false-positive with this

roman-khimov: True. Nice catch. Frankly, we can rework this all into a single error returned, would be much easier this way.

roman-khimov:
Rebased, updated. It was unprepared object handling, see https://github.com/nspcc-dev/neofs-node/compare/f8eed22402d0f45bc25afb8db0a94b0f16755305..7656ba6b6cef9640a900e76bca0d04306a8f79e5. Tests still running in https://github.com/nspcc-dev/neofs-node/actions/runs/19860579457/job/56909340778

roman-khimov:
Green (problems are not related to locks).
https://rest.fs.neo.org/Aku38wpn2xEqfEPxmwyAa7N3VnAkMV8PEVkpLqwKM1z3/4441-1764687654/index.html
https://rest.fs.neo.org/Aku38wpn2xEqfEPxmwyAa7N3VnAkMV8PEVkpLqwKM1z3/4441-1764687878/index.html

PR #3705, opened 2025-11-27
refactor: use strings.Builder to improve performance

bigmoonbit:
strings.Builder has fewer memory allocations and better performance.
More info: [golang/go#75190](https://github.com/golang/go/issues/75190)


roman-khimov (approved):
Some `strings.JoinStringers()` would be so nice to have, but it's not there.

PR #3706, opened 2025-11-27
Fix/ir shutdown

carpawell:


carpawell:
Not sure about the second one. Maybe it is better to panic when `stop` is called twice; it is unexpected for sure.

PR #3709, opened 2025-12-01
.github: decrease macOS shards for system tests

evgeniiz321:


cthulhu-rider:
the commit should explain why this is being done

PR #3717, opened 2025-12-04
Fix shard phy counter

cthulhu-rider:


cthulhu-rider:
updated changes to fix phy counter only. Left all the rest to #3740

roman-khimov (commented):
We need to update counters then, same as for 7->8 metabase version migration.

PR #3718, opened 2025-12-05
EC expiration

cthulhu-rider:
@evgeniiz321 try pls

### DIFF
--- a/pkg/local_object_storage/shard/delete.go
+++ b/pkg/local_object_storage/shard/delete.go
@@ -61,11 +61,7 @@ func (s *Shard) deleteObjs(addrs []oid.Address) error {
 		removedPayload := res.RemovedObjects[i].PayloadLen
 		totalRemovedPayload += removedPayload
 		s.addToContainerSize(res.RemovedObjects[i].Address.Container().EncodeToString(), -int64(removedPayload))
-	}
 
-	s.addToPayloadCounter(-int64(totalRemovedPayload))
-
-	for i := range res.RemovedObjects {
 		err = s.blobStor.Delete(res.RemovedObjects[i].Address)
 		if err == nil {
 			logOp(s.log, deleteOp, res.RemovedObjects[i].Address)
@@ -79,6 +75,8 @@ func (s *Shard) deleteObjs(addrs []oid.Address) error {
 				zap.Error(err))
 		}
 	}
+
+	s.addToPayloadCounter(-int64(totalRemovedPayload))
 
 	return nil
 }


pkg/local_object_storage/shard/delete.go
```diff
@@ -57,14 +57,6 @@ func (s *Shard) deleteObjs(addrs []oid.Address) error {
 
 	var totalRemovedPayload uint64
 
-	for i := range res.RemovedObjects {
```
roman-khimov: Makes the behavior different in case of `IsErrNotFound`, but at the same time I'm not sure what is the correct one here. The object is gone, but was it deleted from counters previously (or the file has evaporated), should it be deleted now (potentially twice)?

pkg/local_object_storage/metabase/delete.go
```diff
@@ -180,3 +196,95 @@ func delUniqueIndexes(tx *bbolt.Tx, cnr cid.ID, oID oid.ID) error {
...
+func supplementRemovedObjects(tx *bbolt.Tx, addrs []oid.Address) ([]RemovedObject, error) {
+	var res []RemovedObject
+	var err error
+
+	cnrMetaBktKey := make([]byte, 1+cid.Size)
+	cnrMetaBktKey[0] = metadataPrefix
+
+	for i := range addrs {
+		cnr := addrs[i].Container()
+		if slices.ContainsFunc(addrs[:i], func(a oid.Address) bool { return a.Container() == cnr }) {
```
roman-khimov: This can bite you with a large number of addrs. You can instead:
 * copy `addrs` into `[]RemovedObject` first (you have to do it anyway)
 * sort the list by containers once (N*log(N))
 * rely on this sorting, subslicing the set if you need to

pkg/local_object_storage/shard/delete.go
cthulhu-rider: noticed this too. Will just revert previous behavior. Other thoughts are outside this issue

pkg/local_object_storage/metabase/delete_test.go
```diff
@@ -117,7 +117,7 @@ func TestDB_Delete(t *testing.T) {
 		require.EqualValues(t, all, res.RawRemoved)
 		require.Len(t, res.RemovedObjects, all)
 
-		require.ElementsMatch(t, res.RemovedObjects[len(res.RemovedObjects)-2:], []meta.RemovedObject{
+		require.ElementsMatch(t, res.RemovedObjects[:2], []meta.RemovedObject{
```
roman-khimov: Can be problematic wrt sorting.

pkg/local_object_storage/metabase/delete.go
```diff
@@ -180,3 +196,95 @@ func delUniqueIndexes(tx *bbolt.Tx, cnr cid.ID, oID oid.ID) error {
 
 	return nil
 }
+
+// forms list of objects from addrs and their missing parts.
```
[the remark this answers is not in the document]
roman-khimov: Data for regular splits is collected the layer above. At this stage it's hard to tell what would happen if we're to change it, but in general each shard can be seen as micronode and in that sense moving split resolution to metabase level can be good. In any event we're OK to have this fix for EC now.

pkg/local_object_storage/metabase/delete.go
```diff
@@ -180,3 +196,95 @@ func delUniqueIndexes(tx *bbolt.Tx, cnr cid.ID, oID oid.ID) error {
...
+		if islices.AllZeros(partID) {
+			return nil, invalidMetaBucketKeyErr(k, oid.ErrZero)
+		}
+
+		if partCrs == nil {
+			partCrs = cnrMetaBkt.Cursor()
+		}
+
+		if ecPref == nil {
+			ecPref = slices.Concat([]byte{metaPrefixIDAttr}, partID, []byte(iec.AttributePrefix)) // any of EC attributes
```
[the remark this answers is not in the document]
carpawell: i meant if it is possible to misuse this code with ~our regular splits~ custom objects that have conflicting attributes

carpawell (approved):

PR #3725, opened 2025-12-08
Fix impassable `addStructs` failure in non-Alphabet case

cthulhu-rider:


pkg/innerring/processors/container/processor.go
```diff
@@ -273,6 +274,11 @@ func (cp *Processor) AddContainerStructs(ctx context.Context) error {
 	for ; ; time.Sleep(5 * time.Second) {
 		txRes, err := fsChain.CallNotary(ctx, cnrContract, fschaincontracts.AddContainerStructsMethod)
 		if err != nil {
+			if errors.Is(err, client.ErrNonAlphabet) {
```
roman-khimov: But you can check `cp.alphabetState.IsAlphabet()` above without any changes to `morph/client`.

cthulhu-rider:
some magic happened, i did not press close button

Issue #3476, opened 2025-07-21
Drop old lock/ts support


roman-khimov:
## Is your feature request related to a problem? Please describe.
I'm always frustrated when we have a lot of legacy code.

## Describe the solution you'd like
Drop support for locks and tombstones with API <2.18. Optimize metabase.

## Additional context
https://github.com/nspcc-dev/neofs-node/pull/3470

PR #3731, opened 2025-12-10
Drop support for old tombstones

roman-khimov:
A part of #3476.

### DIFF
--- a/pkg/services/object/tombstone/verify.go
+++ b/pkg/services/object/tombstone/verify.go
@@ -76,6 +76,13 @@ func (v *Verifier) verifyMember(ctx context.Context, cnr cid.ID, member oid.ID) 
 		return fmt.Errorf("heading object: %w", err)
 	}
 
+	switch header.Type() {
+	case object.TypeLink, object.TypeTombstone, object.TypeLock:
+		return fmt.Errorf("tombstone target is %s", header.Type())
+	default:
+		// Regular one, OK.
+	}
+
 	firstChild, firstSet := header.FirstID()
 	parent := header.Parent()
 	sID := header.SplitID()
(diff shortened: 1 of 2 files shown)


roman-khimov:
All tests (running now): https://github.com/nspcc-dev/neofs-node/actions/runs/20104857374

pkg/core/object/fmt.go
```diff
@@ -149,25 +143,16 @@ func (v *FormatValidator) validate(obj *object.Object, unprepared, isParent bool
...
-			if obj.AssociatedObject().IsZero() {
-				return errors.New("system object has zero associated object")
-			}
-		}
-	case object.TypeLock:
+	case object.TypeLock, object.TypeTombstone:
 		if !unprepared {
 			if !version.SysObjTargetShouldBeInHeader(obj.Version()) {
-				return errors.New("obsolete LOCK object version")
+				return fmt.Errorf("obsolete %s object version", obj.Type())
```
carpawell: can this check be simplified to just a version check? this is literally what this PR does: no TS/LOCK with version 2.18- are not allowed (but in both cases of `unprepared`)...

pkg/core/object/fmt.go
```diff
@@ -343,10 +330,10 @@ func (v *FormatValidator) ValidateContent(o *object.Object) (ContentMeta, error)
 
 var errExpired = errors.New("object has expired")
 
-func (v *FormatValidator) checkExpiration(obj object.Object) error {
+func (v *FormatValidator) checkExpiration(obj object.Object, expirationRequired bool) error {
```
carpawell: add more context to the err? i think error in logs now would be `"object did not pass expiration check: missing expiration epoch attribute"` but it is now clear why it is bad

carpawell (approved):

PR #3727, opened 2025-12-08
Improve object deletion via control service

roman-khimov:

PR #3734, opened 2025-12-11
control: drop useless setters/wrappers around protobuf types

roman-khimov:
Deobfuscate the code somewhat, even though most of it is relatively simple it's not obvious immediately what happens where, are copies made in setters or not, are pointers reused properly in functions or not. This code has exactly zero value, structures can be handled as structures.


roman-khimov:
Inspired by https://rest.fs.neo.org/Aku38wpn2xEqfEPxmwyAa7N3VnAkMV8PEVkpLqwKM1z3/4503-1765405528/index.html#suites/e85a210f8eb4807f0e7a860383537e45/92cc4d5b85ded3af/, I'm still not sure what's the problem there. It's a recent regression (#3727 more specifically), but I can't find the source of it.

PR #3747, opened 2025-12-19
go.mod: upgrade to the latest contracts

roman-khimov:
With a number of fixes.


carpawell (changes requested):
MaintenanceAllowed should be dropped.

PR #3750, opened 2025-12-22
Cli session token v2

End-rey:
Part of #3671.

### DIFF
--- a/cmd/neofs-cli/modules/container/util_session_v2.go
+++ b/cmd/neofs-cli/modules/container/util_session_v2.go
@@ -1,0 +1,72 @@
+package container
+
+import (
+	"crypto/ecdsa"
+	"fmt"
+
+	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/common"
+	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/commonflags"
+	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
+	"github.com/nspcc-dev/neofs-sdk-go/session/v2"
+	"github.com/nspcc-dev/neofs-sdk-go/user"
+	"github.com/spf13/cobra"
+)
+
+func getSessionV2(cmd *cobra.Command) (*session.Token, error) {
+	common.PrintVerbose(cmd, "Trying to read V2 container session from the file...")
+
+	path, _ := cmd.Flags().GetString(commonflags.SessionToken)
+	if path == "" {
+		common.PrintVerbose(cmd, "Session not provided.")
+		return nil, nil
+	}
+
+	common.PrintVerbose(cmd, "Reading V2 container session from the file [%s]...", path)
+
+	var tok session.Token
+
+	err := common.ReadBinaryOrJSON(cmd, &tok, path)
+	if err != nil {
+		return nil, fmt.Errorf("read V2 container session: %w", err)
+	}
+
+	common.PrintVerbose(cmd, "V2 session successfully read.")
+	return &tok, nil
+}
+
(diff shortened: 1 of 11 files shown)


End-rey:
Currently, by default, I am creating new v2 tokens for requests in the CLI, which is causing errors in the tests. Perhaps I should leave the v1 tokens in this PR and create v2 tokens in the PR with node and IR processing?

roman-khimov:
Yes, tests should still pass.

cmd/neofs-cli/modules/object/util_session_v2.go
```diff
@@ -0,0 +1,120 @@
...
+		return nil, fmt.Errorf("invalid V2 session token: %w", err)
+	}
+
+	if !tok.AssertVerb(cmdVerb, cnr) {
+		return nil, fmt.Errorf("v2 session token does not authorize %v for container %s", cmdVerb, cnr.String())
+	}
+
+	signer := user.NewAutoIDSigner(*key)
+	if tok.Issuer() != signer.UserID() {
+		return nil, fmt.Errorf("v2 session token issuer %v does not match user %s", tok.Issuer(), signer.UserID().String())
```
carpawell: `... does not match provided key/wallet`? i think a user with no understanding of source code will not understand what the signer is, and why it does not match. he has provided the key/wallet

cmd/neofs-cli/modules/container/util_session_v2.go
```diff
@@ -0,0 +1,72 @@
...
+		return fmt.Errorf("invalid V2 session token: %w", err)
+	}
+
+	signer := user.NewAutoIDSigner(*key)
+	if tok.Issuer() != signer.UserID() {
+		return fmt.Errorf("v2 session token issuer %v does not match user %s", tok.Issuer(), signer.UserID().String())
+	}
+
+	if !tok.AssertContainer(verb, cnrID) {
+		return fmt.Errorf("v2 session token does not authorize %v for container %s", verb, cnrID.String())
```
carpawell: hm, i believe, we need `session.Verb` to implement `fmt.Stringer`, otherwise it prints digits from SDK

End-rey: I thought about it, but looked at the previous [ObjectVerb](https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go/session#ObjectVerb) and [ContainerVerb](https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go/session#ContainerVerb), which did not have a `String` method, so I did not create one for the new one.

@roman-khimov Do we need it?

cmd/neofs-cli/modules/object/util.go
```diff
@@ -358,6 +383,154 @@ func finalizeSession(cmd *cobra.Command, dst SessionPrm, tok *session.Object, ke
...
+
+			ecdsaPubKey := (*ecdsa.PublicKey)(neoPubKey)
+			userID := user.NewFromECDSAPublicKey(*ecdsaPubKey)
+
+			subjects = append(subjects, sessionv2.NewTargetUser(userID))
+			nodeCount++
+		}
+	}
+
+	common.PrintVerbose(cmd, "Added %d node public keys as subjects", nodeCount)
```
roman-khimov: I'd expect v2 to accept targets as parameters. Can be a single node, can be a set of nodes, can be NNS, can be whatever.

cmd/neofs-cli/modules/container/util_session_v2.go
roman-khimov: Yes, it's very convenient to have one.

carpawell (approved):

PR #3778, opened 2026-01-21
Optimize token encoding

cthulhu-rider:


pkg/services/object/acl/v2/service.go
```diff
@@ -180,7 +188,15 @@ func (b Service) decodeAndVerifySessionTokenCommon(m *protosession.SessionToken)
 		return token, fmt.Errorf("%s: token is invalid at %d epoch)", invalidRequestMessage, currentEpoch)
 	}
 
-	if err := icrypto.AuthenticateToken(&token, historicN3ScriptRunner{
+	body, err := getFirstBytesField(mb)
```
roman-khimov: `mb` has _both_ body and signature, right?

PR #3790, opened 2026-01-28
Policy probe util

### LINKED ISSUE #3626: Provide storage policy probe util
sometimes I have to formulate a non-trivial container policy. I want to verify its correctness, i.e.:
1. it has correct format
2. being applied to the current netmap, it acts as needed to me

currently, all i can do is to
1. create container with this policy. If succeeded, it has correct format
2. call `neofs-cli container nodes`
3. delete the container if SN are not selected as expected

1 and 3 are paid ops. It'd be nice to have a utility for testing policies

provide
```
$ neofs-cli container test-policy -p policy.json -r s01.neofs.devenv:8080
```
with same output as `container nodes`

add `--policy` flag to `container nodes` cmd mutexed w/ currently required `--cid` one. Pros: no new cmd. Cons: input branching

End-rey:
Closes #3626.

### DIFF
--- a/cmd/internal/cmdprinter/netmap.go
+++ b/cmd/internal/cmdprinter/netmap.go
@@ -3,6 +3,7 @@ package cmdprinter
 import (
 	"encoding/hex"
 
+	iec "github.com/nspcc-dev/neofs-node/internal/ec"
 	"github.com/nspcc-dev/neofs-sdk-go/netmap"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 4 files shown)


cmd/neofs-cli/modules/container/test_policy.go
```diff
@@ -0,0 +1,104 @@
+package container
```
roman-khimov: Anything with "test" in name is suspicious by definition, better name it `policy.go`

cmd/neofs-cli/modules/container/test_policy.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/nspcc-dev/neofs-node/cmd/neofs-cli/internal/commonflags"
+	iec "github.com/nspcc-dev/neofs-node/internal/ec"
+	"github.com/nspcc-dev/neofs-sdk-go/client"
+	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
+	"github.com/nspcc-dev/neofs-sdk-go/netmap"
+	"github.com/spf13/cobra"
+)
+
+var testPolicyCmd = &cobra.Command{
+	Use:   "test-policy",
```
roman-khimov: This can also be `policy` to me. Maybe `policy try` like we have `acl basic print`, although here it's application to a particular network that matters. I'm open to other suggestions, but I'd try to avoid `test` as well.

End-rey: Are you suggesting a separate `policy` command with a single `try` command?

roman-khimov: `policy check`

PR #3802, opened 2026-02-04
sn/storage: Fix unclosed compressed files in `FSTree`

cthulhu-rider:


pkg/local_object_storage/blobstor/fstree/util.go
```diff
@@ -0,0 +1,17 @@
+package fstree
+
+import (
+	"io"
+
+	"github.com/klauspost/compress/zstd"
+)
+
+type zstdCloser struct {
+	dec *zstd.Decoder
```
roman-khimov: Likely you don't need `zstd` import here, it's a standard `io.Closer` as well, you just have two of them to care about (but ordered).

cthulhu-rider: done

PR #3803, opened 2026-02-04
sn/policer: add "clean" cycle flag to metrics

### LINKED ISSUE #3795: Policer "clean" metrics flag
Policer can signal if it has done any real changes in the previous cycle which can be helpful to detect if the cluster is in the consistent state.

Add a boolean policer consistency metric, start with false then if policer makes a full cycle without doing any relocations (not counting local deletions) we flag it as true (meaning, consistent). Then once it does any relocation subsequently it sets the flag to false till a full clean cycle is made again.

Theoretically this can be per-container and/or tied to contract-level state, but let's start simple for now.

## Additional context
Related to #3762 and #3765.

carpawell:
If policer passes all the placement checks for every object SN stores, mention it in SN's metrics. Closes #3795.

### DIFF
--- a/pkg/metrics/state.go
+++ b/pkg/metrics/state.go
@@ -5,7 +5,8 @@ import "github.com/prometheus/client_golang/prometheus"
 const stateSubsystem = "state"
 
 type stateMetrics struct {
-	healthCheck prometheus.Gauge
+	healthCheck             prometheus.Gauge
+	policerConsistencyState prometheus.Gauge
 }
 
 func newStateMetrics() stateMetrics {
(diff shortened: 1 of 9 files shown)


pkg/metrics/state.go
```diff
@@ -16,13 +17,28 @@ func newStateMetrics() stateMetrics {
 			Name:      "health",
 			Help:      "Current Node state",
 		}),
+		engineConsistencyState: prometheus.NewGauge(prometheus.GaugeOpts{
+			Namespace: storageNodeNameSpace,
+			Subsystem: stateSubsystem,
+			Name:      "engine_consistency_state",
```
roman-khimov: s/engine/policer/?

roman-khimov (commented):
OK otherwise.

PR #3804, opened 2026-02-09
metabase: put batch during resync metabase

### LINKED ISSUE #3453: Speed up metabase resync process
I'm always frustrated when metabase resync takes too much time. Currently it iterates over all objects and does `metabase.Put()` for all objects as well as has some special handling for locks and tombstones. Lock and TS handling will be simplified with https://github.com/nspcc-dev/neofs-api/issues/312 and https://github.com/nspcc-dev/neofs-api/issues/311, but doing `Put()` for every object is problematic on its own because it's a BoltDB `Batch()` for every object and in absence of concurrent modifications this means `max_batch_delay` for every object. If it's 10ms per object this means ~3h for 1M objects not even counting reads and real writing time.

Rework transaction handling for the resync case, open a transaction, process something like 1K objects, flush the transaction, rinse, repeat.

Keep wasting time for nothing.

End-rey:
Closes #3453.

### DIFF
--- a/pkg/local_object_storage/metabase/put.go
+++ b/pkg/local_object_storage/metabase/put.go
@@ -11,6 +11,7 @@ import (
 	apistatus "github.com/nspcc-dev/neofs-sdk-go/client/status"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	"go.uber.org/zap"
 )
 
 const maxObjectNestingLevel = 2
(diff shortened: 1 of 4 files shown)


pkg/local_object_storage/metabase/put.go
```diff
@@ -47,6 +48,57 @@ func (db *DB) Put(obj *object.Object) error {
...
+			}
+			successIndices = append(successIndices, i)
+		}
+		return nil
+	})
+	if err == nil {
+		for _, i := range successIndices {
+			storagelog.Write(db.log,
+				storagelog.AddressField(objs[i].Address()),
+				storagelog.OpField("metabase PUT batch"))
```
roman-khimov: Just `metabase PUT`?

roman-khimov (commented):
Seems good.

roman-khimov (approved):

PR #3817, opened 2026-02-12
Access session keys by account only

### LINKED ISSUE #3793: Access session keys by account only
I'm always frustrated when we have some duality in the DB that stores session keys, like in https://github.com/nspcc-dev/neofs-node/pull/3671#discussion_r2737394393.

1. Eliminate user-defined keys from the DB interface, put the key and derive an account from it internally, get by account only.
2. For session tokens v1 account can be derived from the session key embedded in the token itself, UUID just isn't needed.
3. Old keys need to be migrated on node upgrade, but that's rather easy.

Bloating the DB doesn't make much sense, so we better just stop it.

## Additional context
https://github.com/nspcc-dev/neofs-node/pull/3671

End-rey:
Closes #3793.

### DIFF
--- a/pkg/services/object/get/service.go
+++ b/pkg/services/object/get/service.go
@@ -105,8 +105,8 @@ type cfg struct {
 	}
 
 	keyStore interface {
-		GetKey(*util.SessionInfo) (*ecdsa.PrivateKey, error)
-		GetKeyBySubjects(user.ID, []sessionv2.Target) (*ecdsa.PrivateKey, error)
+		GetKey(*user.ID) (*ecdsa.PrivateKey, error)
+		GetKeyBySubjects([]sessionv2.Target) (*ecdsa.PrivateKey, error)
 	}
 
 	nnsResolver sessionv2.NNSResolver
--- a/pkg/services/object/server.go
+++ b/pkg/services/object/server.go
@@ -15,7 +15,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/google/uuid"
 	icrypto "github.com/nspcc-dev/neofs-node/internal/crypto"
 	"github.com/nspcc-dev/neofs-node/pkg/core/client"
 	"github.com/nspcc-dev/neofs-node/pkg/core/container"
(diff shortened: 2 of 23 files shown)


End-rey:
Blocked by #3785.

pkg/services/object/server.go
```diff
@@ -123,10 +122,10 @@ type FSChain interface {
 }
 
 type sessions interface {
-	// GetSessionPrivateKey reads private session key by user and session IDs.
+	// GetSessionPrivateKey reads private session key by user and session user-defined public key.
 	// Returns [apistatus.ErrSessionTokenNotFound] if there is no data for the
 	// referenced session.
-	GetSessionPrivateKey(usr user.ID, uid uuid.UUID) (ecdsa.PrivateKey, error)
+	GetSessionPrivateKey(usr, account user.ID) (ecdsa.PrivateKey, error)
```
roman-khimov: Why two accounts? A single one should be sufficient.

End-rey: At first, I didn't understand that it meant literally “by account only”. Now it's done.

pkg/services/object/get/service.go
```diff
@@ -105,8 +105,8 @@ type cfg struct {
 	}
 
 	keyStore interface {
-		GetKey(*util.SessionInfo) (*ecdsa.PrivateKey, error)
-		GetKeyBySubjects(user.ID, []sessionv2.Target) (*ecdsa.PrivateKey, error)
+		GetKey(*user.ID) (*ecdsa.PrivateKey, error)
```
roman-khimov: Why a pointer here?

pkg/util/state/token.go
```diff
@@ -124,22 +113,17 @@ func (p PersistentStorage) RemoveOldTokens(epoch uint64) {
 // FindTokenBySubjects searches for a private token whose public key
 // matches any of the given user ID Targets.
 // Returns nil if no matching non-expired token is found.
-func (p PersistentStorage) FindTokenBySubjects(ownerID user.ID, subjects []sessionv2.Target) *session.PrivateToken {
+func (p PersistentStorage) FindTokenBySubjects(subjects []sessionv2.Target) *session.PrivateToken {
```
roman-khimov: I still don't like how we pass `sessionv2` things here (this package should care much about tokens), but this is easier for now.

pkg/util/state/migratation.go
```diff
@@ -69,3 +74,109 @@ func (p PersistentStorage) MigrateOldTokenStorage(oldPath string) error {
...
+
+	var err error
+	if p.gcm != nil {
+		rawKey, err = p.decrypt(rawKey)
+		if err != nil {
+			return nil, fmt.Errorf("could not decrypt session key: %w", err)
+		}
+	}
+
+	privKey, err := x509.ParseECPrivateKey(rawKey)
```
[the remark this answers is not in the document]
End-rey: What's the problem with them?

pkg/util/state/token.go
```diff
@@ -94,25 +87,21 @@ func (p PersistentStorage) RemoveOldTokens(epoch uint64) {
...
-			// iterating over fixed ownerID's tokens
-			for k, v := c.First(); k != nil; k, v = c.Next() {
-				if epochFromToken(v) <= epoch {
-					err = c.Delete()
-					if err != nil {
-						p.l.Error("could not delete %s token",
-							zap.String("token_id", hex.EncodeToString(k)),
-						)
-					}
+		// iterating over accounts
```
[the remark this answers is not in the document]
End-rey: Is there already such an issue? Or should I create one?

pkg/util/state/migratation.go
```diff
@@ -69,3 +74,116 @@ func (p PersistentStorage) MigrateOldTokenStorage(oldPath string) error {
...
+		if rootBucket == nil {
+			return nil
+		}
+
+		migratedCount := 0
+		deletedCount := 0
+
+		owners := make(map[user.ID]int)
+		c := rootBucket.Cursor()
+		for ownerKeyBytes, v := c.First(); ownerKeyBytes != nil; ownerKeyBytes, v = c.Next() {
```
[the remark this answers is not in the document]
roman-khimov: Ideally you should just read something (no DB changes at all) and move on if there is nothing to do.

cthulhu-rider (approved):

PR #3844, opened 2026-02-24
lens/fstree: get depth automatically

End-rey:
Refs https://github.com/nspcc-dev/neofs-testcases/pull/1287#issuecomment-3949907535


pkg/local_object_storage/blobstor/fstree/control.go
```diff
@@ -65,6 +67,9 @@ func (t *FSTree) checkConfig() error {
 			return fmt.Errorf("read descriptor %q: %w", descPath, err)
 		}
 		if t.readOnly {
+			if t.initDone {
```
roman-khimov: Seems wrong. It won't trigger for Open(true)/Init() since `initDone` is set after the check.

PR #3849, opened 2026-02-25
meta: move meta resync into lens command

### LINKED ISSUE #3823: Move meta resync into CLI command
I'm always frustrated when I think of metabase resynchronization. Currently it's done via `resync_metabase` configuration flag, but:
 * it requires configuration modifications to use
 * it's performed sequentially for shards which is slow (#3416)
 * it's very easy to forget about it once resync is done since it can take quite a lot of time leading to unplanned resync on the next start
 * it's harder to track its state if something goes wrong since the node would just start normally

1. Drop configuration (it's not used normally, so not a problem to do this immediately).
2. Move resync into `lens meta resync` which can take a blobstor and metabase file as parameters (don't forget about shard IDs).

This allows to run as many resynchronizations as needed in separate processes with clear result for each of them.

End-rey:
Closes #3823.


pkg/local_object_storage/shard/control.go
```diff
@@ -140,30 +127,36 @@ func (s *Shard) Init() error {
 	return nil
 }
 
-func (s *Shard) resyncMetabase() error {
-	err := s.metaBase.Reset()
+// ResyncBlobstorToMetabase resets the metabase and repopulates it by iterating
+// over all objects in the blobstor. The shard ID stored in the metabase is
```
roman-khimov: Usually metabase is empty when we're doing it, it's better to take ID from blobstor.

pkg/local_object_storage/shard/control.go
```diff
@@ -140,30 +127,45 @@ func (s *Shard) Init() error {
...
-func (s *Shard) resyncMetabase() error {
-	err := s.metaBase.Reset()
+// ResyncBlobstorToMetabase resets the metabase and repopulates it by iterating
+// over all objects in the blobstor. shardID is the shard ID to associate with
+// the metabase (typically read from the blobstor). If the metabase
+// already contains data for a different shard ID, the function returns an error
+// unless force is true. onIterationError is called for every object that fails
+// to be read or unmarshalled; returning a non-nil error aborts the resync;
+// if onIterationError is nil, iteration errors are ignored.
+func ResyncBlobstorToMetabase(bs common.Storage, db *meta.DB, shardID string, force bool, onIterationError func(oid.Address, error) error) error {
```
roman-khimov: 1. `force` is not appropriate at this level, it's something from CLI, API shouldn't care. Since all data is dropped on `Reset()` you better always push a new ID here and handle check/force logic in lens itself.
2. BTW, why is it a `shard` thing? Looks like it can be a method of `db` since it doesn't need anything from Shard.

CHANGELOG.md
```diff
@@ -28,6 +31,13 @@ SN config option `node.persistent_sessions.path` was deprecated since release 0.
 and has now been removed. Drop this config and associated file, migrate with 0.51.1 if needed.
 Sessions are stored in `node.persistent_state.path` DB now.
 
+Due to the previously stored incorrect shard ID in the FSTree descriptor,
+in this release it is necessary to delete the old descriptor,
+and then the correct one will be created at startup.
```
End-rey: For now, just like that. But maybe we should migrate?

PR #3850, opened 2026-02-26
Go 1.25

roman-khimov:


internal/testutil/rand.go
```diff
@@ -4,12 +4,10 @@ import (
 	"encoding/binary"
 	"math/rand/v2"
 	"time"
-
-	"golang.org/x/exp/constraints"
 )
 
 // RandByteSlice returns randomized byte slice of specified length.
-func RandByteSlice[I constraints.Integer](ln I) []byte {
+func RandByteSlice[I ~int | ~uint64](ln I) []byte {
```
cthulhu-rider: i'd make it `~int64` right away to support `int64` args. For example, this type is often used in `io` interfaces

PR #3854, opened 2026-02-27
replicator: move EC local/remote logic from policer

### LINKED ISSUE #3809: Shard evacuation fault handler doesn't work correctly for EC parts
## Expected Behavior
EC parts treated properly by shard evacuation logic.

## Current Behavior
When pushing objects into local shards it's not a problem, but there is a fault handler that tries to replicate things when local put doesn't succeed. It uses generic object logic that doesn't fit for EC parts.

## Possible Solution
1. Move EC-specific logic from `recreateECPart()` of policer to replicator.
2. Make `recreateECPart()` use generic replicator interface.
3. Drop `PutObjectToNode()` from replicator.

As a bonus, drop `PutObject()` calls as well. Replication should _always_ use replication API. This is totally possible for complete objects evacuation operates with.

Perform network evacuation of EC parts, placement is going to be suboptimal.

## Context
Nothing specific.

## Regression
Not exactly, but this was irrelevant pre-EC era.

## Your Environment
* Version used: 0.51.0

End-rey:
Closes #3809.

### DIFF
--- a/pkg/services/policer/ec.go
+++ b/pkg/services/policer/ec.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	iec "github.com/nspcc-dev/neofs-node/internal/ec"
+	"github.com/nspcc-dev/neofs-node/pkg/services/replicator"
 	apistatus "github.com/nspcc-dev/neofs-sdk-go/client/status"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 	"github.com/nspcc-dev/neofs-sdk-go/netmap"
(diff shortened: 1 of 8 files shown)


pkg/services/policer/ec.go
```diff
@@ -504,40 +505,30 @@ func (p *Policer) recreateECPart(ctx context.Context, parent object.Object, rule
 		return
 	}
 
-	for i := range iec.NodeSequenceForPart(partIdx, int(rule.DataPartNum)+int(rule.ParityPartNum), len(sortedNodes)) {
-		if p.network.IsLocalNodePublicKey(sortedNodes[i].PublicKey()) {
```
roman-khimov: Not sure this check is practical at all. A node can't detect it's own missing part. Even if netmap changes the set of nodes is the same from check to recreate. We also have some part already (which is why we're here). So we should replicate this part (if needed) and check for other parts (on other nodes) and then replicate new parts to that remote nods (if needed).

One thing I can imagine netmap change + a lost part in which case we'd have a part, but a wrong one and we'd like to _both_ push our part to another node and recreate this missing part for ourselves. How practical that is (given that other nodes will recreate/replicate it too) is an open question.

pkg/services/replicator/replicator.go
```diff
@@ -73,3 +82,10 @@ func WithLocalStorage(v *engine.StorageEngine) Option {
 		c.localStorage = v
 	}
 }
+
+// WithNetwork provides LocalNodeKey component.
+func WithNetwork(n LocalNodeKey) Option {
```
roman-khimov: WithNetwork, but LocalNodeKey, seems inconsistent.

pkg/services/replicator/replicator.go
```diff
@@ -8,6 +8,13 @@ import (
 	"go.uber.org/zap"
 )
 
+// Network provides information about the NeoFS network to Replicator for work.
+type Network interface {
+	// IsLocalNodePublicKey checks whether given binary-encoded public key is
+	// assigned in the network map to a local storage node running [Replicator].
+	IsLocalNodePublicKey([]byte) bool
```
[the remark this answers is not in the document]
roman-khimov: Should be changed for all services, here we can follow whatever is done everywhere already.

carpawell (approved):

PR #3856, opened 2026-03-02
Feat/more engine metrics

carpawell:


pkg/local_object_storage/metabase/counter.go
```diff
@@ -124,49 +152,105 @@ func updateCounter(tx *bbolt.Tx, typ objectType, delta uint64, inc bool) error {
...
+
+	c := b.Cursor()
+	for obj := range iterAttrVal(b.Cursor(), object.FilterType, []byte(object.TypeRegular.String())) {
+		if string(getObjAttribute(c, obj, object.FilterPhysical)) == binPropMarker {
+			phyCounter++
+		}
+		if string(getObjAttribute(c, obj, object.FilterRoot)) == binPropMarker {
+			rootCounter++
+		}
+		if inGarbage(c, obj) != statusAvailable {
```
roman-khimov: Count garbage separately, it's easier to me, most of the time objects do not have any marks, so checking each of them is not productive. Can also be more correct wrt splits.

This may also be relevant for other cases, seeking through respective indexes only can be more efficient than jumping from one index to another.

pkg/local_object_storage/metabase/put.go
```diff
@@ -133,36 +147,51 @@ func (db *DB) put(tx *bbolt.Tx, obj *object.Object, nestingLevel int, currEpoch
...
+
+		diff.Link++
+		err = updateCounter(metaBkt, linkCounter, 1, true)
+		if err != nil {
+			return fmt.Errorf("could not increase link objects counter: %w", err)
+		}
+
+		// link is always phy
+		diff.Phy++
+		err = updateCounter(metaBkt, phyCounter, 1, true)
```
roman-khimov: Updating counters one by one doesn't seem to be effective to me (not specific to this particular function, even more relevant for deletion), is there any way to collect all the data and update them once for transaction?

roman-khimov (commented):
I guess it's still WIP, update it please

pkg/local_object_storage/metabase/counter.go
```diff
@@ -17,12 +18,23 @@ var objectLogicCounterKey = []byte("logic_counter")
...
-	Lock int
-	Link int
-	GC   int
+	Phy     int
+	Root    int
+	TS      int
+	Lock    int
+	Link    int
+	GC      int
+	Payload int
```
roman-khimov: int64

PR #3861, opened 2026-03-04
shard: restore ID from FSTree when metabase is missing

End-rey:
Refs https://github.com/nspcc-dev/neofs-testcases/pull/1292#issuecomment-3993564935.

### DIFF
--- a/cmd/neofs-lens/internal/meta/resync.go
+++ b/cmd/neofs-lens/internal/meta/resync.go
@@ -91,7 +91,7 @@ func resyncFunc(cmd *cobra.Command, _ []string) error {
 
 	cmd.Println("Iterating over blobstor objects...")
 
-	err = db.ResyncFromBlobstor(fst, blobstorShardID, func(addr oid.Address, err error) error {
+	err = db.ResyncFromBlobstor(fst, func(addr oid.Address, err error) error {
 		cmd.PrintErrf("warn: object %s: %v\n", addr, err)
 		return nil
 	})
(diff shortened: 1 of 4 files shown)


pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -572,6 +573,22 @@ func (t *FSTree) Path() string {
 
 // ShardID returns the shard ID associated with this FSTree.
 func (t *FSTree) ShardID() string {
+	if !t.shardIDSet {
+		descPath := t.descriptorPath()
+		f, err := os.Open(descPath)
```
roman-khimov: Doesn't seem correct long-term (undefined wrt open/init), but ok for a fix.

cmd/neofs-lens/internal/meta/resync.go
```diff
@@ -79,19 +78,9 @@ func resyncFunc(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("init blobstor: %w", err)
 	}
 
-	blobstorShardID := fst.ShardID()
-	idRaw, err := db.ReadShardID()
-	if err != nil {
-		return fmt.Errorf("read shard ID from metabase: %w", err)
-	}
-	metaShardID := base58.Encode(idRaw)
-	if len(metaShardID) > 0 && metaShardID != blobstorShardID && !vForce {
```
roman-khimov: But this breaks `force`. Check here, use there.

roman-khimov (approved):

PR #3877, opened 2026-03-16
Optimize GET/HEAD forwarding

cthulhu-rider:


pkg/services/object/acl/eacl/v2/object.go
```diff
@@ -108,21 +107,12 @@ func headersFromBinaryObjectHeader(buf []byte, cnr cid.ID, id *oid.ID) ([]eaclSD
...
 
 	var off int
-	var prevNum protowire.Number
 	for {
 		num, typ, n, err := iprotobuf.ParseTag(buf[off:])
 		if err != nil {
 			return nil, err
 		}
 
-		if num < prevNum {
```
roman-khimov, line 119: Seems unrelated, although maybe you guarantee it now at some other place.

pkg/services/object/get/exec.go
```diff
@@ -64,6 +64,12 @@ type execCtx struct {
 
 	localGetBuffer         []byte
 	submitLocalGetStreamFn SubmitStreamFunc
+
+	forwardHeadRequestFn ForwardHeadRequestFunc
+
```
roman-khimov: Don't add useless blank lines.

pkg/services/object/common.go
```diff
@@ -0,0 +1,57 @@
...
+			}
+			err := protoscan.ScanMessage(nested, protoscan.ResponseStatusScheme, opts)
+			if err != nil {
+				return fmt.Errorf("handle status field: %w", err)
+			}
+			return nil
+		}
+	}
+
+	return code, protoscan.ScanMessage(buffers, protoscan.ResponseMetaHeaderScheme, opts)
```
roman-khimov: Call scanner first, return its result afterwards. It modifies `code` which is very important.

pkg/services/object/util_internal_test.go
```diff
@@ -0,0 +1,185 @@
...
+	require.NoError(t, originSessionToken.Sign(usertest.User()))
+
+	sessionToken := newTestUnsignedSessionToken(t)
+	sessionToken.SetOrigin(&originSessionToken)
+	require.NoError(t, sessionToken.Sign(usertest.User()))
+
+	par := *object.New(cidtest.ID(), usertest.ID())
+	par.SetCreationEpoch(10738628592919807909)
+	par.SetPayloadSize(12653732379852397698)
+	par.SetPayloadHomomorphicHash(checksum.NewTillichZemor([64]byte(testutil.RandByteSlice(64))))
```
roman-khimov: Requires nolint at least, not sure if needed at all for these tests (although it's a valid field for old objects).

pkg/services/object/get.go
```diff
@@ -0,0 +1,376 @@
...
+
+	return sent, err
+}
+
+func (x *getProxyContext) handleChunkResponse(streamProg *getStreamProgress, respBuf mem.BufferSlice, chunkBuffers iprotobuf.BuffersSlice) (bool, error) {
+	chunkLen := chunkBuffers.Len()
+
+	from, to := chunkBoundsToSend(x.respondedPayload, streamProg.readPayload, chunkLen)
+	if from == to {
+		streamProg.readPayload += chunkLen
```
roman-khimov, line 241: This one is suspicious to me. We have no data, but still adding something to the counter.

pkg/services/object/get_test.go
```diff
@@ -129,6 +133,129 @@ func TestServer_Get_Local(t *testing.T) {
...
+						ObjectPart: &protoobject.GetResponse_Body_Init_{
+							Init: &protoobject.GetResponse_Body_Init{
+								ObjectId:  objMsg.ObjectId,
+								Signature: objMsg.Signature,
+								Header:    objMsg.Header,
+							},
+						},
+					},
+					VerifyHeader: newAnyVerificationHeader(),
+				},
```
roman-khimov, line 217: Looks like data chunks are not tested enough.

pkg/services/object/get.go
cthulhu-rider: similar to https://github.com/nspcc-dev/neofs-node/blob/ec8b66ea43e4c408f0bcb8fdd0f495e9b22bdec0/pkg/services/object/server.go#L1476-L1478

pkg/services/object/acl/eacl/v2/object.go
cthulhu-rider: since headers from remote nodes may now be here, the fields in them may not be ordered. This change weakens the requirement

pkg/services/object/common.go
cthulhu-rider: done

pkg/services/object/util_internal_test.go
cthulhu-rider: dropped

pkg/services/object/get_test.go
cthulhu-rider: fixed

roman-khimov (approved):

PR #3905, opened 2026-03-24
*: prohibit TZ hashing

### LINKED ISSUE #3847: Drop TZ hash support
I'm always frustrated when we're using broken hashes. TZ hash is exactly that, see:
 * https://eprint.iacr.org/2009/376.pdf
 * https://people.maths.ox.ac.uk/petit/files/SAC2010.pdf

What's even more interesting is that we're not using it practically since #3442. We can calculate/store/display/search for the hash value, but it's not that useful.

Drop support for it completely:
 * remove `HomomorphicHashingDisabled` setting
 * don't calculate it
 * don't store it in metabase and reject search requests using it
 * drop tzhash dependency from SDK and node
 * it can still be retrieved/displayed as some bytes (part of the header) for old objects

1. Keep things as is, works, but suggests that the hash is OK while we know now it's not.
2. Replace with some other homomorphic hash, but there are not a lot of alternatives (maybe Zest, see https://people.maths.ox.ac.uk/petit/files/ZesT.pdf), it breaks compatibility if not stored in a separate field (which complicates adoption) and then we still won't be using it in any way other than to calculate/store/retrieve some bytes. Keep in mind it's also very slow.

carpawell:
~This PR does not mean "homomorphic" word cannot be found in the source code, there are still many tests with it, and i think it is ok for now, at least old objects still should be replicated and hanlded ok. For new objects it is prohibited fully. Full homo clean up may be considered later.~

Closes #3847.


pkg/services/object/put/local.go
```diff
@@ -55,9 +54,9 @@ func (p *Service) ValidateAndStoreObjectLocally(obj object.Object) error {
 	switch csType {
 	default:
 		return errors.New("unsupported payload checksum type")
-	case
-		checksum.SHA256,
-		checksum.TillichZemor:
+	case checksum.TillichZemor:
+		return errors.New("object has Tillich-Zémor checksum")
```
carpawell: @roman-khimov, i want it to be this way, but i am not sure: we had [this](https://github.com/nspcc-dev/neofs-api/issues/304), and this was [solved](https://github.com/nspcc-dev/neofs-api/pull/316/changes/a55e10e2abdc2c37fbbe43276d8af8027adc785d), and i really think that not all of our internals were ever ready for TZ hashing in the `payload_checksum` field but on the other hand, i am not sure, prohibiting it like this (even replication) is a good idea

pkg/core/object/metadata.go
```diff
@@ -603,6 +603,7 @@ func combineValues(attr string, dbVal []byte, fltVal string) ([]byte, []byte, er
 			return dbVal, b, nil
 		}
 		dbVal = []byte(hex.EncodeToString(dbVal))
+	//nolint:staticcheck // this is not DB's responsibility to force API rules, DB still may have these values inside
```
roman-khimov, line 611: 64 bytes per object. Whole mainnet. And new API doesn't support searching by homohash.

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/nspcc-dev/neofs-node
 
 go 1.25.0
 
+replace github.com/nspcc-dev/neofs-sdk-go v1.0.0-rc.17.0.20260325133051-c1dbddc28fdf => ../neofs-sdk-go
```
roman-khimov: :eyes:

pkg/core/object/metadata.go
roman-khimov: Actually, more than 64 since we have two keys for it and these keys have prefixes.

carpawell: you mean drop it? migration?

roman-khimov: Exactly.

carpawell: completed TZ hash serving removal, added a migration. all in a separate commit if you dont mind

pkg/local_object_storage/metabase/version.go
```diff
@@ -402,14 +402,51 @@ func migrateFrom9Version(db *DB) error {
...
+		err = syncCounter(tx, true)
 		if err != nil {
 			return fmt.Errorf("resync object counters: %w", err)
 		}
 		return updateVersion(tx, 11)
 	})
 }
 
+func dropHomomorphicIndexes(tx *bbolt.Tx) error {
+	return tx.ForEach(func(name []byte, b *bbolt.Bucket) error {
```
roman-khimov: We have `updateContainersInterruptable` for a reason here.

pkg/local_object_storage/metabase/version.go
```diff
@@ -402,14 +402,51 @@ func migrateFrom9Version(db *DB) error {
...
+		if name[0] != metadataPrefix {
+			return nil
+		}
+		var (
+			c    = b.Cursor()
+			keys [][]byte
+		)
+		for dbKey, _ := c.First(); dbKey != nil; dbKey, _ = c.Next() {
```
roman-khimov: That's an overkill. Iterate over AttrID prefix for homomo, then add two elements to delete.

carpawell: this is more reliable for me to find garbage indexes and drop them than iterating over objects and dropping the opposite. but ok, will do

pkg/local_object_storage/metabase/version.go
carpawell: hm, too complex func without docs to me, and we are dropping db values there. if we need context, logger and limiting it is easier to implement it there than to adopt `updateContainersInterruptable`. it has some old understanding of `validPrefixes` (we now have a single valid meta prefix for objects)

roman-khimov: It can be changed, it's currently unused, but it was left in code for a reason. This migration will take _a lot_ of time, it can be interrupted, it should be interruptible, it shouldn't try keeping all keys in memory, it needs to flush changes periodically. All of that is there.

pkg/local_object_storage/metabase/version.go
roman-khimov: We trust metabase to be consistent in this case, but walking over all keys in the metabase is too expensive.

pkg/local_object_storage/metabase/version.go
```diff
@@ -410,6 +416,44 @@ func migrateFrom10Version(db *DB) error {
...
+		attrIDPrefix = []byte{metaPrefixAttrIDPlain}
+		attrKeyLen   = len([]byte(object.FilterPayloadHomomorphicHash))
+		keysToDrop   [][]byte
+	)
+	attrIDPrefix = append(attrIDPrefix, []byte(object.FilterPayloadHomomorphicHash)...)
+	if after == nil {
+		after = attrIDPrefix
+	}
+	k, _ = c.Seek(after)
+	if bytes.Equal(k, after) {
```
roman-khimov: Why this condition at all?

pkg/local_object_storage/metabase/version.go
```diff
@@ -410,6 +416,44 @@ func migrateFrom10Version(db *DB) error {
...
+			break
+		}
+	}
+
+	for _, keyToDrop := range keysToDrop {
+		_ = b.Delete(keyToDrop)
+
+		v := keyToDrop[1+attrKeyLen+len(objectcore.MetaAttributeDelimiter) : len(keyToDrop)-(oid.Size+len(objectcore.MetaAttributeDelimiter))]
+		id := keyToDrop[len(keyToDrop)-oid.Size:]
+		revercedKey := slices.Concat([]byte{metaPrefixIDAttr}, id, []byte(object.FilterPayloadHomomorphicHash), objectcore.MetaAttributeDelimiter, v)
```
roman-khimov: reversed

pkg/local_object_storage/metabase/version.go
carpawell: can be dropped, no harm, but as I understand it saves us from cases when we are stopped at positive cases in the previous iteration and `after` now also `HasPrefix(k, attrIDPrefix)`

roman-khimov: For this particular migration `after` function parameter can be ignored at all, we're dropping all keys found anyway.

carpawell: i thought i wanted to keep this universal func and keep interruptions possible

roman-khimov: This doesn't require breaking the function interface. In migrations that only read keys this code does make sense in that we're skipping the latest processed value, but in case where we drop all of them it just looks a bit dangerous (even though it shouldn't ever be triggered, `after` is gone and `k` can't equal `after`).

carpawell: got you, tbh i just copied previous functions that were added with `updateContainersInterruptable`, did not think much. dropped anything related to iteration progress, and added more objects and a container in the test

roman-khimov (approved):

PR #3923, opened 2026-04-03
sn/object: add post-initial placement replication

### LINKED ISSUE #3880: Optimize post-initial placement replication
I'm always frustrated when I realize that post-initial placement is currently totally dependent on policer:
 * it walks through all the objects and it can take days before it picks up our new initially-placed object
 * it will flag the state as inconsistent even though it almost is (initial constraint satisfied)

Make an additional queue for replicator to work on, feed incompletely-placed objects into it.

## Additional context
#3830.

End-rey:
Closes #3880.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Changelog for NeoFS Node
 - `object search` CLI command is now the same as `object searchv2` (#3931)
 - Optimized integer comparison in SearchV2 (#3938)
 - Optimized number of nodes for SEARCH in EC containers (#3940)
+- Optimized post-initial placement replication (#3923)
 
 ### Removed
 - `policer.max_workers` configuration (#3920)
--- a/cmd/neofs-node/object.go
+++ b/cmd/neofs-node/object.go
@@ -65,6 +65,32 @@ type objectSvc struct {
 	delete *deletesvc.Service
 }
 
+type putPostInitialPlacementReplicator struct {
+	log        *zap.Logger
+	replicator *replicator.Replicator
+}
+
+func (r putPostInitialPlacementReplicator) HandlePostPlacement(obj *object.Object, nodes []netmapsdk.NodeInfo) {
+	copies := uint32(len(nodes))
+	if pi, err := iec.GetPartInfo(*obj); err == nil && pi.Index >= 0 {
+		copies = 1
+	}
+
+	var task replicator.Task
+	task.SetObjectAddress(obj.Address())
+	task.SetObject(obj)
+	task.SetNodes(nodes)
+	task.SetCopiesNumber(copies)
+
+	if err := r.replicator.EnqueueTask(task); err != nil {
+		r.log.Warn("could not enqueue post-placement replication",
+			zap.Stringer("object", obj.Address()),
+			zap.Uint32("expected", copies),
+			zap.Error(err),
+		)
+	}
+}
+
 func (c *cfg) MaxObjectSize() uint64 {
 	sz, err := c.nCli.MaxObjectSize()
 	if err != nil {
@@ -212,7 +238,7 @@ func initObjectService(c *cfg) {
 		policer.WithBoostMultiplier(c.appCfg.Policer.BoostMultiplier),
 	)
 
-	c.workers = append(c.workers, c.policer)
+	c.workers = append(c.workers, c.policer, c.replicator)
 
 	nnsResolver := nns.NewResolver(c.cli)
 
@@ -246,6 +272,7 @@ func initObjectService(c *cfg) {
 		putsvc.WithContainerSource(c.cnrSrc),
 		putsvc.WithNetworkState(c.cfgNetmap.state),
 		putsvc.WithRemoteWorkerPool(c.cfgObject.pool.putRemote),
+		putsvc.WithPostPlacementReplicator(putPostInitialPlacementReplicator{log: c.log, replicator: c.replicator}),
 		putsvc.WithLogger(c.log),
 		putsvc.WithSplitChainVerifier(split.NewVerifier(sGet)),
 		putsvc.WithTombstoneVerifier(tombstone.NewVerifier(os)),
--- a/pkg/services/object/put/distributed.go
+++ b/pkg/services/object/put/distributed.go
@@ -79,6 +79,8 @@ type distributedTarget struct {
 	ecPart iec.PartInfo
 
 	initialPolicy *netmap.InitialPlacementPolicy
+
+	postPlacementReplicator PostPlacementReplicator
 }
 
 type nodeDesc struct {
@@ -190,6 +192,7 @@ func (t *distributedTarget) saveObject(obj object.Object, encObj encodedObject) 
 	// TODO: handle rules in parallel. https://github.com/nspcc-dev/neofs-node/issues/3503
 
 	repRules := t.containerNodes.PrimaryCounts()
+	fullRepRules := repRules
 	ecRules := t.containerNodes.ECRules()
 	if typ := obj.Type(); typ == object.TypeTombstone || typ == object.TypeLock || typ == object.TypeLink || len(obj.Children()) > 0 {
 		broadcast := typ != object.TypeLock || !t.localOnly
@@ -213,17 +216,21 @@ func (t *distributedTarget) saveObject(obj object.Object, encObj encodedObject) 
 	initial := t.initialPolicy != nil
 
 	if t.ecPart.RuleIndex >= 0 { // already encoded EC part
+		total := int(ecRules[t.ecPart.RuleIndex].DataPartNum + ecRules[t.ecPart.RuleIndex].ParityPartNum)
+		nodes := objNodeLists[len(repRules)+t.ecPart.RuleIndex]
+
 		// part info should already be verified, so we don't prevent out-of-range panic here
 		if initial && len(t.initialPolicy.ReplicaLimits()) != 0 && t.initialPolicy.ReplicaLimits()[len(repRules)+t.ecPart.RuleIndex] == 0 {
 			// Client can seal objects himself, but he is unlikely to enforce placement policies.
 			// Using SDK slicer as an example, the object is PUT to any SN.
-			// But this object should still not be saved according to initial policy.
-			// So, this is no-op.
+			// But this object should still be replicated according to the full policy.
+			postPlacementNodes := ecNodesForPart(nodes, t.ecPart.Index, total)
+			if len(postPlacementNodes) > 0 {
+				t.postPlacementReplicator.HandlePostPlacement(&obj, postPlacementNodes)
+			}
 			return nil
 		}
 
-		total := int(ecRules[t.ecPart.RuleIndex].DataPartNum + ecRules[t.ecPart.RuleIndex].ParityPartNum)
-		nodes := objNodeLists[len(repRules)+t.ecPart.RuleIndex]
 		return t.saveECPart(obj, encObj, t.ecPart.RuleIndex, t.ecPart.Index, total, nodes, &t.metaCollection)
 	}
 
@@ -240,9 +247,12 @@ func (t *distributedTarget) saveObject(obj object.Object, encObj encodedObject) 
 	var maxReplicas uint
 	var ecLimits []uint32
 	var ruleOrder []int
+	var appliedECRules []bool
 	ruleNum := len(repRules) + len(ecRules)
 
 	if initial {
+		appliedECRules = make([]bool, len(ecRules))
+
 		initialLimits := t.initialPolicy.ReplicaLimits()
 		if t.sessionSigner == nil { // same as above
 			if len(initialLimits) < len(repRules) {
@@ -330,6 +340,9 @@ func (t *distributedTarget) saveObject(obj object.Object, encObj encodedObject) 
 			}
 			t.placementIterator.log.Info("PUT by EC rule failure", zap.Stringer("object", obj.Address()), zap.Error(err))
 			return false, nil
+		}
+		if appliedECRules != nil {
+			appliedECRules[ecRuleIdx] = true
 		}
 		if maxReplicas > 0 {
 			leftReplicas--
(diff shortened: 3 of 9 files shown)


pkg/services/replicator/replicator.go
```diff
@@ -26,6 +33,8 @@ type Option func(*cfg)
 
 type cfg struct {
 	putTimeout time.Duration
+	queueSize  int
+	workers    int
```
End-rey: These parameters can be made configurable if this type of queue is suitable

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ Changelog for NeoFS Node
 ## [Unreleased]
 
 ### Added
+- Post-initial placement replication (#3923)
```
roman-khimov: That's not correct, we do support initial placements already. Here we're optimizing replications for them, it's not a new function.

pkg/services/object/put/distributed.go
```diff
@@ -240,6 +243,7 @@ func (t *distributedTarget) saveObject(obj object.Object, encObj encodedObject)
 	var maxReplicas uint
 	var ecLimits []uint32
 	var ruleOrder []int
+	appliedECRules := make([]bool, len(ecRules))
```
roman-khimov: Only relevant for policies with initial placement.

pkg/services/replicator/queue.go
```diff
@@ -0,0 +1,60 @@
...
+	"go.uber.org/zap"
+)
+
+// EnqueueTask submits replication task for asynchronous execution and blocks
+// until the task is accepted by the queue or the context is canceled.
+func (p *Replicator) EnqueueTask(ctx context.Context, task Task) error {
+	select {
+	case p.taskQueue <- task:
+		return nil
+	case <-ctx.Done():
```
roman-khimov: Rather should have `case default`, we can't _guarantee_ queuing (it's _always_ possible to overflow the queue) and if you leave it this way initial placement optimization won't really work, processes would get blocked here for the ~same amount of time it takes to properly replicate.

pkg/services/replicator/replicator.go
```diff
@@ -8,6 +8,13 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	// defaultQueueSize is the default capacity of an asynchronous replication queue.
+	defaultQueueSize = 128
```
roman-khimov: A 1000 of small objects can be put in a second. This should have some capacity for a number of seconds.

pkg/services/replicator/replicator.go
roman-khimov: Meh, we've just removed `replicator.pool_size`. It'd be nice to not have any configurations, let's have like 8 for now and we'll improve on it later.

pkg/services/object/put/distributed.go
```diff
@@ -428,12 +433,75 @@ nextRule:
...
 	if len(repRules) > 0 {
-		return t.submitMetaCollection(obj.Address(), &t.metaCollection)
+		err = t.submitMetaCollection(obj.Address(), &t.metaCollection)
+		if err != nil {
+			return err
+		}
 	}
 
+	t.replicateRemainingPrimaryNodes(obj, objNodeLists[:len(fullRepRules)], fullRepRules, repProg)
+	t.replicateRemainingECRules(obj, ecRules, objNodeLists[len(fullRepRules):], appliedECRules)
```
roman-khimov: Needs to be checked for various buffer reuses. Although `obj` is not a problem and `objNodeLists` should be fine too.

pkg/services/replicator/queue.go
End-rey: If the queue is full, the object will be skipped, and then the replicator will handle it?

roman-khimov: Exactly.

pkg/services/object/put/distributed.go
End-rey: The only thing I found is that `replicator.Task` now crosses the boundary of an asynchronous call, so it needs to maintain its own list of nodes instead of aliasing caller memory.

CHANGELOG.md
End-rey: Fixed.

pkg/services/object/put/distributed.go
End-rey: Allocate only for initial now.

pkg/services/replicator/queue.go
End-rey: Done.

pkg/services/replicator/replicator.go
End-rey: Made 4096.

roman-khimov (approved):

pkg/services/replicator/queue.go
```diff
@@ -0,0 +1,63 @@
...
+	"context"
+	"errors"
+	"sync"
+
+	"github.com/nspcc-dev/neofs-sdk-go/netmap"
+	"go.uber.org/zap"
+)
+
+// ErrQueueFull means that a replication task could not be queued immediately.
+var ErrQueueFull = errors.New("replication queue is full")
```
cthulhu-rider: used inside the package only, why exported?

pkg/services/replicator/queue.go
```diff
@@ -0,0 +1,63 @@
...
+// Run processes queued replication tasks until the context is canceled.
+func (p *Replicator) Run(ctx context.Context) {
+	var wg sync.WaitGroup
+	for range p.workers {
+		wg.Go(func() {
+			p.runWorker(ctx)
+		})
+	}
+
+	<-ctx.Done()
```
cthulhu-rider: seems excessive, we wait all workes to be interrupted by `ctx` anyway

pkg/services/replicator/queue.go
```diff
@@ -0,0 +1,63 @@
...
+	wg.Wait()
+}
+
+func (p *Replicator) runWorker(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case task := <-p.taskQueue:
+			res := new(countingReplicationResult)
```
cthulhu-rider: can be reused b/w iterations

pkg/services/replicator/task.go
```diff
@@ -34,7 +36,7 @@ func (t *Task) SetObject(obj *object.Object) {
 
 // SetNodes sets a list of potential object holders.
 func (t *Task) SetNodes(v []netmap.NodeInfo) {
-	t.nodes = v
+	t.nodes = slices.Clone(v)
```
cthulhu-rider: why cloning is needed?

pkg/services/replicator/replicator.go
```diff
@@ -34,10 +43,15 @@ type cfg struct {
 	localStorage *engine.StorageEngine
 
 	localNodeKey LocalNodeKey
+
+	taskQueue chan Task
```
cthulhu-rider: lets make this a `Replicator` field. It's configured but initialized internally

pkg/services/replicator/replicator.go
cthulhu-rider: no need to have them as `cfg` fields if they are not configurable. Just use consts

cmd/neofs-node/object.go
```diff
@@ -218,6 +248,7 @@ func initObjectService(c *cfg) {
 		policer.WithBoostMultiplier(c.appCfg.Policer.BoostMultiplier),
 	)
 
+	c.workers = append(c.workers, c.replicator)
```
cthulhu-rider: merge into existing append

pkg/services/object/put/ec.go
```diff
@@ -117,3 +127,13 @@ func (t *distributedTarget) saveECPartOnNode(ruleIdx int, obj object.Object, enc
 
 	return t.sendObject(obj, enc, n, metaC)
 }
+
+func ecNodesForPart(nodeList []netmap.NodeInfo, partIdx, totalParts int) []netmap.NodeInfo {
+	return slices.Collect(func(ni func(netmap.NodeInfo) bool) {
```
cthulhu-rider: looks trickier than it might be. Just make a slice of `len(nodeList)` and fill in `for`

cmd/neofs-node/object.go
```diff
@@ -67,6 +67,36 @@ type objectSvc struct {
 	delete *deletesvc.Service
 }
 
+type putPostPlacementReplicator struct {
+	log        *zap.Logger
+	replicator *replicator.Replicator
+}
+
+func (r putPostPlacementReplicator) HandlePostPlacement(obj *object.Object, nodes []netmapsdk.NodeInfo) {
+	if len(nodes) == 0 {
```
cthulhu-rider: it's senseless to me to call this w/ empty node list. Single current caller prevents this on it's own

pkg/services/object/put/distributed.go
```diff
@@ -428,12 +437,75 @@ nextRule:
...
+
 	return nil
 }
 
+func (t *distributedTarget) replicateRemainingPrimaryNodes(obj object.Object, nodeLists [][]netmap.NodeInfo, repRules []uint, prog *repProgress) {
+	if t.initialPolicy == nil || t.postPlacementReplicator == nil || len(nodeLists) == 0 || prog == nil {
```
cthulhu-rider: i wouldn't call this method in these cases at all. Same for EC

btw in which cases `t.postPlacementReplicator` is nil?

cthulhu-rider (commented):
partially viewed

pkg/services/replicator/task.go
End-rey: I thought about how the queue changed `Task` lifetime: it is no longer executed immediately, so it should own nodes instead of aliasing caller memory.

cthulhu-rider: > instead of aliasing caller memory

why not? It does not mutate it

pkg/services/replicator/queue.go
End-rey: Fixed.

pkg/services/replicator/queue.go
End-rey: Dropped.

pkg/services/replicator/queue.go
End-rey: Done.

pkg/services/replicator/replicator.go
End-rey: Done.

cmd/neofs-node/object.go
End-rey: Done.

pkg/services/object/put/ec.go
End-rey: Done.

cmd/neofs-node/object.go
End-rey: There were two callers, but I understood the point and checked before making the call.

pkg/services/object/put/distributed.go
End-rey: Done.

> btw in which cases t.postPlacementReplicator is nil?

Only in the tests, but I fixed that behavior.

pkg/services/replicator/task.go
End-rey: You're right, I don't currently see a place where that slice is mutated after being passed in. I added the clone as a safety measure because `Task` is now queued and can outlive the caller. If we don't want the extra copy, i can drop it and document that the caller must not modify or reuse the slice after `SetNodes`.

cmd/neofs-node/object.go
```diff
@@ -64,6 +64,32 @@ type objectSvc struct {
 	delete *deletesvc.Service
 }
 
+type putPostPlacementReplicator struct {
```
carpawell: `putPostInitialPlacementReplicator`?

cmd/neofs-node/object.go
```diff
@@ -64,6 +64,32 @@ type objectSvc struct {
...
+	}
+
+	var task replicator.Task
+	task.SetObjectAddress(obj.Address())
+	task.SetObject(obj)
+	task.SetNodes(nodes)
+	task.SetCopiesNumber(copies)
+
+	if err := r.replicator.EnqueueTask(task); err != nil {
+		r.log.Debug("could not enqueue post-placement replication",
```
carpawell: smth harder than `Debug`? at least `Warn`? i think placement replicas problems are quite critical. inside `distributedTarget` it is `Info`

pkg/services/object/put/distributed.go
carpawell, line 223: is this `if` branch handled correctly?

pkg/services/object/put/distributed.go
```diff
@@ -240,9 +243,12 @@ func (t *distributedTarget) saveObject(obj object.Object, encObj encodedObject)
 	var maxReplicas uint
 	var ecLimits []uint32
 	var ruleOrder []int
+	var appliedECRules []bool
```
carpawell, line 250: i would move everything about progress to a more general `replicationProgress` struct, both EC and REP, this func is too complex for me now. do not insist

pkg/services/replicator/task.go
carpawell: > Task is now queued and can outlive the caller

i think it is our common pattern. caller should not change a task he created

pkg/services/object/put/service_test.go
```diff
@@ -1584,81 +1822,61 @@ func testInitialPlacementREP(t *testing.T, repRules []uint, initialRules []uint3
...
-	}
+	env := newPlacementTestEnv(t, cnr, repRules, nil, true, false)
+	nodeIdx := nodeIndexForList(env.nodeLists, localList)
+	obj := newSignedRegularObject(t, creator, creator.ID)
 
-	storeObjectWithSession(t, cluster.nodeServices[0], obj, nil, nil)
+	storeObjectWithSession(t, env.cluster.nodeServices[nodeIdx], obj, nil, nil)
 
-	nodeObjLists := cluster.allStoredObjects()
+	require.Eventually(t, func() bool {
```
carpawell, line 1912: can `synctest` be applied there?

carpawell (commented):
Conflicts.

pkg/services/object/put/service_test.go
End-rey: I don't think so: this test waits for a real async side effect, so `Eventually` fits better here than `synctest`.

cmd/neofs-node/object.go
End-rey: Renamed.

cmd/neofs-node/object.go
End-rey: Done.

pkg/services/object/put/distributed.go
End-rey: This branch was overlooked. Now added the handling and a test for this case.

pkg/services/object/put/distributed.go
End-rey: I think it would be better to do this in another PR.

carpawell (approved):

Issue #1410, opened 2022-05-20
replication: Submit single task per object


cthulhu-rider:
In current implementation `Policer` can send several `Replicator` tasks related to single object per tact. This happens because tasks are submitted inside `processNodes` methods. We can accumulate check results and send no more than one replication job per object.

PR #3930, opened 2026-04-06
policer: drop ants pool for EC checks

roman-khimov:
Same reasons as in 4bcea7d3458e6552fdfbb66e222038e02bab4d10.

### DIFF
--- a/pkg/services/policer/policer.go
+++ b/pkg/services/policer/policer.go
@@ -2,7 +2,6 @@ package policer
 
 import (
 	"context"
-	"fmt"
 	"io"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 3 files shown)


pkg/services/policer/policer.go
```diff
@@ -57,10 +54,6 @@ type Policer struct {
 	hadPlacementMismatch atomic.Bool
 
 	signer neofscrypto.Signer
-
```
carpawell, line 60: can this removal be a problem? in policer, it was added for a certain reason: https://github.com/nspcc-dev/neofs-node/issues/1410

carpawell: i meant `checkECPartsProgressMtx`, GH links a strange line

roman-khimov: This means having at least two parts of the same object on the same node. Not the most probable scenario normally, only temporary in case of object redistribution. Then also these objects should have sufficiently large common OID prefix to appear in the same batch taken from metabase which can also happen, but just not very likely.

carpawell (approved):

PR #3931, opened 2026-04-08
Drop search v1

roman-khimov:


pkg/services/object/server.go
```diff
@@ -1714,192 +1719,8 @@ func continueRangeFromRemoteNode(ctx context.Context, conn *grpc.ClientConn, req
...
-		var id oid.ID
-		for i := range chunk {
-			if err := id.FromProtoMessage(chunk[i]); err != nil {
-				return nil, fmt.Errorf("invalid object ID: %w", err)
-			}
-			res = append(res, id)
-		}
-	}
+func (s *Server) Search(_ *protoobject.SearchRequest, _ protoobject.ObjectService_SearchServer) error {
+	return grpcstatus.Error(grpccodes.Unimplemented, "deprecated")
```
roman-khimov: @cthulhu-rider, not sure what's the most canonical way to do this. NeoFS API error is not applicable here, I want GRPC-level error specifically. Looks like this does the trick, but not tested.

cthulhu-rider: it's OK. I'd clarify the message: `No longer supported, use SearchV2 instead.`

cthulhu-rider (approved):

pkg/services/object/server.go
roman-khimov: Improved.

Issue #3133, opened 2025-02-14
Profile/optimize search result merger


roman-khimov:
## Is your feature request related to a problem? Please describe.
I'm always frustrated when we're not doing things efficiently.

## Describe the solution you'd like
Consider
https://github.com/nspcc-dev/neofs-node/pull/3111#discussion_r1955934791
https://github.com/nspcc-dev/neofs-node/pull/3111#discussion_r1955923040

We need to add some benchmarks with more or less realistic and worst cases before trying them.

## Describe alternatives you've considered
Keep it as is, it works.

## Additional context
https://github.com/nspcc-dev/neofs-node/issues/3058

PR #3932, opened 2026-04-08
core/object: add benchmarks for `MergeSearchResults`

End-rey:
Refs #3133.

As far as I can see, benchmarks show that the main cost comes from numeric attribute comparison `cmpInt` and from large merges with low overlap between result sets.

### DIFF
--- a/pkg/core/object/metadata_bench_test.go
+++ b/pkg/core/object/metadata_bench_test.go
@@ -1,0 +1,370 @@
+package objectcore
+
+import (
+	"fmt"
+	"strconv"
+	"testing"
+
+	"github.com/nspcc-dev/neofs-sdk-go/client"
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	benchSetLen       = 10_000
+	benchDenseDataLen = 30_000
+
+	benchOverlapShift = benchSetLen * 9 / 10
+
+	benchSmallObjectSize        = 4 * 1024
+	benchMediumObjectSize       = 512 * 1024
+	benchLargeObjectSize  int64 = 8 * 1024 * 1024 * 1024
+)
+
+func BenchmarkMergeSearchResults(b *testing.B) {
+	for _, lim := range []uint16{1, 2, 1000, 10000} {
+		b.Run(fmt.Sprintf("lim_%d", lim), func(b *testing.B) {
+			for _, setCount := range []int{2, 4, 8, 16} {
+				b.Run(fmt.Sprintf("set_count_%d", setCount), func(b *testing.B) {
+					b.Run("id_identical", func(b *testing.B) {
+						benchMergeSearchResults(b, mergeBenchCase{
+							lim:     lim,
+							sets:    makeIdenticalSets(setCount, makeIDOnlyItems(benchSetLen)),
+							mores:   make([]bool, setCount),
+							wantLen: min(int(lim), benchSetLen),
+						})
+					})
+
+					b.Run("id_disjoint", func(b *testing.B) {
+						benchMergeSearchResults(b, mergeBenchCase{
+							lim:     lim,
+							sets:    makeDisjointIDOnlySets(setCount, benchSetLen),
+							mores:   make([]bool, setCount),
+							wantLen: min(int(lim), benchSetLen),
+						})
+					})
+
+					b.Run("id_intersecting", func(b *testing.B) {
+						benchMergeSearchResults(b, mergeBenchCase{
+							lim:     lim,
+							sets:    makeSlidingIDOnlySets(setCount, benchSetLen, benchOverlapShift),
+							mores:   make([]bool, setCount),
+							wantLen: min(int(lim), benchSetLen),
+						})
+					})
+
+					b.Run("size_numeric", func(b *testing.B) {
+						benchMergeSearchResults(b, mergeBenchCase{
+							lim:       lim,
+							cmpInt:    true,
+							sets:      makeNumericSizeSets(setCount, benchSetLen, benchOverlapShift),
+							mores:     make([]bool, setCount),
+							firstAttr: "Size",
+							wantLen:   min(int(lim), benchSetLen),
+						})
+					})
+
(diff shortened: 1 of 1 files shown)


pkg/core/object/metadata_bench_test.go
```diff
@@ -0,0 +1,368 @@
...
+	"strconv"
+	"testing"
+
+	"github.com/nspcc-dev/neofs-sdk-go/client"
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	"github.com/stretchr/testify/require"
+)
+
+func BenchmarkMergeSearchResults(b *testing.B) {
+	for _, lim := range []uint16{1, 2, 1000, 10000} {
```
roman-khimov, line 25: 10K isn't practicaly (search API is limited by 1K), but maybe OK for a benchmark, just to see how it goes.

pkg/core/object/metadata_bench_test.go
```diff
@@ -0,0 +1,368 @@
...
+		b.Run(fmt.Sprintf("lim_%d", lim), func(b *testing.B) {
+			for _, setCount := range []int{4, 16} {
```
roman-khimov: 2 is a a good number for shards in our current tests. 8 is a good number of nodes.

pkg/core/object/metadata_bench_test.go
```diff
@@ -0,0 +1,368 @@
...
+							cmpInt:    true,
+							sets:      makeNumericSizeSets(setCount, 10000, 8500),
+							mores:     make([]bool, setCount),
+							firstAttr: "Size",
+						})
+					})
+				})
+			}
+
+			for _, setCount := range []int{3, 6} {
```
roman-khimov: Why different sets here?

pkg/core/object/metadata_bench_test.go
```diff
@@ -0,0 +1,368 @@
...
+	sets      [][]client.SearchResultItem
+	mores     []bool
+}
+
+func benchMergeSearchResults(b *testing.B, tc mergeBenchCase) {
+	b.Helper()
+	b.ReportAllocs()
+	b.ResetTimer()
+
+	for range b.N {
```
roman-khimov: `for b.Loop()`

pkg/core/object/metadata_bench_test.go
```diff
@@ -0,0 +1,368 @@
...
+		sets := make([][]client.SearchResultItem, len(tc.sets))
+		copy(sets, tc.sets)
```
roman-khimov, line 134: This should be done before `ResetTimer()` out of loop.

pkg/core/object/metadata_bench_test.go
```diff
@@ -0,0 +1,368 @@
...
+	shared := makeCommonPrefixItems(30000, setCount)
+	sets := make([][]client.SearchResultItem, setCount)
+	for i := range setCount {
+		switch {
+		case i < setCount-2:
+			sets[i] = shared[:10000]
+		case i == setCount-2:
+			sets[i] = shared[50:10050]
+		default:
+			sets[i] = shared[120:10120]
```
roman-khimov: Many hardcoded constants.

pkg/core/object/metadata_bench_test.go
```diff
@@ -0,0 +1,368 @@
...
+			Attributes: []string{
+				fmt.Sprintf("/bucket/uploads/%03d/part-%06d.bin", i%128, i), "4096",
+			},
+		})
+	}
+	for i := range largeN {
+		idx := smallN + i
+		res = append(res, client.SearchResultItem{
+			ID: benchOID(idx),
+			Attributes: []string{
```
roman-khimov: Add more attributes? Does it change anything? S3 requests a lot of them.

roman-khimov (commented):
But cases look realistic overall.

roman-khimov:
Can we also have a limit of 1K, but just a single real result? This looks realistic for S3 as well (most of the time it doesn't specifically limit the number of results, but then we don't have many `cat.png` results, a single from shard and some number from nodes (3 for REP, 8 for EC)).

pkg/core/object/metadata_bench_test.go
End-rey: Left 2 and 8 then.

pkg/core/object/metadata_bench_test.go
End-rey: I used to distinguish between real use cases and synthetic ones. But I've removed that distinction now.

pkg/core/object/metadata_bench_test.go
End-rey: Done.

pkg/core/object/metadata_bench_test.go
End-rey: I moved the memory allocation itself outside the loop, but I can't move it entirely because `MergeSearchResults` modifies this sets, which means it can't be passed back to the function.

pkg/core/object/metadata_bench_test.go
End-rey: Tried to move some of it into constants and clean it up.

pkg/core/object/metadata_bench_test.go
End-rey: Added 10 attributes for S3.

pkg/core/object/metadata_bench_test.go
roman-khimov: We better keep 4 and 16 as well, this thing should scale predictably, 10-20 shards or nodes are feasible even if we don't have them immediately (but we had 20 shards in one test, btw).

roman-khimov (approved):
OK otherwise. And the numbers suggest that it's not that bad even now, just integer handling that is clearly problematic. But other improvements can be useful too.

roman-khimov:
For integers https://github.com/holiman/uint256 is likely to help, `math.Big` is more generic, but less effective.

pkg/core/object/metadata_bench_test.go
End-rey: Done.

PR #3934, opened 2026-04-10
Search check fix

roman-khimov:
Follow-up to #3931.

### DIFF
--- a/pkg/services/object/server.go
+++ b/pkg/services/object/server.go
@@ -159,7 +159,6 @@ type ACLInfoExtractor interface {
 	HashRequestToInfo(*protoobject.GetRangeHashRequest) (aclsvc.RequestInfo, error)
 	GetRequestToInfo(*protoobject.GetRequest) (aclsvc.RequestInfo, error)
 	RangeRequestToInfo(*protoobject.GetRangeRequest) (aclsvc.RequestInfo, error)
-	SearchRequestToInfo(*protoobject.SearchRequest) (aclsvc.RequestInfo, error)
 	SearchV2RequestToInfo(*protoobject.SearchV2Request) (aclsvc.RequestInfo, error)
 }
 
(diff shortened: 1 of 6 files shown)


pkg/services/object/server.go
```diff
@@ -1961,6 +1961,16 @@ func verifySearchFilter(f *protoobject.SearchFilter) error {
 			return fmt.Errorf("value for attribute %s is prohibited", f.Key)
 		}
 	}
+	//nolint:exhaustive
+	switch f.MatchType {
```
cthulhu-rider: this is correct and match previous behavior. But i also see that we do this in https://github.com/nspcc-dev/neofs-node/blob/943599af26e25130f5d3deb8ac033903e15690d8/pkg/core/object/metadata.go#L712-L715

i'd either apply proposed change there or deduplicate these checks in the future (not a critical point, but still)

roman-khimov: Well, the reality is that we don't get any error in https://rest.fs.neo.org/Aku38wpn2xEqfEPxmwyAa7N3VnAkMV8PEVkpLqwKM1z3/5141-1775829634/index.html#suites/87d8da5f92824d091055170c29079bc0/5e5dbbff96715ad4/ after `search` became and alias to `searchv2`.

roman-khimov: Huh
https://github.com/nspcc-dev/neofs-node/blob/943599af26e25130f5d3deb8ac033903e15690d8/pkg/services/object/server.go#L2047-L2053

cthulhu-rider: don't we test the same for searchv2?

roman-khimov: Reworked a bit, should be better now.

carpawell (approved):
BTW, which cli search is used in test-cases? I would stop using searchV2 starting from now, do we have an issue about it?

roman-khimov:
https://github.com/nspcc-dev/neofs-testcases/issues/1327

PR #3938, opened 2026-04-13
Optimize `MergeSearchResults`

### LINKED ISSUE #3133: Profile/optimize search result merger

End-rey:
Closes #3133.

https://github.com/nspcc-dev/neofs-node/pull/3111#discussion_r1955923040:
`calcMaxUniqueSearchResults` is not just a preallocation helper. Its result also shrinks the effective limit used by `MergeSearchResults`, which lets the merge stop earlier in high-overlap/low-unique-result cases. Replacing it with a coarse capacity estimate reduced code complexity, but benchmarked worse because that early-stop effect was lost.
[baseVSnoMaxUniqueResults.txt](https://github.com/user-attachments/files/26680433/baseVSnoMaxUniqueResults.txt)

https://github.com/nspcc-dev/neofs-node/pull/3111#discussion_r1955934791:
I tried replacing the linear head scan with a classic heap-based k-way merge. Although this looks simpler algorithmically, benchmarks showed it to be slower overall for our real fan-in sizes: the `container/heap` overhead was higher than the cost of scanning a small number of set heads linearly.
[baseVSheap.txt](https://github.com/user-attachments/files/26680431/baseVSheap.txt)

Kept the direct decimal string comparison optimization instead of `big.Int`.

### DIFF
--- a/pkg/core/object/metadata.go
+++ b/pkg/core/object/metadata.go
@@ -43,10 +43,6 @@ func MergeSearchResults(lim uint16, firstAttr string, cmpInt bool, sets [][]clie
 	lim = calcMaxUniqueSearchResults(lim, sets)
 	res := make([]client.SearchResultItem, 0, lim)
 	var more bool
-	var minInt, curInt *big.Int
-	if cmpInt {
-		minInt, curInt = new(big.Int), new(big.Int)
-	}
 	var minOID, curOID oid.ID
 	var minUsr, curUsr user.ID
 	var err error
@@ -58,7 +54,7 @@ func MergeSearchResults(lim uint16, firstAttr string, cmpInt bool, sets [][]clie
 			if minInd < 0 {
 				minInd = i
 				if cmpInt {
-					if _, ok := minInt.SetString(sets[i][0].Attributes[0], 10); !ok {
+					if _, _, err = splitIntString(sets[i][0].Attributes[0]); err != nil {
 						return nil, false, fmt.Errorf("non-int attribute in result #%d", i)
 					}
 				}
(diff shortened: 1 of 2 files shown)


pkg/core/object/metadata.go
```diff
@@ -150,6 +140,78 @@ func MergeSearchResults(lim uint16, firstAttr string, cmpInt bool, sets [][]clie
...
+func splitIntString(s string) (bool, string, error) {
+	if s == "" {
+		return false, "", errors.New("empty")
+	}
+
+	var (
+		i   int
+		neg bool
+	)
+	if s[0] == '-' {
```
roman-khimov: Can have `+` as well.

pkg/core/object/metadata.go
```diff
@@ -43,10 +43,6 @@ func MergeSearchResults(lim uint16, firstAttr string, cmpInt bool, sets [][]clie
 	lim = calcMaxUniqueSearchResults(lim, sets)
 	res := make([]client.SearchResultItem, 0, lim)
 	var more bool
-	var minInt, curInt *big.Int
-	if cmpInt {
-		minInt, curInt = new(big.Int), new(big.Int)
```
roman-khimov, line 48: This means parsing for every other comparison. Does anything change if parsing result is cached?

roman-khimov (commented):
Worth mentioning as optimization in chlog.

pkg/core/object/metadata.go
End-rey: Caching the parsed form of the current min gave a small benefit for tiny limits, but regressed larger limits. The remaining string parsing was cheap enough that extra cache maintenance in the hot loop outweighed the saved reparsing for longer merges.

roman-khimov: OK then

pkg/core/object/metadata.go
End-rey: Done.

roman-khimov (approved):

Issue #3914, opened 2026-03-27
Optimize SEARCH for optimally-placed containers


roman-khimov:
## Is your feature request related to a problem? Please describe.
I'm always frustrated when we're making more requests to other nodes than we could. Consider REP 3 CBF 3 container, it has 9 nodes, but a reply from 7 of them is sufficient if all replicas are present, the other two don't have any additional objects. Consider EC 6/2 CBF 1 container, _any_ single node has all the data and to ensure we're not missing anything any 3 out of 8 are sufficient. 

## Describe the solution you'd like
Think of optimizations in this direction.

roman-khimov:
Safe optimizations can be considered first. For REP there are none that are valid since any node can be a single replica holder, but for EC the optimization works perfectly since  most of the time it's `CBF 1` and `parity parts + 1` number of nodes (+ those over CBF if it's not 1) guarantees 100% that we have an objects. For a typical 7+3 that's 4 requests instead of 10, huge difference. Can be distributed more or less randomly.

PR #3940, opened 2026-04-13
Optimize EC SEARCH

roman-khimov:


pkg/services/object/server.go
```diff
@@ -2085,64 +2086,72 @@ func (s *Server) ProcessSearch(ctx context.Context, req *protoobject.SearchV2Req
...
+			sets            [][]sdkclient.SearchResultItem
+			expectedRes     int
+		)
+
+		req.MetaHeader = &protosession.RequestMetaHeader{Ttl: 1, Origin: req.MetaHeader}
+		if req.VerifyHeader, err = neofscrypto.SignRequestWithBuffer[*protoobject.SearchV2Request_Body](neofsecdsa.Signer(s.signer), req, nil); err != nil {
+			return nil, nil, fmt.Errorf("sign request: %w", err)
+		}
+
+		var allNodes = (len(attrs) == 0) || (!strings.HasPrefix(attrs[0], iec.AttributePrefix))
```
roman-khimov: 1. Maybe worth checking all attributes.
2. Consider `object hash == smth`, non-EC, but can return a part as well, check for all system attributes?

roman-khimov: And it should look at filters, not attributes.

pkg/services/object/server.go
```diff
@@ -104,14 +106,14 @@ type FSChain interface {
 	// found.
 	ForEachContainerNodePublicKeyInLastTwoEpochs(cid.ID, func(pubKey []byte) bool) error
 
-	// ForEachContainerNode iterates over all nodes matching the referenced
+	// ForSearchableContainerNode iterates over all nodes matching the referenced
```
carpawell: correct the description?

pkg/services/object/server.go
```diff
@@ -2104,7 +2106,12 @@ func (s *Server) ProcessSearch(ctx context.Context, req *protoobject.SearchV2Req
 			return nil, nil, fmt.Errorf("sign request: %w", err)
 		}
 
-		err = s.fsChain.ForEachContainerNode(cID, func(node sdknetmap.NodeInfo) bool {
+		var optimizedNodes = (len(body.Filters) != 0) &&
+			slices.ContainsFunc(body.Filters, func(filt *protoobject.SearchFilter) bool {
+				return !strings.HasPrefix(filt.Key, "$Object:") && !strings.HasPrefix(filt.Key, "__NEOFS__")
```
carpawell, line 2112: can you, please, explain the idea of this optimization? i feel like the only case we can safely omit some nodes is a EC container and an `$Object:ROOT` filter

carpawell (commented):
Why it closes #3914. It also has REP suggestion.

roman-khimov:
> It also has REP suggestion.

Those suggestions aren't safe (consider incomplete placement/netmap changes). This one is.

pkg/services/object/server.go
roman-khimov: EC parts don't have "normal" attributes like `FilePath`, so any request for `FilePath=cat.png` is safe to optimize this way. Regarding `$Object:ROOT`, that's the only object-level attribute that can be considered as safe. Can be added into the check, but to me regular attributes are more useful. And for S3 this should be sufficient.

pkg/services/object/server.go
roman-khimov: Sure, fixed.

pkg/services/object/server.go
carpawell: > EC parts don't have "normal" attributes like FilePath

as long as it is true, i am ok

carpawell (approved):

PR #3944, opened 2026-04-15
wc: do not produce logs when deleting a missing object

carpawell:
```
writecache/flush.go:155	can't flush objects	{"shard_id": "4YVJ1aJRfMw8mnWaqcUECj", "substorage": "write-cache", "worker": 12, "first_object": "Fz6GUm1oVtwKBnToZNEnhPUgGRU6SQ1aftUkbY4yGa3W/CihtqZKhvhKJY76XsRpywpUBwXtnzS1wHficeXKnwuSK", "error": "logical error: status: code = 2049 message = object not found"}
```

### DIFF
--- a/pkg/local_object_storage/writecache/flush.go
+++ b/pkg/local_object_storage/writecache/flush.go
@@ -237,7 +237,7 @@ func (c *cache) flushBatch(addrs []oid.Address) error {
 			c.log.Error("can't remove object from write-cache", zap.Error(err))
 		}
 	}
-	return err
+	return nil
 }
 
 func (c *cache) getObject(addr oid.Address) ([]byte, error) {


pkg/local_object_storage/writecache/delete.go
carpawell, line 27: maybe this log should be also skipped

pkg/local_object_storage/writecache/delete.go
```diff
@@ -20,6 +23,10 @@ func (c *cache) Delete(addr oid.Address) error {
 
 func (c *cache) delete(addr oid.Address) error {
 	err := c.fsTree.Delete(addr)
+	if errors.As(err, new(apistatus.ObjectNotFound)) {
```
End-rey: This change seems broader than necessary. I think it would be better to keep `delete` semantics as-is and suppress `ObjectNotFound` locally in `flushSingle/flushBatch` instead of changing `delete` globally, since it is also called from outside.

pkg/local_object_storage/writecache/delete.go
End-rey, line 30: Also, if you ignore the error, the counters will process it again.

pkg/local_object_storage/writecache/delete.go
carpawell: to me, removing smth that is not found, should be `err==nil`

> it is also called from outside.

as i see, it is used outside only to ignore it. i think every client is expecting to ignore such cases, so it can be done in a single place

carpawell: ok, ignore it locally

pkg/local_object_storage/writecache/flush.go
End-rey: I noticed another strange thing, in `flushSingle`, delete errors are only logged and never returned, but in `flushBatch` the function can still return the last delete error. If post-flush delete is considered non-fatal cleanup, `flushBatch` should probably also just log it and return `nil`.

carpawell: also noticed, but did not want to mess with any logic there, let it return `nil` always too

End-rey:
I guess a squash is needed.

End-rey (approved):

roman-khimov (approved):

PR #3946, opened 2026-04-15
Simplify, fix and resync counters for the next version

roman-khimov:


pkg/local_object_storage/metabase/version.go
```diff
@@ -394,6 +395,16 @@ func migrateFrom9Version(db *DB) error {
 	})
 }
 
+func migrateFrom10Version(db *DB) error {
```
carpawell, line 398: just for the sake of order: our migrations always have a test

carpawell (approved):
What really bothers me: none of tests' lines were changed. I would try to improve coverage.

End-rey (approved):

pkg/local_object_storage/metabase/version.go
roman-khimov: It does nothing but a counter sync which is tested in a number of ways, so this doesn't matter much to me. The migration function itself is executed in tests as well and nothing breaks there.

PR #3954, opened 2026-04-17
node: simplify IR key fetcher

roman-khimov:
1. Returning old value from it is safe (similar to MaxObjectSize).
2. It doesn't need TTL cache, it's a single value. This removes exclusive lock as well (1.5% of overall mutex waits).

### DIFF
--- a/cmd/neofs-node/cache.go
+++ b/cmd/neofs-node/cache.go
@@ -9,7 +9,6 @@ import (
 	"github.com/nspcc-dev/neofs-node/pkg/core/container"
 	"github.com/nspcc-dev/neofs-node/pkg/core/netmap"
 	cntClient "github.com/nspcc-dev/neofs-node/pkg/morph/client/container"
-	putsvc "github.com/nspcc-dev/neofs-node/pkg/services/object/put"
 	apistatus "github.com/nspcc-dev/neofs-sdk-go/client/status"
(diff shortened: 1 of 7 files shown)


cmd/neofs-node/cache.go
```diff
@@ -410,77 +409,62 @@ func (s *ttlContainerLister) reset() {
...
-func newCachedIRFetcher(f interface{ InnerRingKeys() ([][]byte, error) }) *cachedIRFetcher {
-	const (
-		irFetcherCacheSize = 1 // we intend to store only one value
-
-		// Without the cache in the testnet we can see several hundred simultaneous
-		// requests (neofs-node #1278), so limiting the request rate solves the issue.
-		//
-		// Exact request rate doesn't really matter because Inner Ring list update
-		// happens extremely rare, but there is no FS chain events for that as
-		// for now (neofs-contract v0.15.0 notary disabled env) to monitor it.
```
carpawell, line 426: i think the key to this issue is to make an event about changing the keys and not having TTL at all, not dropping this comment

carpawell (approved):
Ok as an improvement.

cmd/neofs-node/cache.go
roman-khimov: The event _is_ here since some NeoGo HF (committee change), but that's something to improve in future.

PR #3957, opened 2026-04-20
Improve local requests

roman-khimov:
Shave off some useless actions and cache modifications.

### DIFF
--- a/pkg/services/object/acl/v2/request.go
+++ b/pkg/services/object/acl/v2/request.go
@@ -11,82 +11,20 @@ import (
 // RequestInfo groups parsed version-independent (from SDK library)
 // request information and raw API request.
 type RequestInfo struct {
-	basicACL    acl.Basic
-	requestRole acl.Role
-	operation   acl.Op // put, get, head, etc.
+	BasicACL    acl.Basic
+	RequestRole acl.Role
+	Operation   acl.Op // put, get, head, etc.
 
-	idCnr cid.ID
+	Cnr cid.ID
 
 	// optional for some request
 	// e.g. Put, Search
-	obj *oid.ID
+	Obj *oid.ID
 
-	senderKey     []byte
-	senderAccount *user.ID
+	SenderKey     []byte
+	SenderAccount *user.ID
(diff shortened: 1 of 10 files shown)


roman-khimov:
Forwarded EC requests failed, looks like they have TTL=1 but do not request any specific part. So they're kinda local, but at the same time not so much since EC code will produce additional requests to other nodes. Are TTL=1 non-part requests valid at all, @cthulhu-rider?

pkg/services/object/get/get.go
```diff
@@ -36,6 +36,13 @@ func (s *Service) Get(ctx context.Context, prm Prm) error {
 		return s.copyLocalECPart(prm.objWriter, prm.addr.Container(), prm.addr.Object(), pi)
 	}
 
+	if prm.common.LocalOnly() &&
+		len(prm.container.PlacementPolicy().ECRules()) == 0 && // EC breaks TTL requirements currently.
```
carpawell, line 40: do we have an issue for that?

> currently it does part-specific requests with TTL=1 which is not correct

hard to understand immediately

carpawell: what is the difference b/w split object forwarding and EC object?

roman-khimov: No big difference, secondary requests are being created in both cases. You mean we can have the same thing going on for splits? This just makes TTL even more broken.

pkg/services/object/acl/v2/request.go
```diff
@@ -11,11 +12,11 @@ import (
 // RequestInfo groups parsed version-independent (from SDK library)
 // request information and raw API request.
 type RequestInfo struct {
-	BasicACL    acl.Basic
 	RequestRole acl.Role
 	Operation   acl.Op // put, get, head, etc.
 
-	Cnr cid.ID
+	Container   container.Container
```
cthulhu-rider, line 18: unclear why this change is needed on its own. I'd squash

cthulhu-rider (approved):
seems ok

PR #3959, opened 2026-04-21
sn: add post-placement queue controls to replicator config

### LINKED ISSUE #3956: Add a config value to slowdown or turn off a replication process after an initial placement
See https://github.com/nspcc-dev/neofs-testcases/issues/1337

https://rest.fs.neo.org/Aku38wpn2xEqfEPxmwyAa7N3VnAkMV8PEVkpLqwKM1z3/5170-1776269772/index.html#suites/6154bb8bfc29a4aa44bb883b9246b560/761fad8204e99708/

End-rey:
Closes #3956.


cmd/neofs-node/config/replicator/replicator.go
```diff
@@ -5,9 +5,17 @@ import "time"
 // PutTimeoutDefault is the default timeout of object put request in replicator.
 const PutTimeoutDefault = time.Minute
 
+// PostPlacementQueue contains configuration for the queue of post-placement
+// replication tasks submitted after initial object placement.
+type PostPlacementQueue struct {
+	Disabled bool          `mapstructure:"disabled"`
+	Delay    time.Duration `mapstructure:"delay"`
```
roman-khimov: Do we need both, @evgeniiz321?

cmd/neofs-node/object.go
```diff
@@ -82,13 +92,31 @@ func (r putPostInitialPlacementReplicator) HandlePostPlacement(obj *object.Objec
...
+		}
+	}
+
+	if r.delay <= 0 {
+		enqueue()
+		return
 	}
+
+	go func() {
+		timer := time.NewTimer(r.delay)
```
roman-khimov: Delay seems _so_ artificial here. It could have some meaning as a throttling mechanism in the replicator itself (sleep between tasks).

cmd/neofs-node/config/replicator/replicator.go
evgeniiz321: For tests - no, disabled is ok, delay is also ok. Keep the one that you like the most :)

roman-khimov: Then it's disabled to me and then it doesn't need to be nested in config. `replicator.disable_post_initial_queue` maybe.

End-rey: Done.

roman-khimov (approved):

roman-khimov:
Conflicts

carpawell (approved):

PR #3971, opened 2026-04-29
metabase: store associated object field in bytes 

### LINKED ISSUE #3926: Reconsider associated object field format in metabase
I'm always frustrated when we have to base58 encode OIDs for associated (lock/ts) object check.

1. Check that non-OID contents is not accepted for `__NEOFS__ASSOCIATE` (this should already be true).
2. Convert from base58 to bytes on metabase PUT.
3. Use these bytes when reading, convert back if/when needed.
4. Check performance gains/losses for different cases.
5. Add migration code.

Things work as is and current scheme is simpler, but it seriously affects search/existence checks while it seems like base58-encoded value is not needed as often as unencoded one.

End-rey:
Closes #3926.


pkg/local_object_storage/metabase/get.go
```diff
@@ -163,6 +163,14 @@ func get(metaCursor *bbolt.Cursor, addr oid.Address, checkStatus, raw bool, curr
 		case object.FilterPhysical, object.FilterRoot:
 			// Not real attributes, ignored.
 		default:
+			if string(attrKey) == object.AttributeAssociatedObject && len(attrVal) == oid.Size {
```
roman-khimov: But it's a natural part of the switch above, isn't it?

pkg/local_object_storage/metabase/lock_test.go
```diff
@@ -53,9 +55,11 @@ func TestDB_IsLocked(t *testing.T) {
 		cnr := cidtest.ID()
 
 		o := objecttest.Object()
+		o.ResetRelations()
```
roman-khimov: Maybe we need to fix `objecttest` or have some local regular test object function?

pkg/local_object_storage/metabase/put.go
```diff
@@ -198,7 +198,7 @@ func handleLinkObject(diff *CountersDiff) error {
 func handleObjectWithAssociation(metaBkt *bbolt.Bucket, diff *CountersDiff, currEpoch uint64, obj object.Object) error {
 	target := obj.AssociatedObject()
 	if target.IsZero() {
-		return nil
+		return fmt.Errorf("%s object has zero associated object", obj.Type())
```
roman-khimov, line 201: Somewhat different problem, this `nil` was relevant before 5675b9d799ab1ebf199b68959deff4031fd9904c, now it's obsolete, yes.

pkg/local_object_storage/metabase/version.go
```diff
@@ -445,6 +450,152 @@ func dropHomomorphicIndexes(_ *zap.Logger, _ *bbolt.Tx, b *bbolt.Bucket, _ cid.I
 	return uint(len(keysToDrop)), nil, nil
 }
 
+func migrateAssociatedObjectValueToIDBytes(l *zap.Logger, _ *bbolt.Tx, b *bbolt.Bucket, cnr cid.ID, afterKey []byte, rem uint) (uint, []byte, error) {
+	c := b.Cursor()
+	k, _ := c.First()
```
roman-khimov: Position to associated ID prefix immediately, then you have AttrID and you can get IDAttr to change from it as well. You don't want to scan the whole DB, in fact this attribute is pretty rare.

pkg/local_object_storage/metabase/get.go
End-rey: Fixed.

pkg/local_object_storage/metabase/lock_test.go
End-rey: Used existing helpers.

pkg/local_object_storage/metabase/version.go
End-rey: Done.

pkg/local_object_storage/metabase/version.go
```diff
@@ -445,6 +450,127 @@ func dropHomomorphicIndexes(_ *zap.Logger, _ *bbolt.Tx, b *bbolt.Bucket, _ cid.I
...
+		scanned++
+		nextKey = slices.Clone(k)
+
+		val, idRaw, err := splitAttributeValueObjectID(k[len(pref):])
+		if err != nil {
+			l.Warn("skip malformed associated object metadata entry during migration",
+				zap.Stringer("container", cnr),
+				zap.String("key", hex.EncodeToString(k)),
+				zap.Error(err))
+			if scanned == rem {
```
roman-khimov: This one is not needed, keep going.

pkg/local_object_storage/metabase/version.go
```diff
@@ -445,6 +450,127 @@ func dropHomomorphicIndexes(_ *zap.Logger, _ *bbolt.Tx, b *bbolt.Bucket, _ cid.I
...
+			continue
+		}
+
+		var associated oid.ID
+		if err = associated.DecodeString(string(val)); err != nil {
+			l.Warn("skip malformed associated object metadata entry during migration",
+				zap.Stringer("container", cnr),
+				zap.String("key", hex.EncodeToString(k)),
+				zap.Error(err))
+			if scanned == rem {
```
roman-khimov: As well as this one. We limit the number of _actions done_ to flush things periodically and be prepared for interruptions, but limiting the number of reads doesn't make sense in migration (can lead to a number of transactions changing nothing for example).

roman-khimov:
Changelog, btw.

pkg/local_object_storage/metabase/version.go
```diff
@@ -445,6 +450,121 @@ func dropHomomorphicIndexes(_ *zap.Logger, _ *bbolt.Tx, b *bbolt.Bucket, _ cid.I
...
+		if err != nil {
+			l.Warn("skip malformed associated object metadata entry during migration",
+				zap.Stringer("container", cnr),
+				zap.String("key", hex.EncodeToString(k)),
+				zap.Error(err))
+			continue
+		}
+
+		if _, err = oid.DecodeBytes(val); err == nil {
+			if scanned == rem {
```
roman-khimov: This one too.

pkg/local_object_storage/metabase/version.go
```diff
@@ -445,6 +450,121 @@ func dropHomomorphicIndexes(_ *zap.Logger, _ *bbolt.Tx, b *bbolt.Bucket, _ cid.I
...
+
+	var (
+		scanned uint
+		nextKey []byte
+		buf     keyBuffer
+		updates []keyRewrite
+	)
+
+	for ; k != nil && bytes.HasPrefix(k, pref); k, _ = c.Next() {
+		scanned++
```
roman-khimov: In fact this should be done to the end of the cycle, when you're adding something to `updates`.

roman-khimov (approved):

Issue #3972, opened 2026-04-29
Non-GRPC API


roman-khimov:
## Is your feature request related to a problem? Please describe.
I'm always frustrated when we can't transfer data fast enough. Maybe custom codecs and binary dances can get us to the level we need, but it's not yet obvious.

## Describe the solution you'd like
Consider some APIs that are not GRPC-based. We technically can have HTTP2 directly served from the same interface. We can transfer protobuf things for regular HTTP requests as an easy option.

## Describe alternatives you've considered
QUIC?

## Additional context
https://github.com/grpc/grpc-dotnet/issues/1186
https://github.com/grpc/grpc-dotnet/issues/1944
https://ops.tips/blog/sending-files-via-grpc/

roman-khimov:
Things tried in https://github.com/nspcc-dev/neofs-node/pull/3973 and https://github.com/nspcc-dev/neofs-node/pull/4010 turned out to be not very effective. QUIC is a complete disaster, HTTP/2 shows slight regressions even though it shouldn't, HTTP 1.1 is the only interesting case that is more effective with lower number of threads yet at the same time it's not so impressive with small objects and big number of threads. The gain itself is also not that dramatic to justify the change.

Most likely all of the prototypes can be improved, but the same resources can invested into gRPC and likely improve it as well. So while gRPC isn't designed for our use case it works good enough and we can reach better numbers without any shifts to other protocols.

PR #3973, opened 2026-04-29
Experiment with Go http

roman-khimov:


roman-khimov:
What we need here is to finish this prototype with an end-to-end GET data transport over HTTP, from S3 gateway (additional branch there) to SN. Many things can be improved here performance-wise, but we're interested in a quick prototype first, so many of object service internals are still left as is and just two proto-encoded byte arrays are pushed to the client (that needs to parse them).

End-rey:
Ran load against target (HTTP/2) vs base (gRPC): REP3, 4 KB and 4 MB objects, 1/4/16 threads/node.

Result: at 1 thread everything is within the margin of error, but the gap grows steadily as threads increase - the prototype scales worse under concurrency.

| Test | Size | 1 | 4 | 16 |
|------|-----:|--:|--:|---:|
| PUT | 4 KB | +0.9% | 0.0% | −5.4% |
| PUT | 4 MB | −6.6% | −13.3% | −15.8% |
| GET | 4 KB | −7.3% | −12.4% | −13.2% |
| GET | 4 MB | −1.8% | −5.8% | −9.6% |

PUT wasn't changed - the regression comes from the node now serving gRPC over `net/http` instead of the native gRPC server.

Likely cause of the degradation with more threads: stock `net/http2` serializes concurrent streams on a single connection (no write batching like gRPC's `loopyWriter`), so the more parallel requests there are, the more they queue up on one connection.

End-rey:
| Policy | Test | Size, KB | Threads | Base | HTTP/1.1 |      Δ |
|:------:|:----:|---------:|-------------:|----------:|--------------:|-------:|
|  REP3  | PUT  |        4 |            1 |       235 |           237 |  +0.9% |
|  REP3  | PUT  |        4 |            4 |       686 |           691 |  +0.7% |
|  REP3  | PUT  |        4 |           16 |      1002 |           954 |  −4.8% |
|  REP3  | PUT  |     4096 |            1 |        91 |            86 |  −5.5% |
|  REP3  | PUT  |     4096 |            4 |       180 |           165 |  −8.3% |
|  REP3  | PUT  |     4096 |           16 |       209 |           186 | −11.0% |
|  REP3  | GET  |        4 |            1 |      1429 |          1383 |  −3.2% |
|  REP3  | GET  |        4 |            4 |      2325 |          2110 |  −9.2% |
|  REP3  | GET  |        4 |           16 |      2874 |          2495 | −13.2% |
|  REP3  | GET  |     4096 |            1 |       328 |           358 |  +9.1% |
|  REP3  | GET  |     4096 |            4 |       549 |           617 | +12.4% |
|  REP3  | GET  |     4096 |           16 |       633 |           659 |  +4.1% |

HTTP/1.1 removes the main HTTP/2 bottleneck for large GET requests and outperforms the base in the 4 MB GET cases. Small-object GETs still regress under higher concurrency, so the remaining gap is likely in request/forwarding overhead rather than protocol negotiation itself.

roman-khimov:
Experiment done, won't be merged, see https://github.com/nspcc-dev/neofs-node/issues/3972

PR #3980, opened 2026-05-07
Optimize multishard operations

### LINKED ISSUE #3911: Optimize multishard meta DB operations
I'm always frustrated when I'm thinking of how SEARCH walks over local shards or how `exists()` checks are performed for PUT. There is a set of SN operations that require data from all local shards and currently in all of these cases we process shards sequentially.

There can be some concurrency in the process, maybe a routine per shard, maybe a routine per 2-4. Benchmark different strategies for different sets of shards (1-2-4-8-16-32), choose whatever works better. Dynamic (checked during startup) choice can also be relevant.

Things work as is, but likely suboptimally.

## Additional context
An attempt to make it was made once in #1146, but it failed. At the same time this was using old DB (before #3372), most likely things are very much different now.

End-rey:
Closes #3911.

Mostly regressions, except for `List`. That's something to think about.

### DIFF
--- a/pkg/local_object_storage/engine/bench_test.go
+++ b/pkg/local_object_storage/engine/bench_test.go
@@ -1,0 +1,394 @@
+package engine
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"sync"
+	"testing"
+	"time"
+
+	objectcore "github.com/nspcc-dev/neofs-node/pkg/core/object"
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/shard"
+	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
+	cidtest "github.com/nspcc-dev/neofs-sdk-go/container/id/test"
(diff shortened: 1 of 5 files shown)


roman-khimov:
1. All of the tests are performed with DB in memory. Some `echo 1 > /proc/sys/vm/drop_caches` before performing the operation can change the results significantly and we're also interested in this case (but requires running tests from root).
2. Your current strategy is to push jobs into a static number of routines, which has some good things (slow shard won't block others), but it also doesn't guarantee any real concurrency. Maybe not an issue though, if there is no CPU to run threads static split won't help, still I'd try spinning up as little routines as possible (like in chunks, say you're going sequential for <4 shards, then you spawn two routines till you have 8 shards, then three for 12, then four for 16, etc).
3. Still, for sub-10us operations (in optimal case) this seems to be totally redundant (existence checks and alike).

End-rey:
Removed changes related to `exists`. Reset the cache in the benchmarks. Implemented dynamic workers. Based on the results, as I understand it, the changes to `collectRawWithAttribute` and `Search` may be dropped.

Issue #1354, opened 2022-05-04
Unify network endpoints in single type 


cthulhu-rider:
NeoFS SDK works with network endpoints in one way, Storage nodes with another (`pkg/network` package). After #1353 NeoFS CLI defined internal `NetAddr` type as a mediator.

Proposal
Leave a single type for network address information (`network.Address`) which provides methods for all needs. Maybe we need to place it in SDK?

roman-khimov:
Drop multiaddr remnants here, refs. #3982.

PR #3982, opened 2026-05-08
network: write URIs into network map info

roman-khimov:
Which allows to avoid parsing when using them. Blocked by #3981.

### DIFF
--- a/pkg/network/group.go
+++ b/pkg/network/group.go
@@ -116,7 +116,7 @@ func WriteToNodeInfo(g AddressGroup, ni *netmap.NodeInfo) {
 
 	for addr := range slices.Values(g) {
 		ni.SetNetworkEndpoints()
-		addrs = append(addrs, addr.String())
+		addrs = append(addrs, addr.URIAddr())
 	}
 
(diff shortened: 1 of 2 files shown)


roman-khimov:
Refs. #1354.

pkg/network/group.go
```diff
@@ -116,7 +116,7 @@ func WriteToNodeInfo(g AddressGroup, ni *netmap.NodeInfo) {
 
 	for addr := range slices.Values(g) {
 		ni.SetNetworkEndpoints()
-		addrs = append(addrs, addr.String())
+		addrs = append(addrs, addr.URIAddr())
```
cthulhu-rider, line 119: pleasantly surprised that such a small change was needed

cthulhu-rider (approved):

pkg/network/group.go
roman-khimov: That's because most of the work was done previously in #3981 and tests were updated in https://github.com/nspcc-dev/neofs-testcases/pull/1447. But yeah, looks magic. And more code to drop eventually.

PR #3983, opened 2026-05-12
engine: add fan-out benchmark suite for shard traversal ops

End-rey:
Refs #3911.

### DIFF
--- a/pkg/local_object_storage/engine/bench_test.go
+++ b/pkg/local_object_storage/engine/bench_test.go
@@ -1,0 +1,401 @@
+package engine
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"sync"
+	"testing"
+	"time"
+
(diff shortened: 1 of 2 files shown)


pkg/local_object_storage/engine/bench_test.go
```diff
@@ -0,0 +1,394 @@
...
+
+	attrs := []string{attr}
+	resFS, cursor, err := objectcore.PreprocessSearchQuery(fs, attrs, "")
+	if err != nil {
+		tb.Fatal(err)
+	}
+	return resFS, attrs, cursor
+}
+
+func dropCaches(b *testing.B) {
```
roman-khimov: Should be optional (some const/var in code), I'd like to be able to test things without root privileges by default.

End-rey: Added const.

roman-khimov (approved):

Issue #3409, opened 2025-06-20
Avoid logger over-cloning


cthulhu-rider:
## Is your feature request related to a problem? Please describe.
#3408 showed that https://pkg.go.dev/go.uber.org/zap#Logger.With is quite demanding in terms of resources. Being used frequently, it adds pretty visible alloc overhead. I propose to avoid using it as much as we can

moreover, when messages have lower severity level (e.g.debug in production), this is not just overhead but pure waste of resources

## Describe the solution you'd like
never clone logger in request handlers and any other reoperations

even if cloning isn't done frequently in some cases, try to avoid it just to deduplicate common fields attached to different messages. Pass them explicitly or use var

## Additional context
benchmarks

PR #3991, opened 2026-05-14
Feat/meta/sn

carpawell:


pkg/core/metachain/meta/meta_test.go
```diff
@@ -437,8 +437,9 @@ func (s signer) SignTx(magic netmode.Magic, tx *transaction.Transaction) error {
 
 	// neotest does not support custom scripts and cannot calculate network
 	// fee correctly so change it there manually to smth that is currently
-	// enough for tests
-	tx.NetworkFee = 10_000_000
+	// enough for tests: verification test run showed that meta verification
+	// script takes ~0.03GAS, 0.5GAS should be enough
+	tx.NetworkFee = (1 * native.GASFactor) / 2
```
cthulhu-rider, line 442: multiply by 1 is usually excessive

cmd/neofs-node/meta_new.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/nspcc-dev/neo-go/pkg/core/storage/dbconfig"
+	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
+	"github.com/nspcc-dev/neo-go/pkg/encoding/fixedn"
+	"github.com/nspcc-dev/neofs-node/pkg/core/metachain/gas"
+	meta "github.com/nspcc-dev/neofs-node/pkg/services/meta_new"
+	"github.com/nspcc-dev/neofs-node/pkg/services/sidechain"
+	"go.uber.org/zap"
+)
+
+func initMeta_new(c *cfg) {
```
cthulhu-rider: why not `initMeta` which is dropped?

cmd/neofs-node/meta_new.go
```diff
@@ -0,0 +1,204 @@
...
+	l := c.log.With(zap.String("component", "metadata chain (SN)"))
```
cthulhu-rider: `SN` is suspicious to me here. I expect some other marker that this is an SN log

cmd/neofs-node/meta_new.go
```diff
@@ -0,0 +1,204 @@
...
+	v, err := c.cfgMorph.client.GetVersion()
+	fatalOnErr(err)
+
+	fsChainProtocol := v.Protocol
+	standByCommittee := make([]string, 0, len(v.Protocol.StandbyCommittee))
+	for _, c := range v.Protocol.StandbyCommittee {
+		standByCommittee = append(standByCommittee, c.StringCompressed())
+	}
+
+	if c.appCfg.Meta.SeedPort >= math.MaxUint16 {
```
cthulhu-rider: where does this restriction come from?

cmd/neofs-node/meta_new.go
```diff
@@ -0,0 +1,204 @@
...
+			Relay:             true,
+			Oracle:            config.OracleConfiguration{},
+			P2PNotary:         config.P2PNotary{},
+			StateRoot:         config.StateRoot{},
+			NeoFSBlockFetcher: config.NeoFSBlockFetcher{},
+			NeoFSStateFetcher: config.NeoFSStateFetcher{},
+		},
+	}
+
+	var cfgDB = dbconfig.DBConfiguration{
```
cthulhu-rider: var can be inlined

cmd/neofs-node/meta_new.go
```diff
@@ -0,0 +1,204 @@
...
+	c.workers = append(c.workers, newWorkerFromFunc(func(ctx context.Context) {
+		err = c.metaService.Run(ctx)
+		if err != nil {
+			c.internalErr <- fmt.Errorf("meta data service error: %w", err)
+		}
+	}))
+}
+
+func changePort(addrs []string, port uint16) ([]string, error) {
+	res := slices.Clone(addrs)
```
cthulhu-rider: cloning seems excessive here

cmd/neofs-node/meta_new.go
```diff
@@ -0,0 +1,204 @@
...
+	for _, c := range v.Protocol.StandbyCommittee {
+		standByCommittee = append(standByCommittee, c.StringCompressed())
+	}
+
+	if c.appCfg.Meta.SeedPort >= math.MaxUint16 {
+		fatalOnErr(fmt.Errorf("seed port exceeds max number %d: %d", math.MaxUint16, c.appCfg.Meta.SeedPort))
+	}
+
+	seedList, err := changePort(fsChainProtocol.SeedList, uint16(c.appCfg.Meta.SeedPort))
+	fatalOnErr(err)
```
cthulhu-rider: can be unclear address of what is being parsed. Lets add more context to `err`

cmd/neofs-node/meta_new.go
```diff
@@ -0,0 +1,204 @@
...
+	if cfg.ApplicationConfiguration.P2P.PingTimeout == 0 {
+		cfg.ApplicationConfiguration.P2P.PingTimeout = time.Minute
+	}
+	if cfg.ApplicationConfiguration.RPC.MaxWebSocketClients == 0 {
+		cfg.ApplicationConfiguration.RPC.MaxWebSocketClients = 64
+	}
+	if cfg.ApplicationConfiguration.RPC.SessionPoolSize == 0 {
+		cfg.ApplicationConfiguration.RPC.SessionPoolSize = 20
+	}
+	if cfg.ApplicationConfiguration.RPC.MaxGasInvoke == 0 {
```
cthulhu-rider: `ApplicationConfiguration` is accessed in all cases, i'd simplify code by accepting it

cmd/neofs-node/meta_new.go
```diff
@@ -0,0 +1,204 @@
...
+			TimePerBlock:                    50 * time.Millisecond,
+			MaxTimePerBlock:                 20 * time.Second,
+			ValidatorsCount:                 uint32(len(standByCommittee)),
+			ValidatorsHistory:               nil,
+			VerifyTransactions:              true,
+		},
+		ApplicationConfiguration: config.ApplicationConfiguration{
+			P2P: config.P2P{
+				Addresses:              c.appCfg.Meta.P2PAddresses,
+				MinPeers:               max(1, len(seedList)),
```
cthulhu-rider: are there any cases `fsChainProtocol.SeedList` when is empty?

cmd/neofs-node/meta_new.go
```diff
@@ -0,0 +1,204 @@
...
+			NeoFSStateSyncExtensions:        false,
+			ReservedAttributes:              false,
+			SeedList:                        seedList,
+			StandbyCommittee:                standByCommittee,
+			StateRootInHeader:               false,
+			StateSyncInterval:               0,
+			TimePerBlock:                    50 * time.Millisecond,
+			MaxTimePerBlock:                 20 * time.Second,
+			ValidatorsCount:                 uint32(len(standByCommittee)),
+			ValidatorsHistory:               nil,
```
cthulhu-rider: so many fields explicitly set to zero, why?

pkg/core/object/replicate.go
```diff
@@ -25,7 +35,58 @@ const (
 	typeKey         = "type"
 )
 
-// EncodeReplicationMetaInfo uses NEO's map (strict order) serialized format as a raw
+// Chain provide access to metadata chain state.
```
cthulhu-rider: either `metadata` is redundant (interface is provided by any chain) or type should be kinda `MetaChain`

pkg/core/object/replicate.go
```diff
@@ -25,7 +35,58 @@ const (
...
+type Chain interface {
+	MagicNumber() uint32
+	TransactionTestInvocation(tx *transaction.Transaction) error
+}
+
+var metaHash = state.CreateNativeContractHash("MetaData")
```
cthulhu-rider: object service is not a good place for this to me. Is there any more generic one for meta stuff?

pkg/core/object/replicate.go
```diff
@@ -25,7 +35,58 @@ const (
...
+// and returns any error related to test executing, otherwise error is always
+// nil.
+func EncodeChainMetaInfo(chain Chain, numberOfPlacementVectors int, cID cid.ID, oID, firstPart, previousPart oid.ID, pSize uint64, typ object.Type,
+	deleted, locked oid.ID, vub uint64, magicNumber uint32) (*transaction.Transaction, []byte, error) {
+	metadata := EncodeObjectMetadata(cID, oID, firstPart, previousPart, pSize, typ, deleted, locked, vub, magicNumber)
+
+	verifScript := MetaVerifScript(cID[:], numberOfPlacementVectors)
+	tx, err := objectTransaction(hash.Hash160(verifScript), metadata, uint32(vub))
+	if err != nil {
+		panic(fmt.Errorf("making transaction: %w", err))
```
cthulhu-rider: why panic when func returns an error?

pkg/core/object/replicate.go
```diff
@@ -25,7 +35,58 @@ const (
...
+	}
+	tx.Scripts = append(tx.Scripts[:0], transaction.Witness{
```
cthulhu-rider: isnt `tx.Scripts` empty before instruction? If so, append is excessive

pkg/core/object/replicate.go
```diff
@@ -25,7 +35,58 @@ const (
...
+	emit.Bytes(writer, cID)
+	emit.Int(writer, 2) // number or args
+	emit.Opcodes(writer, opcode.PACK)
+	emit.AppCallNoArgs(writer, metaHash, "verifyPlacementSignatures", callflag.ReadOnly)
+
+	return verifScriptBuf.Bytes()
+}
+
+// EncodeChainMetaInfo creates transaction for metadata and byte slice
+// that should be signed for this transaction. If chain non-nil, verifies it
```
cthulhu-rider: calling it w/ nil chain looks strange to me. I'd leave one function which makes a transaction. Then the only current caller passing chain can handle it on its own. And `Chain` interface wont be needed at all

pkg/core/object/replicate.go
```diff
@@ -77,18 +138,32 @@ func EncodeReplicationMetaInfo(cID cid.ID, oID, firstPart, previousPart oid.ID,
...
+	tx.ValidUntilBlock = vub
+	tx.Signers = append(tx.Signers, transaction.Signer{
+		Account: acc,
+		Scopes:  transaction.Global,
+	})
+	// cold contract takes ~0.02 GAS execution, ~0.03 GAS verification;
+	// 0.5 GAS should be enough and this allows up to 100 transactions in a sigle
+	// block; theses values must be the same for every node constructing transactions
+	// since fees takes part in transaction hashing
+	tx.SystemFee = 1 * native.GASFactor / 2
```
cthulhu-rider: again multiplying by 1

pkg/core/object/replicate.go
```diff
@@ -77,18 +138,32 @@ func EncodeReplicationMetaInfo(cID cid.ID, oID, firstPart, previousPart oid.ID,
...
+	script, err := smartcontract.CreateCallScript(metaHash, "submitObjectPut", metaData)
+	if err != nil {
+		return nil, fmt.Errorf("making transaction script: %w", err)
+	}
+
+	tx := transaction.New(script, 0)
+	tx.Nonce = vub
+	tx.ValidUntilBlock = vub
+	tx.Signers = append(tx.Signers, transaction.Signer{
+		Account: acc,
```
cthulhu-rider, line 131: `acc` protects from tx collision when several SN do this for the same object, right?

pkg/services/meta_new/blocks.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/nspcc-dev/neo-go/pkg/core/block"
+	"github.com/nspcc-dev/neo-go/pkg/core/state"
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	"go.uber.org/zap"
+)
+
+func (m *Meta) blockHandler(ctx context.Context, buff <-chan *block.Header) {
+	for {
+		if len(buff) >= blockBuffSize-1 {
```
cthulhu-rider: doubt it can be `>`

btw is it possible to do by channel writer(s). Would be more natural to me

pkg/services/meta_new/blocks.go
```diff
@@ -0,0 +1,64 @@
...
+			m.l.Warn("block header buffer has been completely filled")
+		}
+
+		select {
+		case <-ctx.Done():
+			return
+		case b := <-buff:
+			h := b.Hash()
+			ind := b.Index
+			m.l.Debug("received block", zap.Stringer("block hash", h), zap.Uint32("index", ind))
```
cthulhu-rider: so we have this worker just for debug log?

pkg/services/meta_new/blocks.go
```diff
@@ -0,0 +1,64 @@
...
+
+			switch n.Name {
+			case objPutEvName:
+				ev, err := parseObjNotification(*n)
+				if err != nil {
+					l.Error("invalid object notification received", zap.Error(err))
+					continue
+				}
+
+				if magic := uint32(ev.network.Uint64()); magic != m.magicNumber {
```
cthulhu-rider: `Uint64()` can return undefined value. Lets assert value is a valid `uint32` within `parseObjNotification` (error otherwise), and just cmp here

pkg/services/meta_new/blocks.go
```diff
@@ -0,0 +1,64 @@
...
+	for {
+		if len(buff) >= notificationBuffSize-1 {
+			m.l.Warn("notification buffer has been completely filled")
+		}
+
+		select {
+		case <-ctx.Done():
+			return
+		case n := <-buff:
+			l := m.l.With(zap.Stringer("tx", n.Container))
```
cthulhu-rider: beware https://github.com/nspcc-dev/neofs-node/issues/3409

pkg/services/meta_new/meta.go
```diff
@@ -0,0 +1,325 @@
...
+	// any blocking or making additional RPC.
+	notificationBuffSize = 10000
+)
+
+type (
+	// NeoFSNetwork describes current NeoFS storage network state.
+	NeoFSNetwork interface {
+		// Head returns actual object header from the NeoFS network (non-local
+		// objects should also be returned). Missing, removed object statuses
+		// must be reported according to API statuses from SDK.
```
cthulhu-rider: looks like caller does not care about API statuses

cthulhu-rider (commented):
not finished

pkg/services/meta_new/meta.go
```diff
@@ -297,7 +300,13 @@ func (m *Meta) SubmitObjectPut(tx *transaction.Transaction, signatures [][]neofs
 
 	tx.Scripts[0].InvocationScript = invokBuff.Bytes()
 
-	return m.chain.AddTx(tx)
+	m.l.Debug("sending transaction to chain...", zap.String("txHash", tx.Hash().StringLE()))
+	now := time.Now()
+	err := m.chain.AddTx(tx)
+	took := time.Since(now)
```
cthulhu-rider: timing is useless if log is not debug. Costs mb small, but nevertheless

why do we actually care about it btw? Any chance to have a context limit?

pkg/services/meta_new/meta.go
```diff
@@ -297,7 +300,13 @@ func (m *Meta) SubmitObjectPut(tx *transaction.Transaction, signatures [][]neofs
...
+	m.l.Debug("sent transaction to chain", zap.String("txHash", tx.Hash().StringLE()), zap.Duration("took", took), zap.Error(err))
```
cthulhu-rider: if `err != nil` then not `sent`, isn't it?

pkg/services/meta_new/blocks.go
```diff
@@ -19,9 +21,16 @@ func (m *Meta) blockHandler(ctx context.Context, buff <-chan *block.Header) {
...
 			return
 		case b := <-buff:
+			blockReceivedAfter := time.Since(prevBlockFetchTime)
+			prevBlockFetchTime = time.Now()
+
 			h := b.Hash()
 			ind := b.Index
 			m.l.Debug("received block", zap.Stringer("block hash", h), zap.Uint32("index", ind))
+
+			m.chainHeigh.Store(ind)
```
cthulhu-rider: Heigh?

pkg/services/meta_new/blocks.go
```diff
@@ -10,6 +11,7 @@ import (
 )
 
 func (m *Meta) blockHandler(ctx context.Context, buff <-chan *block.Header) {
+	prevBlockFetchTime := time.Now()
```
cthulhu-rider: for the first block received it is not representative. I'd not upd metrics for it

pkg/services/meta_new/blocks.go
carpawell: this is a counter (will be reused later) and metrics writer

pkg/core/object/replicate.go
carpawell: this is a hardcoded transaction creating; there should not be any errors ever

cmd/neofs-node/meta_new.go
carpawell: i had problems with running new chain so kept them set explicitly to see what is set and what is a default value. can drop

cmd/neofs-node/meta_new.go
carpawell: i think no machine should have a bigger tcp port, this should be treated as a strange config imo

cmd/neofs-node/meta_new.go
carpawell: do you have a sugggestion? it helped in debug when there are IR and SN on the same host

cmd/neofs-node/meta_new.go
carpawell: my plan was to implement new functionality and then change it in the last commit. though i undrestand that linter is not happy for the first commit

pkg/core/metachain/meta/meta_test.go
carpawell: i just always treat such statements as "(N * unit of measurement) operation" like "one GAS divide by two"

cmd/neofs-node/meta_new.go
cthulhu-rider: i'd just trust net layer to handle this

pkg/core/object/replicate.go
cthulhu-rider: then i dont expect error return from `objectTransaction` func

pkg/services/meta/meta.go
```diff
@@ -93,295 +117,222 @@ func (on *objectNotifier) unsubscribe(addr oid.Address) {
...
+
+	on.m.Unlock()
+
+	if ok {
+		on.metaSvc.taskQueue <- storageTask{addr: addr, o: &sub.objHeader}
+	} else {
+		on.metaSvc.taskQueue <- storageTask{addr: addr}
+	}
+
+	on.metaSvc.l.Debug("object notification handled",
```
cthulhu-rider: seems more `received` than `handled` to me

pkg/services/meta/meta.go
```diff
@@ -93,295 +117,222 @@ func (on *objectNotifier) unsubscribe(addr oid.Address) {
...
-	m.magicNumber = uint32(v.Protocol.Network)
+// TransactionTestInvocation validates transaction.
+func (m *Meta) TransactionTestInvocation(tx *transaction.Transaction) error {
+	return m.chain.TransactionTestInvocation(tx)
+}
 
-	m.stM.RLock()
-	hasContainers := len(m.storages) > 0
-	m.stM.RUnlock()
+// IndexedSignature is undexed signature, pointing at place of signature's
```
cthulhu-rider:
```suggestion
// IndexedSignature is indexed signature, pointing at place of signature's
```

pkg/services/meta/storage.go
```diff
@@ -0,0 +1,53 @@
...
+		case <-ctx.Done():
+			return
+		case n := <-buff:
+			l := m.l.With(zap.Stringer("addr", n.addr))
+
+			cID := n.addr.Container()
+			ok, err := m.net.IsMineWithMeta(cID, nil)
+			if err != nil {
+				l.Error("failed to check container relation to node", zap.Stringer("cid", cID), zap.Error(err))
+				return
```
cthulhu-rider: `continue`? Same below

pkg/services/meta/storage.go
```diff
@@ -0,0 +1,53 @@
...
+			}
+			if !ok {
+				return
+			}
+
+			if n.o == nil {
+				o, err := m.net.Head(ctx, n.addr.Container(), n.addr.Object())
+				if err != nil {
+					l.Error("cannot fetch object header", zap.Error(err))
```
cthulhu-rider: `addr` field is useful

pkg/services/meta/storage.go
```diff
@@ -0,0 +1,53 @@
...
+	for {
+		if len(buff) >= notificationBuffSize-1 {
+			m.l.Warn("storage task queue buffer has been completely filled")
+		}
+
+		select {
+		case <-ctx.Done():
+			return
+		case n := <-buff:
+			l := m.l.With(zap.Stringer("addr", n.addr))
```
cthulhu-rider: again try to avoid logger cloning in frequent code. Pls check all places in this PR

pkg/services/meta/storage.go
```diff
@@ -0,0 +1,53 @@
...
+				if err != nil {
+					l.Error("cannot fetch object header", zap.Error(err))
+					continue
+				}
+
+				n.o = &o
+			}
+
+			err = m.PutObject(n.o)
+			if err != nil {
```
cthulhu-rider, line 65: guess nothing can be done, but what effect comes with this?

pkg/services/object/put/distributed.go
```diff
@@ -609,19 +632,21 @@ func (t *distributedTarget) encodeObjectMetadata(obj object.Object) []byte {
 		firstObj = obj.GetID()
 	}
 
-	var deletedObjs []oid.ID
-	var lockedObjs []oid.ID
+	var (
```
cthulhu-rider: such bracing changes are not rly related

pkg/services/object/put/distributed.go
```diff
@@ -570,16 +585,24 @@ func (t *distributedTarget) submitMetaCollection(addr oid.Address, metaC *metaCo
...
+		objAcceptedCh chan struct{}
+		addr          = o.Address()
+	)
 	if await {
-		objAccepted = make(chan struct{}, 1)
-		t.metaSvc.NotifyObjectSuccess(objAccepted, addr)
+		h := t.metaCollection.metaTransaction.Hash()
+
+		var objCopy object.Object
+		o.CutPayload().CopyTo(&objCopy)
```
cthulhu-rider, line 596: why do we need deep copy?

cthulhu-rider (commented):
finished. Lets fix everything and merge

@AnnaShaleva may u pls check chain configuration?

cmd/neofs-node/meta.go
```diff
@@ -190,3 +164,185 @@ func (c *neofsNetwork) List(e uint64) (map[cid.ID]struct{}, error) {
...
+	if cfg.ApplicationConfiguration.P2P.ProtoTickInterval == 0 {
+		cfg.ApplicationConfiguration.P2P.ProtoTickInterval = 2 * time.Second
+	}
+	if cfg.ProtocolConfiguration.MaxTraceableBlocks == 0 {
+		cfg.ProtocolConfiguration.MaxTraceableBlocks = 17280
+	}
+	if cfg.ProtocolConfiguration.MaxValidUntilBlockIncrement == 0 {
+		cfg.ProtocolConfiguration.MaxValidUntilBlockIncrement = 8640
+	}
+	if cfg.ApplicationConfiguration.P2P.PingInterval == 0 {
```
AnnaShaleva: Enable `DirectRelay` (available starting from v0.119.0).

cmd/neofs-node/meta.go
```diff
@@ -190,3 +164,185 @@ func (c *neofsNetwork) List(e uint64) (map[cid.ID]struct{}, error) {
...
+
+	fsChainProtocol := v.Protocol
+	standByCommittee := make([]string, 0, len(v.Protocol.StandbyCommittee))
+	for _, c := range v.Protocol.StandbyCommittee {
+		standByCommittee = append(standByCommittee, c.StringCompressed())
+	}
+
+	if c.appCfg.Meta.SeedPort >= math.MaxUint16 {
+		fatalOnErr(fmt.Errorf("seed port exceeds max number %d: %d", math.MaxUint16, c.appCfg.Meta.SeedPort))
+	}
```
AnnaShaleva, line 181: It can't be. `SeedPort` should be of type uint16.

cmd/neofs-node/meta.go
```diff
@@ -190,3 +164,185 @@ func (c *neofsNetwork) List(e uint64) (map[cid.ID]struct{}, error) {
...
+				MaxValidUntilBlockIncrement: fsChainProtocol.MaxValidUntilBlockIncrement,
+				Roles: map[noderoles.Role]keys.PublicKeys{
+					noderoles.P2PNotary:     v.Protocol.StandbyCommittee,
+					noderoles.NeoFSAlphabet: v.Protocol.StandbyCommittee,
+				},
+				TimePerBlock: 50 * time.Millisecond,
+			},
+			Magic:                           fsChainProtocol.Network + 1,
+			MemPoolSize:                     0,
+			Hardforks:                       nil,
```
AnnaShaleva: The list is better to be specified explicitly. Otherwise genesis state will be silently mutated after the subsequent hardfork release and NeoGo version update.

cmd/neofs-node/meta.go
```diff
@@ -190,3 +164,185 @@ func (c *neofsNetwork) List(e uint64) (map[cid.ID]struct{}, error) {
...
+			InitialGASSupply:                0,
+			P2PNotaryRequestPayloadPoolSize: 1000,
+			MaxBlockSize:                    0,
+			MaxBlockSystemFee:               0,
+			MaxTraceableBlocks:              fsChainProtocol.MaxTraceableBlocks,
+			MaxTransactionsPerBlock:         0,
```
AnnaShaleva: Derive from `fsChainProtocol` or set explicitly? It's important characteristic, and given the fact that you use short blocks, it's likely shouldn't be set to a default.

Either change it or create an issue for further meta chain parameters tuning, may require some load-testing.

cmd/neofs-node/meta.go
```diff
@@ -190,3 +164,185 @@ func (c *neofsNetwork) List(e uint64) (map[cid.ID]struct{}, error) {
...
+	for i := range res {
+		host, _, err := net.SplitHostPort(res[i])
+		if err != nil {
+			return nil, fmt.Errorf("[%d] address ('%s') cannot be parsed: %w", i, res[i], err)
+		}
+		res[i] = net.JoinHostPort(host, strconv.FormatUint(uint64(port), 10))
+	}
+
+	return res, nil
+}
```
AnnaShaleva, line 291: No code duplication, please. Create a separate package, reuse from both inner-ring and SN.

cmd/neofs-node/meta.go
```diff
@@ -190,3 +164,185 @@ func (c *neofsNetwork) List(e uint64) (map[cid.ID]struct{}, error) {
...
+		cfg.ApplicationConfiguration.P2P.MaxPeers = 100
+	}
+	if cfg.ApplicationConfiguration.P2P.AttemptConnPeers == 0 {
+		cfg.ApplicationConfiguration.P2P.AttemptConnPeers = cfg.ApplicationConfiguration.P2P.MinPeers + 10
+	}
+	if cfg.ApplicationConfiguration.P2P.DialTimeout == 0 {
+		cfg.ApplicationConfiguration.P2P.DialTimeout = time.Minute
+	}
+	if cfg.ApplicationConfiguration.P2P.ProtoTickInterval == 0 {
+		cfg.ApplicationConfiguration.P2P.ProtoTickInterval = 2 * time.Second
```
AnnaShaleva: Not sure about this one given 50ms blocks. We need to check if the nodes have enough peer connections during stress-tests.

cmd/neofs-node/meta.go
```diff
@@ -190,3 +164,185 @@ func (c *neofsNetwork) List(e uint64) (map[cid.ID]struct{}, error) {
...
+	}
+	if cfg.ProtocolConfiguration.MaxTraceableBlocks == 0 {
+		cfg.ProtocolConfiguration.MaxTraceableBlocks = 17280
+	}
+	if cfg.ProtocolConfiguration.MaxValidUntilBlockIncrement == 0 {
+		cfg.ProtocolConfiguration.MaxValidUntilBlockIncrement = 8640
```
AnnaShaleva: ~7 minutes of 50ms-blocks. Is it OK if some NeoFS request will be outdated and cancelled after 7 minutes of awaiting in the pool?

cmd/neofs-node/meta.go
```diff
@@ -190,3 +164,185 @@ func (c *neofsNetwork) List(e uint64) (map[cid.ID]struct{}, error) {
...
+		cfg.ApplicationConfiguration.P2P.PingInterval = 30 * time.Second
+	}
+	if cfg.ApplicationConfiguration.P2P.PingTimeout == 0 {
+		cfg.ApplicationConfiguration.P2P.PingTimeout = time.Minute
+	}
+	if cfg.ApplicationConfiguration.RPC.MaxWebSocketClients == 0 {
+		cfg.ApplicationConfiguration.RPC.MaxWebSocketClients = 64
+	}
+	if cfg.ApplicationConfiguration.RPC.SessionPoolSize == 0 {
+		cfg.ApplicationConfiguration.RPC.SessionPoolSize = 20
```
AnnaShaleva: In another words, not more than 20 parallel requests that use iterator sessions in `invoke*` and don't close it after the first response. Does NeoFS use opened iterator sessions extensively?

cmd/neofs-node/meta.go
AnnaShaleva: Always use it in all nodes of meta chains.

pkg/services/meta/meta.go
```diff
@@ -4,85 +4,109 @@ import (
...
-
-	// rootKey is the key for the last known state root in KV data base
-	// associated with MPT.
-	rootKey = 0xff
-
 	// notificationBuffSize is a nesessary buffer for neo-go's client proper
 	// notification work; it is required to always read notifications without
 	// any blocking or making additional RPC.
-	notificationBuffSize = 100
+	notificationBuffSize = 10000
```
AnnaShaleva, line 35: Is it approved by @roman-khimov?

pkg/services/object/put/streamer.go
```diff
@@ -249,11 +255,41 @@ func (p *Streamer) newCommonTarget(prm *PutInitPrm) internal.Target {
...
+				k1Raw := a.PublicKey()
+				k2Raw := b.PublicKey()
+
+				// Sort by ECPoint's (X, Y) components: compare X first, and then compare Y.
+				cmpX := bytes.Compare(k1Raw[1:], k2Raw[1:])
+				if cmpX != 0 {
+					return cmpX
+				}
+				// The case when X components are the same is extremely rare, thus we perform
+				// key deserialization only if needed. No error can occur.
```
AnnaShaleva, line 278: > No error can occur.

Are you sure? For native Neo it's true because we validate keys prior to putting them to the storage.

pkg/services/sidechain/sidechain.go
```diff
@@ -0,0 +1,186 @@
...
+	t := time.NewTicker(s.core.GetConfig().Genesis.TimePerBlock)
+
+	for {
+		s.logger.Info("waiting for synchronization with the blockchain network...")
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("await state sync: %w", context.Cause(ctx))
+		case <-t.C:
+			if s.netServer.IsInSync() {
+				s.logger.Info("blockchain state successfully synchronized")
```
AnnaShaleva, line 122: Are you sure? This check is dependent on peers, if there's less than MinPeers, the node consider itself as non-synced.

pkg/services/sidechain/sidechain.go
```diff
@@ -0,0 +1,186 @@
...
+// Stop stops the side chain. Must be called only after a successful [SideChain.Run].
+func (s *SideChain) Stop() {
+	s.netServer.Shutdown()
+	s.core.Close()
+	close(s.chErr)
+}
+
+// Height returns current side chaih block height.
+func (s *SideChain) Height() uint32 {
+	return s.core.HeaderHeight()
```
AnnaShaleva, line 138: Header height differs from block height. What exactly do you need here?

pkg/services/sidechain/sidechain.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// AddTx adds transaction to node's transaction pools. Non-nil return
+// value _does not mean_ transaction is included, use
+// [SideChain.SubscribeForNotifications] or [SideChain.SubscribeForBlocks]
+// to be notified about accepted chain changes.
+func (s *SideChain) AddTx(tx *transaction.Transaction) error {
+	return s.netServer.RelayTxn(tx)
```
AnnaShaleva: You likely need RelayTxnDirectly, check out the doc. If this SN is the first node in the network who receives this transaction, then you need RelayTxnDirectly.

cmd/neofs-node/meta.go
AnnaShaleva: Don't enable Gorgon for now. The rest of them are OK.

pkg/services/sidechain/sidechain.go
AnnaShaleva: Available from v0.119.0, it's a significant optimisation made preliminary for meta chain.

pkg/services/object/put/distributed.go
carpawell: likely no real need for that now, _but_ this object will outlive the request handling in most cases (it will be used in addition to index building in a separate routine). if you want, i will drop the copy

pkg/services/meta/storage.go
carpawell: an object was indexed, it is stored in the chain, everybody can know it _must_ be presented, but this exact node will not answer any `Search` call with this object

pkg/services/meta/storage.go
carpawell: i still doubt it is critical. fixed, but i would love to see the results (do understand that is costs smth but how bad is it?)

pkg/services/meta/storage.go
carpawell: it was there but with `zap.With`

pkg/services/meta/storage.go
carpawell: wow, fixed

pkg/services/meta_new/blocks.go
carpawell: always do not like dead code for the first iteration, it will be one of the millions of values in the metrics, that is technically still correct (this is the time required for us to fetch a block). but done

pkg/services/meta_new/blocks.go
carpawell: fixed, strange no linter problems were found

pkg/services/meta_new/meta.go
carpawell: i treat it like "sent transaction and there is the result: transaction was sent with an error"

pkg/services/meta_new/meta.go
carpawell: the whole performance is still questionable; it needs investigation, metrics, logs, etc, can be dropped later. it can also be a metric. do you this all of the above make sense when we are talking about meta performance? we literally do nothing for 25ms at least for every object

pkg/services/meta_new/meta.go
carpawell: i copied desc from the old one: https://github.com/nspcc-dev/neofs-node/blob/3b4a1dd1897c959f233a409df5f1716e86223571/pkg/services/meta/containers.go#L154

it is not clear what to do when the primary index knows an object and the secondary one cannot fetch it

pkg/services/meta_new/blocks.go
carpawell: well, let it be without `With`, but it is sad

pkg/services/meta_new/blocks.go
carpawell: done

pkg/services/meta_new/blocks.go
carpawell: > doubt it can be >

why? we read an item, do some magic that _may take time_, and then check again what size we have now, a new item may have come


> btw is it possible to do by channel writer(s). Would be more natural to me

writer does not belong to this repo, it is sent to neo-go code

pkg/core/object/replicate.go
carpawell: hm, not sure i understand you. acc does not belong to any public key, it is a hash of verification function that puts a certain cID and number of placement vectors on the stack. it is always the same for every object in any container and has nothing in common with storage node keys

pkg/core/object/replicate.go
carpawell: okok, let it be without 1

btw:
```
▶ grep -r --include \*.go --include \*test.go "1 \*" | wc -l
25
```

pkg/core/object/replicate.go
carpawell: it is empty. i feel simple append is more error-prone

pkg/core/object/replicate.go
carpawell: we use neo-go's script builder, it _may_ fall if there are some unsupported args (they are `any` at compilation time). it _may_ fail if args are extended in the future, and no build error will be faced. i see it as a dev error at runtime that cannot be faced by static checkers. if you meant exact place for this panic, i see no big differences

pkg/core/object/replicate.go
carpawell: reused existing from the native meta contract package

cmd/neofs-node/meta_new.go
carpawell: also it helped to review to see what exactly is set and how

cmd/neofs-node/meta_new.go
carpawell: likely not, lets just use `len(seedList)`

cmd/neofs-node/meta_new.go
carpawell: dropped, but why? one-time execution line at the start, new slice just feels more comfortable than changing the existing to me, like built-in `append`

cmd/neofs-node/meta_new.go
carpawell: made this config field uint16 from the beginning

cmd/neofs-node/meta_new.go
carpawell: old one is dropped only in the last commit. it is hard to maintain this commit structure with the number of fixes and rebases, but i still want to keep it this way. i remember you wanted to keep both metas for now in this PR, is it still true?

pkg/services/sidechain/sidechain.go
carpawell: the PR is too old. yes, you are right, `RelayTxnDirectly` is what i need there, fixed. BTW, do you have a list of that what done to improve performance _and_ what cannot be done via updating to 0.119.0 implicitly? i can recheck that I have applied everything

pkg/services/sidechain/sidechain.go
carpawell: > Header height differs from block height

in which way? you mean a 1-block difference in values? i need it for async, but the same VUB calculation

pkg/services/sidechain/sidechain.go
carpawell: i have copied this from the local consensus approach. do have a better suggestion on how to check it?

pkg/services/object/put/streamer.go
carpawell: this is [network map](https://github.com/nspcc-dev/neofs-api/blob/6e280c75a1e56000a2a420c799651a629c33ca94/netmap/types.proto#L221), if it is broken, many things should break. IR nodes should validate it before it is put to the contract

pkg/services/meta/meta.go
carpawell: this is an address + pointer buff channel, what cons may we have? ~703KB of memory

cmd/neofs-node/meta.go
carpawell: metadata chain does not have iterators at all for now, but i may change this value to any more sense value

cmd/neofs-node/meta.go
carpawell: i think it is normal, likely no object request will outlive 7m because of client/server timeouts

cmd/neofs-node/meta.go
carpawell: lets find optimal value in https://github.com/nspcc-dev/neofs-node/issues/4009

cmd/neofs-node/meta.go
carpawell: i dont even know where to place it, it is just a coincidence that two apps use it for their config, it was not like that in the first implementation

cmd/neofs-node/meta.go
carpawell: https://github.com/nspcc-dev/neofs-node/issues/4009

cmd/neofs-node/meta.go
carpawell: can you explain, please? i should mention every hardfork there with zero block? i doubt this is done for our regular FSchain

AnnaShaleva: > i should mention every hardfork there with zero block?

Yes.

> i doubt this is done for our regular FSchain

It's done:
https://github.com/nspcc-dev/neo-go/blob/76fc43a8bb78301d7a81076402abd2377d5a3b60/config/protocol.mainnet.neofs.yml#L31
https://github.com/nspcc-dev/neo-go/blob/76fc43a8bb78301d7a81076402abd2377d5a3b60/config/protocol.testnet.neofs.yml#L32

pkg/services/sidechain/sidechain.go
AnnaShaleva: No.

pkg/services/sidechain/sidechain.go
AnnaShaleva: > you mean a 1-block difference in values?

Not only. It can be large difference if, e.g. the node is in the process of syncing or stale. Headers synchronisation is a separate process, different from blocks synchronisation.

pkg/services/sidechain/sidechain.go
AnnaShaleva: Read the NeoGo changelogs. For 0.119.0 there's no such things except these two.

cmd/neofs-node/meta.go
carpawell: > Yes.

done then.

> It's done:

i meant that as i see we do not fill this field by default for our internal consensus. maybe neofs team does not fully understand "genesis state will be silently mutated after the subsequent hardfork release ". anyway done for this PR

pkg/services/meta/meta.go
carpawell: i can change it to any value. do you have a suggestion? i just had an experience with deadlock with neo-go subscription, so this is why this big value was chosen

pkg/services/sidechain/sidechain.go
carpawell: the only thing i can suggest is to block any operation before the node is in sync

cmd/neofs-node/meta.go
```diff
@@ -190,3 +162,183 @@ func (c *neofsNetwork) List(e uint64) (map[cid.ID]struct{}, error) {
...
+			},
+			Magic:       fsChainProtocol.Network + 1,
+			MemPoolSize: 0,
+			Hardforks: map[string]uint32{
+				config.HFAspidochelone.String(): 0,
+				config.HFBasilisk.String():      0,
+				config.HFCockatrice.String():    0,
+				config.HFDomovoi.String():       0,
+				config.HFEchidna.String():       0,
+				config.HFFaun.String():          0,
```
AnnaShaleva: Add a compatibility unit-test: check that hardforks starting from Aspid ending with `config.HFLatestStable` are enabled in the resulting `chainCfg`. On subsequent NeoGo update if latest stable hardfork is changed, you'll know about it.

pkg/innerring/innerring.go
```diff
@@ -815,8 +814,17 @@ func New(ctx context.Context, log *zap.Logger, cfg *config.Config, errChan chan<
...
-			Hardforks:                   config.Hardforks{},
-			ValidatorsHistory:           config.ValidatorsHistory{},
+			Hardforks: config.Hardforks{
+				Name: map[string]uint32{
+					configneogo.HFAspidochelone.String(): 0,
+					configneogo.HFBasilisk.String():      0,
+					configneogo.HFCockatrice.String():    0,
+					configneogo.HFDomovoi.String():       0,
+					configneogo.HFEchidna.String():       0,
+					configneogo.HFFaun.String():          0,
```
AnnaShaleva: Ditto for this config.

pkg/util/network.go
```diff
@@ -0,0 +1,21 @@
+package util
```
AnnaShaleva: `util` is always a bad name, think about a better one. CC @roman-khimov

AnnaShaleva (commented):
Tests are failing. Other than that looks legit as a prototype.

pkg/services/sidechain/sidechain.go
AnnaShaleva: The decision is up to you, I'm just warning. Either way, you should either adjust the documentation and method name or adjust the return value.

pkg/services/meta/meta.go
AnnaShaleva: That's why I'm asking. Increasing buffer size is not a solution to the deadlock problem. It will make the problem more hidden, that's it.

If there's a deadlock, it should be fixed in some other (explicit) way.

carpawell: > Increasing buffer size is not a solution to the deadlock problem

we are balancing b/w notification handling speed and storage speed. i guess it is another thing for exploring in load testing (yet not answered)

cmd/neofs-node/meta.go
carpawell: done

pkg/util/network.go
carpawell: this is not the package i created, i just extended it. but agree, moved to another `network` package

carpawell:
> Tests are failing. Other than that looks legit as a prototype.

They should be adopted, yes. And there are other problems. Config check and changelog are not accepted to the experimental things.

cmd/neofs-node/meta_new.go
cthulhu-rider: i meant `make()` is enough since we fill the slice from scatch in loop

cmd/neofs-node/meta_new.go
cthulhu-rider: ok for review, but i'd drop before merge

pkg/core/object/replicate.go
cthulhu-rider: i'd expect simple slice init, not append

pkg/core/object/replicate.go
cthulhu-rider: it is not incorrect, just strange

pkg/core/object/replicate.go
cthulhu-rider: so, if multiple SN push such transaction, it'll be the same? If so, won't collision lead to any problems?

pkg/services/meta_new/blocks.go
cthulhu-rider: > and then check again

`buff` can initially have `blockBuffSize-1` which is a capacity. This does not mean queue is full. Wouldnt be a big problem, but still

pkg/services/meta_new/meta.go
cthulhu-rider: > can be dropped later

we need a reminder then

pkg/services/meta_new/meta.go
cthulhu-rider: more like "attempt to send is finished". "Sent" means "sending succeeded" to me. Not a big deal for debug

pkg/services/meta/storage.go
cthulhu-rider: very visible allocs while benefit in 3-5 places in code

pkg/services/object/put/distributed.go
cthulhu-rider: safety is must have, but if there is no mutations deepcopying is just an overhead

pkg/core/object/replicate.go
cthulhu-rider: > if you meant exact place for this panic, i see no big differences

we'd feel the diff if `objectTransaction` code is extended w/ legal error return

pkg/services/meta_new/meta.go
carpawell: added TODO

pkg/core/object/replicate.go
carpawell: hash gonna be the same, and one of the nodes will get an error from the chain. but why is it possible? why two nodes takes the same object (PUT, not REPLICATE)?

pkg/core/object/replicate.go
carpawell: done

pkg/core/object/replicate.go
carpawell: made panic in `objectTransaction`

cmd/neofs-node/meta_new.go
carpawell: fixed

pkg/core/object/replicate.go
cthulhu-rider: parallel PUT into container with REP N > 1

cthulhu-rider (approved):
seems good to me overall

cmd/neofs-node/meta_new.go
```diff
@@ -0,0 +1,189 @@
...
+				MaxTraceableBlocks:          fsChainProtocol.MaxTraceableBlocks,
+				MaxValidUntilBlockIncrement: fsChainProtocol.MaxValidUntilBlockIncrement,
+				Roles: map[noderoles.Role]keys.PublicKeys{
+					noderoles.P2PNotary:     v.Protocol.StandbyCommittee,
+					noderoles.NeoFSAlphabet: v.Protocol.StandbyCommittee,
+				},
+				TimePerBlock: 50 * time.Millisecond,
+			},
+			Magic:                           fsChainProtocol.Network + 1,
+			MemPoolSize:                     0,
```
roman-khimov: Wow.

cmd/neofs-node/meta_new.go
roman-khimov: Better drop them, the list of options can be extended anyway and you won't notice.

pkg/services/meta_new/blocks.go
carpawell: but there is `len`, not `cap`. not sure i understand you

pkg/core/object/replicate.go
carpawell: different objects will have different oids (part of transaction's script), there should not be any conflicts about _different_ PUTs

pkg/services/meta_new/blocks.go
roman-khimov: I don't think this check is useful, you can't separate overflow from normally filled queue on the reader side, only writer can do this on attempt to write.

pkg/services/meta_new/blocks.go
```diff
@@ -0,0 +1,64 @@
...
+			h := b.Hash()
+			ind := b.Index
+			m.l.Debug("received block", zap.Stringer("block hash", h), zap.Uint32("index", ind))
+		}
+	}
+}
+
+func (m *Meta) notificationHandler(ctx context.Context, buff <-chan *state.ContainedNotificationEvent) {
+	for {
+		if len(buff) >= notificationBuffSize-1 {
```
roman-khimov: Same here.

pkg/services/meta_new/meta.go
```diff
@@ -0,0 +1,325 @@
...
+		delete(on.subs, addr)
+
+		timeTook = time.Since(sub.timeSubscriptionStarted)
+		blocksTook = currHeight - sub.blockSubscriptionStarted
+	}
+
+	on.m.Unlock()
+
+	if ok {
+		on.metaSvc.taskQueue <- storageTask{addr: addr, o: &sub.objHeader}
```
roman-khimov: Does it make any difference? Local HEAD should be pretty fast.

pkg/services/meta_new/meta.go
```diff
@@ -0,0 +1,325 @@
...
+}
+
+func (on *objectNotifier) notifyReceived(addr oid.Address) {
+	var (
+		timeTook   time.Duration
+		blocksTook uint32
+		currHeight = on.metaSvc.chainHeigh.Load()
+	)
+
+	on.m.Lock()
```
roman-khimov: Locks are never good in this game.

pkg/services/meta_new/meta.go
```diff
@@ -0,0 +1,325 @@
...
+	m.magicNumber = m.chain.Magic()
+
+	var wg sync.WaitGroup
+
+	wg.Go(func() { m.blockHandler(ctx, m.bCh) })
+	wg.Go(func() { m.notificationHandler(ctx, m.evsCh) })
+	wg.Go(func() { m.storager(ctx, m.taskQueue) })
+
+	m.chain.SubscribeForBlocks(m.bCh)
+	m.chain.SubscribeForNotifications(m.evsCh)
```
roman-khimov: Seems like this can be done via JSON-RPC as well. Can we have this mode (non-local chain) for nodes with IR nearby?

pkg/services/meta/storage.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"context"
+
+	"github.com/nspcc-dev/neofs-sdk-go/object"
+	"go.uber.org/zap"
+)
+
+// PutObject forces [Meta] to index provided object.
+func (m *Meta) PutObject(o *object.Object) error {
+	return m.metabase.Put(o)
```
roman-khimov: Per-object transaction?

pkg/services/object/server.go
```diff
@@ -212,20 +211,18 @@ type Server struct {
 }
 
 // New provides protoobject.ObjectServiceServer for the given parameters.
-func New(hs Handlers, magicNumber uint32, sp *ants.Pool, fsChain FSChain, st Storage, metaSvc *metasvc.Meta, signer ecdsa.PrivateKey, m MetricCollector, ac aclsvc.ACLChecker, rp ACLInfoExtractor, cs ClientConstructor, log *zap.Logger) *Server {
+func New(hs Handlers, sp *ants.Pool, fsChain FSChain, st Storage, metaSvc *metasvc.Meta, signer ecdsa.PrivateKey, m MetricCollector, ac aclsvc.ACLChecker, rp ACLInfoExtractor, cs ClientConstructor, log *zap.Logger) *Server {
 	pubKeyBytes := (*keys.PublicKey)(&signer.PublicKey).Bytes()
 	if len(pubKeyBytes) != compressedECDSAPublicKeyLen {
 		panic(fmt.Sprintf("wrong public key len: expected %d, got %d", compressedECDSAPublicKeyLen, len(pubKeyBytes)))
 	}
-
```
roman-khimov, line 220: :eyes:

pkg/services/sidechain/sidechain.go
roman-khimov: From what I see it's used in meta _signing_ only where header is more appropriate as "the last known current height network it operating with", suggest renaming the method or fixing the comment.

cmd/neofs-node/meta_new.go
```diff
@@ -151,6 +151,17 @@ func initMetaNew(c *cfg) {
 	}))
 }
 
+func metadataChainHardforks() map[string]uint32 {
```
roman-khimov: Should follow FS chain HF config (heights can be different, but just enabling it like this can be problematic as well). At least for 0-height ones.

pkg/services/sidechain/sidechain.go
carpawell: fixed the comment

cmd/neofs-node/meta_new.go
carpawell: ok, dropped everything that has default values

pkg/services/meta_new/blocks.go
carpawell: > normally filled queue on the reader side

what is it? this len means we are nearly possible neo-go deadlock. nobody knows what to do with it, and i have never heard about any solution that can be done

> only writer can do this on attempt to write

i do not see such code on neo-go side, so we will never know. i know for sure that the same code worked (i saw such messages, and tried to new values) in the previous meta service

pkg/services/meta_new/meta.go
carpawell: why HEAD smth, when a node already has header in memory?

carpawell: it also may be non-local, why not?

pkg/services/meta_new/meta.go
carpawell: what do you suggest in this case?

pkg/services/meta/storage.go
carpawell: metabase has batching inside. we may have `PutBatch` here, but why?

cmd/neofs-node/meta_new.go
carpawell: > Should follow FS chain HF config

can you explain? why should different chains be the same in hardforks?

> enabling it like this can be problematic

why?

pkg/services/meta_new/meta.go
carpawell: > Seems like this can be done via JSON-RPC as well

why is it better in any way?

> Can we have this mode (non-local chain) for nodes with IR nearby?

what mode would it be? SN that does not have a local chain? what chain does it put it in then?

this is an interface, so not problem to have any chain source there, but i do not fully inderstand your intention

pkg/services/meta/storage.go
roman-khimov: Per-object DB writes are _very_ expensive. They just can't work with any meaningful ops/s.

pkg/services/meta_new/meta.go
roman-khimov: I don't see any need for them really. You're sharing metrics and header, most likely you either don't need to or can share those in a different manner.

pkg/services/meta_new/meta.go
roman-khimov: Code is more complicated for sure. Whether it's worth this complication or not is an open question.

pkg/services/meta_new/meta.go
carpawell: we either are talking about different lines or do not understand each other. this lock is taken to check whether we have a subscriber just for a just-received object notification (we may not), and we need to drop this subscription (it also stores a channel and is filled in a different place concurrently). time and block height were just an additional part after it was required to investigate performance we did not know

pkg/services/meta_new/meta.go
carpawell: 8 nodes, REP 3: we read (block writing at the same time) our own metabase, OR in 62% go to a different node for a header we had in memory just about ~25ms ago. why?

pkg/services/meta_new/meta.go
carpawell: we either are talking about different lines or do not understand each other. this lock is taken to check whether we have a subscriber just for a just-received object notification (we may not), and we need to drop this subscription (it also stores a channel and is filled in a different place concurrently). time and block height were just an additional part after it was required to investigate performance we did not know

pkg/services/meta/storage.go
carpawell: added batching

pkg/services/meta_new/meta.go
roman-khimov: Pretty easy and already discussed a number of times, IR+SN combo has _two_ chains onboard while it could only have one. We can also experiment with non-local chains, but let's move this to another issue, https://github.com/nspcc-dev/neofs-node/issues/4022

pkg/services/meta_new/blocks.go
roman-khimov: > what is it?

`len(queue) == cap(queue)`, it's a correct behavior and condition, you just don't know if writer has anything else pending to be put into the queue.

cmd/neofs-node/meta_new.go
roman-khimov: > why should different chains be the same in hardforks?

Not exactly the same, but you should follow hardforks in some way to fix various bugs at least, then you can have networks started with different 0-height HFs and you need to deal with that as well. Current code can't do this.

pkg/services/meta_new/meta.go
roman-khimov: Can be improved in future.

pkg/services/meta_new/meta.go
roman-khimov: I don't see any difference in chances, in 3 out of 8 cases it's local, in 5 out of 8 it's networked. Irrespective of this structure. Let's keep it as is for now, this can be checked in future as well.

pkg/services/meta/storage.go
```diff
@@ -0,0 +1,90 @@
...
+
+			ch <- n.o
+		}
+	}
+}
+
+func batchWriter(ctx context.Context, l *zap.Logger, m *metabase.DB, buff <-chan *object.Object) {
+	const (
+		maxBuffSize   = 100
+		maxBatchDelay = 100 * time.Millisecond
```
roman-khimov: 1s here, 100ms is also too optimistic.

pkg/services/meta_new/blocks.go
carpawell: currently, this buffer is 10k size, if its queue is 9999, i would say this is not much further from 10k. this message can be treated "the queue is 9999, we do not expect such load, soon degradation _may_ be expected"

anyway, i am scared of such situations, i would like to find them as fast as possible after load tests. if you do not like such logs, i can suggest metrics about queue length to see it when we are critical. do not agree with simply dropping overflow feedback. writer side is not acceptable in this repo

roman-khimov: OK, I can live with this debug thing for a while.

pkg/services/meta_new/meta.go
carpawell: > 5 out of 8 it's networked

and it means _this_ node collected header, has it already locally in memory, collected every signature, sent tx to the chain, expects the results in ~25ms, and... decided to drop this header to request it soon again from another node. to me, this is "read header from network" vs "read header from memory". i think i do not understand smth

roman-khimov: Are you talking about forwarding in-container node? That'd be 1 out of 5 not storing the object.

pkg/services/meta/storage.go
carpawell: done

cmd/neofs-node/meta_new.go
carpawell: oh, i understand that, yes. but there is no network yet. @AnnaShaleva asked to add a test for new HFs, it is impossible to forget that smth new fixed in neo-go, the test cannot allow it: https://github.com/nspcc-dev/neofs-node/pull/3991#discussion_r3357264847

carpawell: dropped explicit config list, left tests for both applications to be sure the latest known HF is Faun

roman-khimov (approved):

PR #3998, opened 2026-05-21
Replace parallel EC GET with paired stream

cthulhu-rider:
this approach seems more cost-effective overall. However, basic load tests did not show any increase in throughput. Therefore lets resolve this issue separately


roman-khimov:
OK, let's postpone this to 0.54.0 then.

PR #4002, opened 2026-05-27
Fix/new ir does not start

carpawell:
Fixes https://github.com/nspcc-dev/neofs-testcases/pull/1371#issuecomment-4536459808

### DIFF
--- a/pkg/innerring/metadata.go
+++ b/pkg/innerring/metadata.go
@@ -1,0 +1,172 @@
+package innerring
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"math"
+	"math/big"
+	"net"
+	"path"
+	"slices"
+	"strconv"
+	"time"
+
+	"github.com/nspcc-dev/neo-go/pkg/core/storage/dbconfig"
+	"github.com/nspcc-dev/neo-go/pkg/core/transaction"
+	"github.com/nspcc-dev/neo-go/pkg/neorpc"
+	"github.com/nspcc-dev/neo-go/pkg/rpcclient/actor"
+	"github.com/nspcc-dev/neo-go/pkg/rpcclient/gas"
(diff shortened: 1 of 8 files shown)


pkg/innerring/metadata.go
```diff
@@ -1,32 +1,193 @@
...
+		return nil
+	}
+
+	return metaActor
+}
+
+func (nc *metadataNotaryClient) RegisterMetadataContainer(cID cid.ID, nonce uint32) error {
+	act := nc.buildNotaryActor()
+	if act == nil {
+		nc.l.Debug("node is not part of Alphabet list, do not register metadata container", zap.Stringer("cid", cID))
```
cthulhu-rider: we can get here when node is a part of Alphabet, so the message can be misleading. Lets replace it into `buildNotaryActor()`

same below

pkg/innerring/metadata.go
```diff
@@ -1,32 +1,193 @@
...
+			kk = append(kk, k)
+		}
+
+		sort.Sort(kk)
+		cnrVector.Nodes = kk
+
+		placement = append(placement, cnrVector)
+	}
+
+	_, err := act.WaitSuccess(act.Notarize(act.MakeTunedCall(meta.Hash, "updateContainerList", nil, func(r *result.Invoke, t *transaction.Transaction) error {
```
cthulhu-rider: can be shared w/ `RegisterMetadataContainer` somehow? Seems very similar

carpawell: shared

pkg/innerring/metadata.go
carpawell: you mean if there is an error inside `buildNotaryActor` that may lead to returning nil? fixed

cthulhu-rider (approved):

cthulhu-rider:
fixes released stuff? Thinking of changelog

roman-khimov (approved):

PR #3996, opened 2026-05-18
sn/object: Optimize EC GET handling without recovery

cthulhu-rider:
Closes #3952.

PR #4003, opened 2026-05-29
TLS prototype

### LINKED ISSUE #4001: TLS-only prototype
We're running nodes with TLS on testnet/mainnet, however open ports can still be configured and we're still signing all requests to detect who is the author of the request. While the latter is very much useful for forwarded requests many of the things can be simplified if:
 * we're to never expose open ports
 * we're to make mutual TLS auth using node/client keys (CAs don't matter in this case, keys are sufficient)
 * we're to not sign 1:1 requests (TTL=1)

Related to #3978 (secondary queries are exactly 1:1 between SNs).

End-rey:
Closes #4001.

### DIFF
--- a/pkg/services/object/server.go
+++ b/pkg/services/object/server.go
@@ -26,6 +26,7 @@ import (
 	containercore "github.com/nspcc-dev/neofs-node/pkg/core/container"
 	netmapcore "github.com/nspcc-dev/neofs-node/pkg/core/netmap"
 	objectcore "github.com/nspcc-dev/neofs-node/pkg/core/object"
+	"github.com/nspcc-dev/neofs-node/pkg/network/peerauth"
 	metasvc "github.com/nspcc-dev/neofs-node/pkg/services/meta"
 	aclsvc "github.com/nspcc-dev/neofs-node/pkg/services/object/acl/v2"
 	"github.com/nspcc-dev/neofs-node/pkg/services/object/common"
@@ -59,7 +60,9 @@ import (
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
 	grpccodes "google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/mem"
+	"google.golang.org/grpc/peer"
 	grpcstatus "google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
 )
@@ -265,7 +268,9 @@ func (s *Server) sendPutResponse(stream protoobject.ObjectService_PutServer, res
 		resp.MetaHeader = s.makeResponseMetaHeader(util.ToStatus(err))
 	}
 
-	resp.VerifyHeader = util.SignResponseIfNeeded(&s.signer, resp, req)
+	if needSignObjectResponse(stream.Context(), req) {
+		resp.VerifyHeader = util.SignResponse(&s.signer, resp)
+	}
 	return stream.SendAndClose(resp)
 }
 
@@ -335,10 +340,12 @@ func (x *putStream) resignRequest(req *protoobject.PutRequest) (*protoobject.Put
 		Ttl:    meta.GetTtl() - 1,
 		Origin: meta,
 	}
-	var err error
-	req.VerifyHeader, err = neofscrypto.SignRequestWithBuffer(neofsecdsa.Signer(x.signer), req, nil)
-	if err != nil {
-		return nil, err
+	if req.MetaHeader.Ttl > 1 {
+		var err error
+		req.VerifyHeader, err = neofscrypto.SignRequestWithBuffer(neofsecdsa.Signer(x.signer), req, nil)
+		if err != nil {
+			return nil, err
+		}
 	}
 	return req, nil
 }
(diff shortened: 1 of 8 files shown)


End-rey:
Benchmark tests were run on this branch against the 0.53 release. Throughput in ops/s. Very nice improvement for small objects, where the per-message signature dominates and the 1:1-hop skip removes it; large objects are bandwidth-bound, so any difference there is within the test's margin of error (no regression).

**REP3**

| Test | Size, KB | Threads | base | mTLS |   Δ   |
|------|---------:|--------:|-----:|-----:|:-----:|
| PUT  |        4 |       1 |  235 |  241 |  +3%  |
| PUT  |        4 |       4 |  694 |  751 |  +8%  |
| PUT  |        4 |      16 |  998 | 1186 | +19%  |
| PUT  |     4096 |       1 |   89 |   88 |  −1%  |
| PUT  |     4096 |       4 |  171 |  170 |  −1%  |
| PUT  |     4096 |      16 |  210 |  202 |  −4%  |
| GET  |        4 |       1 | 1477 | 2217 | +50%  |
| GET  |        4 |       4 | 2353 | 4373 | +86%  |
| GET  |        4 |      16 | 2877 | 6142 | +114% |
| GET  |     4096 |       1 |  323 |  345 |  +7%  |
| GET  |     4096 |       4 |  535 |  575 |  +7%  |
| GET  |     4096 |      16 |  613 |  669 |  +9%  |

**EC6+2**

| Test | Size, KB | Threads |            base |            mTLS |  Δ   |
|------|---------:|--------:|----------------:|----------------:|:----:|
| PUT  |        4 |       1 |             152 |             167 | +10% |
| PUT  |        4 |       4 |             303 |             369 | +22% |
| PUT  |        4 |      16 |             271 |             316 | +17% |
| PUT  |     4096 |       1 |              86 |              92 | +7%  |
| PUT  |     4096 |       4 |             142 |             160 | +13% |
| PUT  |     4096 |      16 |             173 |             166 | −4%  |
| GET  |        4 |       1 |            1338 |            1655 | +24% |
| GET  |        4 |       4 |            2309 |            3246 | +41% |
| GET  |        4 |      16 | 1240 (4683 err) | 2301 (8646 err) | +86% |
| GET  |     4096 |       1 |             295 |             353 | +20% |
| GET  |     4096 |       4 |             443 |             492 | +11% |
| GET  |     4096 |      16 |   479 (425 err) |   547 (509 err) | +14% |

cthulhu-rider:
very visible improvement for smaller data 🚀 

it seems to me that this approach can be extended to user-SN connections too. If so, we can expect additional performance gains. Are there any blockers? (not for this PR)

cthulhu-rider:
it should also be noted that EC GET [code](https://github.com/nspcc-dev/neofs-node/blob/master/pkg/services/object/get.go) has not been touched yet. Tested version includes #3996 improvement, that's why we see 📈. It may become even faster with same TLS optimization

End-rey:
> it seems to me that this approach can be extended to user-SN connections too. If so, we can expect additional performance gains. Are there any blockers? (not for this PR)

Yes, it can extend to client-SN and would help the read path. In my opinion two problems, though: client authorization can take a trickier path than node-to-node, and it may not be possible to cover all requests this way (but that's not for sure). Worth exploring, but not a drop-in like the node-to-node case.

pkg/services/object/server.go
```diff
@@ -864,10 +881,12 @@ func convertHeadPrm(signer ecdsa.PrivateKey, cnr container.Container, req *proto
 					Ttl:     1,
 				},
 			}
-			var err error
-			req.VerifyHeader, err = neofscrypto.SignRequestWithBuffer(neofsecdsa.Signer(signer), req, nil)
-			if err != nil {
-				return nil, iprotobuf.BuffersSlice{}, err
+			if req.MetaHeader.Ttl > 1 {
```
roman-khimov: It's always 1 now (and in similar cases).

End-rey: Noticed it, fixed it.

End-rey:
After rebase the impact is still there: mTLS keeps a clear positive effect on GET workloads, while PUT remains mostly neutral to slightly negative.

## REP3

| Test | Size, KB | Threads | base | mTLS | Δ|
|------|---------:|--------:|-----:|-----:|:----------------:|
| PUT  |        4 |       1 |  239 |  236 |      -1.3%       |
| PUT  |        4 |       4 |  723 |  729 |      +0.8%       |
| PUT  |        4 |      16 | 1104 | 1207 |      +9.3%       |
| PUT  |     4096 |       1 |   91 |   86 |      -5.5%       |
| PUT  |     4096 |       4 |  182 |  168 |      -7.7%       |
| PUT  |     4096 |      16 |  217 |  202 |      -6.9%       |
| GET  |        4 |       1 | 1755 | 2261 |      +28.8%      |
| GET  |        4 |       4 | 3018 | 4819 |      +59.7%      |
| GET  |        4 |      16 | 3835 | 6793 |      +77.1%      |
| GET  |     4096 |       1 |  344 |  352 |      +2.3%       |
| GET  |     4096 |       4 |  585 |  590 |      +0.9%       |
| GET  |     4096 |      16 |  679 |  692 |      +1.9%       |

## EC6+2

| Test | Size, KB | Threads |           base |            mTLS | Δ |
|------|---------:|--------:|---------------:|----------------:|:----------------:|
| PUT  |        4 |       1 |            164 |             164 |      +0.0%       |
| PUT  |        4 |       4 |            367 |             374 |      +1.9%       |
| PUT  |        4 |      16 |            318 |             311 |      -2.2%       |
| PUT  |     4096 |       1 |             91 |              87 |      -4.4%       |
| PUT  |     4096 |       4 |            159 |             145 |      -8.8%       |
| PUT  |     4096 |      16 |            178 |             172 |      -3.4%       |
| GET  |        4 |       1 |           1311 |            1700 |      +29.7%      |
| GET  |        4 |       4 |           2368 |            3436 |      +45.1%      |
| GET  |        4 |      16 | 695 (4591 err) | 2176 (8881 err) |     +213.1%      |
| GET  |     4096 |       1 |            347 |             368 |      +6.1%       |
| GET  |     4096 |       4 |            477 |             505 |      +5.9%       |
| GET  |     4096 |      16 |  525 (293 err) |   559 (465 err) |      +6.5%       |

roman-khimov:
Superseded by #4086.

PR #4006, opened 2026-06-02
Feat/eacl with container

carpawell:


pkg/innerring/processors/container/process_eacl.go
```diff
@@ -77,7 +79,51 @@ func (cp *Processor) checkSetEACL(req container.PutContainerEACLRequest) error {
 	return nil
 }
 
-func (cp *Processor) approveSetEACL(req container.PutContainerEACLRequest) {
+func (cp *Processor) checkOptionalEacl(req cnrevent.PutContainerEACLRequest, cID cid.ID, cnr container.Container) error {
```
roman-khimov: Why duplicating `checkSetEACL` code? DRY.

pkg/innerring/processors/container/processor.go
```diff
@@ -348,3 +348,12 @@ func (cp *Processor) AddContainerStructs(ctx context.Context) error {
 		cp.log.Info("more containers have been successfully structured in the contract, continue", zap.String("tx", txs))
 	}
 }
+
+func acceptOnlySingleCall(unaryParser func(notaryEvent event.NotaryEvent) (event.Event, error)) event.NotaryParser {
```
roman-khimov: Why can't you have some `p.SetMultiCall()` instead? This affects every request and is duplicated a lot, there is no reason for that.

pkg/morph/event/container/notary_requests.go
```diff
@@ -21,35 +21,58 @@ type CreateContainerV2Request struct {
 	InvocationScript   []byte
 	VerificationScript []byte
 	SessionToken       []byte
+
+	// OptionalEACLTable is set if container eACL was set in the same transaction
+	// container was created. It is nil otherwise.
+	OptionalEACLTable *PutContainerEACLRequest
```
roman-khimov: Many things are optional (session token, for example), not sure for this buys you in the field name.

carpawell: "additional"?

pkg/innerring/processors/container/process_eacl.go
carpawell: they differ a little but in many things (no need to fetch container, we already have it there, different func signature, etc)

pkg/morph/event/container/notary_requests.go
roman-khimov: Just `EACLTable`, it's a pointer, can be `nil` if missing.

carpawell: ok, done

roman-khimov (approved):

PR #4007, opened 2026-06-03
Deduplicate object request fields' handling

cthulhu-rider:


pkg/services/object/common.go
```diff
@@ -60,3 +68,49 @@ func getStatusCodeFromResponseMetaHeader(buffers iprotobuf.BuffersSlice) (uint32
 
 	return code, nil
 }
+
+func checkRequiredObjectAddress(m *protorefs.Address) (cid.ID, oid.ID, string) {
```
carpawell: why returning string, not an error?

pkg/services/object/acl/eacl/v2/headers.go
```diff
@@ -203,16 +203,6 @@ func (h *cfg) readObjectHeaders(dst *headerSource) error {
 					// first object not defined, unexpected, do not attach any header
 				}
 			}
-		case *protoobject.SearchRequest:
```
carpawell, line 219: why dont we have SearchV2Request there? it also has cID

pkg/services/object/acl/eacl/v2/opts.go
```diff
@@ -55,7 +55,7 @@ func WithCID(v cid.ID) Option {
 	}
 }
 
-func WithOID(v *oid.ID) Option {
+func WithOID(v oid.ID) Option {
```
carpawell, line 66: isnt it a separate change?

pkg/services/object/common.go
cthulhu-rider: it's written to status message which is of string type

pkg/services/object/acl/eacl/v2/headers.go
cthulhu-rider: it's never been there 🤷‍♂️ 
refs https://github.com/nspcc-dev/neofs-testcases/issues/1327

pkg/services/object/acl/eacl/v2/opts.go
cthulhu-rider: it could be, but i decided to change everything related to CID/OID in one change

pkg/services/object/acl/v2/service.go
```diff
@@ -433,115 +433,46 @@ func (b Service) verifyBearerTokenAgainstRequest(token bearer.Token, reqCnr cid.
...
-	header := part.Init.Header
-
-	cIDV2 := header.GetContainerId().GetValue()
-	var cID cid.ID
-	err = cID.Decode(cIDV2)
-	if err != nil {
-		return RequestInfo{}, user.ID{}, fmt.Errorf("invalid container ID: %w", err)
-	}
-
+func (b Service) PutRequestToInfo(request *protoobject.PutRequest, initPart *protoobject.PutRequest_Body_Init, cID cid.ID, obj oid.ID) (RequestInfo, user.ID, error) {
```
roman-khimov: I'd prefer `cnr` here, both `cID` and `cnr` were used previously, so some code has to be changed anyway and `cnr` is used in most of other places.

pkg/services/object/common.go
roman-khimov: Still `string` is confusing here and makes using this function unobvious (its declaration reads as error-free), in its essence it's exactly an `error`, let the caller deal with how it's stringified or not.

pkg/services/object/common.go
```diff
@@ -60,3 +68,49 @@ func getStatusCodeFromResponseMetaHeader(buffers iprotobuf.BuffersSlice) (uint32
...
+
+	obj, msg := checkRequiredObjectID(m.ObjectId)
+	if msg != "" {
+		return cid.ID{}, oid.ID{}, "invalid object address: " + msg
+	}
+
+	return cnr, obj, ""
+}
+
+func checkRequiredContainerID(m *protorefs.ContainerID) (cid.ID, string) {
```
roman-khimov: Same for other functions here, it's Go, "smth + error" is what everyone expects.

pkg/services/object/server.go
```diff
@@ -466,18 +466,49 @@ func (s *Server) Put(gStream protoobject.ObjectService_PutServer) error {
...
-			bad.SetMessage("malformed request: empty body")
-			err = bad // defer
+			err = newBadRequestError(missingRequestBodyMessage) // defer
 			return err
 		}
 
-		if reqInfo, objOwner, err := s.reqInfoProc.PutRequestToInfo(req); err != nil {
+		part, ok := req.Body.GetObjectPart().(*protoobject.PutRequest_Body_Init_)
+		if !ok {
+			if err = ps.forwardRequest(req); err != nil {
```
roman-khimov: Just as an idea, maybe it's worth moving init handling outside of the loop, currently it's not immediately clear why we're forwarding here.

pkg/services/object/server.go
```diff
@@ -154,12 +157,12 @@ type Storage interface {
...
+	DeleteRequestToInfo(*protoobject.DeleteRequest, cid.ID, oid.ID, *protosession.SessionTokenV2, *sessionv2.Token, *protosession.SessionToken, *session.Object, *protoacl.BearerToken, *bearer.Token) (aclsvc.RequestInfo, error)
+	HeadRequestToInfo(*protoobject.HeadRequest, cid.ID, oid.ID, *protosession.SessionTokenV2, *sessionv2.Token, *protosession.SessionToken, *session.Object, *protoacl.BearerToken, *bearer.Token) (aclsvc.RequestInfo, error)
+	GetRequestToInfo(*protoobject.GetRequest, cid.ID, oid.ID, *protosession.SessionTokenV2, *sessionv2.Token, *protosession.SessionToken, *session.Object, *protoacl.BearerToken, *bearer.Token) (aclsvc.RequestInfo, error)
+	RangeRequestToInfo(*protoobject.GetRangeRequest, cid.ID, oid.ID, *protosession.SessionTokenV2, *sessionv2.Token, *protosession.SessionToken, *session.Object, *protoacl.BearerToken, *bearer.Token) (aclsvc.RequestInfo, error)
+	SearchV2RequestToInfo(*protoobject.SearchV2Request, cid.ID, *protosession.SessionTokenV2, *sessionv2.Token, *protosession.SessionToken, *session.Object, *protoacl.BearerToken, *bearer.Token) (aclsvc.RequestInfo, error)
```
roman-khimov: I don't like how we're mixing proto/non-proto here (technically they're duplicating) and the overall parameter count becomes bigger than desired. Can we at least avoid proto/non-proto in the same call?

pkg/services/object/common.go
```diff
@@ -114,3 +119,66 @@ func checkObjectID(m *protorefs.ObjectID) (oid.ID, string) {
...
+		var token session.Object
+		if err := token.FromProtoMessage(meta.SessionToken); err != nil {
+			return nil, nil, nil, nil, "invalid session token: " + err.Error()
+		}
+		return nil, nil, meta.SessionToken, &token, ""
+	}
+
+	if meta.SessionTokenV2 != nil {
+		var token sessionv2.Token
+		if err := token.FromProtoMessage(meta.SessionTokenV2); err != nil {
```
roman-khimov: This means we're decoding tokens for all requests instead of taking the decoded one from cache, that's kinda regression.

pkg/services/object/common.go
cthulhu-rider: made error

pkg/services/object/acl/v2/service.go
cthulhu-rider: done

pkg/services/object/server.go

pkg/services/object/common.go
cthulhu-rider: done

pkg/services/object/common.go
```diff
@@ -122,3 +130,49 @@ func callUnary(ctx context.Context, conn *grpc.ClientConn, method string, req an
 
 	return respBuf, nil
 }
+
+func checkRequiredObjectAddress(m *protorefs.Address) (cid.ID, oid.ID, error) {
```
roman-khimov: `s/check/fetch/`? Semantically that's the case.

pkg/services/object/server.go
```diff
@@ -464,18 +464,50 @@ func (s *Server) Put(gStream protoobject.ObjectService_PutServer) error {
...
+
+		var cnrID cid.ID
+		cnrID, err = checkRequiredContainerID(hdr.ContainerId)
+		if err != nil {
+			err = newBadRequestError(invalidPutRequestInitFieldMessage + ": invalid header field: " + err.Error()) // defer
+			return s.sendStatusPutResponse(gStream, err, reqFirst)
+		}
+
+		var objID oid.ID
+		if part.Init.ObjectId != nil {
```
roman-khimov: Can be a part of `checkObjectID`, like `fetchOptionalOID`

pkg/services/object/acl/v2/service.go
```diff
@@ -433,39 +414,39 @@ func (b Service) verifyBearerTokenAgainstRequest(token bearer.Token, reqCnr cid.
 
 // GetRequestToInfo resolves RequestInfo from the request to check it using
 // [ACLChecker].
-func (b Service) GetRequestToInfo(request *protoobject.GetRequest, cnr cid.ID, obj oid.ID) (RequestInfo, error) {
-	return b.findRequestInfo(request, cnr, acl.OpObjectGet, sessionSDK.VerbObjectGet, sessionv2.VerbObjectGet, obj)
+func (b Service) GetRequestToInfo(request *protoobject.GetRequest, cnr cid.ID, tokens common.RequestTokens) (RequestInfo, error) {
```
roman-khimov, line 417: Looks like RequestTokens definition slipped into the previous commit.

roman-khimov (approved):

pkg/services/object/server.go

pkg/services/object/acl/v2/service.go
cthulhu-rider: whoops, fixed

Issue #3978, opened 2026-05-06
Replace request TTL for GET/HEAD/RANGE


cthulhu-rider:
## Is your feature request related to a problem? Please describe.
currently, SN handles https://github.com/nspcc-dev/neofs-api/blob/becf47f13cbb10ae73cd6cd7bb979db07eb47654/session/types.proto#L224 field to determine whether object request should be handled locally-only or it can be forwarded to other SN. If `ttl <= 1`, local storage is accessed only. Otherwise, remote SN can be called as well

TTL also prevents query looping

local storage is always checked first https://github.com/nspcc-dev/neofs-node/blob/2507f0e65b5c60866f6160842f64987a016a39db/pkg/services/object/get/get.go#L333-L340

to make this behave properly, recursive meta header was originally introduced https://github.com/nspcc-dev/neofs-api/blob/aa5f16c4606cb453a92b55db491de026b120ff26/session/types.proto#L240

this approach complicates the request format (i.e. its formation/processing), and also goes beyond the client authentication on which the behavior should be based

## Describe the solution you'd like
get rid of recursive request structure. Instead, authenticate clien (this is done for authorization in any case) and determine what server is

client can be a user, that is anyone outside the current network map

SN can be in the current network map but outside the current container (proxy). In this case, just forward request to first responsive container SN. This simulates redirect mechanism

thus, only `server=container` performs data operations incl. split-chain joining. Also, if `client=container`, request is handled local-only

## Additional context
#3413

roman-khimov:
Rather we need to replace the request when handling in-container things. Non-container nodes can proxy with no changes (original request, one signature, ttl=2), then container nodes _create new requests_ signed by them as appropriate (one signature, ttl=1 directed to replica or EC part holders). This makes ttl kinda binary (same way it is now, only 2/1 can be used) and this leaves the original request handling (wrt security checks) to the first container node, others receive easy to check in-container requests and not bother with the original one. Security-wise this changes nothing since nodes already can do this and nodes already have permissions to GET/HEAD/RANGE anything they'd like to.

PR #4008, opened 2026-06-03
Optimize cryptography for forwarded GET/HEAD/SEARCH requests

cthulhu-rider:


pkg/services/object/server.go
```diff
@@ -1211,18 +1224,43 @@ func convertGetPrm(signer ecdsa.PrivateKey, cnr container.Container, req *protoo
...
+			}
+		}
+
+		if stream.serverInContainer || !proxyCtx.suppressInit {
+			return c.ForAnyGRPCConn(ctx, func(ctx context.Context, conn *grpc.ClientConn) error {
+				return proxyCtx.continueWithConn(ctx, req, conn) // TODO: log error
+			})
+		}
+
+		// TODO: double-check following code is required
```
carpawell, line 1262: what does this todo mean?

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Changelog for NeoFS Node
 
 ### Changed
 - Optimized EC GET request execution (#3996)
+- Optimized GET/HEAD/SEARCH request signing when forwarding by container node  (#XXX)
```
carpawell: should be fixed

pkg/services/object/server.go
```diff
@@ -156,10 +156,10 @@ type Storage interface {
 type ACLInfoExtractor interface {
 	PutRequestToInfo(*protoobject.PutRequest) (aclsvc.RequestInfo, user.ID, error)
 	DeleteRequestToInfo(*protoobject.DeleteRequest) (aclsvc.RequestInfo, error)
-	HeadRequestToInfo(*protoobject.HeadRequest) (aclsvc.RequestInfo, error)
-	GetRequestToInfo(*protoobject.GetRequest) (aclsvc.RequestInfo, error)
+	HeadRequestToInfo(*protoobject.HeadRequest) (aclsvc.RequestInfo, bool, error)
+	GetRequestToInfo(*protoobject.GetRequest) (aclsvc.RequestInfo, bool, error)
 	RangeRequestToInfo(*protoobject.GetRangeRequest) (aclsvc.RequestInfo, error)
-	SearchV2RequestToInfo(*protoobject.SearchV2Request) (aclsvc.RequestInfo, error)
+	SearchV2RequestToInfo(*protoobject.SearchV2Request) (aclsvc.RequestInfo, bool, error)
```
carpawell, line 162: imo, desc should be extended then

carpawell (commented):
First commit has "TBD" body

pkg/services/object/server.go
cthulhu-rider: this is recent code in which the original request is changed. Therefore, this is not exactly forwarding. So i decided to left behavior for this case as is. But maybe smth can be improved for it

pkg/services/object/server.go
cthulhu-rider: by desc u mean `aclsvc.RequestInfo`?

pkg/services/object/acl/v2/service.go
```diff
@@ -433,45 +433,75 @@ func (b Service) verifyBearerTokenAgainstRequest(token bearer.Token, reqCnr cid.
 
 // GetRequestToInfo resolves RequestInfo from the request to check it using
 // [ACLChecker].
-func (b Service) GetRequestToInfo(request *protoobject.GetRequest) (RequestInfo, error) {
+func (b Service) GetRequestToInfo(request *protoobject.GetRequest) (RequestInfo, bool, error) {
```
roman-khimov, line 436: Can be a part of `RequestInfo`? It's a structure anyway.

pkg/services/object/server.go
```diff
@@ -806,10 +806,20 @@ func convertHeadPrm(signer ecdsa.PrivateKey, cnr container.Container, req *proto
...
+					Body: req.Body,
+					MetaHeader: &protosession.RequestMetaHeader{
+						Version: version.Current().ProtoMessage(),
+						Ttl:     1,
+					},
+				}
+			} else {
+				req.MetaHeader = &protosession.RequestMetaHeader{
+					// TODO: #1165 think how to set the other fields
+					Ttl:    meta.GetTtl() - 1,
```
roman-khimov, line 822: Why changing it at all? We don't want ttl=1 request to container node, it should be able to forward it appropriately.

pkg/services/object/acl/v2/service.go
cthulhu-rider: can be, but will be undefined for some requests then. So decided to separate for now

pkg/services/object/server.go
cthulhu-rider: currently this works when server asks container nodes 1-by-1. Not exactly forwarding. And i touched only in-container case. To cover non-container ones, more should be done

pkg/services/object/acl/v2/service.go
roman-khimov: Eventually all of them should get it and then if it's unused in some right now it doesn't hurt either.

cthulhu-rider: prefer to keep as is and change things as needed

roman-khimov (approved):
Conflicts.

pkg/services/object/server.go
roman-khimov: OK for now, keeps the behavior as is. The rest is #3978.

pkg/services/object/acl/v2/service.go
roman-khimov: Can be justified by the fact that this data doesn't relate to the original request (it's node/container relation that is external to request), still I think eventually it'd be more convenient to have it all in `RequestInfo`. Not a blocker for now anyway.

pkg/services/object/server.go
carpawell: interface, bool arg always confuse

carpawell (approved):

roman-khimov:
BTW, authorship needs to be fixed as well.

PR #4011, opened 2026-06-05
morph: make retrying messages more informative

carpawell:


pkg/morph/client/notary.go
```diff
@@ -505,7 +505,11 @@ func (c *Client) NotarySignAndInvokeTX(mainTx *transaction.Transaction, await bo
 		},
 		expBackoff,
 		func(err error, d time.Duration) {
-			c.logger.Info("retrying due to error", zap.Error(err), zap.Duration("retry-after", d))
+			c.logger.Info("retrying to send signed notary request due to error",
```
roman-khimov: `s/due to error//`, we already have `zap.Error` here, so it's obvious

roman-khimov (approved):

pkg/morph/client/notary.go
carpawell: fixed

PR #4013, opened 2026-06-09
EC GET fixes

### LINKED ISSUE #4012: EC GET breaks sometimes
defects of #3996

1. SN can respond w/ incorrect payload of EC object
2. SN shutdown can hang after EC GET requests
3. following errors sometimes occur
```
part#1: /dns4/localhost/tcp/8084: parse ID (nested message) field: parse value (bytes) field: parse field #1 of LEN type: unexpected EOF
```
```
part#1: /dns4/localhost/tcp/8084: handle meta header: parse version (nested message) field: wrong type of field #1: expected LEN, got VARINT
```

## Possible solution
1 and 2 are just code mistakes., i already found a solution

3 is not yet clear. Code seems ok overall, and it happens inconsistently. Local debug revealed dependency on the concurrent requests. Isolated errors begin to pop up at ~90-100 threads. After that, a more or less linear dependence errors<->routines is observed

load testing

i'm using `EC 4/1` policy and sending GET requests to one of container nodes

## Regression
yes

cthulhu-rider:
closes #4012

### DIFF
--- a/pkg/services/object/get.go
+++ b/pkg/services/object/get.go
@@ -418,7 +418,9 @@ func (x *getECTransport) CopyLocalECPartParentHeaderAndPayload(ctx context.Conte
 		if errors.Is(err, apistatus.ErrObjectAlreadyRemoved) || errors.As(err, &splitErr) {
 			return false, 0, 0, 0, err
 		}
-		logError("local storage failure (read EC part)", err)
+		if !errors.Is(err, apistatus.ErrObjectNotFound) {
+			logError("local storage failure (read EC part)", err)
+		}
 		return false, 0, 0, 0, nil
 	}
 
@@ -514,7 +516,9 @@ func (x *getECTransport) CopyLocalECPartRange(ctx context.Context, storage *engi
 		if errors.Is(err, apistatus.ErrObjectAlreadyRemoved) {
 			return 0, err
 		}
-		logError("local storage failure (read EC part range)", err)
+		if !errors.Is(err, apistatus.ErrObjectNotFound) {
+			logError("local storage failure (read EC part range)", err)
+		}
 		return 0, nil
 	}
 
--- a/pkg/services/object/get.go
+++ b/pkg/services/object/get.go
@@ -12,6 +12,7 @@ import (
 	"sync"
 
 	iec "github.com/nspcc-dev/neofs-node/internal/ec"
+	igrpc "github.com/nspcc-dev/neofs-node/internal/grpc"
 	iobject "github.com/nspcc-dev/neofs-node/internal/object"
 	iprotobuf "github.com/nspcc-dev/neofs-node/internal/protobuf"
 	"github.com/nspcc-dev/neofs-node/internal/protobuf/protoscan"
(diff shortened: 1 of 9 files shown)


pkg/services/object/get.go
```diff
@@ -384,6 +384,10 @@ type getECTransport struct {
 	getPartRangeRequests    map[iec.PartInfo]*preparedRangeRequest
 
 	rangeViaGet bool
+
+	// if rangeViaGet is set
+	getPartRequestsMtx sync.RWMutex
+	getPartRequests    map[iec.PartInfo]*mem.Buffer
```
roman-khimov, line 391: Have you checked https://github.com/nspcc-dev/neofs-node/pull/4014? It's pretty much the same problem and that code is simpler (and allocating less).

cthulhu-rider: no, i only focused on current problems. It's closer to #3998 which i wanna revisit after higher impact optimizations

pkg/services/object/get.go
```diff
@@ -1153,6 +1170,49 @@ func (x *getECTransport) makeGetECPartRangeRequest(partInfo iec.PartInfo, off, l
...
+	x.getPartRequestsMtx.Lock()
+
+	reqPtr = x.getPartRequests[partInfo]
+	if reqPtr != nil {
+		x.getPartRequestsMtx.Unlock()
+		return *reqPtr, nil
+	}
+
+	if x.getPartRequests == nil {
+		x.getPartRequests = make(map[iec.PartInfo]*mem.Buffer, 1)
```
roman-khimov: Why 1 here, shouldn't it be the number of parts? Do we need it at all?

pkg/services/object/get.go
```diff
@@ -1153,6 +1170,49 @@ func (x *getECTransport) makeGetECPartRangeRequest(partInfo iec.PartInfo, off, l
...
+
+	reqObj := x.request.GetBody().GetAddress()
+	cnr := reqObj.GetContainerId().GetValue()
+	parent := reqObj.GetObjectId().GetValue()
+
+	reqMetaHdr := x.request.GetMetaHeader()
+	sessionToken := reqMetaHdr.GetSessionTokenV2()
+	sessionTokenV1 := reqMetaHdr.GetSessionToken()
+
+	reqBuf, err := x.server.makeGetECPartRequest(cnr, parent, partInfo, sessionToken, sessionTokenV1)
```
roman-khimov: Notice that it would be the same request with https://github.com/nspcc-dev/neofs-api/issues/348

roman-khimov (approved):

pkg/services/object/get.go
cthulhu-rider: i dont rly wanna accept total number of parts as a parameter just for this

> Do we need it at all?

map? It allows to not recreate requests in cases except the best one (when part is received from the optimal node)

roman-khimov: I mean size parameter, seems to be useless here.

cthulhu-rider: dropped

PR #4014, opened 2026-06-09
object: speed up split streaming

### LINKED ISSUE #2791: Speed up split object streaming
Node can try to collect more child objects at once for streaming to a client, not one by one.

Think and try some alternatives to this [code](https://github.com/nspcc-dev/neofs-node/blob/5c10c503ce7a9be8f25d53972ace416541c80f75/pkg/services/object/get/assemble.go#L165-L172).
At least keeping 2 objects _may_ be more effective. While streaming one object to a client, the second one is being fetched. This way, with the same network speed b/w nodes and clients, a client should not wait for a child object to be received from the next-door node.

Keep it as is. At least, it works and we do not have _yet_ issues about slow object assembling.

End-rey:
Closes #2791.

### DIFF
--- a/pkg/services/object/get/assemble.go
+++ b/pkg/services/object/get/assemble.go
@@ -159,6 +159,21 @@ func (exec *execCtx) initFromChild(obj oid.ID) (*oid.ID, []oid.ID) {
 func (exec *execCtx) overtakePayloadDirectly(children []oid.ID, rngs []object.Range, checkRight bool) {
 	withRng := len(rngs) > 0 && exec.ctxRange() != nil
 
+	if len(children) > 1 && prefetchWindow > 1 {
+		at := func(i int) (oid.ID, *object.Range, bool) {
+			if i >= len(children) {
+				return oid.ID{}, nil, false
+			}
+			if withRng {
+				return children[i], &rngs[i], true
+			}
+			return children[i], nil, true
+		}
+
+		exec.statusError = exec.streamChildrenPipelined(at, !withRng && checkRight)
+		return
+	}
+
 	for i := range children {
 		var r *object.Range
 		if withRng {
--- a/pkg/services/object/get/assemble_pipeline.go
+++ b/pkg/services/object/get/assemble_pipeline.go
@@ -1,0 +1,120 @@
+package getsvc
+
+import (
+	"context"
+	"io"
+	"sync"
+
+	"github.com/nspcc-dev/neofs-sdk-go/object"
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+)
+
+// prefetchWindow is the split-assembly look-ahead: one child streams to the
+// client while up to prefetchWindow-1 others are fetched concurrently.
+// value <=1 disables pipelining and falls back to the serial path.
+var prefetchWindow = 2
+
+// childSlot carries one child's stream and the producer's result.
+type childSlot struct {
+	ready chan struct{}
+	rc    io.ReadCloser
+	hdr   *object.Object
+	se    statusError
+}
+
+// streamChildrenPipelined fetches children concurrently, up to prefetchWindow in flight,
+// and writes their payloads to the client writer strictly in index order. at(i)
+// yields the i-th child ID and its sub-range; ok=false means there are no more
+// children.
+func (exec *execCtx) streamChildrenPipelined(at func(i int) (oid.ID, *object.Range, bool), checkHdr bool) statusError {
+	ring := make([]*childSlot, prefetchWindow)
+
+	gCtx, cancel := context.WithCancel(exec.context())
+	defer cancel()
+
+	var wg sync.WaitGroup
+	launch := func(i int) bool {
+		id, rng, ok := at(i)
+		if !ok {
+			return false
+		}
+
+		s := &childSlot{ready: make(chan struct{})}
+		ring[i%len(ring)] = s
+		wg.Go(func() {
+			defer close(s.ready)
+			s.hdr, s.rc, s.se = exec.fetchChildStream(gCtx, id, rng)
+		})
+		return true
+	}
+
+	launched := 0
+	for launched < prefetchWindow && launch(launched) {
+		launched++
+	}
+	if launched == 0 {
+		return statusError{status: statusOK}
+	}
+	if launched < len(ring) {
+		ring = ring[:launched]
+	}
+
+	buf := bufferPool.Get()
+	defer bufferPool.Put(buf)
+	dst := exec.prm.objWriter
+
+	result := statusError{status: statusOK}
+	for i := 0; i < launched; i++ {
+		if se := exec.streamChildSlot(ring[i%len(ring)], dst, *buf.(*[]byte), checkHdr); se.status != statusOK {
+			result = se
+			break
+		}
+
+		if launch(launched) {
+			launched++
+		}
+	}
+
+	cancel()
+	closeChildSlots(ring)
+	wg.Wait()
+	return result
+}
+
+func (exec *execCtx) streamChildSlot(s *childSlot, dst ChunkWriter, buf []byte, checkHdr bool) statusError {
+	<-s.ready
+
+	if s.se.status != statusOK {
+		return s.se
+	}
+
+	if checkHdr && s.hdr != nil && !exec.isChild(s.hdr) {
+		exec.log.Debug("parent address in child object differs")
+	}
+
+	_, copyErr := copyPayloadStreamBuffer(dst, s.rc, buf)
+	closeErr := s.rc.Close()
+	s.rc = nil
+
+	if copyErr == nil && closeErr != nil {
+		copyErr = closeErr
+	}
+	if copyErr != nil {
+		return statusError{status: statusUndefined, err: copyErr}
+	}
+
+	return statusError{status: statusOK}
+}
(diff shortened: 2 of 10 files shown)


pkg/services/object/get/assemble.go
```diff
@@ -159,6 +159,19 @@ func (exec *execCtx) initFromChild(obj oid.ID) (*oid.ID, []oid.ID) {
 func (exec *execCtx) overtakePayloadDirectly(children []oid.ID, rngs []object.Range, checkRight bool) {
 	withRng := len(rngs) > 0 && exec.ctxRange() != nil
 
+	if len(children) > 1 && prefetchWindow > 1 {
+		var prngs []*object.Range
+		if withRng {
+			prngs = make([]*object.Range, len(children))
```
roman-khimov: Why copying into a list of pointers if you already have the same data in `rngs`?

pkg/services/object/get/assemble_pipeline.go
```diff
@@ -0,0 +1,112 @@
...
+
+// streamChildrenPipelined fetches children concurrently, up to w in flight, and
+// writes their payloads to writer strictly in index order.
+func (exec *execCtx) streamChildrenPipelined(children []oid.ID, rngs []*object.Range, checkHdr bool, w int) statusError {
+	n := len(children)
+	if w < 1 {
+		w = 1
+	}
+
+	slots := make([]*childSlot, n)
```
roman-khimov: For all 100500 children? Ring buffer of `w` size would be more efficient.

pkg/services/object/get/assembly_v2.go
```diff
@@ -152,7 +154,17 @@ func (exec *execCtx) rangeFromLink(link object.Link) bool {
 			}
 		}
 
-		if !exec.copyChild(child.ObjectID(), rngPerChild, false) {
+		ids = append(ids, child.ObjectID())
+		rngs = append(rngs, rngPerChild)
```
roman-khimov: Likely some iterator can be made out of link data to avoid allocations.

pkg/services/object/get/assemble_pipeline.go
```diff
@@ -0,0 +1,112 @@
...
+	"context"
+	"errors"
+	"io"
+	"sync"
+
+	"github.com/nspcc-dev/neofs-sdk-go/object"
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+)
+
+var prefetchWindow = 2
```
roman-khimov, line 15: A comment on why this is a `var` would be appropriate.

pkg/services/object/get/assemble.go
End-rey: Fixed.

pkg/services/object/get/assembly_v2.go
End-rey: Done.

pkg/services/object/get/assemble_pipeline.go
End-rey: Done.

pkg/services/object/get/assemble_pipeline.go
End-rey: Done.

pkg/services/object/get/assemble.go
```diff
@@ -159,6 +159,18 @@ func (exec *execCtx) initFromChild(obj oid.ID) (*oid.ID, []oid.ID) {
...
+			if withRng {
+				return children[i], &rngs[i]
+			}
+			return children[i], nil
+		}
+
+		exec.statusError = exec.streamChildrenPipelined(len(children), at, !withRng && checkRight, prefetchWindow)
+		return
+	}
+
```
roman-khimov, line 176: What worries me a bit is that everything down below effectively becomes a dead code. Most likely it can be dropped in future.

pkg/services/object/get/assemble_pipeline.go
roman-khimov: I mean more that it's not a `const` since it's used in tests, but anyway.

pkg/services/object/get/assemble_pipeline.go
```diff
@@ -0,0 +1,112 @@
...
+
+// prefetchWindow is the split-assembly look-ahead: one child streams to the
+// client while up to prefetchWindow-1 others are fetched concurrently.
+// value <=1 disables pipelining and falls back to the serial path.
+var prefetchWindow = 2
+
+// pipeChunkWriter is the producer-side ChunkWriter: it forwards payload chunks
+// into an io.Pipe so the ordered consumer can drain them.
+type pipeChunkWriter struct {
+	w *io.PipeWriter
```
roman-khimov: Looks like an additional data copy, btw.

pkg/services/object/get/assemble_pipeline.go
```diff
@@ -0,0 +1,112 @@
...
+	pr  *io.PipeReader
+	pw  *io.PipeWriter
+	hdr *object.Object
+	se  statusError
+}
+
+// streamChildrenPipelined fetches n children concurrently, up to w in flight,
+// and writes their payloads to the client writer strictly in index order. at(i)
+// yields the i-th child ID and its sub-range (nil means the whole child).
+func (exec *execCtx) streamChildrenPipelined(n int, at func(i int) (oid.ID, *object.Range), checkHdr bool, w int) statusError {
```
roman-khimov: Please drop `n` parameter, it's not needed (allocate for `w`, fetch as much as you can, trim if there are not enough elements). Also `w` likely can be avoided, it's always the same package-level var.

pkg/services/object/get/assemble.go
```diff
@@ -159,6 +159,18 @@ func (exec *execCtx) initFromChild(obj oid.ID) (*oid.ID, []oid.ID) {
 func (exec *execCtx) overtakePayloadDirectly(children []oid.ID, rngs []object.Range, checkRight bool) {
 	withRng := len(rngs) > 0 && exec.ctxRange() != nil
 
+	if len(children) > 1 && prefetchWindow > 1 {
+		at := func(i int) (oid.ID, *object.Range) {
```
roman-khimov: Can it be `(oid.ID, object.Range)`, without a pointer?

roman-khimov (commented):
Do we have any test results from real cluster for this? It looks ok in general, but I'd like to check that it has some real effect.

pkg/services/object/get/assemble_pipeline.go
End-rey: Yeah, fair point - it’s an extra pass through pipe. Otherwise we’d need to keep the child in memory and write it later in order. I picked pipe as the simpler streaming approach. What other options do you think we should consider?

pkg/services/object/get/assemble.go
End-rey: Not really: `nil` means “whole child” here. With plain `object.Range`, we’d lose that and need an extra flag.

roman-khimov: `0,0` is the same to me, we do have a notion of that for ranged requests.

pkg/services/object/get/assemble_pipeline.go
roman-khimov: Ordinary forwarding code does the same thing effectively, it's just that you're reading from one source at a time here (the other is kinda ready, but you're just not reading from it till it's time).

pkg/services/object/get/assemble.go
End-rey: Why is that better? We're still using a pointer in [RangePrm](https://pkg.go.dev/github.com/nspcc-dev/neofs-node/pkg/services/object/get#RangePrm.SetRange) anyway.

roman-khimov: Well, this sucks as well and can be fixed. In some cases it can lead to microallocations for no real reason, the struct is just 16 bytes. Not critical for now though.

pkg/services/object/get/assemble_pipeline.go
End-rey: Done.

pkg/services/object/get/assemble_pipeline.go
End-rey: Done.

End-rey:
| Policy | Test | Size, GB | Threads per Node |     0.25.0 | speed up split |     Δ |
|--------|------|---------:|-----------------:|-----------:|---------------:|------:|
| REP3   | PUT  |        1 |                1 | 0.54 obj/s |     0.52 obj/s | -3.7% |
| REP3   | PUT  |        1 |                2 | 0.80 obj/s |     0.78 obj/s | -2.5% |
| REP3   | PUT  |        1 |                4 | 0.93 obj/s |     0.92 obj/s | -1.1% |
| REP3   | GET  |        1 |                1 | 1.61 obj/s |     1.65 obj/s | +2.5% |
| REP3   | GET  |        1 |                2 | 2.27 obj/s |     2.27 obj/s |    0%  |
| REP3   | GET  |        1 |                4 | 2.59 obj/s |     2.55 obj/s | -1.5% |

The observed numbers are within the margin of error. Most likely, the change does not show a noticeable improvement in this setup because the potential gain is masked by other, more resource-intensive operations currently dominating the request path.

roman-khimov (approved):

Issue #3689, opened 2025-11-21
IR stops processing (some?) requests after RPC reconnection


roman-khimov:
## Expected Behavior
RPC connection loss -> reconnect -> continue working with notary requests as usual.

## Current Behavior
RPC connection loss -> reconnect -> node bootstrap requests are no longer processed.

## Possible Solution
Fix something.

## Steps to Reproduce (for bugs)
External RPC, drop connection. Most likely it's not 100% reproducible.

## Context
This has happened on mainnet.

## Regression
Doesn't look like.

## Your Environment
* Version used: 0.49.0

roman-khimov:
Reproduced with testnet, RPCs have restarted at the same time.

AnnaShaleva:
Was reproduced 5 Jun 2026 on NeoFS mainnet due to simultaneous NeoGo nodes restart (including all RPCs) after https://github.com/nspcc-dev/neo-go/issues/2552#issuecomment-4630196752. The problem was in epoch tick Notary requests handling.

PR #4020, opened 2026-06-16
morph/client: improve resubscription logs after connection loss

carpawell:
This will allow us to be sure client subscription was restored correctly with INFO logs. Refs #3689.


roman-khimov (approved):
Noisy, but we need it for now.

carpawell:
We have INFO logs for reconnection anyway. This is likely to happen once in a huge time interval compared to any policer error log.

PR #4021, opened 2026-06-16
Optimize forwarding of GET/HEAD/RANGE/SEARCH requests

### LINKED ISSUE #3978: Replace request TTL for GET/HEAD/RANGE

cthulhu-rider:
closes #3978

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ Changelog for NeoFS Node
 
 ### Changed
 - Optimized EC GET request execution (#3996)
-- Optimized GET/HEAD/SEARCH request signing when forwarding by container node  (#4008)
+- Optimized GET/HEAD/SEARCH/RANGE request signing when forwarding  (#4008, #4021)
 - Optimized ranged GET in FSTree (#4016)
 
 ### Removed
--- a/cmd/neofs-node/object.go
+++ b/cmd/neofs-node/object.go
@@ -47,11 +47,9 @@ import (
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	protoobject "github.com/nspcc-dev/neofs-sdk-go/proto/object"
-	protosession "github.com/nspcc-dev/neofs-sdk-go/proto/session"
 	"github.com/nspcc-dev/neofs-sdk-go/reputation"
 	sessionv2 "github.com/nspcc-dev/neofs-sdk-go/session/v2"
 	"github.com/nspcc-dev/neofs-sdk-go/user"
-	"github.com/nspcc-dev/neofs-sdk-go/version"
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
 )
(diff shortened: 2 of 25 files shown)


internal/netmap/netmap.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+func nodeSetContainsPublicKeyFunc(nodeSet []netmap.NodeInfo, checkPubKeyFn func([]byte) bool) bool {
+	return slices.ContainsFunc(nodeSet, func(node netmap.NodeInfo) bool {
+		return checkPubKeyFn(node.PublicKey())
+	})
+}
+
+// ZapEndpoints returns zap log field with node network endpoints.
+func ZapEndpoints(node netmap.NodeInfo) zap.Field {
```
roman-khimov, line 25: Looks somewhat alien here, although it's about `netmap.NodeInfo`.

cmd/neofs-node/object.go
```diff
@@ -676,18 +682,9 @@ func (o objectSource) SearchOne(ctx context.Context, cnr cid.ID, filters object.
...
-			},
 		}
 	)
 
-	req.VerifyHeader, err = neofscrypto.SignRequestWithBuffer[*protoobject.SearchV2Request_Body](o.signer, req, nil)
-	if err != nil {
-		return id, err
-	}
-
 	res, _, err := o.server.ProcessSearch(ctx, req, false)
```
roman-khimov: BTW, not sure this `false` is correct here, non-container nodes don't bother with TS verification anyway (which is kinda proven by removed signature, this request is never truly sent now, only ttl=1 are).

pkg/services/object/search.go
```diff
@@ -37,3 +46,41 @@ func iterateSearchableContainerNodes(nodeSets [][]netmap.NodeInfo, repRules []ui
 		}
 	}
 }
+
+func (s *Server) forwardSearchRequest(ctx context.Context, req any, nodeSets [][]netmap.NodeInfo) (mem.BufferSlice, error) {
+	for i := range nodeSets {
+		for j := range nodeSets[i] {
+			node, err := s.nodeClients.Get(ctx, nodeSets[i][j])
```
roman-khimov: Makes us pick the same node for different requests on different nodes. GET and alike are all sorted, here we need some shuffling to spread the load.

cmd/neofs-node/object.go
cthulhu-rider: flag specifies whether try to forward or not. Since we create request from scratch here, the flag is unset

but dropping meta header was a mistake. Reverted

pkg/services/object/search.go
cthulhu-rider: done

CHANGELOG.md
```diff
@@ -16,7 +16,7 @@ Changelog for NeoFS Node
 - Optimized EC GET request execution (#3996)
 - Optimized GET/HEAD/SEARCH request signing when forwarding by container node  (#4008)
 - Optimized ranged GET in FSTree (#4016)
-- Optimized GET/HEAD/SEARCH request signing when forwarding  (#4008, #4021)
+- Optimized GET/HEAD/RANGE/SEARCH request forwarding  (#4008, #4021)
```
roman-khimov: Duplicated now?

pkg/services/object/get/forward.go
```diff
@@ -4,86 +4,86 @@ import (
...
+			if fin, err := s.forwardRequestToNode(ctx, sortedNodeLists[i][0], op, forwardRequestFn); fin {
 				return err
 			}
+			continue
+		}
 
-			if !errors.Is(err, apistatus.ErrObjectNotFound) {
-				s.log.Info("failed to GET object from remote node", zap.Error(err))
+		// randomize load on equilibrium nodes
+		for _, nodeIdx := range islices.ShuffleIndexes(int(repRules[i])) {
```
roman-khimov, line 35: This seems excessive for GET/HEAD (lists are sorted anyway), although it can give some benefit when everyone is rushing for the same object.

pkg/services/object/search.go
roman-khimov: Still, SEARCH is not spread. For REP it has no meaning, for EC it does (when we only ask a subset of nodes).

pkg/services/object/get/forward.go
cthulhu-rider: u told me about shuffling, i added shuffling

roman-khimov: That was specifically about SEARCH, right in `forwardSearchRequest`.

CHANGELOG.md
cthulhu-rider: fixed

pkg/services/object/search.go
cthulhu-rider: done

roman-khimov (approved):

PR #4023, opened 2026-06-17
Fix/container with eacl things

carpawell:


cmd/neofs-cli/modules/container/create.go
```diff
@@ -184,6 +184,10 @@ It will be stored in FS chain when inner ring will accepts it.`,
 		}
 
 		if flagVarsSetEACL.srcPath != "" {
+			if !cnr.BasicACL().Extendable() {
+				return fmt.Errorf("'%s' flag in not empty, but container does not allow extended ACL", commonflags.EACLTable)
```
roman-khimov: is not

carpawell: fixed

roman-khimov (approved):

PR #4024, opened 2026-06-17
Fix access denied on payload-only GET

End-rey:


roman-khimov (approved):
I'm not excited about this change, but looks like with #3978 it's inevitable, SNs can miss the original request completely and forwarding node has to have the header to ensure EACL.

carpawell (approved):
Can this be more flexible, like doing this only if we know for sure there are eACL rules based on any header info?

End-rey:
> Can this be more flexible, like doing this only if we know for sure there are eACL rules based on any header info?

In theory yes: we already have the [recheckEACL](https://github.com/nspcc-dev/neofs-node/blob/a09f9da34d9e82b1ded721da1838da823fbc6058/pkg/services/object/server.go#L979) flag, so if we propagate it carefully through this path, we should be able to do this only for requests that may actually need the recheck, not for every request.

roman-khimov:
What also bugs me wrt this is default S3 EACL set, to me it should not require any EACL rechecking. @smallhive, can you provide us with the set of rules created by the gateway?

PR #4029, opened 2026-06-19
object/get: make internal payload-only GETs conditional on eACL recheck

### LINKED ISSUE #4028: Request header only when needed for EACL recheck
> Can this be more flexible, like doing this only if we know for sure there are eACL rules based on any header info?

In theory yes: we already have the [recheckEACL](https://github.com/nspcc-dev/neofs-node/blob/a09f9da34d9e82b1ded721da1838da823fbc6058/pkg/services/object/server.go#L979) flag, so if we propagate it carefully through this path, we should be able to do this only for requests that may actually need the recheck, not for every request.

_Originally posted by @End-rey in https://github.com/nspcc-dev/neofs-node/issues/4024#issuecomment-4743018691_

End-rey:
Closes #4028.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Changelog for NeoFS Node
 - SN panic on start without Control service (#4017)
 - Incorrect protobuf scanning (#4013)
 - Data race in RANGE response buffer (#4013)
-- Incorrect access denial during EACL rechecks for payload-only GET (#4024)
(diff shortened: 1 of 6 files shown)


roman-khimov (commented):
It looks good, but let's wait for #4021 to get in first.

carpawell (approved):

roman-khimov:
Conflicts.

PR #4030, opened 2026-06-19
sn/object: move placement logic into shared service

### LINKED ISSUE #1803: Share functionality of placement building between components

End-rey:
Closes #1803.

### DIFF
--- a/pkg/services/object/placement/service.go
+++ b/pkg/services/object/placement/service.go
@@ -1,4 +1,4 @@
-package main
+package placement
 
 import (
 	"fmt"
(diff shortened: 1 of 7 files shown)


pkg/services/object/placement/service.go
```diff
@@ -13,13 +13,18 @@ import (
...
-// particular container and network map incl. error.
-type storagePolicyRes struct {
-	nodeSets  [][]netmap.NodeInfo
-	repCounts []uint
-	ecRules   []iec.Rule
-	err       error
+// Placement stores storage policy application result for a particular
+// container and network map.
+type Placement struct {
+	Epoch     uint64
```
roman-khimov, line 19: Likely can be avoided in this structure (it's used in _keys_, so it's somewhere around anyway and we can avoid duplicating it in value).

roman-khimov (approved):
Neat.

There are also calls to `ContainerNodes()` in IR and CLI, but CLI is a one-shot thing and IR events are rather rare, per-epoch ones are useless, load processing can benefit from this cache, but it's not critical there either.

pkg/services/object/placement/service.go
End-rey: I see the point, but this value is passed outside of the cache and later used for object-specific sorting. Keeping `Epoch` here preserves the link between the placement and the netmap it was built from. Otherwise we'd need to pass epoch separately or add another wrapper, which seems less clear for this local case.

roman-khimov:
Rebase, please.

cthulhu-rider (approved):

PR #4033, opened 2026-06-22
sn: return first found EC part when part index is missing

carpawell:
Implementation of https://github.com/nspcc-dev/neofs-api/pull/393.
How other nodes should behave is a subject to be discussed.

### DIFF
--- a/pkg/local_object_storage/engine/ec.go
+++ b/pkg/local_object_storage/engine/ec.go
@@ -54,26 +54,42 @@ func (e *StorageEngine) ReadECPart(_ context.Context, cnr cid.ID, parent oid.ID,
 // If object is marked as garbage (e.g. via [StorageEngine.MarkGarbage]),
 // GetECPart returns [apistatus.ErrObjectNotFound].
 //
+// If allowReturnAnyPart is true, in cases when specified object part ID cannot be
+// found, StorageEngine returns any part it has for this EC rule. Returned part ID
+// should be found in returned header's attributes.
+//
 // If object is locked (e.g. via [StorageEngine.Lock] or stored locker object),
 // GetECPart ignores expiration, tombstone and garbage marks.
-func (e *StorageEngine) GetECPart(_ context.Context, cnr cid.ID, parent oid.ID, pi iec.PartInfo) (object.Object, io.ReadCloser, error) {
+func (e *StorageEngine) GetECPart(_ context.Context, cnr cid.ID, parent oid.ID, pi iec.PartInfo, allowReturnAnyPart bool) (object.Object, io.ReadCloser, error) {
 	if e.metrics != nil {
 		defer elapsed(e.metrics.AddGetECPartDuration)()
 	}
 
-	var hdr object.Object
-	var stream io.ReadCloser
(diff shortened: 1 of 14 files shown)


pkg/local_object_storage/engine/ec.go
```diff
@@ -54,26 +54,42 @@ func (e *StorageEngine) ReadECPart(_ context.Context, cnr cid.ID, parent oid.ID,
...
 // GetECPart returns [apistatus.ErrObjectNotFound].
 //
+// If allowReturnAnyPart is true, in cases when specified object part ID cannot be
+// found, StorageEngine returns any part it has for this EC rule. Returned part ID
+// should be found in returned header's attributes.
+//
 // If object is locked (e.g. via [StorageEngine.Lock] or stored locker object),
 // GetECPart ignores expiration, tombstone and garbage marks.
-func (e *StorageEngine) GetECPart(_ context.Context, cnr cid.ID, parent oid.ID, pi iec.PartInfo) (object.Object, io.ReadCloser, error) {
+func (e *StorageEngine) GetECPart(_ context.Context, cnr cid.ID, parent oid.ID, pi iec.PartInfo, allowReturnAnyPart bool) (object.Object, io.ReadCloser, error) {
```
roman-khimov, line 63: Why not encoding this in `iec.PartInfo`? It's an int there, can be -1.

pkg/local_object_storage/engine/ec.go
```diff
@@ -54,26 +54,42 @@ func (e *StorageEngine) ReadECPart(_ context.Context, cnr cid.ID, parent oid.ID,
...
+				return err
+			})
 		}
-		return err
-	})
+	)
+
+	err := gFunc()
+	if errors.Is(err, apistatus.ErrObjectNotFound) && allowReturnAnyPart {
+		pi.Index = -1
```
roman-khimov, line 88: You're even using it here, can be used by method caller as well.

pkg/services/object/server.go
```diff
@@ -1039,22 +1061,22 @@ func (s *Server) Get(req *protoobject.GetRequest, gStream protoobject.ObjectServ
...
+		payloadOnly:             req.GetBody().GetPayloadOnly(),
+		sendECPartIndInResponse: sendECPartIdxInResponse(req),
 	}, cnrID, objID, reqMD)
 	if err != nil {
 		if !errors.Is(err, apistatus.Error) {
 			err = newBadRequestError(err.Error()) // defer
 		}
 		return s.sendStatusGetResponse(gStream, err, needSignResp)
 	}
-
```
roman-khimov, line 1057: :eyes:

pkg/local_object_storage/engine/ec.go
carpawell: On engine level -1 is never used, node always tries to find an appropriate part (substitute part index if missing), but then it (upper node level) also asks to return anything the engine has

pkg/local_object_storage/engine/ec.go
carpawell: then anything below engine retried with -1, meaning asking to return anything metabase has. but engine is never called with -1, this is less optimal, node always has a number to try at least

roman-khimov:
CHANGELOG

carpawell:
Rebased, fixed small bug, fixed go.sum.

roman-khimov (approved):

PR #4036, opened 2026-06-22
Meta fixes

carpawell:
There is no need to delay for so long while metabase has its own batching. Closes ##4032.

### DIFF
--- a/cmd/neofs-node/meta.go
+++ b/cmd/neofs-node/meta.go
@@ -14,6 +14,7 @@ import (
 	"github.com/nspcc-dev/neo-go/pkg/core/storage/dbconfig"
 	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
 	"github.com/nspcc-dev/neo-go/pkg/encoding/fixedn"
+	blobstorconfig "github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/engine/shard/blobstor"
 	containercore "github.com/nspcc-dev/neofs-node/pkg/core/container"
 	"github.com/nspcc-dev/neofs-node/pkg/core/metachain/gas"
 	netmapcore "github.com/nspcc-dev/neofs-node/pkg/core/netmap"
@@ -254,6 +255,10 @@ func initMeta(c *cfg) {
 		c.sidechain.Stop()
 	})
 
+	var batchFlushInterval = blobstorconfig.DefaultFlushInterval
+	if c.appCfg.Storage.Default.Blobstor.FlushInterval > 0 {
+		batchFlushInterval = c.appCfg.Storage.Default.Blobstor.FlushInterval
+	}
 	c.metaService, err = meta.New(meta.Parameters{
 		Logger: c.log.With(zap.String("component", "metadata service")),
 		Chain:  c.sidechain,
@@ -265,6 +270,7 @@ func initMeta(c *cfg) {
 			network:    c.netMapSource,
 			header:     c.cfgObject.getSvc,
 		},
+		FlushBatchInterval: batchFlushInterval,
 	})
 	fatalOnErr(err)
 
--- a/pkg/services/meta/storage.go
+++ b/pkg/services/meta/storage.go
@@ -16,7 +16,7 @@ func (m *Meta) PutObjects(o []*object.Object) error {
 
 func (m *Meta) storager(ctx context.Context, buff <-chan storageTask) {
 	ch := make(chan *object.Object, cap(buff))
-	go batchWriter(ctx, m.l, m.metabase, ch)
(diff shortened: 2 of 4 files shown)


pkg/services/meta/storage.go
```diff
@@ -47,46 +36,9 @@ func (m *Meta) storager(ctx context.Context, buff <-chan storageTask) {
...
-	const (
-		maxBuffSize   = 100
-		maxBatchDelay = time.Second
-	)
-	var (
-		t          = time.NewTicker(maxBatchDelay)
-		batch      = make([]*object.Object, 0, maxBuffSize)
-		writeBatch = func() {
-			err := m.PutBatch(batch)
+			err = m.metabase.Put(n.o)
```
roman-khimov: It just occurred to me that you can process objects here one by one in which case BoltDB batching doesn't help at all. Is that the case?

carpawell: > process objects here one by one

in this exact routine -- yes, notifications are the source of these objects. i dont know what else can be done there, any idea we considered has cons

roman-khimov: PutBatch is the only thing that works with the current DB then.

roman-khimov: Of course it has to be more aggressive in timing in this case (take the batching time from shard, it's rather small)

carpawell: this requires external batch collecting, that makes this PR irrelevant (or we need to adjust timings)

carpawell: > take the batching time from shard, it's rather small

made it 10ms

roman-khimov: Then we're back to the question of why things are failing now. Also currently it doesn't follow engine config.

carpawell: > Also currently it doesn't follow engine config.

do you mean it should use the same value, not just hardcode the same default?

carpawell: i can suggest reusing [this const](https://github.com/nspcc-dev/neofs-node/blob/47d122f122f348bbbd58e81004b2286b3fe0cde8/cmd/neofs-node/config/engine/shard/blobstor/blobstor.go#L24)

carpawell: > Then we're back to the question of why things are failing now

fixed split objects, have no idea how that happened, latest SN PR did not break it

roman-khimov: > do you mean it should use the same value, not just hardcode the same default?

Sure.

carpawell: done

pkg/services/meta/storage.go
```diff
@@ -55,7 +56,7 @@ func (m *Meta) storager(ctx context.Context, buff <-chan storageTask) {
 func batchWriter(ctx context.Context, l *zap.Logger, m *metabase.DB, buff <-chan *object.Object) {
 	const (
 		maxBuffSize   = 100
-		maxBatchDelay = time.Second
+		maxBatchDelay = blobstorconfig.DefaultFlushInterval
```
roman-khimov: Not following real configuration.

roman-khimov (approved):

pkg/services/meta/storage.go
carpawell: done

cmd/neofs-node/meta.go
```diff
@@ -254,6 +255,10 @@ func initMeta(c *cfg) {
 		c.sidechain.Stop()
 	})
 
+	var batchFlushTheshold = blobstorconfig.DefaultFlushInterval
+	if c.appCfg.Storage.Default.Blobstor.CombinedSizeThreshold > 0 {
+		batchFlushTheshold = time.Duration(c.appCfg.Storage.Default.Blobstor.CombinedSizeThreshold)
```
roman-khimov: What's the relation between size and time?

pkg/services/meta/search_test.go
```diff
@@ -33,10 +34,11 @@ func (n *singleObjectMeta) Put(obj *object.Object) error {
...
 	m, err := New(Parameters{
-		Logger:  zaptest.NewLogger(t),
-		Chain:   &sidechain.SideChain{},
-		Path:    t.TempDir(),
-		Network: noopNetwork{},
+		Logger:              zaptest.NewLogger(t),
+		Chain:               &sidechain.SideChain{},
+		Path:                t.TempDir(),
+		Network:             noopNetwork{},
+		FlushBatchThreshold: 10 * time.Millisecond,
```
roman-khimov: Should have the default one to avoid this specification.

carpawell: ok

cmd/neofs-node/meta.go
carpawell: whoops

PR #4042, opened 2026-06-24
Fix split streaming

End-rey:
Closes #4041.

PR #4043, opened 2026-06-24
Fix split streaming 2

End-rey:
Ref #4042.

### DIFF
--- a/pkg/services/object/get/exec.go
+++ b/pkg/services/object/get/exec.go
@@ -299,30 +299,57 @@ func (exec *execCtx) headOnly() bool {
 	return exec.head
 }
 
+// childFetchCtx is an immutable snapshot used by prefetch goroutines.
+type childFetchCtx struct {
+	svc *Service
+
+	prm RangePrm
+	cnr cid.ID
+
+	payloadOnly bool
+	legacyRange bool
+
+	log *zap.Logger
+}
+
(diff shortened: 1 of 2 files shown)


pkg/services/object/get/exec.go
```diff
@@ -299,30 +299,57 @@ func (exec *execCtx) headOnly() bool {
...
+
+		log: exec.log,
+	}
+}
+
 // fetchChildStream fetches a single physical child stream. Virtual children are
 // returned as statusVIRTUAL for the caller to handle.
-func (exec *execCtx) fetchChildStream(ctx context.Context, id oid.ID, rng *object.Range) (*object.Object, io.ReadCloser, statusError) {
-	log := exec.log
+func (c childFetchCtx) fetchChildStream(ctx context.Context, id oid.ID, rng *object.Range) (*object.Object, io.ReadCloser, statusError) {
```
roman-khimov, line 331: Have you considered just making it `func (exec execCtx)` (although some might argue that it mixes pointer/non-pointer methods)?

roman-khimov (approved):

pkg/services/object/get/exec.go
End-rey: A value receiver would fix the aliasing, but `childFetchCtx` makes the prefetch concurrency boundary explicit and prevents accidentally pulling mutable `execCtx` state back in later.

PR #4054, opened 2026-06-30
blobstor: drop compressor

roman-khimov:
1. It's very error-prone.
2. It's very ineffective for ranged requests.
3. It spends quite some CPU and memory.


cmd/neofs-node/config/engine/shard/shard.go
```diff
@@ -8,16 +8,12 @@ import (
...
 	metabaseconfig "github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/engine/shard/metabase"
 	writecacheconfig "github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/engine/shard/writecache"
-	"github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/internal"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/shard/mode"
 )
 
 // ShardDetails contains configuration for a single shard of a storage node.
 type ShardDetails struct {
-	Mode                           mode.Mode `mapstructure:"mode"`
-	Compress                       *bool     `mapstructure:"compress"`
```
carpawell, line 18: this hard? no deprecation window? like that

pkg/local_object_storage/blobstor/fstree/fstree_test.go
```diff
@@ -30,30 +29,10 @@ func TestAddressToString(t *testing.T) {
...
-		comp := compression.Config{Enabled: true}
-		require.NoError(t, comp.Init())
-
-		fst := setupFSTree(t)
-		fst.SetCompressor(&comp)
-
-		testGetRangeStream(t, fst)
-	})
-
 	testGetRangeStream(t, setupFSTree(t))
```
carpawell, line 32: likely this func only exists for two versions of tests (compressed and raw data), can be inlined now. same in other places

carpawell (approved):

carpawell:
Conflicts.

pkg/local_object_storage/blobstor/fstree/fstree_test.go
roman-khimov: That's true, but I don't see a lot of value in inlining now.

cmd/neofs-node/config/engine/shard/shard.go
roman-khimov: Yeah, this only affects testnet/mainnet, so can be adopted easily, no other network has it enabled.

End-rey (approved):

PR #4056, opened 2026-07-01
sn: fix session v2 tokens for container+eACL RPC

carpawell:
Fix's reasoning: simple, clear for review, not a magic one, AND it only skips `checkSessionIssuer` which is the only container owner check inside. Session v2 cannot be issued with different original issuers for container creation and eacl setting, and (in practice) should always have a wildcard container. It is still possible to find strange cases when a token was reissued (has a chain) with a more narrow context, and setEacl is impossible by the last part of token's chain; when container creation was issued for _a certain container ID_, and there is no setEacl for this container ID; but these are not practical cases.
@roman-khimov, @End-rey

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Changelog for NeoFS Node
 ### Added
 
 ### Fixed
+- Session v2 token was not supported in the new container + eACL RPC (#4056)
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


roman-khimov (approved):
For container put specifically this seems reasonable.

End-rey (approved):

PR #4058, opened 2026-07-03
object: skip stream read for buffered GET payload

### LINKED ISSUE #3888: Check whether payload is buffered by payload length header in GET
I think this case can be detected from protobuf we have in the buffer, it has payload length and either it's all in the same buffer or not.

_Originally posted by @roman-khimov in https://github.com/nspcc-dev/neofs-node/pull/3882#discussion_r2959788789_
            
see TODO

End-rey:
Closes #3888.

### DIFF
--- a/pkg/services/object/server.go
+++ b/pkg/services/object/server.go
@@ -1129,17 +1129,26 @@ func (s *Server) Get(req *protoobject.GetRequest, gStream protoobject.ObjectServ
 
 func (s *Server) copyGetStream(gStream grpc.ServerStream, hdrRespBuf *iprotobuf.MemBuffer, hdrBuf []byte,
 	prefixLen, hdrTo int, stream io.Reader, pldFldOff int, needSignResp bool) error {
-	var chunkRespBuf *iprotobuf.MemBuffer
-	var chunkBuf []byte
+	var (
+		chunkRespBuf        *iprotobuf.MemBuffer
+		chunkBuf            []byte
+		bufferedPldLen      int
+		bufferedPldTagLen   int
+		fullPayloadBuffered bool
+	)
 
 	prereadPldLen := prefixLen - pldFldOff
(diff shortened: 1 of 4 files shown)


pkg/services/object/server.go
```diff
@@ -1133,13 +1133,20 @@ func (s *Server) copyGetStream(gStream grpc.ServerStream, hdrRespBuf *iprotobuf.
 	var chunkBuf []byte
 
 	prereadPldLen := prefixLen - pldFldOff
+	var bufferedPldLen int
+	var bufferedPldTagLen int
+	var fullPayloadBuffered bool
```
roman-khimov: Have a `var` block then and put earlier definitions (like chunkBuf) there as well.

pkg/services/object/server.go
```diff
@@ -1179,6 +1186,25 @@ func (s *Server) copyGetStream(gStream grpc.ServerStream, hdrRespBuf *iprotobuf.
...
+			}
+			bodyf.To += n
+		}
+
+		chunkRespBuf.SetBounds(bodyf.From, bodyf.To)
+		if err := gStream.SendMsg(chunkRespBuf); err != nil {
+			return fmt.Errorf("%w: %w", getsvc.ErrResponseStreamFailure, err)
+		}
+		sent += bufferedPldLen
+		return nil
```
roman-khimov, line 1207: Close `stream`?

roman-khimov (commented):
Changelog can mention this as optimization as well.

pkg/services/object/server.go
End-rey: It's closed by the caller's defer.

pkg/services/object/server.go
End-rey: Done.

roman-khimov (approved):

Issue #4057, opened 2026-07-02
Add __NEOFS__EC_PART_HASHES into EC parts, use it for verifications


roman-khimov:
## Is your feature request related to a problem? Please describe.
We've got https://github.com/nspcc-dev/neofs-api/pull/351 specified, but it seems to be unused.

## Describe the solution you'd like
1. Add this attribute to parent header.
2. Use it to check EC part objects _instead_ of part signature. Parts go unsigned, but because we know all of their payload hashes from confirmed parent header we can easily calculate all the proper OIDs of parts for this object (or the one that needs to be verified).

PR #4059, opened 2026-07-03
sn/meta: submit only parent object for EC split

carpawell:


carpawell:
@roman-khimov, taking into account [this](https://github.com/nspcc-dev/neofs-node/issues/4057), it means there will be as many objects in meta as there are small splits for REPs _plus_ an object for _every_ EC rule.

carpawell:
Would like to wait for full tests.

pkg/services/object/put/streamer.go
```diff
@@ -319,6 +319,22 @@ func (p *Streamer) newCommonTarget(prm *PutInitPrm) internal.Target {
...
+func metaRequired(metaAttrValue string, containerNodes ContainerNodes, ecRule iec.PartInfo) bool {
+	if metaAttrValue == "" {
+		return false
+	}
+	if ecRule.RuleIndex < 0 {
+		// not an EC part, meta is required
+		return true
+	}
+
+	// submit meta only for last EC part in every rule
```
roman-khimov: A single signature?

pkg/services/object/put/distributed.go
```diff
@@ -642,7 +642,7 @@ func (t *distributedTarget) encodeObjectMetadata(obj object.Object) (*transactio
 	default:
 	}
 
-	tx, dataToSign := objectcore.EncodeChainMetaInfo(len(t.containerNodes.PrimaryCounts()), obj.GetContainerID(), obj.GetID(), firstObj, obj.GetPreviousID(),
+	tx, dataToSign := objectcore.EncodeChainMetaInfo(len(t.containerNodes.PrimaryCounts())+len(t.containerNodes.ECRules()), obj.GetContainerID(), obj.GetID(), firstObj, obj.GetPreviousID(),
```
cthulhu-rider: how was this problem discovered?

pkg/services/object/server.go
```diff
@@ -2215,15 +2210,21 @@ func (s *Server) metaInfoSignature(o object.Object) ([]byte, error) {
 	default:
 	}
 
+	_, repRules, ecRules, err := s.fsChain.SelectContainerNodes(o.GetContainerID())
```
cthulhu-rider: couldn't they have been taken from `cnr`?

pkg/services/object/put/streamer.go
cthulhu-rider: what's so special about the last parity part?

pkg/services/object/server.go
```diff
@@ -2214,17 +2214,39 @@ func (s *Server) metaInfoSignature(o object.Object) ([]byte, error) {
...
-	firstBlock := (uint64(currentBlock)/currentEpochDuration + 1) * currentEpochDuration
-	secondBlock := firstBlock + currentEpochDuration
-	thirdBlock := secondBlock + currentEpochDuration
-	placementVectorsNum := len(repRules) + len(ecRules)
-
-	_, firstMeta := objectcore.EncodeChainMetaInfo(placementVectorsNum, o.GetContainerID(), o.GetID(), firstObj, prevObj, o.PayloadSize(), typ, deleted, locked, firstBlock, s.meta.MagicNumber())
-	_, secondMeta := objectcore.EncodeChainMetaInfo(placementVectorsNum, o.GetContainerID(), o.GetID(), firstObj, prevObj, o.PayloadSize(), typ, deleted, locked, secondBlock, s.meta.MagicNumber())
-	_, thirdMeta := objectcore.EncodeChainMetaInfo(placementVectorsNum, o.GetContainerID(), o.GetID(), firstObj, prevObj, o.PayloadSize(), typ, deleted, locked, thirdBlock, s.meta.MagicNumber())
+	var (
+		currentBlock         = s.meta.Height()
```
cthulhu-rider, line 2244: so much code shifts just to pack into `var ()`. I'd do such changes in separate commits

pkg/services/object/server.go
```diff
@@ -2214,17 +2214,39 @@ func (s *Server) metaInfoSignature(o object.Object) ([]byte, error) {
...
+			parentID          = par.GetID()
+			zeroObj           oid.ID
+		)
+		_, firstMeta = objectcore.EncodeChainMetaInfo(placementVectorsNum, cID, parentID, zeroObj, zeroObj, parentPayloadSize, typ, zeroObj, zeroObj, firstBlock, magic)
+		_, secondMeta = objectcore.EncodeChainMetaInfo(placementVectorsNum, cID, parentID, zeroObj, zeroObj, parentPayloadSize, typ, zeroObj, zeroObj, secondBlock, magic)
+		_, thirdMeta = objectcore.EncodeChainMetaInfo(placementVectorsNum, cID, parentID, zeroObj, zeroObj, parentPayloadSize, typ, zeroObj, zeroObj, thirdBlock, magic)
+	} else {
+		_, firstMeta = objectcore.EncodeChainMetaInfo(placementVectorsNum, cID, oID, firstObj, prevObj, payloadSize, typ, deleted, locked, firstBlock, magic)
+		_, secondMeta = objectcore.EncodeChainMetaInfo(placementVectorsNum, cID, oID, firstObj, prevObj, payloadSize, typ, deleted, locked, secondBlock, magic)
+		_, thirdMeta = objectcore.EncodeChainMetaInfo(placementVectorsNum, cID, oID, firstObj, prevObj, payloadSize, typ, deleted, locked, thirdBlock, magic)
```
cthulhu-rider, line 2276: 6 very similar lines, so it's hard to me to see the diff. Suggest to either func out or conditionally parameterize args before

pkg/services/object/server.go
```diff
@@ -2214,17 +2214,39 @@ func (s *Server) metaInfoSignature(o object.Object) ([]byte, error) {
...
+		cID         = o.GetContainerID()
+		magic       = s.meta.MagicNumber()
+		payloadSize = o.PayloadSize()
+
+		firstMeta, secondMeta, thirdMeta []byte
+		placementVectorsNum              = len(repRules) + len(ecRules)
+	)
+	if mustSignParent(o, ecRules) {
+		par := o.Parent()
+		if par == nil {
```
cthulhu-rider, line 2262: inobvious requirement for `mustSignParent` condition. Suggest to call it as is `isLastECPart`

carpawell: reworked (and had to rename your way in the new version)

pkg/services/object/server.go
carpawell: they are almost the same calls in every `if-else` branch, yes, only one arg and returned value differ. but they do different things in different branches, so i prefer to keep it this way, not `for` loops, easy to read for me

pkg/services/object/server.go
carpawell: i needed a new block of vars here so it is either a new `var ()` and a separate commit that merges them, or a new block for everything from the beginning. i dont care mush about `var := var := var :=` or a single `var ()` but i prefer to make it similar with others, i know @roman-khimov prefer `var ()`

pkg/services/object/put/streamer.go
carpawell: reworked, data parts + parity parts number of signs is used now

pkg/services/object/server.go
carpawell: reverted but both versions should be cached, so i see no differences

pkg/services/object/put/distributed.go
carpawell: in the updated version i would say this is just part of an updated understanding of what EC info should be put to metadata chain, not a fix

pkg/services/object/put/streamer.go
carpawell: reworked

roman-khimov (approved):

roman-khimov:
Conflicts.

pkg/innerring/metadata.go
```diff
@@ -135,11 +131,19 @@ func replicas(policy netmap.PlacementPolicy) []uint32 {
 		return initial.ReplicaLimits()
 	}
 
-	netmapReps := policy.Replicas()
-	res := make([]uint32, len(netmapReps))
-	for i := range netmapReps {
+	var (
+		reps    = policy.Replicas()
+		repsLen = len(reps)
```
cthulhu-rider: we could set it to `policy.NumberOfReplicas()`, so `reps` wouldnt be needed

pkg/services/object/put/meta.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/nspcc-dev/neo-go/pkg/core/transaction"
+	objectcore "github.com/nspcc-dev/neofs-node/pkg/core/object"
+	"github.com/nspcc-dev/neofs-sdk-go/object"
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	"go.uber.org/zap"
+)
+
+func (t *distributedTarget) encodeMetaIfRequested(obj object.Object) error {
+	var err error
+	if t.metaCollection == nil || !t.localNodeInContainer {
```
cthulhu-rider: can we get here when `t.localOnly` is true? Seems like this differs from check in `t.submitMetaCollection`

in total, i suggest to review all meta-or-not conditions and reduce them to smth simple. For example, in streamer.go field where we init `t.MetaCollection` field (other cond there). Or already mentioned submission method. Doable?

pkg/services/object/server.go
```diff
@@ -2282,6 +2308,18 @@ func (s *Server) metaInfoSignature(o object.Object) ([]byte, error) {
...
+// ecPart checks if EC exception for parent signature instead of
+// the object itself is applicable to this object.
+func ecPart(o object.Object, ecRules []netmap.ECRule) bool {
+	if len(ecRules) == 0 {
+		return false
+	}
+
+	// this is the latest part of request handling, EC attributes must be valid
+	pi, _ := iec.GetPartInfo(o)
+	return pi.RuleIndex >= 0
```
cthulhu-rider, line 2323: zero happens on `iec.GetPartInfo` error. Execution should not be continued in this case

pkg/services/object/put/meta.go
```diff
@@ -0,0 +1,111 @@
...
+	typ := obj.Type()
+	switch typ {
+	case object.TypeTombstone:
+		deletedObj = obj.AssociatedObject()
+	case object.TypeLock:
+		lockedObj = obj.AssociatedObject()
+	default:
+	}
+
+	tx, dataToSign := objectcore.EncodeChainMetaInfo(len(t.containerNodes.PrimaryCounts())+len(t.containerNodes.ECRules()), obj.GetContainerID(), obj.GetID(), firstObj, obj.GetPreviousID(),
```
cthulhu-rider, line 49: just wondering is it possible to teach the contract itself to understand how many signatures need to be collected? Seems like we always pass REP+EC number, right?

pkg/innerring/metadata.go
carpawell: funny that for server-side I decided to do this exactly your way. fixed

pkg/services/object/put/meta.go
carpawell: > can we get here when t.localOnly is true?

yes. i dont know why it may happen, such requests will fail in most cases (exceptions are REP 1 with local node included in a container). but it is a broader problem for me of PUTing requests with localOnly. but client decided to do it for some reason, so we will try

> reduce them to smth simple

made metacollection when `distributedTarget` is created and used only its pointer as the only condition for meta submission

pkg/services/object/server.go
carpawell: i may add a check there, but already added a comment that at this step it should have been checked already. a may add a panic there, but it is a repetative check anyway

pkg/services/object/put/meta.go
carpawell: we pass it to be included in the verification script (and make an account based on it). it should be used as a PACK's arg for [verification script](https://github.com/nspcc-dev/neofs-node/blob/612aa28c6c230aad9f17b252094c647fd10c912d/pkg/core/object/replicate.go#L44), so its number is a const for any container, and every container's placement vector (a rule in more recent terms) should be checked independently.

verification check contract code: https://github.com/nspcc-dev/neofs-node/blob/612aa28c6c230aad9f17b252094c647fd10c912d/pkg/core/metachain/meta/method_objects.go#L299-L343

pkg/services/object/server.go
cthulhu-rider: hmm, now i understand the comment. Then ideally we should pre-parse EC info and pass it both to fmt validator and this func. Or return EC info from fmt validator Code would be safer and no repetitive actions. But this requires much more code changes and the optimization wouldnt be felt

so, i'd panic for now

pkg/services/object/put/meta.go
cthulhu-rider: > PACK's arg for [verification script](https://github.com/nspcc-dev/neofs-node/blob/612aa28c6c230aad9f17b252094c647fd10c912d/pkg/core/object/replicate.go#L44)

so it cannot be accessed from within contract method. Thx for explanation

pkg/services/object/put/streamer.go
```diff
@@ -310,14 +281,53 @@ func (p *Streamer) newCommonTarget(prm *PutInitPrm) internal.Target {
...
+	if !localNodeInContainer {
+		// if node is not container's part, it will send objects using PUT,
+		// not REPLICATE, it will be other nodes' responsibility to submit
+		// meta information
+		return nil
+	}
+
+	cnrNodes := containerNodes.Unsorted()
+	for _, vector := range cnrNodes {
+		slices.SortFunc(vector, func(a, b netmap.NodeInfo) int {
```
cthulhu-rider, line 305: just noticed this sorting. `vector` is [cached](https://github.com/nspcc-dev/neofs-node/blob/55e09238e848954a79dab37f6814d1c00a0ac22d/pkg/services/object/put/service.go#L85), so sorting it directly is very unsafe. We should definitely revise this. Unrelated though

carpawell: hm, that is bad, added commit that clones unsorted nodes

pkg/services/object/server.go
carpawell: > pre-parse EC

too problematic currently; validator is an interface and returning EC info is strange imo

> so, i'd panic for now

no problem

pkg/services/object/put/streamer.go
cthulhu-rider: nice, at leas code is safe now

for the future: do we have any options to avoid these extra allocs?

cthulhu-rider (approved):

PR #4067, opened 2026-07-07
Replace RANGE with GET and improve in-container GET/HEAD

cthulhu-rider:
i'd like to drop `GetRange` support completely. Any objections?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Changelog for NeoFS Node
 - Optimized GET for fully buffered payloads (#4058)
 - Proxied GET data is not checked against SHA256 hash from header (#4081)
 - SN requests EC part ranges using GET API instead of deprecated RANGE one (#4067)
+- In-container SN no longer attaches session/bearer tokens to spawned GET/HEAD requests (#4067)
 
 ### Removed
 - Compression support from FSTree (#4054)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Changelog for NeoFS Node
 ### Changed
 - Optimized GET for fully buffered payloads (#4058)
 - Proxied GET data is not checked against SHA256 hash from header (#4081)
+- SN requests EC part ranges using GET API instead of deprecated RANGE one (#4067)
 
 ### Removed
 - Compression support from FSTree (#4054)
--- a/pkg/services/object/get/util.go
+++ b/pkg/services/object/get/util.go
@@ -17,7 +17,6 @@ import (
 	"github.com/nspcc-dev/neofs-sdk-go/netmap"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
-	"github.com/nspcc-dev/neofs-sdk-go/session"
 	sessionv2 "github.com/nspcc-dev/neofs-sdk-go/session/v2"
 	"github.com/nspcc-dev/neofs-sdk-go/user"
 )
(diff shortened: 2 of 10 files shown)


roman-khimov:
> i'd like to drop GetRange support completely. Any objections?

Not yet. It's still used.

pkg/services/object/get/util.go
```diff
@@ -487,6 +492,21 @@ func (c *clientCacheWrapper) InitGetObjectStream(ctx context.Context, node netma
 	// TODO: SkipChecksumVerification() turns off checking all object checksums. Better to keep checking
 	//  OID against header and payload checksum.
 
+	if rng != nil {
+		b := []byte{0}
+		if _, err = io.ReadFull(rc, b); err != nil {
```
roman-khimov, line 463: Why this one-byte read?

pkg/services/object/acl/v2/service.go
```diff
@@ -414,14 +414,36 @@ func (b Service) verifyBearerTokenAgainstRequest(token bearer.Token, reqCnr cid.
...
 // [ACLChecker].
-func (b Service) GetRequestToInfo(request *protoobject.GetRequest, cnr cid.ID, tokens common.RequestTokens) (RequestInfo, error) {
-	return b.findRequestInfo(request, cnr, acl.OpObjectGet, tokens)
+func (b Service) GetRequestToInfo(request *protoobject.GetRequest, cnr cid.ID, tokens common.RequestTokens) (RequestInfo, bool, error) {
+	reqInfo, err := b.findRequestInfo(request, cnr, acl.OpObjectGet, tokens)
+	if err != nil {
+		return RequestInfo{}, false, err
+	}
+
+	// TODO: consider merging with client check
```
roman-khimov: Let's do this right away, we've already done #4007, so this looks like a regression now.

pkg/services/object/server.go
```diff
@@ -1038,6 +1043,11 @@ func (s *Server) Get(req *protoobject.GetRequest, gStream protoobject.ObjectServ
 		recheckEACL = true
 	}
 
+	if serverInContainer {
+		// in-container node does not need tokens for processing
+		reqMD.tokens = common.RequestTokens{}
```
roman-khimov: Can't this be decided when forwarding? This can be problematic is any other code assumes info to always have all the data.

pkg/services/object/get/util.go
cthulhu-rider: it was done before, and kept. Now it's needed to check status error. This could be tidied up a bit, will see

pkg/services/object/server.go
cthulhu-rider: yes for sure. Code will be complicated, but i'll try to make this change along with the others

pkg/services/object/get/util.go
carpawell: do we need this check? what is the difference b/w returning a status error from `InitGetObjectStream` and returning a status from the reader later?

cthulhu-rider: reviewed this one more time and decided to keep as is. Callers expect status error to be returned as `error`, not after first read

> returning a status from the reader later

issued this

pkg/services/object/acl/v2/service.go
cthulhu-rider: as it turned out, this part is not needed at all

pkg/services/object/server.go
cthulhu-rider: as it turns out, new requests are sent ONLY from container nodes, so this is fixed everywhere before calling the API

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Changelog for NeoFS Node
 ### Changed
 - Optimized GET for fully buffered payloads (#4058)
 - SN requests EC part ranges using GET API instead of deprecated RANGE one (#4067)
+- - In-container SN no longer attaches session/bearer tokens to spawned GET/HEAD requests (#4067)
```
roman-khimov: Double dash.

roman-khimov (approved):
Much better this way.

Issue #4071, opened 2026-07-08
GET does not stream meta headers


carpawell:
I am not sure what exacly should be done.

## Expected Behavior
Meta headers are defined in API, they should be streamed.

## Current Behavior
They are not streamed.

## Possible Solution
Attache them.

## Steps to Reproduce (for bugs)
Look at:
https://github.com/nspcc-dev/neofs-node/blob/5df4d7b7cb07ee80a99c422eb96b8d70cc12b6db/pkg/services/object/server.go#L967-L999 for local buffered cases

and https://github.com/nspcc-dev/neofs-node/blob/5df4d7b7cb07ee80a99c422eb96b8d70cc12b6db/pkg/services/object/proto.go#L129 usage (HEAD only) for EC object collecting cases.

## Context
https://github.com/nspcc-dev/neofs-api/issues/397.

## Regression
I think yes, before buffering era, they supposed to be filled for every response, but did not check.

cthulhu-rider:
if nothing is broken, maybe this is redundant data?

carpawell:
API issue should resolved this way before closing this issue then.

roman-khimov:
> if nothing is broken, maybe this is redundant data?

Looks like. Let's review the API definitions, `ResponseMetaHeader` can just be optional.

roman-khimov:
Still relevant? https://github.com/nspcc-dev/neofs-api/issues/397 is solved.

carpawell:
Obsolete considering https://github.com/nspcc-dev/neofs-node/pull/4072, https://github.com/nspcc-dev/neofs-node/pull/4110 and https://github.com/nspcc-dev/neofs-api/issues/397.

PR #4072, opened 2026-07-08
sn: do not respond with epoch in metaheader

carpawell:
<s>This affects only
1. any REP objects;
2. EC parts;
3. EC full object with RANGE.
This does not affect case implemented in #3996 (Get service's Prm with `WithECTransport()` called). Refs #4071.</s>

### DIFF
--- a/pkg/services/object/proto.go
+++ b/pkg/services/object/proto.go
@@ -127,24 +127,12 @@ func (s *Server) signResponse(buf, body, metaHdr []byte) (int, error) {
 }
 
 func (s *Server) writeMetaHeaderToResponseBuffer(buf []byte) (int, int) {
-	epoch := s.fsChain.CurrentEpoch()
-
 	ln := len(currentVersionResponseMetaHeader)
-	if epoch > 0 {
-		ln += 1 + protowire.SizeVarint(epoch)
-	}
 
 	buf[0] = iprotobuf.TagBytes2
 	off := 1 + binary.PutUvarint(buf[1:], uint64(ln))
-
 	valOff := off
-
 	off += copy(buf[off:], currentVersionResponseMetaHeader)
-
-	if epoch > 0 {
-		buf[off] = iprotobuf.TagVarint2
-		off += 1 + binary.PutUvarint(buf[off+1:], epoch)
-	}
 
 	return valOff, off
 }
--- a/pkg/services/session/server.go
+++ b/pkg/services/session/server.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/google/uuid"
 	icrypto "github.com/nspcc-dev/neofs-node/internal/crypto"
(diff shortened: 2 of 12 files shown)


roman-khimov (dismissed):
Something to tell about in chlog?

carpawell:
PR adds a metaheader to every message. This is how i read the spec currently. But i think @cthulhu-rider thinks differently. Once we decide i will update chlog.

carpawell:
Attaching version is OK to me now.

But if we want to "Must be attached only if server's version is lower than client's one" then it will require more adoption: current code is just a number of helper functions that always attach the version; there is no local request in these funcs.

roman-khimov (approved):

pkg/services/session/server.go
```diff
@@ -51,7 +51,6 @@ func (s *server) makeCreateResponse(body *protosession.CreateResponse_Body, st *
 		Body: body,
 		MetaHeader: &protosession.ResponseMetaHeader{
 			Version: version.Current().ProtoMessage(),
-			Epoch:   s.net.CurrentEpoch(),
```
cthulhu-rider, line 54: this makes `s.net` unused so it's worth to be deparameterized

pls review other cases. Same can be done or at least narrow the dependency interface

pkg/services/object/proto.go
```diff
@@ -127,25 +127,13 @@ func (s *Server) signResponse(buf, body, metaHdr []byte) (int, error) {
...
-	epoch := s.fsChain.CurrentEpoch()
-
 	ln := len(currentVersionResponseMetaHeader)
-	if epoch > 0 {
-		ln += 1 + protowire.SizeVarint(epoch)
-	}
 
 	buf[0] = iprotobuf.TagBytes2
 	off := 1 + binary.PutUvarint(buf[1:], uint64(ln))
-
```
cthulhu-rider, line 139: separately, although it is hardly necessary

cthulhu-rider (commented):
`Version` behavior does not yet correspond to https://github.com/nspcc-dev/neofs-api/pull/404

pkg/services/object/proto.go
carpawell: did not understand, what exactly is unnecessary?

pkg/services/session/server.go
carpawell: dropped when unused

carpawell:
> Version behavior does not yet correspond to nspcc-dev/neofs-api#404

As i said, i would leave it as is, since it is "optional" in API, so attaching it always is ok to me. But branched returning version.

carpawell:
Left https://github.com/nspcc-dev/neofs-node/pull/4110. Only epoch drop for now.

roman-khimov (approved):

PR #4074, opened 2026-07-08
sidechain: change pooling time to 1s

carpawell:
In most cases meta chain will see fractions of seconds as a config, that is too small period that make logs full of:
```
info blockchain/blockchain.go:365 waiting for synchronization with the blockchain network... {"component": "metadata chain (IR)"}
info sidechain/sidechain.go:116 waiting for synchronization with the blockchain network... {"component": "metadata chain (SN)"}
```

### DIFF
--- a/pkg/services/sidechain/sidechain.go
+++ b/pkg/services/sidechain/sidechain.go
@@ -110,7 +110,7 @@ func (s *SideChain) Run(ctx context.Context) error {
 	go s.core.Run()
 	go s.netServer.Start()
 
-	t := time.NewTicker(s.core.GetConfig().Genesis.TimePerBlock)
+	t := time.NewTicker(time.Second)
 
 	for {
 		s.logger.Info("waiting for synchronization with the blockchain network...")
--- a/pkg/innerring/internal/blockchain/blockchain.go
+++ b/pkg/innerring/internal/blockchain/blockchain.go
@@ -359,7 +359,7 @@ func (x *Blockchain) Run(ctx context.Context) (err error) {
 	go x.netServer.Start()
 
 	// await synchronization with the network
-	t := time.NewTicker(x.core.GetConfig().TimePerBlock)
+	t := time.NewTicker(time.Second)
 
 	for {
 		x.logger.Info("waiting for synchronization with the blockchain network...")


pkg/services/sidechain/sidechain.go
```diff
@@ -110,7 +110,7 @@ func (s *SideChain) Run(ctx context.Context) error {
 	go s.core.Run()
 	go s.netServer.Start()
 
-	t := time.NewTicker(s.core.GetConfig().Genesis.TimePerBlock)
+	t := time.NewTicker(time.Second)
 
 	for {
```
roman-khimov, line 115: What's the purpose of this loop?

carpawell: notify if we are in sync. the same way for FSchain: https://github.com/nspcc-dev/neofs-node/blob/5df4d7b7cb07ee80a99c422eb96b8d70cc12b6db/pkg/innerring/internal/blockchain/blockchain.go#L364-L375. how do i know there is a problem if not this way? you saw logs, they told you smth was wrong

roman-khimov: Usually `netServer` has something to say on its own when it's in sync.

roman-khimov: Then this can be also changed for FS chain.

carpawell: so drop both loops? for neofs-dev it is clearer than neo-go internal logs

carpawell: i often searched for `blockchain state successfully synchronized` to be sure everything was ok

roman-khimov: To me they're useless, if you're running a netServer it will log the moment it detects sync state.

carpawell: neo-go's logger can be suppressed, it can be changed independently. current solution is neofs app level to me

roman-khimov: I still think it's useless, but if you need it we can leave it. It has to be consistent with FS chain though.

pkg/services/sidechain/sidechain.go
```diff
@@ -110,7 +110,7 @@ func (s *SideChain) Run(ctx context.Context) error {
 	go s.core.Run()
 	go s.netServer.Start()
 
-	t := time.NewTicker(s.core.GetConfig().Genesis.TimePerBlock)
+	t := time.NewTicker(time.Second)
```
cthulhu-rider, line 113: if frequent log records is the only problem, i'd just move `waiting...` record before loop

carpawell: as for me: if something is wrong, it would be better to repeat it until smth is done

cthulhu-rider: as i understand this is a background job. If observer does not care about `synchronized` message, it does not matter how many `waiting...` there was. If cares, again does not matter

overall, any spam is annoying, so im ok with this change

cthulhu-rider (approved):

pkg/services/sidechain/sidechain.go
carpawell: > as i understand this is a background job

yes it is. meta was enabled and configured incorrectly, there are no blocks/objects on the node, and this log is the only way to understand that smth must be done (ok, @roman-khimov, also says that there is _smth_ inside neo-go that would also tell you that there are no new blocks, but i do not want to rely on it)

PR #4079, opened 2026-07-09
fstree: add compressed object decompression tool

End-rey:


pkg/local_object_storage/blobstor/fstree/rewrite_compressed.go
```diff
@@ -0,0 +1,341 @@
...
+	id   oidsdk.ID
+	data []byte
+}
+
+// RewriteCompressed scans FSTree physical files, decompresses zstd-compressed
+// object bytes and stores canonical uncompressed object bytes back.
+//
+// The method preserves combined physical files by rewriting all hard links to
+// the same new combined file. The storage node must not write to this FSTree
+// while the migration is running.
```
roman-khimov: Wrong assumption, it will. And the idea is to let it do so.

pkg/local_object_storage/blobstor/fstree/rewrite_compressed.go
```diff
@@ -0,0 +1,341 @@
...
+func (t *FSTree) RewriteCompressed() (RewriteCompressedStats, error) {
+	if t.readOnly {
+		return RewriteCompressedStats{}, common.ErrReadOnly
+	}
+	if err := t.writer.finalize(); err != nil {
+		return RewriteCompressedStats{}, fmt.Errorf("flush pending writes: %w", err)
+	}
+
+	files, err := t.compressedRewriteFiles()
```
roman-khimov: Please don't create these registries. Our mainnet shards have ~16M objects routinely, just listing them is already a problem and it takes a lot of time. You need to do the job while iterating over the tree.

pkg/local_object_storage/blobstor/fstree/rewrite_compressed.go
```diff
@@ -0,0 +1,341 @@
...
+		binary.BigEndian.PutUint32(pref[combinedLengthOff:], uint32(len(parts[i])))
+		data = append(data, pref[:]...)
+		data = append(data, parts[i]...)
+	}
+
+	return data, nil
+}
+
+func rewriteHardlinkedObjectFile(group []rewriteCompressedFile, data []byte) error {
+	tmp, err := os.CreateTemp(filepath.Dir(group[0].path), ".rewrite-compressed-*")
```
roman-khimov: Use O_TMPFILE, just like regular FSTree does.

pkg/local_object_storage/blobstor/fstree/rewrite_compressed.go
```diff
@@ -0,0 +1,341 @@
...
+		return fmt.Errorf("close temporary object file %q: %w", tmpPath, err)
+	}
+
+	for i := range group {
+		linkPath := group[i].path + ".rewrite-compressed"
+		_ = os.Remove(linkPath)
+		if err = os.Link(tmpPath, linkPath); err != nil {
+			return fmt.Errorf("link temporary object file to %q: %w", linkPath, err)
+		}
+		if err = os.Rename(linkPath, group[i].path); err != nil {
```
roman-khimov: Be prepared for the node deleting the same file as you walk over the FSTree.

pkg/local_object_storage/blobstor/fstree/rewrite_compressed.go
```diff
@@ -0,0 +1,341 @@
...
+	data []byte
+}
+
+// RewriteCompressed scans FSTree physical files, decompresses zstd-compressed
+// object bytes and stores canonical uncompressed object bytes back.
+//
+// The method preserves combined physical files by rewriting all hard links to
+// the same new combined file. The storage node must not write to this FSTree
+// while the migration is running.
+func (t *FSTree) RewriteCompressed() (RewriteCompressedStats, error) {
```
carpawell, line 43: if it is an FSTree method, why do we need this utility then? it can be a background routing for the next release

roman-khimov: Theoretically, yes, I also had this thought, but:
 * either it needs to be synchronized with other activities (deletion first of all to avoid object resurrection) which requires some effort
 * or it keeps the same logic as an outsider, but then having a separate process is easier, can be controlled by shard and can even be performed before the release

pkg/local_object_storage/blobstor/fstree/rewrite_compressed.go
End-rey: The migration can now run while the storage node is running on Linux. On other platforms, the node must be stopped because safe concurrent deletion handling is not available.

pkg/local_object_storage/blobstor/fstree/rewrite_compressed.go
End-rey: Fixed.

pkg/local_object_storage/blobstor/fstree/rewrite_compressed.go
End-rey: Done.

pkg/local_object_storage/blobstor/fstree/rewrite_compressed.go
End-rey: Done.

pkg/local_object_storage/blobstor/fstree/rewrite_compressed.go
```diff
@@ -0,0 +1,210 @@
...
+	"errors"
+	"fmt"
+	"io"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/blobstor/common"
+	oidsdk "github.com/nspcc-dev/neofs-sdk-go/object/id"
```
roman-khimov: Just `oid`, it's never imported as `oidsdk`.

pkg/local_object_storage/blobstor/fstree/rewrite_compressed.go
```diff
@@ -0,0 +1,210 @@
...
+//
+// On Linux, the storage node may write to the FSTree while the migration is
+// running. Other platforms must stop the storage node before migration.
+// Combined physical files are split into individual object files as they are
+// encountered during the walk.
+func (t *FSTree) RewriteCompressed() (RewriteCompressedStats, error) {
+	if t.readOnly {
+		return RewriteCompressedStats{}, common.ErrReadOnly
+	}
+	if err := t.writer.finalize(); err != nil {
```
roman-khimov: Should be irrelevant for the tool (fstree just inited), but doesn't hurt either.

pkg/local_object_storage/blobstor/fstree/rewrite_compressed_linux.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+	if err != nil {
+		return fmt.Errorf("exchange rewritten object file %q: %w", path, err)
+	}
+
+	oldInfo, err := os.Lstat(tmpPath)
+	if err != nil {
+		return fmt.Errorf("stat displaced object file %q: %w", path, err)
+	}
+	if !os.SameFile(source, oldInfo) {
```
roman-khimov: Related to #2563.

pkg/local_object_storage/blobstor/fstree/rewrite_compressed_linux.go
```diff
@@ -0,0 +1,139 @@
...
+	if err != nil {
+		return err
+	}
+	if currentOwner != owner {
+		if err = f.Chown(owner.uid, owner.gid); err != nil {
+			return fmt.Errorf("chown unnamed temporary object file: %w", err)
+		}
+	}
+	if !noSync {
+		if err = f.Sync(); err != nil {
```
roman-khimov: Not needed, O_DSYNC is sufficient.

pkg/local_object_storage/blobstor/fstree/rewrite_compressed.go
```diff
@@ -0,0 +1,210 @@
...
+		return fmt.Errorf("parse object file %q: %w", path, err)
+	}
+
+	st.Scanned++
+	if !isCompressed(entry.data) {
+		st.Skipped++
+		if !combined {
+			return nil
+		}
+		err = rewriteCompressedObjectFile(path, info, owner, entry.data, t.noSync)
```
roman-khimov: Why rewriting here?

pkg/local_object_storage/blobstor/fstree/rewrite_compressed.go
End-rey: Fixed.

pkg/local_object_storage/blobstor/fstree/rewrite_compressed.go
End-rey: Dropped.

pkg/local_object_storage/blobstor/fstree/rewrite_compressed_linux.go
End-rey: Dropped.

pkg/local_object_storage/blobstor/fstree/rewrite_compressed.go
End-rey: Fixed.

roman-khimov (approved):

PR #4081, opened 2026-07-09
object: drop hash check for proxied data

roman-khimov:
It can waste some CPU and limit performance, better checks can be implemented in future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Changelog for NeoFS Node
 
 ### Changed
 - Optimized GET for fully buffered payloads (#4058)
+- Proxied GET data is not checked against SHA256 hash from header (#4081)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Changelog for NeoFS Node
 
 ### Changed
 - Optimized GET for fully buffered payloads (#4058)
+- Proxied GET data is not checked against SHA256 hash from header (#4080)
```
cthulhu-rider:
```suggestion
- Proxied GET data is not checked against SHA256 hash from header (#4081)
```

cthulhu-rider (approved):

CHANGELOG.md
roman-khimov: ok-ok

PR #4082, opened 2026-07-09
sn: cache session tokens authentication results in REPLICATE

carpawell:


pkg/core/time/time.go
```diff
@@ -0,0 +1,10 @@
+package timecore
```
cthulhu-rider: `internal/time` to me

pkg/core/time/time.go
```diff
@@ -0,0 +1,10 @@
...
+
+import "time"
+
+// TimeProvider supplies current FS chain time without calling the chain.
```
cthulhu-rider: interface may have wider applications than just FS chain (even the name hints at it), so I'd give him a more general description

internal/crypto/object.go
```diff
@@ -5,17 +5,60 @@ import (
...
+func newSessionV2Cache() *sessionV2Checker {
+	cache, _ := lru.New[[sha256.Size]byte, error](1024)
+	return &sessionV2Checker{cache: cache}
+}
+
+type sessionV2Checker struct {
+	cache *lru.Cache[[sha256.Size]byte, error]
+}
+
+func (ch sessionV2Checker) authenticateV2Token(s *sessionv2.Token, fsChain HistoricN3ScriptRunner, tp timecore.TimeProvider) error {
```
cthulhu-rider: implementation is close to already existing session cache. I expect single component for all cases

internal/crypto/object.go
```diff
@@ -5,17 +5,60 @@ import (
...
+	k := sha256.Sum256(s.Marshal())
+
+	resErr, ok := ch.cache.Get(k)
+	if !ok {
+		resErr = AuthenticateTokenV2(s, fsChain)
+		defer func() {
```
cthulhu-rider: further panic is possible or why `defer`?

internal/crypto/object.go
```diff
@@ -5,17 +5,60 @@ import (
...
+			ch.cache.Add(k, resErr)
+		}()
+	}
+	if resErr != nil {
+		return resErr
+	}
+
+	currentTime := tp.Now().Round(time.Second)
+	if s.Exp().Before(currentTime) {
```
cthulhu-rider: have these checks been done before?

cthulhu-rider:
missed that PR is still in draft

carpawell:
Reworked.

internal/crypto/object.go
```diff
@@ -6,16 +6,18 @@ import (
 	"errors"
 	"fmt"
 
+	isessions "github.com/nspcc-dev/neofs-node/internal/sessions"
 	versioncore "github.com/nspcc-dev/neofs-node/pkg/core/version"
 	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
 	neofsecdsa "github.com/nspcc-dev/neofs-sdk-go/crypto/ecdsa"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
+	sessionSDK "github.com/nspcc-dev/neofs-sdk-go/session"
```
roman-khimov: Can't be `sessionSDK`.

internal/crypto/object.go
```diff
@@ -55,11 +57,20 @@ func AuthenticateObject(obj object.Object, fsChain HistoricN3ScriptRunner, ecPar
...
 			return errors.New("session token is not for object's signer")
 		}
-		if err := AuthenticateToken(sessionToken, fsChain); err != nil {
-			return fmt.Errorf("session token: %w", err)
-		}
-		if sessionToken.Issuer() != obj.Owner() && versioncore.OwnerSignatureMatchRequired(obj.Version()) {
-			return errors.New("different object owner and session issuer")
+
+		k := sha256.Sum256(sessionToken.Marshal())
+		_, err := sTokenCache.AuthenticateTokenV1(k, func() (sessionSDK.Object, error) {
```
roman-khimov: I'd try to make `sTokenCache` optional. Not critical.

internal/sessions/cache.go
```diff
@@ -0,0 +1,68 @@
...
+	t   sessionv2.Token
+	err error
+}
+
+// ObjectSessionsCache caches object session tokens based on their hashes, both
+// v1 and v2 versions. Caches are LRU, there is no garanty a put token will be
+// available later.
+type ObjectSessionsCache struct {
+	cacheV1 *lru.Cache[[sha256.Size]byte, sV1Results]
+	cacheV2 *lru.Cache[[sha256.Size]byte, sV2Results]
```
roman-khimov: Why changing the structure? It was a single cache previously.

internal/sessions/cache.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+// NewObjectSessionsCache creates ObjectSessionsCache with a provided size.
+// Caches for v1 and v2 tokens are independent. Size must be positive.
+func NewObjectSessionsCache(cacheSize int) *ObjectSessionsCache {
+	cacheV1, err := lru.New[[sha256.Size]byte, sV1Results](cacheSize)
+	if err != nil {
+		panic(err)
+	}
+	cacheV2, _ := lru.New[[sha256.Size]byte, sV2Results](cacheSize)
```
cthulhu-rider: should also panic

internal/crypto/object.go
```diff
@@ -55,11 +57,20 @@ func AuthenticateObject(obj object.Object, fsChain HistoricN3ScriptRunner, ecPar
...
+			if err := AuthenticateToken(sessionToken, fsChain); err != nil {
+				return *sessionToken, fmt.Errorf("session token: %w", err)
+			}
+			if sessionToken.Issuer() != obj.Owner() && versioncore.OwnerSignatureMatchRequired(obj.Version()) {
```
cthulhu-rider: isn't this check request-dependent? Same below

internal/crypto/object.go
```diff
@@ -55,11 +57,20 @@ func AuthenticateObject(obj object.Object, fsChain HistoricN3ScriptRunner, ecPar
...
-		if err := AuthenticateToken(sessionToken, fsChain); err != nil {
-			return fmt.Errorf("session token: %w", err)
-		}
-		if sessionToken.Issuer() != obj.Owner() && versioncore.OwnerSignatureMatchRequired(obj.Version()) {
-			return errors.New("different object owner and session issuer")
+
+		k := sha256.Sum256(sessionToken.Marshal())
+		_, err := sTokenCache.AuthenticateTokenV1(k, func() (sessionSDK.Object, error) {
+			if err := AuthenticateToken(sessionToken, fsChain); err != nil {
+				return *sessionToken, fmt.Errorf("session token: %w", err)
```
cthulhu-rider: dereference is redundant, same below

internal/crypto/object.go
carpawell: IDE did that, fixed

internal/crypto/object.go
carpawell: i considered it, but decided that currently it is more like a must for us. once optional usage is found, a few more lines can be added

internal/sessions/cache.go
carpawell: i did not understand why v1 requests may affect the execution of v2 requests. if you find it strrange i may merge them back

internal/sessions/cache.go
carpawell: why? func and arg are the same

internal/crypto/object.go
carpawell: fixed

internal/crypto/object.go
carpawell: it depends on what we want to cache: a pointer or a value, i chose value

internal/sessions/cache.go
roman-khimov: Keys are cryptographically strong hashes, as long as we trust them v1 shouldn't affect v2 and vice versa.

carpawell: i meant LRU cache size is limited, more v1 requests will make more FSChain requests for v2 tokens and vice versa

roman-khimov: Not an issue to me, v1 is almost a history (present in old objects, but not currently used a lot).

carpawell: but still, they are different caches to me. at least, v1 request will take a lock for v2 request. also, an undefined case when inside a summarizing result struct there are two non-zero fields (for some reason), with two separate caches, it is impossible

roman-khimov: As you can see the old structure worked just fine in its old place.

internal/sessions/cache.go
cthulhu-rider: the same call does not guarantee presence or absence of an error in general

internal/crypto/object.go
cthulhu-rider: i mean value is unused when there is an error, but u return both of them. For some reason or just like that?

internal/sessions/cache.go
carpawell: there is a single cache now

internal/crypto/object.go
carpawell: ah, in this way. fixed

internal/sessions/cache.go
carpawell: cant understand why it can be/not be important. fixed

cthulhu-rider (approved):

roman-khimov (approved):
Conflicts

PR #4095, opened 2026-07-17
node: pin inter-node TLS certificates to node keys

### LINKED ISSUE #4087: Don't check global certificate validity for inter-node connections
We know exactly the key of the node we're connecting to (this one must be checked of course), CA validation is irrelevant here (connections are always direct). This allows to use self-signed certificates in many cases without any security compromises (somewhat similar to SSH local key lists).

End-rey:
Closes #4087.

### DIFF
--- a/pkg/network/cache/clients.go
+++ b/pkg/network/cache/clients.go
@@ -3,6 +3,9 @@ package cache
 import (
 	"bytes"
 	"context"
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/tls"
 	"encoding/hex"
 	"errors"
 	"fmt"
@@ -13,6 +16,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
 	igrpc "github.com/nspcc-dev/neofs-node/internal/grpc"
 	"github.com/nspcc-dev/neofs-node/internal/uriutil"
 	clientcore "github.com/nspcc-dev/neofs-node/pkg/core/client"
@@ -212,7 +216,11 @@ func (x *Clients) initConnection(ctx context.Context, pub []byte, uri string) (*
 	}
 	var transportCreds credentials.TransportCredentials
 	if withTLS {
-		transportCreds = credentials.NewTLS(nil)
+		expectedKey, err := keys.NewPublicKeyFromBytes(pub, elliptic.P256())
+		if err != nil {
+			return nil, fmt.Errorf("parse node public key: %w", err)
+		}
+		transportCreds = credentials.NewTLS(newNodeTLSConfig((*ecdsa.PublicKey)(expectedKey)))
 	} else {
 		transportCreds = insecure.NewCredentials()
 	}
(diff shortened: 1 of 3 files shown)


pkg/network/cache/clients.go
```diff
@@ -254,6 +257,32 @@ func (x *Clients) initConnection(ctx context.Context, pub []byte, uri string) (*
 	return res, nil
 }
 
+func newNodeTLSConfig(expectedPub []byte) *tls.Config {
+	expectedPub = bytes.Clone(expectedPub)
```
cthulhu-rider: i'd prefer to `elliptic.UnmarshalCompressed()` first and cmp curve points. This would protect from garbage expectedPub values

but normally this will work too

pkg/network/cache/clients.go
```diff
@@ -254,6 +257,32 @@ func (x *Clients) initConnection(ctx context.Context, pub []byte, uri string) (*
...
+
+	return &tls.Config{
+		InsecureSkipVerify: true,
```
cthulhu-rider, line 267: why do we need it?

End-rey: To skip default CA and hostname verification; otherwise self-signed certificates are rejected before VerifyConnection runs.

pkg/network/cache/clients.go
End-rey: `PublicKey.X/Y` access is deprecated in go, so I used `keys.NewPublicKeyFromBytes` to validate the compressed key and `PublicKey.Equal` for comparison instead.

pkg/network/cache/clients.go
```diff
@@ -254,6 +258,35 @@ func (x *Clients) initConnection(ctx context.Context, pub []byte, uri string) (*
 	return res, nil
 }
 
+func newNodeTLSConfig(expectedPub []byte) *tls.Config {
+	expectedKey, expectedKeyErr := keys.NewPublicKeyFromBytes(expectedPub, elliptic.P256())
+
+	return &tls.Config{
+		InsecureSkipVerify: true,
+		VerifyConnection: func(state tls.ConnectionState) error {
+			if expectedKeyErr != nil {
```
roman-khimov: I'd try to handle it externally (pass a key into `newNodeTLSConfig`).

End-rey: Done.

pkg/network/cache/clients.go
cthulhu-rider: > PublicKey.X/Y access is deprecated in go

we could use https://pkg.go.dev/crypto/ecdsa#PublicKey.Equal. But `keys` package is ok too, it's widely used in codebase

pkg/network/cache/clients.go
```diff
@@ -212,7 +216,11 @@ func (x *Clients) initConnection(ctx context.Context, pub []byte, uri string) (*
 	}
 	var transportCreds credentials.TransportCredentials
 	if withTLS {
-		transportCreds = credentials.NewTLS(nil)
+		expectedKey, err := keys.NewPublicKeyFromBytes(pub, elliptic.P256())
```
cthulhu-rider, line 219: `initConnection` is called iteratively, so multiple URIs can require `expectedKey`. Although re-parsing is not expected in practice, there is mostly 1 endpoint and cache should accommodate all nodes. Ok leave like this

cthulhu-rider (approved):

roman-khimov (approved):

PR #4096, opened 2026-07-22
Feat/ec part hashes in parent object

carpawell:


pkg/services/object/put/distributed.go
```diff
@@ -375,11 +452,7 @@ nextRule:
 				continue
 			}
 
-			payloadParts, err := iec.Encode(ecRules[ecRuleIdx], obj.Payload())
```
roman-khimov, line 378: That's what I meant in SDK PR, btw. There is a simple `[]byte` in the end of the process anyway...

pkg/services/object/put/distributed.go
```diff
@@ -109,6 +120,63 @@ func newMaxReplicasError(maxReplicas, done uint, lastRule int, lastRuleErr error
...
+		b          = getPayload()
+		payloadLen = int(hdr.PayloadSize())
+	)
+	if cap(b) < payloadLen {
+		putPayload(b)
+		b = make([]byte, 0, payloadLen)
+	}
+	buff := bytes.NewBuffer(b)
+
+	n, err := io.Copy(buff, reader)
```
roman-khimov, line 150: ...but now there is a new buffer we're allocating and copying into in addition to the one that holds the resulting `Payload()` (managed by SDK).

pkg/services/object/put/distributed.go
carpawell: that is right, it has to be a byte slice somewhere anyway, currently copy is inevitable. BUT on this side we can reuse bytes pool, and on the sdk side there will always be a resulting allocation that can not be reused

roman-khimov: What prevents us from reusing the SDK slice here?

carpawell: SDK slice does not currently exist for every case we may have (it may be a set of cached slices of different sizes). on the node side, it will be shared via sync.Pool slice, on the SDK side it will be a new slice every time

i do not insist; it is possible to use every data slicer has and put it into a single slice for the node, but i am afraid it will be slower compared to what we have now

carpawell: we also may consider https://github.com/nspcc-dev/neofs-sdk-go/blob/3f753464c49b422f9319f1cd24fe9c4bd18bc605/object/slicer/options.go#L76 from node side (currently we do not use it), but it is not always clear what size it should be and to which pool to return this memory

pkg/services/object/put/distributed.go
```diff
@@ -109,6 +120,69 @@ func newMaxReplicasError(maxReplicas, done uint, lastRule int, lastRuleErr error
...
+	var (
+		b          = getPayload()
+		payloadLen = int(hdr.PayloadSize())
+	)
+	if cap(b) < payloadLen {
+		putPayload(b)
+		b = make([]byte, 0, payloadLen)
+	}
+	if cap(b) != payloadLen {
+		// prohibit using additional slice's capacity for EC library, it may lead
```
carpawell, line 144: took some time to understand it. currently it is the only way, it may steal some shared memory in practice

cmd/neofs-cli/modules/object/head.go
```diff
@@ -82,6 +82,8 @@ func getObjectHeader(cmd *cobra.Command, _ []string) error {
 		prm.MarkRaw()
 	}
 
+	prm.SkipChecksumVerification()
```
carpawell: this is arguable but required for tests to run ok. make it configurable?

internal/ec/ec.go
```diff
@@ -27,35 +30,58 @@ func (x Rule) String() string {
...
 // [Rule.DataPartNum], last data part is aligned with zeros.
 //
 // If data is empty, all parts are nil.
-func Encode(rule Rule, data []byte) ([][]byte, error) {
+func Encode(rule Rule, data []byte) ([][]byte, []string, error) {
 	if len(data) == 0 {
-		return make([][]byte, rule.DataPartNum+rule.ParityPartNum), nil
+		var (
+			partsNumber = rule.DataPartNum + rule.ParityPartNum
+			hashes      = make([]string, 0, partsNumber)
```
cthulhu-rider, line 51: worth making it once. Although empty data is rare

internal/ec/objects.go
```diff
@@ -135,6 +135,6 @@ func decodeUint8StringToInt(s string) (int, error) {
 // ObjectWithAttributes checks whether obj contains at least one EC attribute.
 func ObjectWithAttributes(obj object.Object) bool {
 	return slices.ContainsFunc(obj.Attributes(), func(a object.Attribute) bool {
-		return strings.HasPrefix(a.Key(), AttributePrefix)
+		return strings.HasPrefix(a.Key(), AttributePrefix) && a.Key() != AttributePartsHashes
```
cthulhu-rider: docs should be updated

pkg/core/object/ec.go
```diff
@@ -47,12 +49,8 @@ func checkEC(hdr object.Object, rules []netmap.ECRule, blank bool, isParent bool
 		}
 	case object.TypeRegular:
 		if isParent {
-			if ecAttr != "" {
-				return false, fmt.Errorf("parent object has EC attribute %s", ecAttr)
-			}
 			return false, nil
 		}
-
```
cthulhu-rider, line 55: hardly necessary

pkg/core/object/ec.go
```diff
@@ -13,7 +15,7 @@ import (
 func checkEC(hdr object.Object, rules []netmap.ECRule, blank bool, isParent bool) (bool, error) {
 	attrs := hdr.Attributes()
 	var ecAttr string
-	if len(attrs) > 0 {
+	if !isParent && len(attrs) > 0 {
```
cthulhu-rider: this also affects following `len(rules) == 0` which still should work

pkg/core/object/ec.go
```diff
@@ -119,5 +117,50 @@ func checkECPart(part object.Object, rules []netmap.ECRule) error {
...
+			hashAttr = attr.Value()
+		}
+	}
+	if hashAttr == "" {
+		return fmt.Errorf("missing %s EC attribute in parent object", iec.AttributePartsHashes)
+	}
+
+	cs, ok := part.PayloadChecksum()
+	if !ok {
+		return fmt.Errorf("EC part does not have payload checksum")
```
cthulhu-rider: no need to fmt

pkg/services/object/put/distributed.go
```diff
@@ -109,6 +120,72 @@ func newMaxReplicasError(maxReplicas, done uint, lastRule int, lastRuleErr error
...
+
+	var (
+		b          = getPayload()
+		payloadLen = int(hdr.PayloadSize())
+	)
+	if cap(b) < payloadLen {
+		putPayload(b)
+		b = make([]byte, 0, payloadLen)
+	}
+	if cap(b) != payloadLen {
```
cthulhu-rider, line 143: `else` to previous condition

pkg/services/object/put/distributed.go
```diff
@@ -109,6 +120,72 @@ func newMaxReplicasError(maxReplicas, done uint, lastRule int, lastRuleErr error
...
+	if err != nil {
+		return fmt.Errorf("reading EC object payload from slicer: %w", err)
+	}
+	if int(n) != payloadLen {
+		return fmt.Errorf("read %d bytes from slicer, but claimed payload len is %d", n, payloadLen)
+	}
+	t.payloadAlreadyRead = true
+	t.objectPayload = buff.Bytes()
+
+	var hashes []string
```
cthulhu-rider: worth preallocation. And seems that `strings.Builder` can be used

pkg/services/object/put/distributed.go
```diff
@@ -109,6 +120,72 @@ func newMaxReplicasError(maxReplicas, done uint, lastRule int, lastRuleErr error
...
+
+	var (
+		foundAttr bool
+		attrs     = hdr.Attributes()
+		hashesStr = strings.Join(hashes, ",")
+	)
+	for _, attr := range attrs {
+		if attr.Key() == iec.AttributePartsHashes {
+			foundAttr = true
+			attr.SetValue(hashesStr)
```
cthulhu-rider: isnt `attr` a copy that leaves the object unchanged?

pkg/services/object/put/distributed.go
```diff
@@ -140,7 +217,14 @@ func (t *distributedTarget) WriteHeader(hdr *object.Object) error {
 }
 
 func (t *distributedTarget) Write(p []byte) (n int, err error) {
-	t.encodedObject.b = append(t.encodedObject.b, p...)
+	if t.payloadAlreadyRead {
+		if t.objectPayload != nil {
+			t.encodedObject.b = append(t.encodedObject.b, t.objectPayload...)
```
cthulhu-rider, line 241: do we rly concat here? What exactly if so?

pkg/services/object/put/distributed.go
```diff
@@ -150,6 +234,17 @@ func (t *distributedTarget) Close() (oid.ID, error) {
...
 		t.encodedObject.b = nil
 		t.resetMetaCollection()
+		t.payloadAlreadyRead = false
+	outer:
+		for _, encodedRule := range t.encodedECParts {
+			for _, part := range encodedRule {
+				putPayload(part)
+				// encoded EC parts share memory, so only a single
+				// slice can be reused
+				break outer
```
cthulhu-rider: seems like `for`s can be just `if`s

pkg/core/object/ec.go
```diff
@@ -71,6 +71,10 @@ func checkEC(hdr object.Object, rules []netmap.ECRule, blank bool, isParent bool
 }
 
 func checkECPart(part object.Object, rules []netmap.ECRule) error {
+	if part.Signature() != nil {
```
cthulhu-rider, line 82: ideally it should be version-based, but ok

pkg/core/object/fmt.go
```diff
@@ -254,11 +254,13 @@ func (v *FormatValidator) validate(ctx context.Context, obj *object.Object, unpr
 			return fmt.Errorf("could not validate header fields: invalid identifier: %w", err)
 		}
 
-		if err := icrypto.AuthenticateObject(*obj, historicN3ScriptRunner{
-			FSChain:        v.fsChain,
-			NetmapContract: v.netmapContract,
-		}, isEC, v.sCache, v.resolver); err != nil {
-			return fmt.Errorf("authenticate: %w", err)
+		if !isEC {
+			if err := icrypto.AuthenticateObject(*obj, historicN3ScriptRunner{
```
cthulhu-rider, line 258: so `isEC` can be deparameterized now

pkg/services/policer/ec.go
```diff
@@ -67,7 +67,7 @@ func (p *Policer) processECPartByRule(ctx context.Context, rule iec.Rule, addr o
 		}
 
 		callCtx, cancel := context.WithTimeout(ctx, headTimeout)
-		_, err := p.apiConns.headObject(callCtx, nodes[i], addr, true, nil)
+		_, err := p.apiConns.headObject(callCtx, nodes[i], addr, false, nil)
```
cthulhu-rider: why did this become necessary?

cmd/neofs-cli/modules/object/head.go
carpawell: obsolete thread

pkg/core/object/ec.go
roman-khimov, line 55: :eyes:

pkg/core/object/ec.go
```diff
@@ -119,5 +117,50 @@ func checkECPart(part object.Object, rules []netmap.ECRule) error {
 		return fmt.Errorf("wrong part payload len: expected %d, got %d, parent %d", exp, partPldLen, parentPldLen)
 	}
 
+	return checkECParent(*parent, part, rules, pi)
+}
+
+func checkECParent(parent, part object.Object, rules []netmap.ECRule, pi iec.PartInfo) error {
+	var hashAttr string
+	for _, attr := range parent.Attributes() {
```
roman-khimov, line 137: Why do you need a second pass over attributes? This can be avoided.

pkg/local_object_storage/metabase/exists.go
```diff
@@ -258,7 +258,7 @@ loop:
 			splitInfo = object.NewSplitInfo()
 		}
 		for attrKey, attrVal := range iterIDAttrs(metaCursor.Bucket().Cursor(), objID) {
-			if strings.HasPrefix(string(attrKey), iec.AttributePrefix) {
+			if strings.HasPrefix(string(attrKey), iec.AttributePrefix) && string(attrKey) != iec.AttributePartsHashes {
```
roman-khimov: BTW, not sure we need hashes stored in metabase. We don't seem to have any scenario for their usage.

pkg/core/object/ec.go
roman-khimov: EC still has experimental status, so at this stage we're free to change it like this.

pkg/local_object_storage/metabase/exists.go
carpawell: overall i agree, but this will be the first case for skipping an attribute in metabase. also, reconstructing an object from metabase will work incorrect (only used in inspect utilities for now)

pkg/services/object/put/distributed.go
carpawell: `cap` can be bigger than `payloadLen` and that is exactly where EC encoding breaks itself. we need strict equality here, it is not `else`

pkg/local_object_storage/metabase/exists.go
roman-khimov: Reconstruction isn't possible for a long time, metabase does not have all the data from header and it doesn't need to.

carpawell: we still have `DB.Get` method and it is used, that is the only thing i meant

internal/ec/ec.go
carpawell: you mean do not duplicate code or what?

pkg/core/object/ec.go
carpawell: why is it the second one? we are checking EC part, it first needs to validate part's attributes, then it needs to find a parent's checksum (another header with different attributes)

pkg/core/object/ec.go
carpawell: i would also leave it experimental for the next release too

pkg/services/object/put/distributed.go
carpawell: preallocated. about Builder: to me, preallocation + strings.Join (has a builder inside) is enough

pkg/services/object/put/distributed.go
carpawell: fixed

pkg/services/object/put/distributed.go
carpawell: if an object was pre-read, it can be written once here, no need to do it with multiple `Write` calls. if a container has both EC and REP rules `t.encodedObject.b` will be used for REP, while EC parts will be taken from `encodedECParts` differenlty

internal/ec/objects.go
carpawell: reverted, no need for this animore

pkg/core/object/ec.go
carpawell: moved this check to deeper levels

pkg/services/policer/ec.go
carpawell: left from the previous version, reverted

pkg/local_object_storage/metabase/exists.go
carpawell: skipped new attribute for metabase

roman-khimov: Adjust the spec then.

pkg/services/object/put/service_test.go
```diff
@@ -1494,8 +1494,13 @@ func splitMembersCount(limit, ln uint64) int {
 	return int(res)
 }
 
-func assertObjectIntegrity(t *testing.T, obj object.Object) {
-	require.NoError(t, obj.CheckVerificationFields())
+func assertObjectIntegrity(t *testing.T, obj object.Object, unsingedObject bool) {
+	if unsingedObject {
```
roman-khimov: unsigned

internal/ec/ec.go
cthulhu-rider: nah i missed that len is dynamic. Forget

pkg/services/object/put/distributed.go
cthulhu-rider: this is definitely `false` if previous condition is met, so it's `else` to me. Ok, not critical

pkg/services/object/get_test.go
```diff
@@ -496,13 +496,13 @@ func assertGetOK(t *testing.T, srv *Server, mtrc *metricsCollector, obj object.O
 
 func assertGetOKVersioned(t *testing.T, srv *Server, mtrc *metricsCollector, obj object.Object, signer neofscrypto.Signer, ver version.Version) []*protoobject.GetResponse {
 	req := newLocalGetRequest(t, ver, obj.Address(), signer)
-	return assertGetRequest(t, srv, mtrc, req, obj)
+	return assertGetRequest(t, srv, mtrc, req, obj, false)
 }
 
-func assertGetRequest(t *testing.T, srv *Server, mtrc *metricsCollector, req *protoobject.GetRequest, obj object.Object) []*protoobject.GetResponse {
+func assertGetRequest(t *testing.T, srv *Server, mtrc *metricsCollector, req *protoobject.GetRequest, obj object.Object, unsingedObject bool) []*protoobject.GetResponse {
```
roman-khimov: Here as well.

pkg/services/object/put/distributed.go
cthulhu-rider: it'd be even shorter w/o `attr` at all. Not critical

pkg/services/object/put/streamer.go
```diff
@@ -77,7 +77,7 @@ func (p *Streamer) initTarget(prm *PutInitPrm) error {
 		return fmt.Errorf("(%T) could not obtain max object size parameter", p)
 	}
 
-	if prm.hdr.Signature() != nil {
+	if prm.hdr.Signature() != nil || prm.ecPart.RuleIndex >= 0 {
```
roman-khimov, line 80: Questionable, btw. Implies an EC chunk from client. Theoretically possible, but practically...

internal/ec/ec.go
```diff
@@ -11,9 +13,10 @@ import (
 
 // Erasure coding attributes.
 const (
-	AttributePrefix  = "__NEOFS__EC_"
-	AttributeRuleIdx = AttributePrefix + "RULE_IDX"
-	AttributePartIdx = AttributePrefix + "PART_IDX"
+	AttributePrefix      = "__NEOFS__EC_"
+	AttributeRuleIdx     = AttributePrefix + "RULE_IDX"
+	AttributePartIdx     = AttributePrefix + "PART_IDX"
+	AttributePartsHashes = AttributePrefix + "PART_HASHES"
```
roman-khimov, line 19: These are available from SDK.

pkg/core/object/ec.go
```diff
@@ -119,5 +125,50 @@ func checkECPart(part object.Object, rules []netmap.ECRule) error {
...
+		if strings.HasPrefix(attr.Key(), iec.AttributePrefix) {
+			if attr.Key() != iec.AttributePartsHashes {
+				return fmt.Errorf("parent object has prohibited EC %s attribute", attr.Key())
+			}
+			if hashAttr != "" {
+				return fmt.Errorf("multiple %s EC attributes in parent object", iec.AttributePartsHashes)
```
roman-khimov: Don't we have any generic checks for duplicate attributes?

pkg/services/object/put/distributed.go
cthulhu-rider: i meant whole string can be preallocated and filled iteratively. Or not?

pkg/local_object_storage/metabase/exists.go
carpawell: what do you call "spec" in this case? tried to update method docs, but there is not much to say

pkg/services/object/put/streamer.go
carpawell: there are many code parts that are expecting ready EC parts via PUT. moreover, unit tests die without this line, i tried to do the absolute minimum for the new logic to work. also, if an SN node is not a container's part, it will "replicate" parts via regular PUT request IINM

internal/ec/ec.go
carpawell: i only added one that is not available from SDK (and not sure if it should)

pkg/core/object/ec.go
carpawell: agree, we have, dropped

pkg/services/object/put/distributed.go
carpawell: added. but immo, it did not worth it

internal/ec/ec.go
roman-khimov: They're all known and should be in SDK. Not critical now.

pkg/services/object/put/streamer.go
roman-khimov: OK, needs some untangling in future.

roman-khimov (approved):

cthulhu-rider (approved):

PR #4097, opened 2026-07-22
node: send certificates for inter-node TLS connections

### LINKED ISSUE #4088: Send client certificate when connecting to another SN
Make node accept these certificates (no validity wrt CA required), treat unsigned requests as coming from the specified key.

End-rey:
Closes #4088.

### DIFF
--- a/cmd/neofs-node/mtls.go
+++ b/cmd/neofs-node/mtls.go
@@ -1,0 +1,27 @@
+package main
+
+import (
+	"crypto/tls"
+	"fmt"
+
+	grpcconfig "github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/grpc"
+)
+
+func clientCertificateProvider(cfgs []grpcconfig.GRPC) func(*tls.CertificateRequestInfo) (*tls.Certificate, error) {
+	for i := range cfgs {
+		if !cfgs[i].TLS.Enabled {
+			continue
+		}
+
+		certFile, keyFile := cfgs[i].TLS.Certificate, cfgs[i].TLS.Key
+		return func(*tls.CertificateRequestInfo) (*tls.Certificate, error) {
+			cert, err := tls.LoadX509KeyPair(certFile, keyFile)
+			if err != nil {
+				return nil, fmt.Errorf("reload TLS client certificate: %w", err)
+			}
+			return &cert, nil
+		}
+	}
+
+	return nil
+}
--- a/pkg/services/object/acl/v2/service.go
+++ b/pkg/services/object/acl/v2/service.go
@@ -1,6 +1,8 @@
 package v2
 
 import (
+	"context"
+	"crypto/ecdsa"
 	"crypto/sha256"
 	"errors"
 	"fmt"
@@ -9,6 +11,7 @@ import (
 	lru "github.com/hashicorp/golang-lru/v2"
 	"github.com/nspcc-dev/neo-go/pkg/core/block"
 	"github.com/nspcc-dev/neo-go/pkg/core/transaction"
+	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
 	"github.com/nspcc-dev/neo-go/pkg/neorpc/result"
 	"github.com/nspcc-dev/neo-go/pkg/smartcontract/trigger"
 	"github.com/nspcc-dev/neo-go/pkg/util"
@@ -17,6 +20,7 @@ import (
 	containercore "github.com/nspcc-dev/neofs-node/pkg/core/container"
 	netmapcore "github.com/nspcc-dev/neofs-node/pkg/core/netmap"
 	nnscore "github.com/nspcc-dev/neofs-node/pkg/core/nns"
+	"github.com/nspcc-dev/neofs-node/pkg/network/peerauth"
 	"github.com/nspcc-dev/neofs-node/pkg/services/object/common"
 	"github.com/nspcc-dev/neofs-sdk-go/bearer"
 	apistatus "github.com/nspcc-dev/neofs-sdk-go/client/status"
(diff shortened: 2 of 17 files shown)


cmd/neofs-node/mtls.go
```diff
@@ -0,0 +1,27 @@
+package main
+
+import (
+	"crypto/tls"
+	"fmt"
+
+	grpcconfig "github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/grpc"
+)
+
+func clientCertificateProvider(cfgs []grpcconfig.GRPC) func(*tls.CertificateRequestInfo) (*tls.Certificate, error) {
```
roman-khimov, line 10: This means any of configured certificates, right? Should be documented then.

pkg/services/object/common/request.go
```diff
@@ -11,4 +11,8 @@ type RequestTokens struct {
 	Session   *sessionv2.Token
 	SessionV1 *session.Object
 	Bearer    *bearer.Token
+
+	// AuthenticatedPeerPublicKey is the compressed ECDSA public key of an
+	// inter-node TLS peer authenticated by the transport layer.
+	AuthenticatedPeerPublicKey []byte
```
roman-khimov: Why not `keys.PublicKey`? Getting bytes from it (when needed) is rather trivial, but going the other way around requires touching the cache or even doing some real math.

pkg/services/object/server.go
```diff
@@ -642,8 +647,10 @@ func (s *Server) HeadBuffered(ctx context.Context, req *protoobject.HeadRequest)
 
 	needSignResp := needSignGetResponse(req)
 
-	if err := icrypto.VerifyRequestSignaturesN3(req, s.fsChain); err != nil {
-		return s.makeStatusHeadResponse(err, needSignResp)
+	if requestNeedsSignature(ctx, req) {
+		if err := icrypto.VerifyRequestSignaturesN3(req, s.fsChain); err != nil {
```
roman-khimov: If it all goes via `icrypto` why not handling `requestNeedsSignatures` internally there? Like `icrypto.VerifyRequestSignaturesN3(ctx, req, s.fsChain)`

pkg/services/object/server.go
```diff
@@ -1907,7 +1920,7 @@ func (s *Server) SearchV2Buffered(ctx context.Context, req *protoobject.SearchV2
 		return s.signSearchResponse(nil, err, req)
 	}
 
-	reqInfo, err := s.reqInfoProc.SearchV2RequestToInfo(req, cnrID, reqMD.tokens)
+	reqInfo, err := s.reqInfoProc.SearchV2RequestToInfo(req, cnrID, requestTokensWithPeer(ctx, req, reqMD.tokens))
```
roman-khimov: Same here, passing `ctx` can solve it.

cmd/neofs-node/mtls.go
End-rey: Documented.

pkg/services/object/common/request.go
End-rey: Done,

pkg/services/object/server.go
End-rey: Done.

pkg/services/object/server.go
End-rey: Done.

roman-khimov (approved):

pkg/services/object/acl/v2/service.go
```diff
@@ -278,6 +282,30 @@ func getCredentialsFromSessionToken(token sessionv2.Token) (user.ID, []byte, err
...
 
+func getCredentialsFromPeerPublicKey(key *keys.PublicKey) (user.ID, []byte, error) {
+	return user.NewFromECDSAPublicKey(ecdsa.PublicKey(*key)), key.Bytes(), nil
+}
+
+func getRequestCredentials(ctx context.Context, req interface {
+	GetMetaHeader() *protosession.RequestMetaHeader
+	GetVerifyHeader() *protosession.RequestVerificationHeader
+}, tokens common.RequestTokens) (user.ID, []byte, error) {
+	if req.GetVerifyHeader() == nil && req.GetMetaHeader().GetTtl() == 1 {
```
carpawell, line 293: what is so special about TTL? what if `req.GetVerifyHeader()` == nil and TTL == 2? isnt such a request still a possible Node-Node? can you point me, please, to a place where non-container node sends a node-to-node request with TTL=2?

pkg/services/object/server.go
```diff
@@ -421,7 +421,7 @@ func (s *Server) Put(gStream protoobject.ObjectService_PutServer) error {
 			s.metrics.AddPutPayload(len(c))
 		}
 
-		if err = icrypto.VerifyRequestSignaturesN3(req, s.fsChain); err != nil {
+		if err = icrypto.VerifyRequestSignaturesN3(gStream.Context(), req, s.fsChain); err != nil {
```
carpawell: `gStream.Context()` can be called once in the beginning of these funcs in this file

pkg/services/object/acl/v2/service.go
End-rey: > can you point me, please, to a place where non-container node sends a node-to-node request with TTL=2? (i guess)

There is no such path in the current object service. A request with TTL=2 may arrive from a client at a non-container node, but when this node forwards it to a container node, it creates a new request with TTL=1. Thus, all node-to-node requests produced by the forwarding paths use TTL=1. TLS authentication replaces request signatures only for this final local hop; unsigned requests with TTL>1 are not trusted as forwarded requests.

pkg/services/object/server.go
End-rey: Done.

pkg/services/object/acl/v2/service.go
roman-khimov: > unsigned requests with TTL>1 are not trusted as forwarded requests.

The key. We can't forward unsigned requests.

carpawell: if it is strictly prohibited, i suggest returning a clear error message. currently, an empty verify header with TTL=2 just skips this branch, and it will fail later casually, like a request with no verification header (even if a client tried to use TLS correctly and just used TTL=2 for some reason)

carpawell (approved):

PR #4099, opened 2026-07-27
Validate eACL record comments

End-rey:


cmd/neofs-cli/modules/util/acl.go
```diff
@@ -383,9 +384,17 @@ func eaclOperationsFromString(s string) ([]eacl.Operation, error) {
...
 // ValidateEACLTable validates eACL table:
 //   - eACL table must not modify [eacl.RoleSystem] access.
+//   - Record comments must be valid UTF-8 without zero bytes.
 func ValidateEACLTable(t eacl.Table) error {
 	var b big.Int
 	for _, record := range t.Records() {
+		if !utf8.ValidString(record.Comment()) {
+			return errors.New("invalid UTF-8 comment")
+		}
+		if strings.ContainsRune(record.Comment(), 0) {
```
roman-khimov, line 394: Can we make an EACL that has comments now via CLI? JSON?

End-rey: Yes, it's possible via JSON. But I mainly added this here because it's an exact duplicate validation from IR.

roman-khimov (approved):

PR #4100, opened 2026-07-28
node: skip TTL=1 request signatures over inter-node mTLS

End-rey:
Would it make sense to add an SDK option for unsigned requests?
Some inter-node TTL=1 paths still use SDK `ObjectHead/ObjectGetInit/ObjectRangeInit/ObjectPutInit`, which always sign requests. An opt-in parameter such as `DisableRequestSigning` could let the node skip signatures after it has confirmed mTLS, while keeping the SDK default safe.

### DIFF
--- a/cmd/neofs-node/grpc.go
+++ b/cmd/neofs-node/grpc.go
@@ -19,7 +19,6 @@ import (
 	"go.uber.org/zap"
 	"golang.org/x/net/netutil"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/resolver"
 )
@@ -196,7 +195,7 @@ func buildSingleGRPCServer(c *cfg, sc grpcconfig.GRPC, maxRecvMsgSizeOpt grpc.Se
 		}
 
 		// read certificate from disk on each handshake to pick up renewals automatically.
-		creds := credentials.NewTLS(&tls.Config{
+		creds := trustedPeerTLSCredentials(&tls.Config{
 			GetConfigForClient: func(*tls.ClientHelloInfo) (*tls.Config, error) {
 				cert, err := tls.LoadX509KeyPair(certFile, keyFile)
 				if err != nil {
--- a/cmd/neofs-node/mtls.go
+++ b/cmd/neofs-node/mtls.go
@@ -1,13 +1,47 @@
 package main
 
 import (
+	"bytes"
 	"crypto/tls"
+	"errors"
 	"fmt"
+	"net"
 
 	grpcconfig "github.com/nspcc-dev/neofs-node/cmd/neofs-node/config/grpc"
+	"github.com/nspcc-dev/neofs-node/pkg/network/peerauth"
+	"google.golang.org/grpc/credentials"
 )
 
-func clientCertificateProvider(cfgs []grpcconfig.GRPC) func(*tls.CertificateRequestInfo) (*tls.Certificate, error) {
+type trustedPeerCredentials struct {
+	credentials.TransportCredentials
+}
+
+func trustedPeerTLSCredentials(config *tls.Config) credentials.TransportCredentials {
+	return trustedPeerCredentials{TransportCredentials: credentials.NewTLS(config)}
+}
+
+func (x trustedPeerCredentials) Clone() credentials.TransportCredentials {
+	return trustedPeerCredentials{TransportCredentials: x.TransportCredentials.Clone()}
+}
+
+func (x trustedPeerCredentials) ServerHandshake(conn net.Conn) (net.Conn, credentials.AuthInfo, error) {
+	conn, authInfo, err := x.TransportCredentials.ServerHandshake(conn)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	tlsInfo, ok := authInfo.(credentials.TLSInfo)
+	if !ok || len(tlsInfo.State.PeerCertificates) == 0 {
+		return conn, authInfo, nil
+	}
+	trustedInfo, err := peerauth.NewAuthInfo(tlsInfo)
+	if err != nil {
+		return conn, authInfo, nil
+	}
+	return conn, trustedInfo, nil
+}
+
+func clientCertificateProvider(cfgs []grpcconfig.GRPC, expectedPublicKey []byte) func(*tls.CertificateRequestInfo) (*tls.Certificate, error) {
 	for i := range cfgs {
 		if !cfgs[i].TLS.Enabled {
 			continue
@@ -19,9 +53,23 @@ func clientCertificateProvider(cfgs []grpcconfig.GRPC) func(*tls.CertificateRequ
 			if err != nil {
 				return nil, fmt.Errorf("reload TLS client certificate: %w", err)
 			}
+			if err := verifyTLSCertificatePublicKey(&cert, expectedPublicKey); err != nil {
+				return nil, err
+			}
 			return &cert, nil
 		}
 	}
 
+	return nil
+}
+
+func verifyTLSCertificatePublicKey(cert *tls.Certificate, expected []byte) error {
+	pub, err := peerauth.CertificatePublicKeyFromRaw(cert.Certificate)
+	if err != nil {
+		return fmt.Errorf("parse TLS client certificate public key: %w", err)
+	}
+	if !bytes.Equal(pub, expected) {
+		return errors.New("TLS client certificate public key differs from node public key")
+	}
 	return nil
 }
--- a/pkg/core/client/client.go
+++ b/pkg/core/client/client.go
@@ -56,3 +56,10 @@ type MultiAddressClient interface {
 	// this client (min(server, client) effectively).
 	APIVersion() *protorefs.Version
 }
+
+// IsMutuallyAuthenticated reports whether all connections of c use mutual TLS.
+// Clients that do not expose this property are treated as unauthenticated.
+func IsMutuallyAuthenticated(c any) bool {
+	x, ok := c.(interface{ IsMutuallyAuthenticated() bool })
+	return ok && x.IsMutuallyAuthenticated()
+}
--- a/pkg/network/cache/clients.go
+++ b/pkg/network/cache/clients.go
@@ -272,6 +272,9 @@ func (x *Clients) initConnection(ctx context.Context, pub []byte, uri string) (*
 		_ = grpcConn.Close()
 		return res, nil, fmt.Errorf("init NeoFS API client from gRPC client conn: %w", err)
 	}
+	if res.mtls.Load() {
+		res.SkipSignatureForLocalRequests()
+	}
 
 	ctx, cancel := context.WithTimeout(ctx, x.streamMsgTimeout)
 	defer cancel()
--- a/pkg/network/cache/clients.go
+++ b/pkg/network/cache/clients.go
@@ -14,6 +14,7 @@ import (
 	"maps"
 	"slices"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
@@ -153,7 +154,7 @@ func (x *Clients) syncWithNetmapSN(ctx context.Context, sn netmap.NodeInfo) erro
 	conns.mtx.Lock()
 	defer conns.mtx.Unlock()
 
-	maps.DeleteFunc(conns.m, func(ma string, c *client.Client) bool {
+	maps.DeleteFunc(conns.m, func(ma string, c *endpointClient) bool {
 		if slices.Contains(as, ma) {
 			return false
 		}
@@ -181,12 +182,13 @@ func (x *Clients) syncWithNetmapSN(ctx context.Context, sn netmap.NodeInfo) erro
 		conns.m[ma] = c
 		x.log.Info("connection to new SN address in the new network map successfully initialized", zap.String("address", ma))
 	}
+	conns.updateMutualAuthenticationStatus()
 
 	return nil
 }
 
 func (x *Clients) initConnections(ctx context.Context, pub []byte, addrs iter.Seq[string]) (*connections, error) {
-	m := make(map[string]*client.Client)
+	m := make(map[string]*endpointClient)
 	l := x.log.With(zap.String("public key", hex.EncodeToString(pub)))
 	var ver *protorefs.Version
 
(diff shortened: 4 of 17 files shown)


roman-khimov:
Yes, SDK should be updated to do/not do this based on client status.

cmd/neofs-node/mtls.go
```diff
@@ -1,12 +1,48 @@
...
+func verifyInterNodePeer(c *cfg) func([][]byte, [][]*x509.Certificate) error {
+	return func(rawCerts [][]byte, _ [][]*x509.Certificate) error {
+		pub, err := peerauth.CertificatePublicKeyFromRaw(rawCerts)
+		if err != nil {
+			return err
+		}
+		if isNetmapNode(c, pub) {
+			return nil
+		}
+		c.log.Warn("reject inter-node TLS peer absent from network map", zap.Binary("public key", pub))
```
roman-khimov: Anyone can try this (and theoretically it can work for non-node cases as well, it's just that TTL=1 is less relevant in this context), so it's Info to me.

cmd/neofs-node/mtls.go
```diff
@@ -1,12 +1,48 @@
...
+	"go.uber.org/zap"
 )
 
+func verifyInterNodePeer(c *cfg) func([][]byte, [][]*x509.Certificate) error {
+	return func(rawCerts [][]byte, _ [][]*x509.Certificate) error {
+		pub, err := peerauth.CertificatePublicKeyFromRaw(rawCerts)
+		if err != nil {
+			return err
+		}
+		if isNetmapNode(c, pub) {
```
roman-khimov: Can be changed in future. Client-side TTL=1 requests are rare, but still valid ones.

cmd/neofs-node/grpc.go
```diff
@@ -197,14 +198,20 @@ func buildSingleGRPCServer(c *cfg, sc grpcconfig.GRPC, maxRecvMsgSizeOpt grpc.Se
...
+			GetConfigForClient: func(hello *tls.ClientHelloInfo) (*tls.Config, error) {
 				cert, err := tls.LoadX509KeyPair(certFile, keyFile)
 				if err != nil {
 					return nil, fmt.Errorf("reload TLS certificate: %w", err)
 				}
+				if hello.ServerName == peerauth.TLSServerName {
+					return &tls.Config{
+						Certificates:          []tls.Certificate{cert},
+						ClientAuth:            tls.RequireAnyClientCert,
+						VerifyPeerCertificate: verifyInterNodePeer(c),
```
roman-khimov: This means non-node client certificates aren't possible at all and connection won't be established, right?

pkg/core/client/client.go
```diff
@@ -56,3 +56,10 @@ type MultiAddressClient interface {
 	// this client (min(server, client) effectively).
 	APIVersion() *protorefs.Version
 }
+
+// IsMutuallyAuthenticated reports whether all connections of c use mutual TLS.
+// Clients that do not expose this property are treated as unauthenticated.
+func IsMutuallyAuthenticated(c any) bool {
```
roman-khimov, line 62: Overall, this part seems like a separate commit at least. It's not about TTL=1 requests, it's all about establishing trust.

pkg/network/cache/clients.go
```diff
@@ -172,13 +175,15 @@ func (x *Clients) syncWithNetmapSN(ctx context.Context, sn netmap.NodeInfo) erro
...
-		c, _, err := x.initConnection(ctx, pub, ma)
+		mtls := new(atomic.Bool)
+		c, _, err := x.initConnection(ctx, pub, ma, mtls)
 		if err != nil {
 			x.log.Info("failed to init connection to new SN address in the new network map",
 				zap.String("address", ma), zap.Error(err))
 			continue
 		}
 		conns.m[ma] = c
+		conns.mtls[ma] = mtls
```
roman-khimov: Why can't it be a part of `client.Client`?

pkg/network/cache/clients.go
```diff
@@ -306,17 +326,49 @@ func newNodeTLSConfig(expectedKey *ecdsa.PublicKey, getClientCertificate func(*t
...
 	apiVersion *protorefs.Version
 }
 
+func (x *connections) IsMutuallyAuthenticated() bool {
+	if !x.hasClientCertificate {
+		return false
+	}
+
+	x.mtx.RLock()
+	defer x.mtx.RUnlock()
```
roman-khimov: More overhead than I'd like to per-request, atomic should be sufficient.

pkg/network/cache/clients.go
```diff
@@ -306,17 +326,49 @@ func newNodeTLSConfig(expectedKey *ecdsa.PublicKey, getClientCertificate func(*t
...
+	}
+	for uri := range x.m {
+		if authenticated, ok := x.mtls[uri]; !ok || !authenticated.Load() {
+			return false
+		}
+		var a network.Address
+		if err := a.FromString(uri); err != nil {
+			return false
+		}
+		_, withTLS, err := uriutil.Parse(a.URIAddr())
```
roman-khimov: Parsing is also irrelevant here, all of this should be decided in connection management phase, client requests should just take some ready to use variable.

pkg/network/cache/clients.go
```diff
@@ -306,17 +326,49 @@ func newNodeTLSConfig(expectedKey *ecdsa.PublicKey, getClientCertificate func(*t
 			return nil
 		},
 	}
+	if getClientCertificate != nil {
+		cfg.ServerName = peerauth.TLSServerName
```
roman-khimov: That's a misuse of server names, it can't be static for all hosts.

pkg/services/object/server.go
```diff
@@ -422,11 +422,6 @@ func (s *Server) Put(gStream protoobject.ObjectService_PutServer) error {
 			s.metrics.AddPutPayload(len(c))
 		}
 
-		if err = icrypto.VerifyRequestSignaturesN3(ctx, req, s.fsChain); err != nil {
```
roman-khimov, line 424: Chunks are signed too.

pkg/services/object/server.go
```diff
@@ -1392,31 +1388,28 @@ func convertGetPrm(signer ecdsa.PrivateKey, cnr container.Container, req *protoo
 		proxyCtx.resolveRange = p.ResolveRange
 	}
 
-	var updatedRequest bool
```
roman-khimov, line 1405: This is still relevant, request should be signed once (if needed).

pkg/services/object/server.go
```diff
@@ -1392,31 +1388,28 @@ func convertGetPrm(signer ecdsa.PrivateKey, cnr container.Container, req *protoo
...
+			},
+		}
+		if proxyCtx.suppressInit {
+			outReq.Body = proto.Clone(outReq.Body).(*protoobject.GetRequest_Body)
+			outReq.Body.PayloadOnly = false
+		}
+		if shouldSignOutgoingRequest(c, outReq) {
 			var err error
-			req.VerifyHeader, err = neofscrypto.SignRequestWithBuffer(neofsecdsa.Signer(signer), req, nil)
+			outReq.VerifyHeader, err = neofscrypto.SignRequestWithBuffer(neofsecdsa.Signer(signer), outReq, nil)
```
roman-khimov: Why renaming, btw?

pkg/services/object/server.go
```diff
@@ -1653,28 +1646,28 @@ func convertRangePrm(signer ecdsa.PrivateKey, cnr container.Container, req *prot
 		return p, nil
 	}
 
-	var onceResign sync.Once
```
roman-khimov, line 1666: Same here (and likely below)

cmd/neofs-node/grpc.go
End-rey: Fixed: the server now requests client certificates normally and does not reject non-node certificates during the TLS handshake.

pkg/core/client/client.go
End-rey: Split into a separate preceding commit.

pkg/network/cache/clients.go
End-rey: Moved endpoint TLS state next to the SDK client into `endpointClient`.

pkg/network/cache/clients.go
End-rey: Fixed.

pkg/network/cache/clients.go
End-rey: Fixed.

pkg/network/cache/clients.go
End-rey: Fixed.

pkg/services/object/server.go
End-rey: Fixed.

pkg/services/object/server.go
End-rey: Fixed.

pkg/services/object/server.go
End-rey: `outReq` is kept separate from the inbound request because forwarding changes its meta header and, for Get, may adjust `PayloadOnly`.

pkg/services/object/server.go
End-rey: Fixed.

cmd/neofs-node/mtls.go
```diff
@@ -1,12 +1,63 @@
...
+	}
+}
+
+func trustInterNodePeerStream(c *cfg) grpc.StreamServerInterceptor {
+	return func(srv any, stream grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
+		return handler(srv, trustedPeerServerStream{ServerStream: stream, ctx: trustInterNodePeer(c, stream.Context())})
+	}
+}
+
+func trustInterNodePeer(c *cfg, ctx context.Context) context.Context {
```
carpawell: `trustInterNodePeerContext`?

pkg/services/object/server.go
```diff
@@ -823,30 +823,33 @@ func convertHeadPrm(signer ecdsa.PrivateKey, cnr container.Container, req *proto
...
+	var (
+		outReq  *protoobject.HeadRequest
+		onceReq sync.Once
+		reqErr  error
+	)
 	p.SetTransportFunc(func(ctx context.Context, c clientcore.MultiAddressClient) (mem.BufferSlice, iprotobuf.BuffersSlice, error) {
-		if !updatedRequest {
-			updatedRequest = true
-			req = &protoobject.HeadRequest{
+		onceReq.Do(func() {
```
carpawell: why `Once` now? here and below

pkg/network/cache/clients.go
```diff
@@ -306,15 +325,44 @@ func newNodeTLSConfig(expectedKey *ecdsa.PublicKey, getClientCertificate func(*t
...
 type connections struct {
-	log    *zap.Logger
-	nodeID string
+	log                  *zap.Logger
+	nodeID               string
+	hasClientCertificate bool
+
+	mtx                   sync.RWMutex
+	m                     map[string]*endpointClient // keys are multiaddrs
+	mutuallyAuthenticated atomic.Bool
```
carpawell, line 337: so not signing stops working once there is a node with multiple addresses? even if we are using a secure connection to it (for a certain request)?

carpawell (commented):
I was confused a little with the number of "client" word in the code while reviewing. Sometimes it is about server's TLS status, sometimes it is about a remote node but they all are "clients" in the same code. But a can live with it.

pkg/network/cache/clients.go
End-rey: > so not signing stops working once there is a node with multiple addresses?

It still works when a node has multiple addresses, provided all of them use mTLS. 
Do we expect mixed mTLS and plaintext endpoints in practice? Supporting that case needs lazy signing if retry falls back to plaintext. Is the extra complexity worth it now?
@roman-khimov

cmd/neofs-node/mtls.go
End-rey: Renamed.

pkg/services/object/server.go
End-rey: Fixed.

pkg/network/cache/clients.go
carpawell: > a node has multiple addresses, provided all of them use mTLS

yeah, that is what i meant (if at least one with no TLS)

roman-khimov: Public addresses are expected to all be either TLS on non-TLS. In fact we don't yet have any real case for multiaddress node in map (local vs public listeners are different).

cmd/neofs-node/grpc.go
```diff
@@ -176,6 +176,8 @@ func buildSingleGRPCServer(c *cfg, sc grpcconfig.GRPC, maxRecvMsgSizeOpt grpc.Se
 			PermitWithoutStream: true,
 		}),
 		grpc.ForceServerCodecV2(iprotobuf.BufferedCodec{}),
+		grpc.UnaryInterceptor(trustInterNodePeerUnary(c)),
+		grpc.StreamInterceptor(trustInterNodePeerStream(c)),
```
roman-khimov: I hope this doesn't cost a lot.

pkg/services/object/server.go
```diff
@@ -1649,11 +1653,11 @@ func convertRangePrm(signer ecdsa.PrivateKey, cnr container.Container, req *prot
 		return p, nil
 	}
 
-	var onceResign sync.Once
 	meta := req.GetMetaHeader()
 	if meta == nil {
 		return getsvc.RangePrm{}, errors.New("missing meta header")
 	}
+	var onceResign sync.Once
```
roman-khimov: Moving this variable is not very helpful.

pkg/services/object/server.go
```diff
@@ -2096,10 +2103,6 @@ func (s *Server) ProcessSearch(ctx context.Context, req *protoobject.SearchV2Req
 				Ttl:     1,
 			},
 		}
-		if req.VerifyHeader, err = neofscrypto.SignRequestWithBuffer[*protoobject.SearchV2Request_Body](neofsecdsa.Signer(s.signer), req, nil); err != nil {
```
roman-khimov, line 2113: In non-TLS case this makes us sign the request multiple times instead of doing it once, that shouldn't be the case.

cmd/neofs-node/grpc.go
End-rey: Made netmap key lookup O(1) to reduce the per-request overhead, but it is indeed still there.
Handshake-time trust would be slightly cheaper, but requires tracking and closing inbound connections on netmap updates to revoke trust from already connected removed nodes.

pkg/services/object/server.go
End-rey: Fixed.

pkg/services/object/server.go
End-rey: Fixed.

cmd/neofs-node/grpc.go
roman-khimov: But why? We know there is a key X on the other side, OK. Then what matters is which permissions this key has and that's done relative to _container_ node list. Even if the node stays in netmap its status wrt different containers can change and this is not a problem, we're checking against the current known list anyway. So to me handshake-only key/account derivation is sufficient.

End-rey: Got it, sounds reasonable. Reworked it around handshake-time peer authentication.

roman-khimov (approved):

End-rey:
Found that `RoleContainer` is not suitable for `Range` ACL checks, so I excluded `Range` from the mTLS signature optimization.

PR #4102, opened 2026-07-28
Optimize ranged EC GET

cthulhu-rider:
noticed that GET EC part w/ range parameters works w/o buffering optimizations recently added to regular GET and almost deprecated RANGE

https://github.com/nspcc-dev/neofs-node/blob/c9eccda4d2313d167f13ac6dbda19c02e9a1f352/pkg/services/object/server.go#L1095-L1097

i guess these exclusions were added just for simplicity's sake. However, with the transition of secondary part receiving from `RANGE` to `GET(range, payload_only)`, this resulted in a regression of processing speed and allocations


pkg/services/object/server.go
```diff
@@ -1097,7 +1117,15 @@ func (s *Server) Get(req *protoobject.GetRequest, gStream protoobject.ObjectServ
 		return s.sendStatusGetResponse(gStream, err, needSignResp)
 	}
 
-	if hdrLen < 0 {
+	if stream == nil {
```
roman-khimov, line 1120: How can it happen?

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -539,50 +541,77 @@ func (t *FSTree) GetRangeStream(addr oid.Address, rng common.PayloadRange, readH
...
+	hf, err := iprotobuf.GetLENFieldBounds(prefix, protoobject.FieldObjectHeader)
 	if err != nil {
-		if stream != nil {
-			stream.Close()
+		return nil, 0, nil, fmt.Errorf("seek header field: %w", err)
+	}
+
+	var pldLen uint64
+	if !hf.IsMissing() {
+		hdrBin := prefix[hf.ValueFrom:hf.To]
```
roman-khimov, line 577: Not checking bounds.

pkg/services/object/server.go
cthulhu-rider: when data is non-local or optimization is not applied

pkg/local_object_storage/blobstor/fstree/fstree.go
cthulhu-rider: `GetLENFieldBounds` does

roman-khimov (approved):

PR #4103, opened 2026-07-28
Remote version and no origin for requests

roman-khimov:
Related to https://github.com/nspcc-dev/neofs-sdk-go/pull/833 and https://github.com/nspcc-dev/neofs-api/pull/402.

### DIFF
--- a/pkg/core/client/client.go
+++ b/pkg/core/client/client.go
@@ -10,6 +10,7 @@ import (
 	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	protorefs "github.com/nspcc-dev/neofs-sdk-go/proto/refs"
 	"github.com/nspcc-dev/neofs-sdk-go/reputation"
 	"github.com/nspcc-dev/neofs-sdk-go/user"
 	"google.golang.org/grpc"
@@ -50,4 +51,8 @@ type MultiAddressClient interface {
 	// ForAnyGRPCConn continues. If this happens on all endpoints, ForAnyGRPCConn
 	// returns [ErrAllConnectionsSkipped].
 	ForAnyGRPCConn(ctx context.Context, f func(context.Context, *grpc.ClientConn) error) error
+
+	// APIVersion returns current protocol version to use in requests using
+	// this client (min(server, client) effectively).
+	APIVersion() *protorefs.Version
 }
--- a/pkg/network/cache/clients.go
+++ b/pkg/network/cache/clients.go
@@ -239,7 +239,7 @@ func (x *Clients) initConnection(ctx context.Context, pub []byte, uri string) (*
 	if err != nil { // should never happen
 		return nil, fmt.Errorf("init gRPC client conn: %w", err)
 	}
-	res, err := client.NewGRPC(ctx, pub, grpcConn, x.signBufPool, x.streamMsgTimeout)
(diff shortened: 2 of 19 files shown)


pkg/core/client/client.go
```diff
@@ -50,4 +51,6 @@ type MultiAddressClient interface {
 	// ForAnyGRPCConn continues. If this happens on all endpoints, ForAnyGRPCConn
 	// returns [ErrAllConnectionsSkipped].
 	ForAnyGRPCConn(ctx context.Context, f func(context.Context, *grpc.ClientConn) error) error
+
+	APIVersion() *protorefs.Version
```
carpawell, line 57: no docs?

pkg/network/cache/clients.go
```diff
@@ -251,15 +259,24 @@ func (x *Clients) initConnection(ctx context.Context, pub []byte, uri string) (*
...
 
 	if got := resp.NodeInfo().PublicKey(); !bytes.Equal(got, pub) {
 		_ = grpcConn.Close()
-		return nil, clientcore.ErrWrongPublicKey
+		return nil, nil, clientcore.ErrWrongPublicKey
+	}
+
+	var (
+		nodeAPI   = resp.LatestVersion()
+		clientAPI = version.Current()
```
carpawell, line 272: what if node's version is higher? making a request that we cannot process potentially?

pkg/services/object/server.go
```diff
@@ -1647,10 +1647,12 @@ func convertRangePrm(signer ecdsa.PrivateKey, cnr container.Container, req *prot
 	p.SetTransportFunc(func(ctx context.Context, c clientcore.MultiAddressClient) error {
 		var err error
 		onceResign.Do(func() {
-			req.MetaHeader = &protosession.RequestMetaHeader{
-				// TODO: #1165 think how to set the other fields
-				Ttl:    meta.GetTtl() - 1,
-				Origin: meta,
+			req = &protoobject.GetRangeRequest{
```
carpawell, line 1625: why do we override the whole `req`, not just its header now?

pkg/services/object/put/distributed.go
```diff
@@ -673,7 +673,7 @@ func (t *distributedTarget) sendObject(obj object.Object, encObj encodedObject,
 		return nil
 	}
 
-	if t.relay != nil {
+	if !t.localNodeInContainer && t.relay != nil {
```
carpawell: cant we do not init `t.relay` on the upper levels and do not change this line? i think `nil` `t.relay` equals `!t.localNodeInContainer`

pkg/network/cache/clients.go
roman-khimov: Why? We'll just use SDK version, whatever that is. Supposedly server knows how to handle older requests and we don't know anything newer anyway.

pkg/services/object/server.go
roman-khimov: It's a different request now and that's what GET does for example. This works for RANGE anyway.

pkg/services/object/put/distributed.go
roman-khimov: Let's try.

pkg/core/client/client.go
roman-khimov: Fixed.

pkg/network/cache/clients.go
carpawell: > We'll just use SDK version, whatever that is

that is what i meant

> Supposedly server knows how to handle older requests

but likely that is all we can do, yes

carpawell (approved):

PR #4104, opened 2026-07-29
Optimize PUT forwarding signatures

cthulhu-rider:


cthulhu-rider:
everything is ok for chunks. Doing the same for headers leads to a request loop, apparently due to the logic being tied to TTL

roman-khimov:
How about https://github.com/nspcc-dev/neofs-node/pull/4103/commits/41bdb9e8c2af0cceb702ca6bd299eddd73e62aa3?

cthulhu-rider:
missed this

PR #4105, opened 2026-07-29
Ignore PUT chunk signatures

cthulhu-rider:


roman-khimov (changes requested):
This can be done only for sliced objects, if we're slicing in node signatures is the only way to ensure the data is what client intended to send.

roman-khimov:
Also, PUT performance is a different story, we have _a lot_ to optimize there.

PR #4106, opened 2026-07-30
sn/ec: do not encode original object if there are only EC rules

carpawell:


roman-khimov (approved):

carpawell:
https://rest.fs.neo.org/Aku38wpn2xEqfEPxmwyAa7N3VnAkMV8PEVkpLqwKM1z3/5724-1785425851/index.html

PR #4110, opened 2026-07-31
Fix/respond with api version only if server is older

carpawell:


cmd/neofs-node/reputation.go
```diff
@@ -256,16 +257,20 @@ type reputationServer struct {
 	routeBuilder       reputationrouter.Builder
 }
 
-func (s *reputationServer) makeResponseMetaHeader(st *protostatus.Status) *protosession.ResponseMetaHeader {
+func (s *reputationServer) makeResponseMetaHeader(st *protostatus.Status, needVersionInResponse bool) *protosession.ResponseMetaHeader {
```
roman-khimov: Better pass request meta header into it (and others) and handle all the other things internally. Theoretically there can be more of version-dependent logic here like different status codes or fields, not just a boolean. And it's easier to pass `req.MetaHeader` anyway.

pkg/services/accounting/server.go
```diff
@@ -44,10 +45,14 @@ func New(s *ecdsa.PrivateKey, c BalanceContract) protoaccounting.AccountingServi
 }
 
 func (s *server) makeBalanceResponse(body *protoaccounting.BalanceResponse_Body, st *protostatus.Status, req *protoaccounting.BalanceRequest) (*protoaccounting.BalanceResponse, error) {
+	var v *protorefs.Version
+	if util.NeedVersionInResponse(req.MetaHeader) {
```
roman-khimov, line 49: Like here, btw.

roman-khimov (approved):

PR #4113, opened 2026-07-31
Fix SIGHUP

End-rey:
Noticed some issues when adding a new configuration to gRPC.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Changelog for NeoFS Node
 ### Fixed
 - SN could panic on forwarding GET/HEAD/RANGE request (#4120)
 - Storage node shuts down when SIGHUP configuration reload fails (#4113)
+- Storage node shuts down gracefully on invalid node attributes during SIGHUP (#4113)
 
 ### Changed
 - SNs exchange TLS certificates on inter-node connections (#4097)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Changelog for NeoFS Node
 
 ### Fixed
 - SN could panic on forwarding GET/HEAD/RANGE request (#4120)
+- Storage node shuts down when SIGHUP configuration reload fails (#4113)
 
 ### Changed
 - SNs exchange TLS certificates on inter-node connections (#4097)
--- a/cmd/neofs-node/config.go
+++ b/cmd/neofs-node/config.go
@@ -624,7 +624,6 @@ func (c *cfg) needBootstrap() bool {
 }
 
 func (c *cfg) configWatcher(ctx context.Context) {
-	var err error
 	ch := make(chan os.Signal, 1)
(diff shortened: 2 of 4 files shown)


roman-khimov (commented):
Create an issue for neofs-testcases to test these scenarios.

cmd/neofs-node/config.go
```diff
@@ -642,73 +641,78 @@ func (c *cfg) configWatcher(ctx context.Context) {
...
+	oldProfiler := writeProfilerConfig(oldCfg)
 
-			// nolint:contextcheck
-			c.reloadMetricsAndPprof(oldMetrics, oldProfiler)
+	newCfg, err := config.New(config.WithConfigFile(oldCfg.Path()))
+	if err != nil {
+		return fmt.Errorf("read configuration: %w", err)
+	}
+	if err := validateConfig(newCfg); err != nil {
+		return fmt.Errorf("validate configuration: %w", err)
```
carpawell, line 667: not sure if running with an incorrect config is ok. maybe app should die? the only feedback we have there is log message, otherwise: app is healthy, and systemd thinks we are "Ready"

roman-khimov: Aren't we running with an old config in this case?

carpawell: yes, but we have a new config file (envs), and the app is still running with already unknown old config

End-rey: Agree. A failed reload can leave the runtime state out of sync with the configuration, so reporting `READY=1` is misleading. I think stopping the node on reload failure would be a good idea.

carpawell: i am fully ok with stopping

carpawell (approved):
Conflicts

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Changelog for NeoFS Node
 - SN could panic on forwarding GET/HEAD/RANGE request (#4120)
 - gRPC SIGHUP reload no longer stops a running public API before replacement configuration is verified (#4113)
 - Storage node shuts down when SIGHUP configuration reload fails (#4113)
+- Invalid node attributes on SIGHUP no longer terminate a storage node (#4113)
```
roman-khimov: This looks _so_ inconsistent... We're shutting down the node on SIGHUP reload failure and we're not terminating it on SIGHUP in case of incorrect attributes. So what do we want from SIGHUP, really?

roman-khimov (commented):
Create an issue for neofs-testcases to test these scenarios.

CHANGELOG.md
End-rey: Reworded accordingly.

End-rey:
The gRPC-specific fixes no longer make sense because any SIGHUP reload failure now triggers a graceful node shutdown.

roman-khimov (approved):

PR #4114, opened 2026-07-31
node: add option to use self signed certs

### LINKED ISSUE #4090: Add an option to use self-signed certificate for TLS gRPC endpoint
Currently we require key/certificate pair, add an option to use self-signed certificate with node ECDSA key.

End-rey:
Closes #4090.


End-rey:
I think #4131 should go first.

cmd/neofs-node/config/grpc/grpc.go
```diff
@@ -12,6 +12,7 @@ type GRPC struct {
 type TLS struct {
 	Enabled     bool   `mapstructure:"enabled"`
 	Certificate string `mapstructure:"certificate"`
+	SelfSigned  bool   `mapstructure:"self_signed"`
```
roman-khimov: This can be made implicit, no `certificate` with `enabled=true` is exactly that. Of course node should log the fact that self-signed certificate is used in this case.

End-rey: Done.

roman-khimov (approved):

Issue #4005, opened 2026-06-02
Clean up code for writing service messages


cthulhu-rider:
why all these func are in the `service/object`? why not somewhere in `iprotobuf` lib?

_Originally posted by @carpawell in https://github.com/nspcc-dev/neofs-node/pull/3996#discussion_r3319877041_
            
make a lib with reusable generic code. Likely it will go to SDK. In particular, https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go@v1.0.0-rc.19/client pkg definitely needs it

cthulhu-rider:
inited by #4116. Apply the same to GET

cthulhu-rider:
take care of responses as well, like https://github.com/nspcc-dev/neofs-node/blob/c5da094d8f6895ea3390790b1663be4c68a72656/pkg/services/object/get.go#L1073

PR #4116, opened 2026-08-05
Optimize buffering of forwarded SEARCH requests

cthulhu-rider:


pkg/services/object/search.go
```diff
@@ -85,3 +110,62 @@ func (s *Server) forwardSearchRequest(ctx context.Context, req any, nodeSets [][
...
+func writeSearchRequestFields(buf []byte, bodyLen int, body *protoobject.SearchV2Request_Body, metaHdrLen int, metaHdr *protosession.RequestMetaHeader, verifHdrLen int, verifHdr *protosession.RequestVerificationHeader) int {
+	off := writeStablyMarshalledField(buf, protobuf.TagBytes1, bodyLen, body)
+	off += writeStablyMarshalledField(buf[off:], protobuf.TagBytes2, metaHdrLen, metaHdr)
+	off += writeStablyMarshalledField(buf[off:], protobuf.TagBytes3, verifHdrLen, verifHdr)
+	return off
+}
+
+func (s *Server) writeLocalSearchObjectsRequest(buf []byte, bodyLen int, body *protoobject.SearchV2Request_Body, metaHdrLen int) (int, error) {
+	// TODO: can be calculated once and reused
+	originSig, err := neofsecdsa.Signer(s.signer).Sign(nil)
```
roman-khimov: Not needed for API 2.25.

roman-khimov (commented):
We need to share this code with SDK, it's unmaintainable otherwise.

pkg/services/object/search.go
cthulhu-rider: right. Lets handle #4118 first then

cthulhu-rider:
> We need to share this code with SDK, it's unmaintainable otherwise.

this is planned to be done in #4005

roman-khimov:
Let's solve #4005 first then.

cthulhu-rider:
> Let's solve #4005 first then.

i don't think it's worth blocking optimizations with refactorings

roman-khimov:
It's too easy to lose control over the codebase this way, we need to structure our changes appropriately as early as possible.

cthulhu-rider:
> It's too easy to lose control over the codebase this way, we need to structure our changes appropriately as early as possible.

agree. But at the same time the more usecases are created, the easier it will be to form libs

i'll try my best to synthesize SDK packages asap

pkg/services/object/search.go
cthulhu-rider: done

pkg/services/object/get.go
```diff
@@ -888,7 +888,7 @@ func (s *Server) makeGetECPartRequest(remoteServerAPIVersion *protorefs.Version,
 		return nil, err
 	}
 	if n != len(buf) {
-		return nil, fmt.Errorf("got wrong request length: expected %d, got %d", n, len(buf))
+		return nil, newWrongRequestLengthError(len(buf), n)
```
carpawell: why in this commit?

pkg/services/object/search.go
```diff
@@ -48,7 +54,26 @@ func iterateSearchableContainerNodes(nodeSets [][]netmap.NodeInfo, repRules []ui
...
+	metaHdrLen := req.MetaHeader.MarshaledSize()
+	verifHdrLen := req.VerifyHeader.MarshaledSize()
+
+	reqLen := calculateSearchRequestLength(bodyLen, metaHdrLen, verifHdrLen)
+
+	// default gRPC buffer pool used to serialize generated message structures
+	bufferPool := mem.DefaultBufferPool()
+	poolItem := bufferPool.Get(reqLen)
+	defer bufferPool.Put(poolItem)
+	buf := (*poolItem)[:reqLen]
```
carpawell: it can be bigger even after `bufferPool.Get(reqLen)` call?

pkg/services/object/get.go
cthulhu-rider: one more usecase appears

pkg/services/object/search.go
cthulhu-rider: no it should not, dropped

pkg/services/object/proto.go
```diff
@@ -29,6 +29,9 @@ const (
 	maxChunkOffsetInGetResponse  = 1 + maxGetResponseChunkVarintLen + // 1 for iprotobuf.TagBytes1
 		1 + maxGetResponseChunkVarintLen // 1 for iprotobuf.TagBytes2
 	getResponseChunkBufferLen = maxChunkOffsetInGetResponse + maxGetResponseChunkLen + maxResponseVerificationHeaderLen
+
+	modernRequestVerificationSignatureCount = 2
```
carpawell: make it api dependent? we will have https://github.com/nspcc-dev/neofs-api/issues/406 soon, it will then be "the most modern"

pkg/services/object/proto.go
```diff
@@ -349,3 +352,14 @@ func writeECDSAWithSHA512Signature(buf []byte, pubKey, sig []byte) int {
 	off++
 	return off + copy(buf[off:], sig) // scheme is 0
 }
+
+type stablyMarshalledFIeld interface {
+	MarshalStable([]byte)
+}
+
+func writeStablyMarshalledField(buf []byte, tag byte, ln int, fld stablyMarshalledFIeld) int {
```
carpawell: can `ln` and `fld` merged into a single interface? likely every our struct will handle it ok

pkg/services/object/get.go
carpawell: what this error means for a user btw? isnt it a panic?

pkg/services/object/server.go
```diff
@@ -2141,6 +2204,7 @@ func (s *Server) ProcessSearch(ctx context.Context, req *protoobject.SearchV2Req
 			sets = append(sets, searchRes.set)
 			mores = append(mores, searchRes.more)
 		}
+		freeBuffers()
```
carpawell: why need it if we have `defer`? to free asap?

carpawell (approved):

pkg/services/object/proto.go
cthulhu-rider: it is API dependent. To support new API feature, future change is inevitable

pkg/services/object/server.go
cthulhu-rider: > to free asap?

exactly

pkg/services/object/proto.go
cthulhu-rider: length of embedded messages is always needed twice: to calc whole message len and to write field prefix. So, i dont like the idea to calc it twice

pkg/services/object/get.go
cthulhu-rider: internal server error. Can be a panic. Should never happen anyway

roman-khimov (changes requested):
Needs to be shared with SDK code.

cthulhu-rider:
blocked by:
- https://github.com/nspcc-dev/neofs-sdk-go/pull/858
- #4144

pkg/services/object/search.go
```diff
@@ -5,19 +5,32 @@ import (
...
+	protosession "github.com/nspcc-dev/neofs-sdk-go/proto/session"
+	"github.com/nspcc-dev/neofs-sdk-go/version"
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/mem"
 )
 
+const defaultSearchRequestBufferLength = 1 << 10
+
+var searchRequestBufferPool = sync.Pool{
```
roman-khimov: Can't you use the default grpc buffer pool here?

pkg/services/object/search.go
```diff
@@ -85,3 +119,16 @@ func (s *Server) forwardSearchRequest(ctx context.Context, req any, nodeSets [][
 
 	return nil, nil
 }
+
+func writeLocalSearchRequestMetaHeader(buf []byte, apiVersion version.Version) int {
+	return protosession.WriteRequestMetaHeaderToRequest(buf, apiVersion.Major(), apiVersion.Minor(), 1, 0, nil, nil, 0, nil, 0, nil, 0, 0, nil)
```
roman-khimov: Looks like Morse code.

pkg/services/object/server.go
```diff
@@ -2139,15 +2140,111 @@ func (s *Server) ProcessSearch(ctx context.Context, req *protoobject.SearchV2Req
...
+
+			// need to lower API version
+
+			metaHdrLen := calculateLocalSearchRequestMetaHeaderLength(ver)
+
+			verCmp := ver.Compare(version.New(2, 25))
+
+			var reqLen int
+			switch verCmp {
+			case 1: // unsigned
```
roman-khimov: It's a single signature, not unsigned. But it seems to be a dead code given that `localVersion` is 2.26.

pkg/services/object/server.go
```diff
@@ -2139,15 +2140,111 @@ func (s *Server) ProcessSearch(ctx context.Context, req *protoobject.SearchV2Req
...
+			if err != nil {
+				return nil, fmt.Errorf("sign request body: %w", err)
+			}
+			metaSig, err := signECDSAWithSHA512(s.signer, reqBuf[off-metaHdrLen:off])
+			if err != nil {
+				return nil, fmt.Errorf("sign request meta header: %w", err)
+			}
+			var originSig []byte
+			if verCmp < 0 { // with origin signature
+				originSig, err = signECDSAWithSHA512(s.signer, nil)
```
roman-khimov: Why all of this code is not performed via some function accepting body/meta slices and version?

pkg/services/object/server.go
```diff
@@ -2139,15 +2140,111 @@ func (s *Server) ProcessSearch(ctx context.Context, req *protoobject.SearchV2Req
...
+		var reqBufItem *[]byte
+		var reqBuf []byte
+		if reqLen <= defaultSearchRequestBufferLength {
+			reqBufItem = searchRequestBufferPool.Get().(*[]byte)
+			reqBuf = (*reqBufItem)[:reqLen]
+		} else {
+			reqBuf = make([]byte, reqLen)
 		}
+
+		free := func() {
```
roman-khimov: `free` is too generic to me.

pkg/services/object/server.go
```diff
@@ -2185,6 +2282,7 @@ func (s *Server) ProcessSearch(ctx context.Context, req *protoobject.SearchV2Req
 			sets = append(sets, searchRes.set)
 			mores = append(mores, searchRes.more)
 		}
+		free()
```
roman-khimov: Isn't it deferred?

pkg/services/object/server.go
cthulhu-rider: u mean split body and meta buffers?

pkg/services/object/server.go
cthulhu-rider: dropped

pkg/services/object/server.go
cthulhu-rider: dropped

pkg/services/object/server.go
roman-khimov: Splitting isn't needed, subslicing is easy to pass appropriate chunks into a function. Can be slice + a pair of offsets, but a pair of slices is more convenient to me. In any event, it's like 184th time I'm looking at code that checks version and adds signatures accordingly with all of the associated things like error checks.

pkg/services/object/search.go
cthulhu-rider: done

pkg/services/object/server.go
cthulhu-rider: yeah, this part will die soon and i'd like to drop it then

roman-khimov:
Linter.

pkg/services/object/server.go

pkg/services/object/get.go
```diff
@@ -835,7 +835,8 @@ func handleGetECPartResponseInit(buffers iprotobuf.BuffersSlice) (iprotobuf.Buff
 func (x *getECTransport) CopyRemoteECPartRange(ctx context.Context, conn clientcore.MultiAddressClient, partInfo iec.PartInfo, off uint64, ln uint64, full bool, controlCh <-chan bool) (uint64, error) {
 	var copiedPld uint64
 
-	connAPIVersion := conn.APIVersion()
+	connAPIVersionMsg := conn.APIVersion()
+	connAPIVersion := version.New(connAPIVersionMsg.GetMajor(), connAPIVersionMsg.GetMinor())
```
roman-khimov, line 839: Previous API was made to optimize out version->protorefs conversion, now we have protorefs->version. Maybe we need to change the API then.

roman-khimov (approved):

Issue #4019, opened 2026-06-16
Policer work can break availability of objects (EC)


cthulhu-rider:
when objects are PUT on reserve nodes using , some objects cannot be GET

this seems to happen at the moment when:
- GET server iterates over nodes in descending order of weight
- Policer moves object to nodes of greater weight
concurrently

## Expected Behavior
no GET errors

## Current Behavior
```
error	handler/util.go:35	call method	{"status": 500, "request_id": "53b75c45-e3da-44ee-931a-6f2d57142fe7", "method": "GetObject", "bucket": "6a2829a8-ec-size4k-threads128-srv1", "object": "ix(P2rTj/66.hHNY9THBXPcH0jbD.rnd", "description": "could write object output", "error": "status: code = 2052 message = object already removed"}
```
while neither DELETE nor expirations happen

and sometimes
```
error	handler/util.go:35	call method	{"status": 404, "request_id": "af4987ce-0a7b-42fe-9f6c-1cf6cae5ea0c", "method": "GetObject", "bucket": "6a2829a8-ec-size4096k-threads128-srv1", "object": "j2yjyvAN/48.d7JnKkSRr246sq2J.rnd", "description": "could not find object", "error": "NoSuchKey: 404 => The specified key does not exist."}
```
which is less strange but also unexpected 

## Possible Solution
1. when redundant object copy is dropped, do not mark it as DELETE/expired (if so)
2. return pure 404 if and only if all nodes responded 404. If at least one SN is unavailable, reflect this in the status

## Steps to Reproduce (for bugs)
happens while using S3 gateway after concurrent PUT

errors do not occur when Policer is disabled (synthesynthetically)

## Regression
no info

## Your Environment
* Version used: https://github.com/nspcc-dev/neofs-node/releases/tag/v0.53.0
* Server setup and configuration: 8xSN, EC6/2

cthulhu-rider:
guess REP placements have the same problem

roman-khimov:
1. We can have different garbage marks ("reason") and respond with proper object if it's not tombstoned or expired (it can be detected directly for simple objects, but split ones complicate things somewhat and we still don't want to return improper objects).
2. Object assembly code can handle errors better to retry with other nodes.
3. https://github.com/nspcc-dev/neofs-api/issues/348 can help somewhat.

PR #4119, opened 2026-08-07
Restore EC objects during policer relocation

End-rey:
Refs #4019.

### DIFF
--- a/pkg/services/object/get/ec.go
+++ b/pkg/services/object/get/ec.go
@@ -32,6 +32,12 @@ import (
 
 type tooManyPartsUnavailableError int
 
+type partialECObject struct {
+	hdr    object.Object
+	gotHdr bool
+	parts  [][]byte
+}
+
 func (x tooManyPartsUnavailableError) Error() string {
 	return strconv.Itoa(int(x)) + " data parts unavailable"
 }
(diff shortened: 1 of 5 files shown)


pkg/services/object/get/ec.go
```diff
@@ -370,9 +384,181 @@ nextPart:
...
+	for nodeIdx := range sortedNodes {
+		var (
+			hdr object.Object
+			rc  io.ReadCloser
+			err error
+		)
+		if s.neoFSNet.IsLocalNodePublicKey(sortedNodes[nodeIdx].PublicKey()) {
+			hdr, rc, err = s.localObjects.GetECPart(ctx, cnr, parent, iec.PartInfo{RuleIndex: ruleIdx, Index: nodeIdx}, true)
+		} else {
+			hdr, rc, err = s.getAnyECPartFromNode(ctx, cnr, parent, ruleIdx, sortedNodes[nodeIdx])
```
roman-khimov: Sequential? Slow.

pkg/services/object/get/ec.go
```diff
@@ -370,9 +384,181 @@ nextPart:
...
+	partial := partialECObject{parts: make([][]byte, rule.DataPartNum+rule.ParityPartNum)}
+	obj, err := s.restoreFromECPartsByRuleStrict(ctx, cnr, parent, rule, ruleIdx, sortedNodes, &partial)
+	if err == nil || errors.Is(err, apistatus.ErrObjectAlreadyRemoved) || errors.Is(err, apistatus.ErrObjectAccessDenied) || errors.Is(err, ctx.Err()) || errors.As(err, new(*object.SplitInfoError)) {
+		return obj, err
+	}
+
+	s.log.Info("failed to restore EC object using exact part placement, trying any part per node",
+		zap.Stringer("container", cnr), zap.Stringer("object", parent), zap.Stringer("rule", rule), zap.Int("ruleIdx", ruleIdx), zap.Error(err))
+
+	return s.restoreFromAnyECPartsByRule(ctx, cnr, parent, rule, ruleIdx, sortedNodes, partial)
```
roman-khimov, line 397: I hope we can deduplicate these strategies in future, non-indexed one is more reliable in general, indexed one can't solve the case of part moving from node B to node A (with complete deletion on B so that garbage marks won't help) while object getter moves from A to B. But this requires more performance tests.

pkg/services/object/get/ec.go
End-rey: Done, made concurrent.

roman-khimov:
Duplicates #4125.

PR #4121, opened 2026-08-07
Feat/versioned netmap

carpawell:


carpawell:
Fixed things here and there, should be ready. But maybe parsing neofs-cli's output will fail from now.

carpawell:
@evgeniiz321, do you know why deploying storage nodes keeps `Error: can't fetch private key: missing key flag` error message?

roman-khimov:
https://github.com/nspcc-dev/neofs-testcases/issues/1430

cmd/neofs-node/netmap.go
```diff
@@ -176,19 +181,30 @@ func initNetmapService(c *cfg) {
...
+				ni, err = c.netmapLocalNodeState(uint64(v), epoch)
 				if errors.Is(err, rpcclient.ErrWSConnLost) {
 					return backoff.Permanent(err)
 				}
 				return err
 			},
 			expBackoff,
 			func(err error, d time.Duration) {
-				c.log.Info("retrying due to error", zap.Uint64("epoch", e), zap.Error(err), zap.Duration("retry-after", d))
+				c.log.Info("retrying due to error", zap.Error(err), zap.Duration("retry-after", d))
```
roman-khimov: log map version here?

cmd/neofs-node/netmap.go
roman-khimov: Heartbeats still needed per epoch.

roman-khimov (changes requested):

cmd/neofs-node/netmap.go
carpawell: added

roman-khimov (approved):

roman-khimov:
Seems like node removal can be problematic with the new code, see test reports.

carpawell:
Code that decides what to do with local state/actual netmap state/state that should be shown for controls svc's user is bad. Fixed somehow.

PR #4122, opened 2026-08-10
Differentiate garbage marks

### LINKED ISSUE #4019: Policer work can break availability of objects (EC)

End-rey:
Closes #4019.

### DIFF
--- a/pkg/local_object_storage/engine/delete.go
+++ b/pkg/local_object_storage/engine/delete.go
@@ -4,19 +4,29 @@ import (
 	"context"
 	"slices"
 
+	meta "github.com/nspcc-dev/neofs-node/pkg/local_object_storage/metabase"
 	"github.com/nspcc-dev/neofs-node/pkg/local_object_storage/shard"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"go.uber.org/zap"
 )
 
+// GarbageMark specifies why an object is marked for garbage collection.
+type GarbageMark = meta.GarbageMark
+
+const (
+	GarbageMarkDefault   = meta.GarbageMarkDefault
+	GarbageMarkRedundant = meta.GarbageMarkRedundant
+)
+
 // Delete marks the objects to be removed. This is a forced removal, it
 // overrides any locks or other reasons to keep the object, but it doesn't
 // delete the object immediately (like [StorageEngine.Drop] does), instead it
 // just marks the address for later removal by GC according to GC settings.
+// Redundant objects remain readable until they are physically removed by GC.
 //
 // Returns an error if executions are blocked (see BlockExecution).
-func (e *StorageEngine) Delete(_ context.Context, addr oid.Address) error {
+func (e *StorageEngine) Delete(_ context.Context, addr oid.Address, mark GarbageMark) error {
 	if e.metrics != nil {
 		defer elapsed(e.metrics.AddDeleteDuration)()
 	}
@@ -28,7 +38,7 @@ func (e *StorageEngine) Delete(_ context.Context, addr oid.Address) error {
 		return e.blockErr
 	}
 	return e.processAddrDelete(addr, func(sh *shard.Shard, cnr cid.ID, addrs []oid.ID) error {
-		return sh.MarkGarbage(cnr, addrs)
+		return sh.MarkGarbage(cnr, addrs, mark)
 	})
 }
 
@@ -83,7 +93,7 @@ func (e *StorageEngine) DeleteRedundantCopies(_ context.Context, addr oid.Addres
 		e.log.Info("removing redundant local object copy",
 			zap.String("keeper_shard", keeperShard),
 			zap.Stringer("redundant_shard", sh.ID()))
-		return sh.MarkGarbage(cnr, addrs)
+		return sh.MarkGarbage(cnr, addrs, GarbageMarkRedundant)
 	})
 }
 
--- a/pkg/local_object_storage/metabase/exists.go
+++ b/pkg/local_object_storage/metabase/exists.go
@@ -227,9 +227,9 @@ func inGarbage(metaCursor *bbolt.Cursor, id oid.ID) uint8 {
 	}
 
 	var garbageMark = mkGarbageKey(id)
-	k, _ := metaCursor.Seek(garbageMark)
+	k, v := metaCursor.Seek(garbageMark)
 
-	if bytes.Equal(k, garbageMark) {
+	if bytes.Equal(k, garbageMark) && !bytes.Equal(v, redundantGarbageMark) {
 		// object has been marked with GC
 		return statusGCMarked
 	}
--- a/pkg/local_object_storage/metabase/inhume.go
+++ b/pkg/local_object_storage/metabase/inhume.go
@@ -22,8 +22,32 @@ type ContainerGarbageDiff struct {
 	PayloadDiff int64
 }
 
-// MarkGarbage marks objects to be physically removed from shard.
-func (db *DB) MarkGarbage(cnr cid.ID, addrs []oid.ID) (ContainerGarbageDiff, error) {
+// GarbageMark specifies why an object is marked for garbage collection.
+type GarbageMark byte
+
+const (
+	GarbageMarkDefault GarbageMark = iota
+	GarbageMarkRedundant
+)
+
+var redundantGarbageMark = []byte{byte(GarbageMarkRedundant)}
+
+func (m GarbageMark) valid() bool {
+	switch m {
+	case GarbageMarkDefault, GarbageMarkRedundant:
+		return true
+	default:
+		return false
+	}
+}
+
+// MarkGarbage marks objects to be physically removed from shard. Redundant
+// objects remain readable until they are physically removed by GC.
+func (db *DB) MarkGarbage(cnr cid.ID, addrs []oid.ID, mark GarbageMark) (ContainerGarbageDiff, error) {
+	if !mark.valid() {
+		return ContainerGarbageDiff{}, fmt.Errorf("invalid garbage mark %d", mark)
+	}
+
 	db.modeMtx.RLock()
 	defer db.modeMtx.RUnlock()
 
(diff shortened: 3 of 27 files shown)


pkg/local_object_storage/engine/delete.go
```diff
@@ -17,6 +17,16 @@ import (
 //
 // Returns an error if executions are blocked (see BlockExecution).
 func (e *StorageEngine) Delete(_ context.Context, addr oid.Address) error {
+	return e.delete(addr, (*shard.Shard).MarkGarbage)
+}
+
+// DeleteForPolicer marks a redundant policer copy for physical removal while
+// keeping it readable until GC removes its payload.
+func (e *StorageEngine) DeleteForPolicer(_ context.Context, addr oid.Address) error {
```
roman-khimov: What other callers of `Delete()` are known? It looks like we can make this the default `Delete()` behavior since tombstones and expirations are processed in a different manner anyway and things deleted via `Delete()` can safely be available for some time.

pkg/local_object_storage/metabase/exists.go
```diff
@@ -134,6 +134,9 @@ func objectStatusDirect(metaCursor *bbolt.Cursor, oID oid.ID, currEpoch uint64)
 	}
 
 	garbageStatus := inGarbage(metaCursor, oID)
+	if garbageStatus == statusGCMarked && isPolicerGarbage(metaCursor, oID) {
```
roman-khimov: That's two lookups for the same key, should be optimized to a single one (`inGarbage()` can have more statuses if needed or it can be `statusAvailable` since it really is available and only GC should care about the respective object.

pkg/local_object_storage/metabase/inhume.go
```diff
@@ -22,8 +22,20 @@ type ContainerGarbageDiff struct {
 	PayloadDiff int64
 }
 
+var policerGarbageMark = []byte("policer")
```
roman-khimov: Too big, a single byte is sufficient (for other reasons as well, we can have more marks). Can also be named like `redundant` (vs. `deleted` for tombstone-marked ones).

pkg/local_object_storage/metabase/inhume.go
```diff
@@ -22,8 +22,20 @@ type ContainerGarbageDiff struct {
...
+
 // MarkGarbage marks objects to be physically removed from shard.
 func (db *DB) MarkGarbage(cnr cid.ID, addrs []oid.ID) (ContainerGarbageDiff, error) {
+	return db.markGarbage(cnr, addrs, nil)
+}
+
+// MarkGarbageForPolicer marks redundant policer copies for physical removal.
+// Such copies remain readable until they are removed by GC.
+func (db *DB) MarkGarbageForPolicer(cnr cid.ID, addrs []oid.ID) (ContainerGarbageDiff, error) {
```
roman-khimov: Same question, can we have a single method for all purposes?

pkg/local_object_storage/engine/delete.go
End-rey: Unified via explicit `GarbageMark`.

pkg/local_object_storage/metabase/exists.go
End-rey: Done.

pkg/local_object_storage/metabase/inhume.go
End-rey: Done.

pkg/local_object_storage/metabase/inhume.go
End-rey: Done.

pkg/services/object/get/ec.go
carpawell: what is "an availability error"?

carpawell, line 372: docs can be adjusted now?

carpawell: is it expected? we requested EC part but got a non-regular object

carpawell: received object ID and other err context may be useful. it is just `part index is set, rule index is not` or smth like that now

carpawell: i think it is the second time you decode attributes for the remote nodes case (`getAnyECPartFromNode`). can it be done in a single place?

carpawell: as i see, in the end there is always `dst ObjectWriter` but even for the happy path we are collecting all the EC parts in memory without streaming them, right? and after collecting them, we are even concatenating them into a single slice?

pkg/services/object/get/get.go
carpawell: what this part do? i think it is better to extend/change `checkECPartInfoGetRequest` to get what you need there without repeating

pkg/local_object_storage/metabase/inhume.go
```diff
@@ -22,8 +22,19 @@ type ContainerGarbageDiff struct {
...
+const (
+	GarbageMarkDefault GarbageMark = iota
+	GarbageMarkRedundant
+)
+
+var redundantGarbageMark = []byte{byte(GarbageMarkRedundant)}
+
+// MarkGarbage marks objects to be physically removed from shard. Redundant
+// objects remain readable until they are physically removed by GC.
+func (db *DB) MarkGarbage(cnr cid.ID, addrs []oid.ID, mark GarbageMark) (ContainerGarbageDiff, error) {
```
carpawell, line 46: i would check that the mark is known. otherwise, we may end up with a broken metabase with deleted objects no one knows why

carpawell (commented):
We have implemented https://github.com/nspcc-dev/neofs-api/issues/348 but not used anywhere as its usage was questionable. Then we have an optimistic https://github.com/nspcc-dev/neofs-node/pull/4125 that always tries to GET without indexes. Then we have this PR that drops indexes only as a fallback when normal placement does not work. So what do we do? @roman-khimov

carpawell:
Conflicts btw.

roman-khimov:
GC marks are separate from #4119 to me. And #4119 itself has conceptual problems, think of REP handling wrt the same issue and object moved from B to A while fetcher goes from A to B. Nothing can help in this case, only some larger window of availability on B relative to GC doing its job. And it's exactly the same problem as with EC, so EC _should not_ do any retries. Then #4125 is a different beast, it tries a different strategy and if it works better --- it should just be used as a replacement for current one.

pkg/local_object_storage/metabase/inhume.go
End-rey: Done.

pkg/local_object_storage/metabase/exists.go
```diff
@@ -227,9 +227,12 @@ func inGarbage(metaCursor *bbolt.Cursor, id oid.ID) uint8 {
 	}
 
 	var garbageMark = mkGarbageKey(id)
-	k, _ := metaCursor.Seek(garbageMark)
+	k, v := metaCursor.Seek(garbageMark)
 
 	if bytes.Equal(k, garbageMark) {
```
roman-khimov: Can be just `equal && equal`, simpler this way and more future-proof (imagine adding more checks after garbage).

pkg/services/object/get/remote.go
```diff
@@ -31,6 +31,9 @@ func (exec *execCtx) processNode(info netmap.NodeInfo) bool {
 	default:
 		exec.status = statusUndefined
 		exec.err = apistatus.ErrObjectNotFound
+		if errors.Is(err, ErrUnavailableNode) {
```
roman-khimov: 1. Can be handled in `switch`
2. Why this difference matters to you? It's still 404 to client and we still should try more.
3. This change is not related to metabase marks, so should at least be committed separately.

pkg/services/object/get/container.go
```diff
@@ -78,4 +84,9 @@ func (exec *execCtx) executeOnContainer() {
 			i = -1
 		}
 	}
+
+	if firstUnavailableErr != nil {
```
roman-khimov: Also not sure how it's related.

pkg/services/object/get/remote.go
End-rey: There was a slightly different change. I checked it again and confirmed that it is not needed yet, so I removed it.

pkg/local_object_storage/metabase/exists.go
End-rey: Done.

roman-khimov: Not what I meant, previous version is better then.

roman-khimov (approved):

pkg/local_object_storage/metabase/exists.go
End-rey: Ah, right, I understand now. Fixed.

PR #4123, opened 2026-08-10
Update SDK and NeoGo

End-rey:


CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ Changelog for NeoFS Node
 ### Removed
 
 ### Updated
+- `github.com/nspcc-dev/neo-go` module to `v0.122.1-0.20260807115931-cfee8827ddfd` (#4123)
```
roman-khimov, line 18: Why an unreleased one? Should be 0.122.0.

internal/crypto/object_test.go
```diff
@@ -237,7 +237,7 @@ func getUnsignedObject() object.Object {
 	par.SetOwner(mainAcc)
 	par.SetCreationEpoch(creationEpoch)
 	par.SetType(typ)
-	par.SetPayloadHomomorphicHash(checksum.NewTillichZemor(tz.Sum(payload))) //nolint:staticcheck // this is a test and such objects are possible
+	par.SetPayloadHomomorphicHash(testutil.TillichZemorChecksum(tz.Sum(payload))) //nolint:staticcheck // this is a test and such objects are possible
```
roman-khimov: Can use New() with type. Also, we don't want to import tzhash, so can be some static data and not a real hash (nothing checks it now).

CHANGELOG.md
End-rey: v0.122.0 still references the SDK’s removed ObjectHash API, so it does not compile with the updated SDK.

internal/crypto/object_test.go
End-rey: Done.

CHANGELOG.md
roman-khimov: Ah, ok

roman-khimov (approved):

PR #4125, opened 2026-08-10
EC GET without part idxs

cthulhu-rider:
best case when all parts residue on dedicated nodes should work w/ these changes. But violations of order are not yet properly handled

### DIFF
--- a/pkg/services/object/get/get.go
+++ b/pkg/services/object/get/get.go
@@ -16,13 +16,8 @@ import (
 
 // Get serves a request to get an object by address, and returns Streamer instance.
 func (s *Service) Get(ctx context.Context, prm Prm) error {
-	var neofsNet NeoFSNetwork
-	// range requests do not support fetching additional info about EC parts
-	if !prm.payloadRange.IsSet() {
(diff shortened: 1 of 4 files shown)


pkg/services/object/get/get.go
```diff
@@ -16,13 +16,8 @@ import (
...
 func (s *Service) Get(ctx context.Context, prm Prm) error {
-	var neofsNet NeoFSNetwork
-	// range requests do not support fetching additional info about EC parts
-	if !prm.payloadRange.IsSet() {
-		neofsNet = s.neoFSNet
-	}
-
-	pi, err := checkECPartInfoGetRequest(neofsNet, prm)
+	// TODO: upd API
+	pi, err := checkECPartInfoGetRequest(s.neoFSNet, prm)
```
cthulhu-rider, line 20: this is wrong for any case except the best one. For example, if SN#1 stores part#2, payload will be streamed 0-2-...

so, the only thing we can optimize for now is header+payload GET which is used to request part#0 only. For optimal case, optimization won't be visible

PR #4126, opened 2026-08-10
client: Support GET response header with payload

cthulhu-rider:
- PoC  based on https://github.com/nspcc-dev/neofs-sdk-go/pull/851


roman-khimov:
Zero performance gains, useless otherwise, see https://github.com/nspcc-dev/neofs-api/pull/333.

PR #4130, opened 2026-08-13
Tune GRPC buffers

roman-khimov:


carpawell (approved):
OK, but, TBH, i would expect testing on real machines rather than local k6 runs with some magic numbers.
BTW, can we have more memory problems with small object loads using bigger buffers?

carpawell:
Chlog?

roman-khimov:
These buffers are per-connection.

PR #4144, opened 2026-08-25
sn/object: Support new request signatures

cthulhu-rider:
could not upgrade to the latest SDK to test patches above it w/o this


carpawell (approved):
Chlog about new signature scheme?

roman-khimov (approved):
Chlog can be extended a bit, yes.

PR #4149, opened 2026-08-26
fstree: support online depth reshaping

### LINKED ISSUE #3774: Online FSTree reshaping
I'm always frustrated when I can't change the depth of the FSTree.

Nothing prevents us from adding online reshaping functionality:
 * add a secondary depth to FSTree config
 * put with the new depth
 * try get operations with the new one, fallback to old if not present
 * background threads can iterate over storage, add dirs if needed, drop empty if possible, `rename` files appropriately (not even looking into contents)
 * once threads are done drop the old depth from the config

Offline reshaping is somewhat easier, but doesn't really work for big object collections.

## Additional context
I'd really like to change the depth of 5 to 3 for mainnet nodes. But I don't want to lose any data.

End-rey:
Closes #3774.


cmd/neofs-node/config.go
```diff
@@ -679,6 +679,9 @@ func (c *cfg) reloadConfig() error {
 	if err := validateConfig(newCfg); err != nil {
 		return fmt.Errorf("validate configuration: %w", err)
 	}
+	if blobstorLayoutChanged(oldCfg, newCfg) {
+		return errors.New("changing blobstor depth requires node restart")
```
roman-khimov, line 685: Why can't it be done without it? Triggered via control service?

docs/storage-node-configuration.md
```diff
@@ -207,11 +208,14 @@ relevant for big volumes with very high number of stored objects.
 | `path`                    | `string`  |               | Path to the root of the blobstor.                                                                                            |
 | `perm`                    | file mode | `0640`        | Default permission for created files and directories.                                                                        |
 | `depth`                   | `int`     | `4`           | File-system tree depth. Optimal value depends on the number of objects stored in this shard, the number of lower-level directories used by FSTree is 58^depth, with depth 3 this is ~200K, with 4 --- ~11M |
+| `secondary_depth`         | `int`     |               | Previous FSTree depth during an online layout reshape. New writes use `depth`; reads fall back to `secondary_depth`. Used for migration to new depth. |
```
roman-khimov: I'd rather have `allow_depth_change` boolean, then you have the old depth in fstree json and only new one here. In general `depth` is dead data already, it's used only once during initialization, but I don't have any better scheme for now.

pkg/local_object_storage/blobstor/fstree/control.go
```diff
@@ -170,6 +179,12 @@ func (t *FSTree) checkConfig() error {
 		t.shardID = id
 		t.shardIDSet = !id.IsZero()
 	}
+	if reshapeDepth {
+		d.Depth = t.Depth
```
roman-khimov: Insufficient. old/new should both be stored in fstree descriptor as well as some progress mark updated periodically. FSTree should always keep its internal consistency and be aware of its state and basic parameters like depth irrespective of external configuration. Think of 5/3 fstree and 4/2 config, this can and should be detected.

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -156,7 +164,16 @@ func addressFromString(s string) (*oid.Address, error) {
 
 // Iterate iterates over all stored objects.
 func (t *FSTree) Iterate(objHandler func(addr oid.Address, data []byte) error, errorHandler func(addr oid.Address, err error) error) error {
-	return t.iterate(0, []string{t.RootPath}, objHandler, errorHandler, nil, nil)
+	seen := make(map[oid.Address]struct{})
+	return t.iterateLayouts(func(treeDepth uint64) error {
```
roman-khimov: Iterate is more complex, it should (at least try to) keep the object list ordered, so going over one depth and then over another is problematic.

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -156,7 +164,16 @@ func addressFromString(s string) (*oid.Address, error) {
...
+		return t.iterate(0, treeDepth, []string{t.RootPath}, func(addr oid.Address, data []byte) error {
+			if _, ok := seen[addr]; ok {
+				return nil
+			}
+			seen[addr] = struct{}{}
```
roman-khimov: Suppose you have 16M objects in a shard. That's 0.5GB of data.

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -286,28 +338,38 @@ func (t *FSTree) treePath(addr oid.Address) string {
 	return filepath.Join(dirs...)
 }
 
+func (t *FSTree) treePaths(addr oid.Address) []string {
```
roman-khimov: You can avoid allocating a slice here by providing an iterator.

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -286,28 +338,38 @@ func (t *FSTree) treePath(addr oid.Address) string {
...
 
-	p, err := t.getPath(addr)
-	if err != nil {
-		if errors.Is(err, fs.ErrNotExist) {
-			err = logicerr.Wrap(apistatus.ObjectNotFound{})
-		}
-		return err
-	}
+	t.reshapeMtx.Lock()
+	defer t.reshapeMtx.Unlock()
```
roman-khimov: Not sure what you're trying to protect from.

pkg/local_object_storage/blobstor/fstree/reshape.go
```diff
@@ -0,0 +1,246 @@
...
+	if oldPath == newPath {
+		return false, nil
+	}
+
+	t.reshapeIterMtx.Lock()
+	defer t.reshapeIterMtx.Unlock()
+	t.reshapeMtx.Lock()
+	defer t.reshapeMtx.Unlock()
+
+	if _, err := os.Lstat(oldPath); err != nil {
```
roman-khimov: Why? You'll get "not exist" error on rename anyway and you need to process it accordingly (it's OK, object can be deleted).

pkg/local_object_storage/blobstor/fstree/reshape.go
```diff
@@ -0,0 +1,246 @@
...
+		if errors.Is(err, fs.ErrNotExist) {
+			return false, nil
+		}
+		return false, fmt.Errorf("stat old file %q: %w", oldPath, err)
+	}
+	if err := util.MkdirAllX(filepath.Dir(newPath), t.Permissions); err != nil {
+		return false, fmt.Errorf("create destination directory for %q: %w", newPath, err)
+	}
+	if _, err := os.Lstat(newPath); err == nil {
```
roman-khimov: Why? Things can change between lstat and rename. You're trying to avoid this with `reshapeMtx`, but this doesn't matter much anyway, objects with the same address are the same objects and rename can be done safely. Then `reshapeIterMtx` will drastically affect performance, this background process will almost always hold the lock and regular operations will suffer badly.

pkg/local_object_storage/blobstor/fstree/reshape.go
```diff
@@ -0,0 +1,246 @@
...
+	t.reshapeMtx.Lock()
+	defer t.reshapeMtx.Unlock()
+
+	if _, err := os.Lstat(oldPath); err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			return false, nil
+		}
+		return false, fmt.Errorf("stat old file %q: %w", oldPath, err)
+	}
+	if err := util.MkdirAllX(filepath.Dir(newPath), t.Permissions); err != nil {
```
roman-khimov: Can be more optimal, if new depth is lower then we have the directory already for sure. Syscalls matter here.

pkg/local_object_storage/blobstor/fstree/reshape.go
```diff
@@ -0,0 +1,246 @@
...
+					zap.Uint64("old_depth", t.SecondaryDepth),
+					zap.Uint64("new_depth", t.Depth),
+					zap.Int("moved_files", *moved),
+					zap.String("path", t.RootPath),
+				)
+			}
+		}
+	}
+
+	if depth > 0 {
```
roman-khimov: Consider 3->4 reshape, we shouldn't delete directories in this case.

cmd/neofs-node/config.go
End-rey: Runtime layout changes require more than updating configuration: the node must locate the active FSTree instance, persist the new reshape state, atomically switch primary/fallback layouts, and synchronize this transition with concurrent PUT, DELETE, GET and iteration operations. It must also define what happens if configuration reload succeeds only for some shards or the reshape worker cannot start.

   A restart performs this transition during storage initialization, before the node starts serving requests. Reshaping itself is still online, and the persistent descriptor state makes it safe to resume after later restarts. Given that, adding a separate runtime control path does not seem worth the additional complexity and risk.

docs/storage-node-configuration.md
End-rey: Done.

pkg/local_object_storage/blobstor/fstree/control.go
End-rey: Done.

pkg/local_object_storage/blobstor/fstree/fstree.go
End-rey: Merged ordered iteration.

pkg/local_object_storage/blobstor/fstree/fstree.go
End-rey: Map allocation removed.

pkg/local_object_storage/blobstor/fstree/fstree.go
End-rey: Iterator allocation removed.

pkg/local_object_storage/blobstor/fstree/reshape.go
End-rey: Removed.

pkg/local_object_storage/blobstor/fstree/reshape.go
End-rey: Avoided unnecessary mkdirs.

pkg/local_object_storage/blobstor/fstree/reshape.go
End-rey: Done.

pkg/local_object_storage/blobstor/fstree/fstree.go
End-rey: It prevents a delete/rename race that could resurrect a deleted object.

pkg/local_object_storage/blobstor/fstree/reshape.go
End-rey: Done.

pkg/local_object_storage/blobstor/fstree/reshape.go
```diff
@@ -0,0 +1,270 @@
...
+					zap.Uint64("new_depth", t.Depth),
+					zap.Int("moved_files", *moved),
+					zap.String("path", t.RootPath),
+				)
+			}
+		}
+	}
+
+	if t.secondaryDepth > t.Depth && depth > 0 {
+		if err := os.Remove(dir); err != nil && !errors.Is(err, fs.ErrNotExist) && !errors.Is(err, fs.ErrExist) {
```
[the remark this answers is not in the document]
End-rey: Done.

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -181,102 +189,194 @@ func (t *FSTree) IterateSizes(f func(addr oid.Address, size uint64) error, ignor
...
+				return err
+			}
+		} else if sizeHandler != nil {
+			info, statErr := os.Stat(entry.path)
+			if statErr != nil {
+				return fmt.Errorf("stat file %q: %w", entry.path, statErr)
+			}
+			if err = sizeHandler(entry.addr, uint64(info.Size())); err != nil {
+				return err
+			}
```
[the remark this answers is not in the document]
End-rey: Done.

pkg/local_object_storage/blobstor/fstree/control.go
```diff
@@ -170,6 +197,86 @@ func (t *FSTree) checkConfig() error {
...
+	if err = f.Close(); err != nil {
+		return fmt.Errorf("close descriptor %q: %w", descPath, err)
+	}
+	if d.Reshape == nil || d.Reshape.FromDepth != t.secondaryDepth || d.Reshape.ToDepth != t.Depth {
+		return errors.New("FSTree reshape state changed unexpectedly")
+	}
+	d.Depth = d.Reshape.ToDepth
+	d.Reshape = nil
+	if err = writeDescriptor(descPath, d); err != nil {
+		return fmt.Errorf("write completed reshape descriptor: %w", err)
```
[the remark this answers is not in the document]
End-rey: Done.

pkg/local_object_storage/blobstor/fstree/control.go
```diff
@@ -170,6 +221,86 @@ func (t *FSTree) checkConfig() error {
...
 	}
+	if d.Reshape != nil && !t.readOnly {
+		if err := writeDescriptor(descPath, d); err != nil {
+			return fmt.Errorf("write reshaped descriptor: %w", err)
+		}
+	}
+	return nil
+}
+
+func (t *FSTree) reshapeLastProcessedPath() (string, error) {
```
roman-khimov, line 216: I'd expect it to be read once on FSTree init, parsing it again just adds some potential for errors.

pkg/local_object_storage/blobstor/fstree/control.go
```diff
@@ -170,6 +221,86 @@ func (t *FSTree) checkConfig() error {
...
+	}
+	if d.Reshape == nil || d.Reshape.FromDepth != t.secondaryDepth || d.Reshape.ToDepth != t.Depth {
+		return "", errors.New("FSTree reshape state changed unexpectedly")
+	}
+	return d.Reshape.LastProcessedPath, nil
+}
+
+func (t *FSTree) updateReshapeProgress(lastProcessedPath string) error {
+	descPath := t.descriptorPath()
+	f, err := os.Open(descPath)
```
roman-khimov: You can also keep the current version in mem, update last path there and flush it to disk, without reading the file.

pkg/local_object_storage/blobstor/fstree/fstree.go
roman-khimov: `rename` is atomic. Proper sequencing can prevent this. Delete old path first, then delete new path. If old path is deleted before reshaper rename --- reshaper gets a rename error. If reshaper is first to rename then deleting old path fails, but deleting new path does.

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -330,14 +448,24 @@ func (t *FSTree) Exists(addr oid.Address) (bool, error) {
 // checks whether file for the given object address exists and returns path to
 // the file if so. Returns [fs.ErrNotExist] if file is missing.
 func (t *FSTree) getPath(addr oid.Address) (string, error) {
-	p := t.treePath(addr)
+	t.reshapeMtx.RLock()
+	defer t.reshapeMtx.RUnlock()
```
roman-khimov: Same with all other operations, old first, new afterwards --- no locking needed.

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -368,6 +499,9 @@ func (t *FSTree) PutBatch(objs map[oid.Address][]byte) error {
 		return common.ErrReadOnly
 	}
 
+	t.reshapeMtx.Lock()
+	defer t.reshapeMtx.Unlock()
```
roman-khimov: Puts don't care at all, they always use new path.

pkg/local_object_storage/blobstor/fstree/reshape.go
```diff
@@ -0,0 +1,290 @@
...
+			return err
+		}
+		if movedFile {
+			*moved = *moved + 1
+			if *moved%reshapeProgressStep == 0 {
+				if err := t.syncReshape(*dirty); err != nil {
+					return err
+				}
+				*dirty = false
+				if err := t.updateReshapeProgress(relativePath); err != nil {
```
roman-khimov, line 213: You can write this file and have a single `syncfs` call for it and other meta (avoid `syncReshape`).

pkg/local_object_storage/blobstor/fstree/reshape.go
roman-khimov: I still think this lstat can be avoided.

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -181,102 +189,198 @@ func (t *FSTree) IterateSizes(f func(addr oid.Address, size uint64) error, ignor
...
+			}
+		}
+		if advanceSecondary {
+			secondaryEntry, secondaryOK, err = nextLayoutEntry(secondary, errorHandler)
+			if err != nil {
+				return err
+			}
 		}
 	}
-
```
roman-khimov, line 269: :eyes:

pkg/local_object_storage/blobstor/fstree/control.go
End-rey: Done.

pkg/local_object_storage/blobstor/fstree/control.go
End-rey: Done.

pkg/local_object_storage/blobstor/fstree/fstree.go
End-rey: Done.

pkg/local_object_storage/blobstor/fstree/fstree.go
End-rey: Removed.

pkg/local_object_storage/blobstor/fstree/reshape.go
End-rey: Done.

pkg/local_object_storage/blobstor/fstree/fstree.go
End-rey: Fixed.

pkg/local_object_storage/blobstor/fstree/control.go
```diff
@@ -170,6 +204,54 @@ func (t *FSTree) checkConfig() error {
...
+}
+
+func (t *FSTree) updateReshapeProgress(lastProcessedPath string) error {
+	if _, err := t.reshapeLastProcessedPath(); err != nil {
+		return err
+	}
+	d := t.descriptor
+	reshape := *d.Reshape
+	reshape.LastProcessedPath = lastProcessedPath
+	d.Reshape = &reshape
```
roman-khimov, line 230: Any concurrent access to this structure?

pkg/local_object_storage/blobstor/fstree/control.go
```diff
@@ -170,6 +204,54 @@ func (t *FSTree) checkConfig() error {
...
+	if err := writeDescriptor(t.descriptorPath(), d); err != nil {
+		return fmt.Errorf("write reshape progress descriptor: %w", err)
+	}
+	if err := t.syncReshape(true); err != nil {
+		return err
+	}
+	t.descriptor = d
```
roman-khimov, line 237: And this one? Can be atomic.

pkg/local_object_storage/blobstor/fstree/reshape.go
```diff
@@ -0,0 +1,268 @@
...
+	if err != nil {
+		return moved, err
+	}
+	if remaining {
+		return moved, errors.New("secondary layout still contains object files")
+	}
+	return moved, nil
+}
+
+func (t *FSTree) hasSecondaryFiles(ctx context.Context, dir string, depth uint64, prefix string) (bool, error) {
```
roman-khimov: Return error only?

pkg/local_object_storage/blobstor/fstree/reshape.go
```diff
@@ -0,0 +1,268 @@
...
+			}
+			found, err := t.hasSecondaryFiles(ctx, filepath.Join(dir, entry.Name()), depth+1, prefix+entry.Name())
+			if err != nil || found {
+				return found, err
+			}
+			continue
+		}
+		if !entry.IsDir() {
+			if _, err := addressFromString(prefix + entry.Name()); err == nil {
+				return true, nil
```
roman-khimov: This would allow to return file path in the error as well (and print/check/whatever).

pkg/local_object_storage/blobstor/fstree/reshape.go
```diff
@@ -0,0 +1,268 @@
...
+			return false, fmt.Errorf("create destination directory for %q: %w", newPath, err)
+		}
+	}
+	if err := os.Rename(oldPath, newPath); err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			return false, nil
+		}
+		return false, fmt.Errorf("rename file %q to %q: %w", oldPath, newPath, err)
+	}
+	if err := os.Remove(oldPath); err != nil && !errors.Is(err, fs.ErrNotExist) {
```
roman-khimov, line 248: `err == nil` is "renamed", why trying to remove in this case? Old path doesn't exist anyway and no other code should create it (put writes to new depth).

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -181,102 +188,196 @@ func (t *FSTree) IterateSizes(f func(addr oid.Address, size uint64) error, ignor
...
 			}
 		}
 
-		if err != nil {
-			return err
+		if advancePrimary {
+			primaryEntry, primaryOK, err = nextLayoutEntry(primary, errorHandler)
+			if err != nil {
+				return err
+			}
```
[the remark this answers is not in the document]
roman-khimov: Names (OID effectively) are collected in advance and can be tried accordingly.

pkg/local_object_storage/blobstor/fstree/fstree_write_specific.go
```diff
@@ -2,6 +2,17 @@
...
 
+import "os"
+
+func syncFS(path string) error {
+	f, err := os.Open(path)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+	return f.Sync()
```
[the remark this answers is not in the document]
roman-khimov: Not practical to me. And we do have a (nice) check for leftovers when finishing the process.

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -330,14 +442,21 @@ func (t *FSTree) Exists(addr oid.Address) (bool, error) {
 // checks whether file for the given object address exists and returns path to
 // the file if so. Returns [fs.ErrNotExist] if file is missing.
 func (t *FSTree) getPath(addr oid.Address) (string, error) {
-	p := t.treePath(addr)
-
-	_, err := os.Stat(p)
-	if err != nil {
-		return "", fmt.Errorf("get filesystem path for object by address: get file stat %q: %w", p, err)
+	primary, secondary := t.treePaths(addr)
+	for _, p := range [...]string{secondary, primary} {
```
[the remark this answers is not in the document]
roman-khimov: Intentional.

pkg/local_object_storage/blobstor/fstree/control.go
End-rey: No concurrent access: the state is initialized before reshaping starts and is then used only by its single background task. On shutdown, the node waits for that task to finish.

pkg/local_object_storage/blobstor/fstree/reshape.go
End-rey: rename can succeed without removing `oldPath` when `oldPath` and an existing `newPath` are hard links to the same inode. In that case it is a no-op on Linux, so the secondary path would remain and reshaping could never complete. The extra removal handles this case; `ErrNotExist` is ignored for the usual rename case.

pkg/local_object_storage/blobstor/fstree/fstree.go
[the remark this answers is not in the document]
End-rey: Reworked iteration to snapshot each secondary prefix group before scanning the corresponding primary subtree, preventing omissions during rename without global locking or a full object list.

pkg/local_object_storage/blobstor/fstree/reshape.go
End-rey: Done.

pkg/local_object_storage/blobstor/fstree/reshape.go
End-rey: Done.

roman-khimov (approved):

PR #4153, opened 2026-08-28
Improve object GET proto

cthulhu-rider:


pkg/services/object/get.go
```diff
@@ -869,27 +868,37 @@ func (x *getECTransport) CopyRemoteECPartRange(ctx context.Context, conn clientc
...
+	metaHdrLen := calculateRequestMetaHeaderLen(remoteServerAPIVersion, 0, xHdrs)
 
 	verifHdrFldLen := calculateRequestVerificationHeaderFieldLen(remoteServerAPIVersion)
 
-	reqLen := 1 + 1 + getByAddressRequestBodyLen + // first 1 for iprotobuf.TagBytes1
-		1 + protowire.SizeBytes(metaHdrLen) + // 1 for iprotobuf.TagBytes2
-		verifHdrFldLen
+	reqLen := protoencoding.CalculateRequestLength(bodyLen, metaHdrLen, verifHdrFldLen)
 
 	// TODO: try with sync.Pool
```
roman-khimov, line 889: We're doing to for SEARCH already, can try here as well. But can be a separate PR.

roman-khimov (approved):

PR #4158, opened 2026-08-31
Use gRPC buffer pool for GET EC requests

cthulhu-rider:


pkg/services/object/get/ec.go
```diff
@@ -1518,7 +1518,7 @@ func (s *Service) streamFirstECPart(ctx context.Context, transport GetECRequestT
 
 		if !copiedHdr {
 			if local {
-				copiedHdr, parentPldLen, partPldLen, copiedPartPld, err = transport.CopyLocalECPartParentHeaderAndPayload(ctx, s.localObjects.(*engine.StorageEngine), partInfo)
+				copiedHdr, parentPldLen, partPldLen, copiedPartPld, err = transport.CopyECParentHeaderAndPayloadFromLocalFirstPart(ctx, s.localObjects.(*engine.StorageEngine), ruleIdx)
```
roman-khimov, line 1521: This is not related to buffer pools, needs to be separate.

cthulhu-rider: waste of time to me, but ok will do

roman-khimov (approved):

PR #4159, opened 2026-08-31
GET EC TLS

cthulhu-rider:


internal/crypto/requests.go
```diff
@@ -59,8 +59,7 @@ func requestNeedsSignature[B protoencoding.Message](ctx context.Context, req neo
 	if req.GetVerifyHeader() != nil {
 		return true
 	}
-	meta := req.GetMetaHeader()
-	if meta == nil || meta.GetTtl() != 1 {
+	if req.GetMetaHeader().GetTtl() > 1 { // NPE-safe
```
roman-khimov: Why not using ttl=1 for EC part requests?

cthulhu-rider: it works well w/o TTL at all

roman-khimov: This implies missing TTL is treated like TTL=1. In general, it's safer to assume this, but I'm not sure other parts of the code treat it the same way. If we ensure that then it's OK and we can even update spec wrt it.

cthulhu-rider: got ur point. Well, seems like it's easier to set TTL=1 (2 bytes isn't that much after all) in order to keep non-EC and other RPC cases untouched

roman-khimov (approved):

Issue #2796, opened 2024-04-01
Split verification and storage code for incoming objects


cthulhu-rider:
_Originally posted in https://github.com/nspcc-dev/neofs-node/pull/2674#discussion_r1540004148_

it's proposed to split `VerifyObject` and `StoreObject` functionality, both independent and reusable

PR #4164, opened 2026-09-01
Implement streaming replication

cthulhu-rider:


cthulhu-rider:
code is rdy for review. Gonna make performance runs and add unit tests in the background

cmd/neofs-node/object.go
```diff
@@ -627,8 +627,19 @@ func (x storageForObjectService) SearchObjects(ctx context.Context, cID cid.ID,
 	return x.local.Search(ctx, cID, fs, attrs, cursor, count)
 }
 
-func (x storageForObjectService) VerifyAndStoreObjectLocally(ctx context.Context, obj object.Object) error {
```
roman-khimov: Fixes #2796?

pkg/services/object/server.go
```diff
@@ -1768,41 +1819,88 @@ func (s *Server) replicate(ctx context.Context, objMsg *protoobject.Object, sig
...
+	var gotHash []byte
+
+	isStream := recvChunkFn != nil && obj.Type() == object.TypeRegular
+	var streamAbortFn func()
+	var streamCloser io.Closer
+
+	if isStream {
+		// TODO: can be optimized from two sides:
+		//  1. header structure decoding can be done without unmarshaling (e.g. via protoscan funcs)
+		//  2. since header is already serialized in the original request, io.WriterTo can be implemented around it
```
roman-khimov: 2 should be easy to do, isn't it? Although PUTs are relatively slow (and streamed ones are relatively big), so not likely to be noticeable.

pkg/services/object/server.go
```diff
@@ -1768,41 +1819,88 @@ func (s *Server) replicate(ctx context.Context, objMsg *protoobject.Object, sig
...
-		}
-		if st != nil {
-			return nil, st, nil
+			return nil, newStoreLocalObjectFailureStatus(err), nil
 		}
 
-		if len(chunk) == 0 {
-			return nil, newBadRequestStatus("empty payload chunk"), nil
+		var st *protostatus.Status
+		gotHash, gotPayloadLen, st, err = readReplicatedObjectPayload(obj.PayloadSize(), recvChunkFn, func(chunk []byte) *protostatus.Status {
```
roman-khimov: Why not passing a simple (optional) `io.Writer` there? This callback doesn't do a lot, can be simplified this way.

pkg/services/object/server.go
```diff
@@ -1768,41 +1819,88 @@ func (s *Server) replicate(ctx context.Context, objMsg *protoobject.Object, sig
...
+		if hdr.PayloadLength < uint64(len(objMsg.Payload)) { // also prevents slices.Grow() panic below
 			return nil, newWrongReplicatedObjectPayloadLengthStatus(), nil
 		}
 
-		payload = append(payload, chunk...)
+		payload := slices.Grow(objMsg.Payload, int(hdr.PayloadLength)-len(objMsg.Payload))
+
+		var st *protostatus.Status
+		gotHash, gotPayloadLen, st, err = readReplicatedObjectPayload(obj.PayloadSize(), recvChunkFn, func(chunk []byte) *protostatus.Status {
+			payload = append(payload, chunk...)
```
roman-khimov: `bytes.Buffer` can be used here to provide io.Writer.

pkg/services/object/server.go
cthulhu-rider: yes, it's not very difficult, but given the general case of unordered fields, the logic will have to branch. Therefore, I still suggest doing this separately

pkg/services/object/server.go
cthulhu-rider: sounds good also taking into account TODOed buffering optimization and having https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go@v1.0.0-rc.22/proto/protobuf#BuffersSlice.WriteTo. Done

cmd/neofs-node/object.go
cthulhu-rider: oh, yep. Tagged

pkg/services/object/server.go

pkg/services/object/server.go
```diff
@@ -1659,7 +1669,45 @@ func newWrongReplicatedObjectPayloadLengthStatus() *protostatus.Status {
...
+		}
+
+		if totalLen+uint64(len(chunk)) > payloadLen {
+			return nil, 0, newWrongReplicatedObjectPayloadLengthStatus(), nil
+		}
+
+		_, err = w.Write(chunk)
+		if err != nil {
+			return nil, 0, nil, fmt.Errorf("%w: %w", errWriteStream, err)
+		}
```
roman-khimov: Nothing updates totalLen now, the check is gone.

cthulhu-rider: fixed

PR #4168, opened 2026-09-08
Initial support for ReplicateV2

cthulhu-rider:


pkg/services/object/server.go
```diff
@@ -1755,111 +1755,154 @@ func (s *Server) Search(_ *protoobject.SearchRequest, _ protoobject.ObjectServic
...
+		return nil, nil, err
+	}
+
+	reqInit, ok := firstReq.StreamPart.(*protoobject.ReplicateV2Request_Init_)
+	if !ok {
+		return nil, newBadRequestStatus("first request does not contain init field"), nil
+	}
+
+	initPart := reqInit.Init
+	if initPart == nil { // not expected to ever happen, but better to keep safe
```
roman-khimov, line 1564: Peers can send any kind of garbage in general, so it's absolutely necessary.

pkg/services/object/server.go
```diff
@@ -1755,111 +1755,154 @@ func (s *Server) Search(_ *protoobject.SearchRequest, _ protoobject.ObjectServic
 	return grpcstatus.Error(grpccodes.Unimplemented, "no longer supported, use SearchV2")
 }
 
+func readFirstReplicateV2Request(stream protoobject.ObjectService_ReplicateV2Server) (*protoobject.ReplicateV2Request_Init, *protostatus.Status, error) {
```
roman-khimov, line 1552: Can return `(*protoobject.ReplicateV2Request_Init, error)` and use `apistatus.Error` where appropriate.

pkg/core/client/client.go
```diff
@@ -63,3 +64,12 @@ func IsMutuallyAuthenticated(c any) bool {
 	x, ok := c.(interface{ IsMutuallyAuthenticated() bool })
 	return ok && x.IsMutuallyAuthenticated()
 }
+
+// CompareAPIVersion performs three-way comparison of API server version against
+// the given one.
+func CompareAPIVersion(c MultiAddressClient, major uint32, minor uint32) int {
+	cv := c.APIVersion()
```
roman-khimov: Can have version.Version here as well.

pkg/core/client/client.go
```diff
@@ -63,3 +64,12 @@ func IsMutuallyAuthenticated(c any) bool {
 	x, ok := c.(interface{ IsMutuallyAuthenticated() bool })
 	return ok && x.IsMutuallyAuthenticated()
 }
+
+// CompareAPIVersion performs three-way comparison of API server version against
+// the given one.
+func CompareAPIVersion(c MultiAddressClient, major uint32, minor uint32) int {
```
roman-khimov: Can accept version.Version here and have a static var somewhere.

pkg/services/object/put/remote.go
```diff
@@ -119,26 +128,119 @@ func (s *RemoteSender) ReplicateObjectToNode(ctx context.Context, id oid.ID, src
...
+		return protorefs.WriteSignature(buf, pubKey, sig, sigScheme)
+	}
+	protoobject.WriteReplicateV2InitRequest(initReqBuf, hdrLen, writeHdrFn, sigLen, writeSigFn, signObjectMeta)
+
+	chunkReqNum := len(payload) / maxReplicateV2PayloadChunkLen
+	if len(payload)%maxReplicateV2PayloadChunkLen != 0 {
+		chunkReqNum++
+	}
+
+	chunkReqBufItems := make([]*[]byte, chunkReqNum)
```
roman-khimov: A lot of memory. Should use less.

pkg/services/object/server.go
cthulhu-rider: i mean gRPC internals here. In practice `reqInit.Init` is always non-nil when `reqInit` is

pkg/services/object/server.go
cthulhu-rider: this would lead to a "there and back" in most cases. Don't think it's worth it

pkg/core/client/client.go
cthulhu-rider: done

pkg/core/client/client.go
cthulhu-rider: done

pkg/services/object/put/remote.go
cthulhu-rider: improved

roman-khimov (approved):

PR #4170, opened 2026-09-09
policer: balance local shard data

### LINKED ISSUE #4147: Shard data balancing
I'm always frustrated when disk replacements leave nodes in unbalanced state, new disks store less data than they should. Same thing happens if extending node with new shards, old shards keep old data and never share it with new ones. This makes the load uneven and breaks redundancy assumptions since loss of one drive is not equal to loss of another in terms of lost object count (even though the basic redundancy unit is SN we can assume that shards are equal).

Policer has a clear picture of local data balance since #3862 (which is leveraged by #3908), the same data can be used to trigger PUT into more appropriate shard and DELETE from inappropriate one.

This has to be solved anyway.

End-rey:
Closes #4147.

One edge case to consider: a copy marked with `GarbageMarkRedundant` is still visible to `Exists,` but remains queued for GC. If shard topology changes and this shard becomes the HRW-preferred target again before GC runs, relocation sees that the object already exists and does not clear the redundant mark. GC may then remove the current target copy.

Should we handle this by adding a targeted operation similar to `ClearRedundantMark` for the selected target shard, without reviving regular GC or tombstone marks?

### DIFF
--- a/pkg/local_object_storage/engine/optimize.go
+++ b/pkg/local_object_storage/engine/optimize.go
@@ -1,0 +1,114 @@
+package engine
+
+import (
+	"context"
+	"errors"
+
+	apistatus "github.com/nspcc-dev/neofs-sdk-go/client/status"
+	"github.com/nspcc-dev/neofs-sdk-go/object"
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	"go.uber.org/zap"
+)
+
+// OptimizeShardLocation moves an object to its most preferred shard according
+// to HRW ordering and marks its copies on all other listed shards as redundant.
+//
+// The source copies are marked only after the object is known to exist on the
+// preferred shard. The operation therefore never removes the last local copy
+// when the destination cannot accept the object.
+//
+// It returns true if it copied the object to the preferred shard. It returns
+// false when the object was already there.
+//
+// Returns an error if executions are blocked (see BlockExecution), no listed
+// source shard contains the object, or optimization/marking fails.
+func (e *StorageEngine) OptimizeShardLocation(_ context.Context, addr oid.Address, shardIDs []string) (bool, error) {
+	e.blockMtx.RLock()
+	defer e.blockMtx.RUnlock()
+
+	if e.blockErr != nil {
+		return false, e.blockErr
+	}
+	if len(shardIDs) == 0 {
+		return false, errShardNotFound
+	}
+
+	shards := e.sortedShards(addr.Object())
+	if len(shards) == 0 {
+		return false, errShardNotFound
+	}
+
+	target := shards[0]
+	targetID := target.ID().String()
+	if len(shardIDs) == 1 && shardIDs[0] == targetID {
+		return false, nil
+	}
+
+	exists, err := target.Exists(addr, false)
+	if err != nil && !errors.Is(err, apistatus.ErrObjectNotFound) {
+		return false, err
+	}
+
+	moved := false
+	if !exists {
+		var (
+			obj           *object.Object
+			sourceShardID string
+		)
+		for _, sourceID := range shardIDs {
+			if sourceID == targetID {
+				continue
+			}
+
+			source := e.getShard(sourceID)
+			if source.Shard == nil {
+				continue
+			}
+
+			obj, err = source.Get(addr, false)
+			if err == nil {
+				sourceShardID = sourceID
+				break
+			}
+			if !errors.Is(err, apistatus.ErrObjectNotFound) {
+				return false, err
+			}
+		}
+		if obj == nil {
+			return false, apistatus.ObjectNotFound{}
+		}
+
+		err = e.putToShard(target, addr, obj, nil)
+		if err != nil && !errors.Is(err, errExists) {
+			return false, err
+		}
+		moved = err == nil
+		if moved {
+			e.log.Info("moved local object to preferred shard",
+				zap.Stringer("object", addr),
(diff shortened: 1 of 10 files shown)


pkg/services/policer/check.go
```diff
@@ -403,12 +402,19 @@ func (p *Policer) dropRedundantLocalCopies(ctx context.Context, obj objectcore.A
...
 	if err != nil {
-		p.log.Warn("could not mark redundant local shard copies as garbage",
+		p.log.Warn("could not relocate local object to its preferred shard",
 			zap.Stringer("object", obj.Address),
 			zap.Strings("shards", obj.ShardIDs),
 			zap.Error(err))
+		return
+	}
+	if moved {
+		p.metrics.IncPolicerObjectRelocated(isEC)
```
roman-khimov, line 414: Why separating EC and non-EC here? These are all physical objects.

pkg/services/policer/policer.go
```diff
@@ -29,7 +29,7 @@ type replicatorIface interface {
 type localStorage interface {
 	ListWithCursor(context.Context, uint32, *engine.Cursor, ...string) ([]objectcore.AddressWithAttributes, *engine.Cursor, error)
 	Delete(context.Context, oid.Address, engine.GarbageMark) error
-	DeleteRedundantCopies(context.Context, oid.Address, []string) error
+	RelocateObject(context.Context, oid.Address, []string) (bool, error)
```
roman-khimov, line 32: It's more like `OptimizeShardLocation` to me. `Relocate` implies movement, but it's entirely optional here and in most cases this callback does nothing.

pkg/local_object_storage/engine/relocate.go
```diff
@@ -0,0 +1,95 @@
...
+	defer e.blockMtx.RUnlock()
+
+	if e.blockErr != nil {
+		return false, e.blockErr
+	}
+	if len(shardIDs) == 0 {
+		return false, errShardNotFound
+	}
+
+	shards := e.sortedShards(addr.Object())
```
roman-khimov, line 35: Thought of avoiding getting this list again, but looks like it requires `ListWithCursor` changes. Shouldn't be performance-critical. At the same time we can optimize the scheme by checking placement in `ListWithCursor` and returning shard list only if there is something to fix there.

pkg/local_object_storage/engine/relocate.go
```diff
@@ -0,0 +1,95 @@
...
+			}
+			if !errors.Is(err, apistatus.ErrObjectNotFound) {
+				return false, err
+			}
+		}
+		if obj == nil {
+			return false, apistatus.ObjectNotFound{}
+		}
+
+		err = e.putToShard(target, addr, obj, nil)
```
roman-khimov, line 72: Do we have any log message for it?

pkg/local_object_storage/engine/relocate.go
```diff
@@ -0,0 +1,95 @@
...
+		if sourceID == targetID {
+			continue
+		}
+
+		source := e.getShard(sourceID)
+		if source.Shard == nil {
+			continue
+		}
+
+		if err = source.MarkGarbage(addr.Container(), []oid.ID{addr.Object()}, GarbageMarkRedundant); err != nil {
```
roman-khimov, line 89: Log?

roman-khimov (commented):
> If shard topology changes and this shard becomes the HRW-preferred target again before GC runs, relocation sees that the object already exists and does not clear the redundant mark. GC may then remove the current target copy.

This will be fixed by the next policer iteration. Too rare to care.

pkg/local_object_storage/engine/optimize.go
```diff
@@ -0,0 +1,95 @@
...
+	defer e.blockMtx.RUnlock()
+
+	if e.blockErr != nil {
+		return false, e.blockErr
+	}
+	if len(shardIDs) == 0 {
+		return false, errShardNotFound
+	}
+
+	shards := e.sortedShards(addr.Object())
```
End-rey: Considered it, but that would sort shards for every listed object, including remotely redundant ones. Current code sorts only objects policer keeps locally and skips the target existence check for objects already listed only on their preferred shard.

pkg/services/policer/check.go
End-rey: Fixed.

pkg/services/policer/policer.go
End-rey: Renamed.

pkg/local_object_storage/engine/optimize.go
```diff
@@ -0,0 +1,95 @@
...
+			}
+			if !errors.Is(err, apistatus.ErrObjectNotFound) {
+				return false, err
+			}
+		}
+		if obj == nil {
+			return false, apistatus.ObjectNotFound{}
+		}
+
+		err = e.putToShard(target, addr, obj, nil)
```
End-rey: Done.

pkg/local_object_storage/engine/optimize.go
```diff
@@ -0,0 +1,95 @@
...
+		if sourceID == targetID {
+			continue
+		}
+
+		source := e.getShard(sourceID)
+		if source.Shard == nil {
+			continue
+		}
+
+		if err = source.MarkGarbage(addr.Container(), []oid.ID{addr.Object()}, GarbageMarkRedundant); err != nil {
```
End-rey: Done.

roman-khimov (approved):

Issue #2925, opened 2024-08-28
Store objects in "header+payload" format


roman-khimov:
## Is your feature request related to a problem? Please describe.
I'm always frustrated when I look at how we allocate a lot of memory to fetch data from storage. #2316, #2178, #1398. #2814 also reminds about it.

## Describe the solution you'd like
#1932 has some valid input for these problems, although it's localized to CLI, whereas I'm more concerned about SN. We can store/fetch objects in a bit different manner, serializing header and payload separately right at the FSTree level. Then object get operation would return a header and an `io.ReaderCloser` pointing to the file (if it's not _too_ small). This can then be streamed to the outside as needed (but likely API changes are required as well).

## Describe alternatives you've considered
Keep pulling MBs of data from the storage is no fun.

roman-khimov:
We've made a lot of memory-related optimizations, but this costs a lot in terms of code to work(around) with protobuf. Changing the storage scheme can simplify this code a lot and make `blobstor` really be a blob storage.

PR #4171, opened 2026-09-10
Support streaming write into FSTree

cthulhu-rider:


cthulhu-rider:
looking at #2925, #3451 i realized that initially proposed interface is too abstract. Will rework

pkg/local_object_storage/blobstor/fstree/fstree_write_generic.go
```diff
@@ -281,8 +281,8 @@ func handleFileError(tmpPath string, err error) error {
 		switch {
 		case errors.Is(pe.Err, syscall.ENOSPC):
 			err = common.ErrNoSpace
-		case errors.Is(pe.Err, syscall.EEXIST):
-			return syscall.EEXIST
+		case errors.Is(pe.Err, fs.ErrExist):
+			return fs.ErrExist
```
roman-khimov: But it's `fs.ErrExist` already, maybe with some additional context.

internal/testutil/os.go
```diff
@@ -0,0 +1,63 @@
+package testutil
```
roman-khimov: `filetest` or `fstest` at least, `testutil` means nothing.

pkg/local_object_storage/blobstor/fstree/fstree.go
```diff
@@ -511,12 +512,83 @@ func (t *FSTree) getPath(addr oid.Address) (string, error) {
...
+//
+// Either [io.Closer.Close] or abort function must be finally called. All
+// functions must not be called concurrently.
+//
+// If the object already exists, InitPut or resulting stream calls return
+// [fs.ErrExist].
+//
+// If the device runs out of space, InitPut or resulting stream calls return
+// [common.ErrNoSpace].
+func (t *FSTree) InitPut(addr oid.Address, headerLen uint64, payloadLen uint64, headerW io.WriterTo) (io.WriteCloser, func(), error) {
```
roman-khimov, line 536: Why not `header []byte`? We normally have it already when trying to `InitPut()`?

pkg/local_object_storage/blobstor/fstree/fstree_write_generic.go
```diff
@@ -43,6 +46,138 @@ func (w *genericWriter) writeBatch(objs []writeDataUnit) error {
...
+		return x.handleAlreadyExists()
+	}
+	if errors.Is(err, common.ErrNoSpace) {
+		x.abort()
+	}
+	return err
+}
+
+func (x *genericFileWriteStream) openTmpFile() error {
+	for {
```
roman-khimov: Can't be unlimited.

pkg/local_object_storage/blobstor/fstree/fstree_write_generic.go
```diff
@@ -43,6 +46,138 @@ func (w *genericWriter) writeBatch(objs []writeDataUnit) error {
...
+	}
+}
+
+func (x *genericFileWriteStream) Close() error {
+	if x.aborted {
+		return errStreamAborted
+	}
+
+	for {
+		err := x.openTmpFileIfNeeded()
```
roman-khimov: We're closing here, this doesn't seem useful (just check `x.tmpFile`).

pkg/local_object_storage/blobstor/fstree/fstree_write_linux.go
```diff
@@ -192,20 +193,115 @@ func (w *linuxWriter) finalize() error {
 	return nil
 }
 
+type unixFileWriteStream struct {
```
roman-khimov, line 196: It's more of linux than unix.

pkg/local_object_storage/blobstor/fstree/fstree_write_linux.go
```diff
@@ -192,20 +193,115 @@ func (w *linuxWriter) finalize() error {
...
+
+type unixCombinedFileWriteStream struct {
+	linuxWriter *linuxWriter
+	targetPath  string
+	objectID    oid.ID
+	data        []byte
+	aborted     bool
+}
+
+func newUnixCombinedFileWriteStream(w *linuxWriter, targetPath string, objID oid.ID, dataLen uint64) *unixCombinedFileWriteStream {
```
roman-khimov, line 241: Really tricky thing. I tend to think that keeping replicatev1 for small objects allows us to avoid combined writer logic for streams. Combined writer is efficient when it has all the data, otherwise a single slow writer (and it depends on caller, it has to pass the next message before we could write anything) slows down the whole batch. So I'd rather avoid combining for this case irrespective of the size.

pkg/local_object_storage/blobstor/fstree/fstree_write_generic.go
cthulhu-rider: yes. But `fs.ErrExists` is more widely used and OS-independent

pkg/local_object_storage/blobstor/fstree/fstree.go
cthulhu-rider: it can be framed/unsorted in general. Dont mind to start w/ `[]byte`

internal/testutil/os.go
cthulhu-rider: made `fstest` package

pkg/local_object_storage/blobstor/fstree/fstree_write_linux.go
cthulhu-rider: renamed

pkg/local_object_storage/blobstor/fstree/fstree_write_generic.go
cthulhu-rider: it had break condition. Rewritten now

pkg/local_object_storage/blobstor/fstree/fstree_write_generic.go
cthulhu-rider: done

pkg/local_object_storage/blobstor/fstree/fstree_write_linux.go
cthulhu-rider: rly makes sense! Done

pkg/local_object_storage/blobstor/fstree/fstree_write_generic.go
roman-khimov: I mean the sequence of
```
case errors.Is(pe.Err, fs.ErrExist):
			return fs.ErrExist
```
just strips the outer context (if it's present), otherwise it's pretty useless, we already have `fs.ErrExist`.

pkg/local_object_storage/blobstor/fstree/fstree.go
roman-khimov: Framing can interfere, ok.

pkg/local_object_storage/blobstor/fstree/fstree_write_generic.go
cthulhu-rider: yes, and prevents wrapping into new context below. This was the case with `syscall.EEXIST`. Please elaborate on your point

roman-khimov (approved):

PR #4175, opened 2026-09-11
writecache: migrate FSTree layout to depth 2

### LINKED ISSUE #3632: Tune write cache FSTree depth
I'm always frustrated when we have too many objects in FSTree directories. We use `depth: 1` in write cache FSTree for some time, the presumption was that it never has too many objects, while in fact it can if we're pushing small ones. 1G cache for 1K objects is 1M objects or 20K per dir. Larger caches are also possible, 10-100G are all fine.

Switch to FSTree depth 2 for write cache, migrate.

Switch to 3, but this seems to add too much overhead for more common case of 1-10M objects when we have 1-10K of them in the cache.

## Additional context
#2337

End-rey:
Closes #3632

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Changelog for NeoFS Node
 ### Changed
 - SN requests network map only when it has been changed (#4121)
 - SN pushes URI-style endpoints to netmap instead of multiaddr (#3982)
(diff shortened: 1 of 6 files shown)


roman-khimov (commented):
Chlog

roman-khimov (approved):

PR #4176, opened 2026-09-11
Fix and improve EC part range request buffer management

cthulhu-rider:


pkg/services/object/get.go
```diff
@@ -683,7 +683,7 @@ func (x *getECTransport) copyRemotePartRange(ctx context.Context, conn *grpc.Cli
 	}
 
 	copied, err := x.copyRemotePartRangeWithRequest(ctx, conn, request, ln, controlCh)
-	if err != nil || copied == ln {
+	if err != nil || (ln > 0 && copied == ln) {
```
roman-khimov: IIUC this is only relevant for `0,0` (whole payload) requests. Don't we have a leak in this case (`request` not returned to pool)?

pkg/services/object/get.go
```diff
@@ -683,7 +683,7 @@ func (x *getECTransport) copyRemotePartRange(ctx context.Context, conn *grpc.Cli
 	}
 
 	copied, err := x.copyRemotePartRangeWithRequest(ctx, conn, request, ln, controlCh)
-	if err != nil || (ln > 0 && copied == ln) {
+	if err != nil || copied > 0 {
```
roman-khimov: Makes more sense.

pkg/services/object/get.go
```diff
@@ -782,11 +782,15 @@ func (x *getECTransport) copyRemotePartRangeWithRequest(ctx context.Context, con
 			return 0, errors.New("none of the supported oneof fields are specified")
 		}
 
+		copied += uint64(chunkLen)
```
roman-khimov: Technically not yet copied here, so can check without modifying the variable.

roman-khimov (approved):

pkg/services/object/get.go
cthulhu-rider: done

PR #4178, opened 2026-09-14
Support streaming write into write cache

cthulhu-rider:
- based on and blocked by #4171 (last 2 commits)


pkg/metrics/writecache.go
```diff
@@ -75,6 +76,10 @@ func (m writecacheMetrics) AddWCPutDuration(shardID string, d time.Duration) {
 	m.putDuration.With(prometheus.Labels{shardIDLabelKey: shardID}).Observe(d.Seconds())
 }
 
+func (m writecacheMetrics) AddWCInitPutDuration(shardID string, d time.Duration) {
```
roman-khimov: More like `WCStreamingPutDuration` to me, `InitPut` implies that this time is just for the `InitPut()` call only, while in fact it includes all of data receiving/writing time.

pkg/local_object_storage/writecache/put.go
```diff
@@ -45,6 +61,106 @@ func (c *cache) Put(addr oid.Address, _ *object.Object, data []byte) error {
...
+}
+
+func (x *objectPayloadWriteStream) Write(p []byte) (int, error) {
+	if x.aborted {
+		return 0, logicerr.ErrStreamAborted
+	}
+
+	n, err := x.fsTreeStream.Write(p)
+	if err != nil {
+		x.finish()
```
roman-khimov, line 131: `x.abort()` here?

cthulhu-rider: abort() calls FS tree's abort func, but we don't need this after stream error

pkg/metrics/writecache.go
cthulhu-rider: reasonable, done

cthulhu-rider: for Put, i noticed that we submit both OK and error durations into one metric. Although failures happen fast usually, while writing can require more time. This diff may be even more noticeable in InitPut due to reading. Seems like it'd be worthwhile to have separate metrics for OKs and fails. At the same time, Put metric doesn't depend on the object size, which could also complicate its analysis. Let's consider in a separate issue if there are consonants

roman-khimov (approved):

PR #4180, opened 2026-09-15
*: drop local EC attributes key consts, use SDK instead

carpawell:


roman-khimov (approved):

carpawell:
Fixed commit, it was amended to copilot's one.

cthulhu-rider (approved):

Issue #3745, opened 2025-12-19
Do not try to put to blobstor on full write cache


roman-khimov:
## Is your feature request related to a problem? Please describe.
I'm always frustrated when I'm thinking of write cache and blobstor relations. Currently we always try to put object to blobstor if putting to write cache fails.

## Describe the solution you'd like
Don't do that in case write cache is full. If that's the case it's supposed to be flushing to blobstor at the maximum speed possible and additional threads pushing objects to blobstor directly can only slow it down. If it's full because it can't flush objects to blobstor trying this direct put doesn't help either.

## Describe alternatives you've considered
Keep things as is. Current behavior is somewhat better for the case when write cache can't put objects to blobstor, but direct put somehow can. I don't see this happening, but who knows.

## Additional context
https://github.com/nspcc-dev/neofs-node/pull/3744#discussion_r2635320251

PR #4182, opened 2026-09-17
Support streaming write into storage shard

cthulhu-rider:


cthulhu-rider:
code is rdy for review, gonna add some tests

pkg/local_object_storage/shard/put.go
```diff
@@ -34,15 +36,39 @@ func (s *Shard) Put(obj *object.Object, objBin []byte) error {
 		objBin = obj.Marshal()
 	}
 
-	var (
-		addr      = obj.Address()
-		cachedPut bool
-	)
+	var addr = obj.Address()
+
+	writeCacheFn := func(writeCache writecache.Cache) error {
```
carpawell, line 43: tbh, i have never understood this pattern: inside `T`'s method creating a func locally to pass it to another `T`'s method

pkg/local_object_storage/blobstor/common/storage.go
```diff
@@ -144,6 +144,7 @@ type Storage interface {
 	ReadObjectParts(buf []byte, addr oid.Address, rng PayloadRange, interceptHeaderBinaryFn func([]byte) error) (int, io.ReadCloser, error)
 	Exists(oid.Address) (bool, error)
 	Put(oid.Address, []byte) error
+	InitPut(oid.Address, uint64, uint64, io.WriterTo) (io.WriteCloser, func(), error)
```
carpawell, line 147: we have `Get` and `GetStream`, why not `Put` and `PutStream`?

pkg/local_object_storage/shard/put.go
```diff
@@ -62,6 +64,173 @@ func (s *Shard) Put(obj *object.Object, objBin []byte) error {
...
+// [writecache.Cache.InitPut]. In this case, if InitPut or resulting stream
+// fails, fallback to the main storage is performed if possible.
+//
+// If s is in read-only mode, InitPut instantly returns [ErrReadOnlyMode]. If
+// underlying [common.Storage] is in read-only mode, InitPut returns
+// [common.ErrReadOnly].
+//
+// If underlying device runs out of space, InitPut or resulting stream calls
+// return [common.ErrNoSpace].
+func (s *Shard) InitPut(hdr object.Object, hdrLen uint64, hdrW io.WriterTo) (io.WriteCloser, func(), error) {
```
carpawell, line 78: it is not clear to me, why there are `hdr` and `hdrW`. docs also do not explain it

pkg/local_object_storage/shard/put.go
```diff
@@ -119,3 +287,11 @@ func (s *Shard) putToMetabaseLocked(addr oid.Address, hdr object.Object, cachedP
 
 	return nil
 }
+
+func (s *Shard) logPutWriteCacheError(err error) {
+	s.log.Debug("can't put object to the write-cache, trying blobstor", zap.Error(err))
```
carpawell, line 252: WARN to me

pkg/local_object_storage/shard/put.go
cthulhu-rider: there is a need to share the code, and this allows to do it

pkg/local_object_storage/blobstor/common/storage.go
cthulhu-rider: well, maybe PutStream would be better. It does not rly matter for now, i follow internal calls

pkg/local_object_storage/shard/put.go
cthulhu-rider: `hdrW` supports io, `hdr` just provides access to some fields for handling

pkg/local_object_storage/shard/put.go
cthulhu-rider: nothing new, it's been debug for a long time

roman-khimov: Currently it's normal, exactly what happens with full write cache which is inherently limited, so debug is OK. Refs. #3745.

pkg/local_object_storage/shard/put.go
```diff
@@ -34,15 +36,39 @@ func (s *Shard) Put(obj *object.Object, objBin []byte) error {
...
 	}
 
-	var (
-		addr      = obj.Address()
-		cachedPut bool
-	)
+	var addr = obj.Address()
+
+	writeCacheFn := func(writeCache writecache.Cache) error {
+		return s.writeCache.Put(addr, obj, objBin)
```
roman-khimov: Should use `writeCache` parameter here then.

pkg/local_object_storage/shard/put.go
```diff
@@ -34,15 +36,39 @@ func (s *Shard) Put(obj *object.Object, objBin []byte) error {
...
+	}
+
+	blobStorageFn := func(blobStorage common.Storage) error {
+		return s.blobStor.Put(addr, objBin)
```
roman-khimov: Same.

pkg/local_object_storage/shard/put.go
```diff
@@ -62,6 +64,173 @@ func (s *Shard) Put(obj *object.Object, objBin []byte) error {
...
+
+func (x *payloadWriteStream) putToMetabase() error {
+	if !x.cachedPut {
+		logOp(x.shard.log, putOp, x.header.Address())
+	}
+
+	return x.shard.putToMetabaseLocked(x.header.Address(), x.header, x.cachedPut)
+}
+
+func (x *payloadWriteStream) trySwitchToBLOBStorage(err error) error {
```
roman-khimov: Not worth it. It can only handle the first `Write()` failure which is not very likely, there is a file handle already opened, header is already written into it and now the next (but not some other!) `Write()` fails. Not a lot of people would complain if this write fails completely (it's much more likely to fail on `Open()`, the first real `Write()` which is done for the header or any of subsequent ones). So this fragile code can be avoided completely.

cthulhu-rider: this is the closest implementation to Put. Ok, dont mind simplifying it

roman-khimov: Put is very much different because it has a complete payload. It can retry with blobstor in case of _any_ error with write cache. This specific path only does it in case of a _second_ `write()` (as in syscall) failure. Failure on open/first write is much more likely and these are handled properly already.

pkg/local_object_storage/shard/put.go
cthulhu-rider: done

pkg/local_object_storage/shard/put.go
cthulhu-rider: done

pkg/local_object_storage/shard/put.go
cthulhu-rider: this means it's also worth implementing for engine with multiple shards

pkg/local_object_storage/shard/put.go
carpawell: but all you need is addr and payload size. on the other levels, you use it as `addr oid.Address, headerLen uint64, payloadLen uint64, headerW io.WriterTo`. just another thing to consider when reviewing and checking the logic. the difference is not clear

pkg/local_object_storage/shard/put.go
carpawell: different named methods are clearer to me than sharing via anon callbacks

pkg/local_object_storage/shard/put.go
carpawell: > nothing new

for new methods it could be WARN

> exactly what happens with full write cache which is inherently limited\

depends on whether we consider this a normal situation. ok

carpawell (approved):

pkg/local_object_storage/shard/put.go
cthulhu-rider: deeper components just store binaries (at least they are moving towards it). Address is a storage key, lengths are needed for cache control, internal layout, etc. And header is separated from payload. Otherwise, it's a plain i/o. But shard also makes metabase write which currently requires structured header. That's the diff

pkg/local_object_storage/shard/put.go
cthulhu-rider: ur suggesting to not have putFunc sharing writecache-or-blobstorage logic at all? Otherwise, what signatures would u suggest?

roman-khimov (approved):

PR #4187, opened 2026-09-21
Drop nonce attribute from metabase

roman-khimov:


pkg/local_object_storage/metabase/version.go
```diff
@@ -344,3 +345,42 @@ func splitAttributeValueObjectID(b []byte) ([]byte, []byte, error) {
 
 	return b[:valEnd], b[valEnd+1:], nil
 }
+
+func migrateFrom11Version(db *DB) error {
+	err := updateContainersInterruptable(db, []byte{metadataPrefix}, dropNonceIndexes)
+	if err != nil {
+		return fmt.Errorf("drop __NEOFS__NONCE indexes: %w", err)
```
carpawell: can be SDK's const?

pkg/local_object_storage/metabase/version.go
```diff
@@ -344,3 +345,42 @@ func splitAttributeValueObjectID(b []byte) ([]byte, []byte, error) {
...
+	return db.boltDB.Update(func(tx *bbolt.Tx) error {
+		return updateVersion(tx, 12)
+	})
+}
+
+func dropNonceIndexes(_ *zap.Logger, _ *bbolt.Tx, b *bbolt.Bucket, _ cid.ID, _ []byte, limit uint) (uint, []byte, error) {
+	var (
+		c            = b.Cursor()
+		k            []byte
+		attrIDPrefix = []byte{metaPrefixAttrIDPlain}
```
carpawell, line 363: if there are int values, this code skips them

roman-khimov: Makes sense. Currently S3 places b64-encoded things, but it's defined as any data, so can be numeric.

pkg/local_object_storage/metabase/version.go
roman-khimov: OK

carpawell: `"str"+"str"+"str"` in `Errorf`?

carpawell (approved):

pkg/local_object_storage/metabase/version.go
roman-khimov: It's more effective than "%s".

